PR #55, opened 2021-04-30
Remove container after run

andreasjansson:


bfirsh:
I think the main issue is all the tagged Docker images. Not only are they not removed, they also don't automatically get removed by `docker system prune`.

PR #64, opened 2021-05-06
GPU build

andreasjansson:
To facilitate GPU builds, I've refactored the builder to use an
in-process queue. This allows us to throttle the number of concurrent
GPU builds to 1.

Also added logic to exit the build early if the client exists.


bfirsh:
marvelous

PR #66, opened 2021-05-06
Fix concurrency issues

andreasjansson:
* Running cpu and gpu build in parallel, there was a race condition when saving example output
* Throttling was broken, fixed with semaphore


bfirsh:
✨

PR #67, opened 2021-05-07
Refactoring

bfirsh:
Opening for visibility.


andreasjansson:
Controversial maybe, but I agree with this guy about util packages: https://about.sourcegraph.com/go/idiomatic-go/#Package-util. Is the reason for the util package to avoid cluttering the root tree?

PR #74, opened 2021-05-13
Update web hooks

andreasjansson:
* Run "post-build" hook before "post-build-primary" hook
* Include image URI and arch


bfirsh:
Looks great! I'm not sure I fully understand what the hooks are for without looking into it in more detail though. (e.g. What is the difference between a "primary" and "non-primary" hook?) Perhaps needs documentation?

Also needs a rebase sorry :)

bfirsh:
Still missing docs. Unless you think we don't need it?

andreasjansson:
@bfirsh there's a TODO in there to add docs, i'll add that later. it might still move around a bit.

PR #76, opened 2021-05-14
Tail image build logs on push with -l param

andreasjansson:

### DIFF
--- a/end-to-end-test/end_to_end_test/test_server.py
+++ b/end-to-end-test/end_to_end_test/test_server.py
@@ -34,15 +34,10 @@ def cog_server_port_dir():
     server_proc.kill()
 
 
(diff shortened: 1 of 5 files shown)


bfirsh:
Nice!

A few thoughts:
- We should avoid adding single character flags without a lot of thought. The process (which we should document) should probably be add a long flag, then over time add short flags after we tease out usage. User interface also could do with improvement here but I will be tearing apart user interface so no problem.
- Would it be too hard to add basic tests for this behavior? I'm wondering now we are in a "beta" phase, we should be adding tests for any behavior.

If tests are hard or you don't think it's important feel free to merge!

andreasjansson:
@bfirsh Added test and removed `-l` shorthand.

PR #81, opened 2021-05-16
Local prep work

bfirsh:
Extracting a few preparatory bits of work from my work on local mode so as not to make it a nightmare to merge. Feel free to merge at any point, and I will continue to add to this.


andreasjansson (changes requested):
Test failures are legit, otherwise looks good!

bfirsh:
Thanks! I swear they passed locally 🤔

PR #83, opened 2021-05-18
Fix concurrent map read issue in queue

andreasjansson:


bfirsh:
I have no idea what is going on here but maybe there is a way to make this work with channels instead of shared state?

PR #84, opened 2021-05-19
Lint everything

bfirsh:
giphy

Somewhat hesitant to put it in `make test` because that slows down the development cycle. Mayeb we can set up pre-commit or something.


bfirsh:
Also I removed `--local` from goimports because that can't be configured from vscode. Weirdly though it does it correctly if there are already some local imports, so in practice it'll keep the local ones sorted locally. This just avoids some annoying back and forths I had in Keepsake where the CI was doing goimports differently to how my editor was.

andreasjansson:
👍

PR #82, opened 2021-05-17
Batch inference on the CLI

andreasjansson:
Adds --file and --output-dir arguments to cog infer.

## Example usage

Given the following infer.py script that prefixes file contents with "hello ":
```python
# infer.py
from pathlib import Path
import cog

class Model(cog.Model):
    def setup(self):
        self.prefix = "hello"

    @cog.input("file", type=Path, help="File containing text that will get prefixed by 'hello '")
    def run(self, file):
        with file.open() as f:
            return self.prefix + " " + f.read()
```

...you can process an arbitrary number of input files by creating a file containing the filenames process. In this case, we call the file "inputs.txt":

```
# inputs.txt
file=@inputs/file1.txt
file=@inputs/file2.txt
file=@inputs/file3.txt
# [...]
```

The input file contains one line per example, where each example is in the form `<key>=<value>[,<key>=<value>[, ...]]`. As in the single-example `cog infer --input ...` command, filenames are prefixed with `@`.

To process the files in inputs.txt, invoke cog infer with:

```
cog infer --file=inputs.txt --output-dir=outputs my-name/my-model:version
```

This will put outputs in a directory called `outputs`, and the files will be named `outputs/output.0000002.txt`, `outputs/output.0000003.txt`, `outputs/output.0000004.txt`, etc. The number (e.g. `0000002`) corresponds to the line number in inputs.txt. In this case the line number starts at 2 since line 1 is a comment.

Examples that fail to process will output a warning to stderr, but processing will continue.

---

_This PR doesn't add batch infer capabilities to the Python library._

Signed-off-by: andreasjansson <andreas@replicate.ai>

Closes #29

PR #85, opened 2021-05-20
Handle todos

andreasjansson:
Handle various TODOs, broken up into individual commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,10 +23,21 @@ clean:
 generate:
 	go generate ./...
 
-.PHONY: test
-test: check-fmt vet lint
+.PHONY: test-go
+test-go: check-fmt vet lint
 	go get gotest.tools/gotestsum
 	go run gotest.tools/gotestsum -- -timeout 1200s -parallel 5 ./... $(ARGS)
+
+.PHONY: test-end-to-end
+test-end-to-end: install
+	cd end-to-end-test/ && $(MAKE)
+
+.PHONY: test-cog-library
+test-cog-library:
+	cd pkg/docker/ && pytest cog_test.py
+
+.PHONY: test
+test: test-go test-end-to-end test-cog-library
 
 .PHONY: install
 install:
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ test-cog-library:
 	cd pkg/docker/ && pytest cog_test.py
 
 .PHONY: test
-test: test-go test-end-to-end test-cog-library
+test: test-go test-cog-library test-end-to-end
 
 .PHONY: install
 install:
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -28,6 +28,17 @@ test:
...
+.PHONY: test-end-to-end
+test-end-to-end: install
+	cd end-to-end-test/ && $(MAKE)
+
+.PHONY: test-cog-library
+test-cog-library:
+	cd pkg/docker/ && pytest cog_test.py
+
+.PHONY: test-all
+test-all: test test-end-to-end test-cog-library
```
bfirsh, line 40: `make test` should probably test everything, right?

pkg/docker/cog.py
```diff
@@ -504,6 +504,11 @@ def validate_and_convert_inputs(
     return inputs
 
 
+def run_model(model, inputs, cleanup_functions):
+    result = model.run(**inputs)
+    if isinstance(result, Path):
+        cleanup_functions.append(lambda: shutil.rmtree(result))
+    return result
```
bfirsh, line 509: Not important, but the passing by reference took some mental gymnastics to figure out what was going on. Maybe a clearer implementation would be to return `result, cleanup_function` where the latter is either a function or `None`.

pkg/docker/cog.py
```diff
@@ -158,7 +158,7 @@ def handle_request():
                         )
                     except InputValidationError as e:
                         return jsonify({"error": str(e)})
-                    results.append(self.model.run(**instance))
+                    results.append(run_model(self.model, instance, cleanup_functions))
```
bfirsh, line 161: Looks like `cleanup_functions` are never called here.

pkg/docker/cog.py
```diff
@@ -506,7 +506,7 @@ def validate_and_convert_inputs(
 def run_model(model, inputs, cleanup_functions):
     result = model.run(**inputs)
     if isinstance(result, Path):
-        cleanup_functions.append(lambda: shutil.rmtree(result))
+        cleanup_functions.append(lambda: result.unlink())
```
bfirsh, line 508: Can be squashed.

bfirsh (approved):
Great! Some minor things, but nothing blocking merging.

This is based on #82 which I'm not sure is intentional or not. Happy to include that if you want to get it in and we can iterate, otherwise might want to base this on `main` if we want to review that separately.

pkg/docker/cog.py
andreasjansson: 👍

pkg/docker/cog.py
andreasjansson: I added documentation to this function.

Makefile
andreasjansson: Yep, done. The old `make test` is now `make test-go`.

PR #86, opened 2021-05-20
Handle symlinks in uploaded models

andreasjansson:

### DIFF
--- a/pkg/util/zip/write.go
+++ b/pkg/util/zip/write.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


bfirsh:
Do we want to follow symlinks? This is 1) could be a security issue and 2) does not produce consistent results on different machines. I feel like it should be a property of Cog that two directories that are the same on different machines should produce the same version.

For context, Docker and git don't follow symlinks in any way. tar and zip don't unless explicitly told to.

bfirsh:
Test failures are real, too!

bfirsh:
I suppose another way of thinking about this – it should probably behave like doing a Docker build, and Docker doesn't follow symlinks.

andreasjansson:
@bfirsh The use case I had in mind is when you have some small symlinked test set that's shared between several models. But I agree that repeatability probably trumps that use case. I'll scrap this and print warnings when there are symlinks in the directory.

PR #87, opened 2021-05-21
Handle symlinks in uploaded models

andreasjansson:
Print warning if there are symlinks in the model directory and ignore them. I added a TODO to maybe support symlinks that point to files within the tree, but that's not supported in this PR.


bfirsh:
Marvellous! ✨

PR #90, opened 2021-05-27
Throw error if user tries to upload the same version twice

### LINKED ISSUE #73: Throw error if version already exists

andreasjansson:

Closes #73


bfirsh:
Test failures are real!

bfirsh:
Looks good otherwise!

Not important because it seems unlikely, I think two pushes simultaneously might cause weird things to happen. Maybe it will just silently get overwritten with the one that gets written last, which seems reasonable.

andreasjansson:
> Looks good otherwise!
> 
> Not important because it seems unlikely, but I don't think handles the case where two pushes happen simultaneously. Maybe it will just silently get overwritten with the one that gets written last, which seems reasonable, but this will cause two webhooks to get sent for the same ID which will throw an error.

Yeah I think we can live with that for now at least.

Fixed tests.

bfirsh:
✨

Issue #94, opened 2021-05-31
Design prediction API versioning


bfirsh:
I want to rename `/infer` to `/predict` but we can't change it because all the old models have it and there is no way for Cog to detect what version it is and what it should call.

Alternatively, perhaps it is the _client_ which should detect what version of Cog the model has been made with, and adjust its API calls as appropriate.

evilstreak:
Checking the version of Cog is how we're handling similar issues for Replicate.

Renaming an endpoint feels easy to sequence, in a way that a caller wouldn't have to check Cog versions:

1. Release a new version of Cog that adds `/predict` and has a deprecation warning on `/infer`
2. Rebuild all of your active models with the latest version of Cog
3. Update your caller to use `/predict` instead of `/infer`
4. At some point in the future, we release a new version of Cog that drops `/infer`, and everything keeps working.

The case where we want to make a breaking change to the behaviour of an endpoint is trickier, and is maybe an argument for versioning. Versioning would turn a breaking-change problem into something you could sequence like the endpoint renaming situation above. It also breaks more cleanly: you get a `no-such-endpoint /v1/foo` error when upgrading Cog too far without having changed your calls, rather than it crashing out somewhere because something about the request or response body is different.

zeke:
> We could defer this decision. `/predictions` could just be equivalent to the first version, and if we ever want to upgrade the API in a breaking way, we just add `/vwhatever/predictions` and remove `/predictions`.

👍🏼

bfirsh:
My inclination is to close this issue, then. If we feel the pain of prediction API versioning, then we know where to find the thinking around this!

PR #98, opened 2021-05-31
Rename "infer" and "run" to "predict"

bfirsh:
Opened up a fun versioning problem #94. Until that is resolved, `cog predict` actually calls `/infer`.

This will break pushing old models -- there is no backwards compatibility for `run()`. Predictions on existing models that have been built will keep on working, because we are still calling `/infer`.


bfirsh:
@andreasjansson Will merge and we can review post-hoc. :)

andreasjansson:
This all looks good! Thanks for keeping the deprecated `/infer` endpoint around.

PR #99, opened 2021-05-31
Local mode: `cog run` and local `cog predict`

bfirsh:
Based on #98


pkg/cli/run.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/replicate/cog/pkg/util/terminal"
+	"github.com/spf13/cobra"
+)
+
+func newRunCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run <command> <args>",
+		Short: "Run a command inside a Docker environment",
+		RunE:  run,
+		Args:  cobra.MinimumNArgs(1),
```
andreasjansson, line 22: @bfirsh should this not be `2`?

pkg/cli/run.go
```diff
@@ -0,0 +1,92 @@
...
+
+	flags := cmd.Flags()
+	// Flags after first argment are considered args and passed to command
+	flags.SetInterspersed(false)
+
+	return cmd
+}
+
+func run(cmd *cobra.Command, args []string) error {
+	arch := "cpu"
```
andreasjansson, line 33: `TODO: support multiple run architectures, or automatically select arch based on host`

pkg/docker/generate.go
```diff
@@ -176,14 +180,17 @@ RUN apt-get update -q && apt-get install -qy --no-install-recommends \
 }
 
 func (g *DockerfileGenerator) installCog() (string, error) {
-	cogFilename := "cog." + ksuid.New().String() + ".py"
+	cogFilename := ".cog/tmp/cog." + ksuid.New().String() + ".py"
```
andreasjansson, line 183: nice!

pkg/docker/local_builder.go
```diff
@@ -233,19 +231,15 @@ func buildPipe(pipe io.ReadCloser, logWriter logger.Logger) (lastLogs []string,
...
 		line := scanner.Text()
 		logWriter.Debug(line)
 
 		if strings.Contains(line, sectionPrefix) {
 			currentSection = strings.SplitN(line, sectionPrefix, 2)[1]
-			currentLogLines = []string{}
-			logWriter.Infof("  * %s", currentSection)
-		} else {
-			currentLogLines = append(currentLogLines, line)
+			logWriter.Infof(currentSection)
```
andreasjansson, line 242: How does this display stack traces if some build step fails? The previous behavior made it quite clear what exactly failed. Are we now only showing the last _n_ lines of the error message? Is there a way to see the full failure trace from docker build?

pkg/model/config.go
```diff
@@ -125,14 +125,13 @@ func (c *Config) ValidateAndCompleteConfig() error {
...
-		return fmt.Errorf("'model' in cog.yaml must be in the form 'model.py:ModelClass")
-	}
-	if strings.Contains(c.Model, "/") {
-		return fmt.Errorf("'model' in cog.yaml cannot be in a subdirectory. It needs to be in the same directory as cog.yaml, in the form 'model.py:ModelClass")
+	if c.Model != "" {
+		if len(strings.Split(c.Model, ".py:")) != 2 {
+			return fmt.Errorf("'model' in cog.yaml must be in the form 'model.py:ModelClass")
+		}
+		if strings.Contains(c.Model, "/") {
+			return fmt.Errorf("'model' in cog.yaml cannot be in a subdirectory. It needs to be in the same directory as cog.yaml, in the form 'model.py:ModelClass")
```
andreasjansson, line 133: We do actually kind of support model scripts in subdirectories. If you use `workdir` the model script will be loaded from that directory.

andreasjansson (commented):
Awesome! Just some minor nits, and a question about stack traces in failed builds.

pkg/model/config.go
bfirsh: I didn't change this behavior -- it has shown up as an addition just because I reindented.

pkg/docker/local_builder.go
bfirsh: Yes -- behavior is the same but this is now handled by the consumer of the log rather than the writer.

bfirsh:
Thanks for the review!

pkg/cli/run.go
bfirsh: Nope, I don't think so. It just needs a command to run. Usage string could perhaps be clearer.

pkg/model/config.go
andreasjansson: Ah right, then that error message should be fixed.

PR #105, opened 2021-06-01
Make end to end tests parallel

bfirsh:
In theory this should make end to end tests much faster. However -- `find_free_port()` has a race condition. When port is closed, another parallel process can take the port. Perhaps it can defer closing.


bfirsh:
Kind of pointless now they're so fast. Let's revisit this if they're slow again.

Issue #102, opened 2021-05-31
Cog should work in subdirectories


bfirsh:
Cog should search up the file tree for `cog.yaml`, like Keepsake does.

For example, if `/home/ben/hotdog-detector/cog.yaml` exists, then I should be able to run `cog predict` in `/home/ben/hotdog-detector/subdir/` and it should do what I expect.

There is some nuance here with `cog run`. Should the working directory be the relative current directory inside the container?

synek:
Hey I've got a WIP branch for this and will open a PR soon for you to review.

bfirsh:
Thanks for your help here @synek! Will consider this closed and open another issue for the `cog run` thing.

Issue #93, opened 2021-05-30
Concept of "workdir" doesn't map to Docker's concept of "workdir"


bfirsh:
Cog's is relative to `/code`, Docker's is absolute. This is quite confusing, particularly when `cog run` gets involved.

As far as I can see, the intention of cog.yaml's workdir is two-fold: to set PYTHONPATH correctly, and [as a shortcut to set the directory for `post_install`](https://github.com/replicate/cog-examples/blob/3da7867253f66097b0ea83f3e98b4d72a13a08d5/StyleCLIP/cog.yaml).

PR #118, opened 2021-06-07
refactor: re-usable config loading, and basic error definitions

synek:
#102 

### What does this PR do?
* Standardises config-related functionality in `utils/config.go`
* Eliminates some repetitious config loading code
* Adds a func to walk the parent directories in search of a `cog.yaml`
* Adds some basic error creation code in `pkg/errors/errors.go`

### Next Steps/Questions
* Is this a reasonable approach?
* ~~How do we want `cog run` to work - relative to `cwd` or relative to the root (see #102)?~~
* ~~What else needs to change to allow `cog` usage in subdirectories?~~
* What's the best way to test this? Unit tests (where?), some addition to the e2e tests?

(If you like the config-loading (`utils/config.go`) refactor, we could merge that as a separate PR first to keep PRs small and lean)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ Untitled*.ipynb
 __pycache__
 .cog
 dist
+# Used by a vim plugin (projectionist)
+.projections.json
+.venv/
--- a/end-to-end-test/end_to_end_test/test_queue_worker.py
+++ b/end-to-end-test/end_to_end_test/test_queue_worker.py
@@ -1,18 +1,19 @@
 import json
-import redis
-from contextlib import contextmanager
 import multiprocessing
-from flask import Flask, request, jsonify, Response
+from contextlib import contextmanager
+
+import redis
+from flask import Flask, Response, jsonify, request
 
 from .util import (
+    docker_run,
+    find_free_port,
+    get_bridge_ip,
+    push_with_log,
     random_string,
     set_model_url,
     show_version,
-    push_with_log,
-    find_free_port,
-    docker_run,
-    get_local_ip,
-    wait_for_port,
+    wait_for_port
 )
 
 
@@ -28,9 +29,9 @@ def test_queue_worker(cog_server, project_dir, redis_port, tmpdir_factory):
     input_queue = multiprocessing.Queue()
     output_queue = multiprocessing.Queue()
     controller_port = find_free_port()
-    local_ip = get_local_ip()
-    upload_url = f"http://{local_ip}:{controller_port}/upload"
-    redis_host = local_ip
+    bridge_ip = get_bridge_ip()
+    upload_url = f"http://{bridge_ip}:{controller_port}/upload"
+    redis_host = bridge_ip
     worker_name = "test-worker"
     predict_queue_name = "predict-queue"
     response_queue_name = "response-queue"
@@ -67,7 +68,7 @@ def test_queue_worker(cog_server, project_dir, redis_port, tmpdir_factory):
                             "path": {
                                 "file": {
                                     "name": "myinput.txt",
-                                    "url": f"http://{local_ip}:{controller_port}/download",
+                                    "url": f"http://{bridge_ip}:{controller_port}/download",
                                 }
                             },
                         },
@@ -93,7 +94,7 @@ def test_queue_worker(cog_server, project_dir, redis_port, tmpdir_factory):
                             "path": {
                                 "file": {
                                     "name": "myinput.txt",
-                                    "url": f"http://{local_ip}:{controller_port}/download",
+                                    "url": f"http://{bridge_ip}:{controller_port}/download",
                                 }
                             },
                         },
--- a/pkg/cli/model.go
+++ b/pkg/cli/model.go
@@ -3,15 +3,13 @@ package cli
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"github.com/spf13/cobra"
 
 	"github.com/replicate/cog/pkg/client"
-	"github.com/replicate/cog/pkg/global"
 	"github.com/replicate/cog/pkg/settings"
+	"github.com/replicate/cog/pkg/util/config"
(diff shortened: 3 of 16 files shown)


bfirsh:
Amazing, thank you @synek!

> Is this a reasonable approach?

Looks good to me! We have a similar feature in Keepsake, and we can probably lift the code used for that: https://github.com/replicate/keepsake/blob/d9a7922556748e9e913e83c48e5378b7324eeac7/go/pkg/config/load.go#L133-L161

It has some failsafes about max number of directories to search through, etc.

Another nit: This should probably just be `pkg/config`. `pkg/util/config` is intended for things which are just standalone packages not related to Cog. [This is documented here.](https://github.com/replicate/cog/blob/main/CONTRIBUTING.md#project-structure)

> How do we want cog run to work - relative to cwd or relative to the root (see #102)?

I'm not sure, to be honest. It feels like it should be relative to cwd, but I think this might also be incompatible with workdir in some weird ways. See #93.

I'm happy to leave this half finished and get this piece of work in, then we can iterate.

> What else needs to change to allow cog usage in subdirectories?

I'm not sure off the top of my head, but hopefully using it will raise some issues. ;)

> What's the best way to test this? Unit tests (where?), some addition to the e2e tests?

Unit tests of the config loading make sense, I think. Then, maybe a simple end to end test that checks everything works as expected in a subdirectory.

Thanks again!

synek:
Thanks for the reply @bfirsh.

> We have a similar feature in Keepsake, and we can probably lift the code used for that

Done, and I took the liberty of porting over some of the `errors.go` code as well - yay or nay?

> This should probably just be `pkg/config`

Done.

> I'm happy to leave this half finished and get this piece of work in, then we can iterate.

Agreed. I renamed the PR to match.

> Unit tests of the config loading make sense, I think. Then, maybe a simple end to end test that checks everything works as expected in a subdirectory.

Sounds good. I'm going to bed now, but will look into that tomorrow. Presumably we can borrow from Keepsake again?

pkg/cli/model.go
```diff
@@ -53,24 +51,18 @@ func modelSet(cmd *cobra.Command, args []string) error {
...
-		console.Warnf("%s does not exist in %s. Are you in the right directory?", global.ConfigFilename, cwd)
-	}
 
 	cli := client.NewClient()
 	if err := cli.CheckRead(model); err != nil {
 		return err
 	}
 
-	userSettings, err := settings.LoadProjectSettings(cwd)
+	userSettings, err := settings.LoadProjectSettings(projectDir)
```
synek, line 65: Does it make sense to pass in `projectDir` here instead of `cwd`? I'm not 100% clear on what the intent is/was.

bfirsh: Yep, I think this was a bug!

pkg/server/build.go
```diff
@@ -212,18 +212,13 @@ func (s *Server) saveExamples(result *BuildResult, dir string, config *model.Con
...
 }
 
+// Load the Cog config
 func (s *Server) ReadConfig(dir string) (*model.Config, error) {
-	configRaw, err := os.ReadFile(filepath.Join(dir, global.ConfigFilename))
-	if err != nil {
-		return nil, fmt.Errorf("Failed to read %s: %w", global.ConfigFilename, err)
-	}
-	config, err := model.ConfigFromYAML(configRaw)
+	config, _, err := config.GetConfig("") // Read from cwd by default
```
bfirsh, line 217: I think `dir` here is the temporary directory on the server with the config inside it, so this needs to read from there instead of the working directory.

bfirsh:
> Done, and I took the liberty of porting over some of the errors.go code as well - yay or nay?

Nice, yeah, sounds good!

Thanks so much for your help here. Left some comments inline. Looks like there are some linting/test failures too. Otherwise looks good I think.

synek:
Cool thanks for the review. 

I'm working on an e2e test now (just hit some trouble running the tests), and will get all checks passing. I'll ping you again @bfirsh when I think it's ready for a final review.

pkg/server/build.go
synek: Fixed, thanks for pointing it out!

pkg/cli/model.go
synek: Oh I definitely knew that and intended to fix it! :roll_eyes:

synek:
@bfirsh I added an e2e test and fixed linting (`make lint fmt test` passes), so I think this is ready for final review.

end-to-end-test/end_to_end_test/test_queue_worker.py
```diff
@@ -28,9 +29,9 @@ def test_queue_worker(cog_server, project_dir, redis_port, tmpdir_factory):
     input_queue = multiprocessing.Queue()
     output_queue = multiprocessing.Queue()
     controller_port = find_free_port()
-    local_ip = get_local_ip()
-    upload_url = f"http://{local_ip}:{controller_port}/upload"
-    redis_host = local_ip
+    bridge_ip = get_bridge_ip()
```
synek, line 32: I'm not 100% sure this change will be stable on other machines, so would appreciate if someone else could checkout this branch and run the e2e tests to check!

bfirsh (approved):
Boom, thanks! This is fantastic.

Unfortunately looks like the redis fix doesn't work on macOS. We should probably be running these tests inside a consistent Docker environment anyway so maybe let's just leave this broken for the time being and we can fix up. #131 

It'd be great to get commits that fix your own commits squashed too, but not a big deal. :)

.gitignore
```diff
@@ -3,4 +3,7 @@ release
 Untitled*.ipynb
 __pycache__
 .cog
-dist
\ No newline at end of file
+dist
+# Used by a vim plugin (projectionist)
+.projections.json
```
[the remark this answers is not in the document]
synek: I wasn't aware that that was a thing! Thanks for the heads up.

zeke:
@all-contributors please add @synek for code and tests

PR #119, opened 2021-06-08
Attempt to delete :latest tag before tagging with :latest

### LINKED ISSUE #18: Server eats up all disk space very quickly

andreasjansson:
Following up from https://github.com/replicate/cog/pull/88, this will
keep Cog from leaving dangling Docker images that take lots of disk
space.

Fixes #18

### DIFF
--- a/end-to-end-test/end_to_end_test/test_docker_cleanup.py
+++ b/end-to-end-test/end_to_end_test/test_docker_cleanup.py
@@ -1,0 +1,48 @@
+import os
+import subprocess
+from typing import List
+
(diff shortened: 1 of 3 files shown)


bfirsh:
Great!

Unfortunately this doesn't fix #80 because images are not tagged locally. This can be a separate piece of work though, so let's consider #18 fixed but I'll upgrade #80 to a bug because it's flat out broken. (This model I'm working on consumes all the disk space on my machine after ~5 predictions, lol.)

PR #145, opened 2021-07-14
Send logs to queue

andreasjansson:
Queue worker "tees" stdout/stderr to a Redis queue, such that each log
line becomes a message on the queue. The ID is the model URL during
setup, and the prediction ID during the run stage.

### DIFF
--- a/pkg/docker/cog.py
+++ b/pkg/docker/cog.py
@@ -439,6 +439,7 @@ class RedisQueueWorker:
         """
 
         class QueueLogger:
+        # TODO(bfirsh): maybe this should be a subclass of io.TextIOWrapper?
             def __init__(self, redis, queue, old_out):
                 self.redis = redis
                 self.queue = queue
--- a/pkg/docker/cog.py
+++ b/pkg/docker/cog.py
@@ -1,3 +1,4 @@
+import contextlib
 import signal
 import requests
 from io import BytesIO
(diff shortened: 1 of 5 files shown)


pkg/docker/cog.py
```diff
@@ -416,6 +431,51 @@ def upload_to_temp(self, path: Path) -> str:
...
         return resp.json()["url"]
 
+    @contextlib.contextmanager
+    def capture_log(self, stage, id):
+        """
+        Send each log line to a redis RPUSH queue in addition to an
+        existing output stream.
+        """
+
+        class QueueLogger:
```
bfirsh, line 441: `sys.stdout` and `sys.stderr` have a bunch of other properties/methods which code might depend on. They are `io.TextIOWrapper` classes, so maybe `QueueLogger` should subclass from that?

```suggestion
        class QueueLogger:
        # TODO(bfirsh): maybe this should be a subclass of io.TextIOWrapper?
```

This might not be a problem though, so I'm happy to get this in, then we can just see if it breaks code. :)

pkg/docker/generate.go
```diff
@@ -235,7 +235,7 @@ import os
 os.chdir("` + g.getWorkdir() + `")
 sys.path.append("` + g.getWorkdir() + `")
 from ` + module + ` import ` + class + `
-cog.RedisQueueWorker(` + class + `(), redis_host=sys.argv[1], redis_port=sys.argv[2], input_queue=sys.argv[3], upload_url=sys.argv[4], consumer_id=sys.argv[5]).start()`
+cog.RedisQueueWorker(` + class + `(), redis_host=sys.argv[1], redis_port=sys.argv[2], input_queue=sys.argv[3], upload_url=sys.argv[4], consumer_id=sys.argv[5], model_id=sys.argv[6], log_queue=sys.argv[7]).start()`
```
bfirsh, line 238: I wonder how we can make this "API" backward / future compatible. I expect we'll keep on adding new options, and old models should work with the new runner (and new models with old runners?).

bfirsh (approved):
Wonderful! Not merged myself for visibility on the subclass comment, but looks great.

PR #172, opened 2021-07-29
Andreas/docker registry credentials

andreasjansson:


bfirsh:
As we discussed on discord, adding a credentials helper means we need to add a new install method, which is not ideal. I did some cursory browsing and it looks like that prefix is [very hardcoded](https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/config/credentials/native_store.go#L10) and I can't think of a simple workaround. We can't install a binary in `/usr/local/bin` at runtime because that is not writeable on Linux. `.local/bin/` is in `$PATH` on Mac, but not Linux, and besides that seems flimsy.

Another thought: Could we just use Docker's credential store?

E.g. `cog login` could retrieve API token, then run `docker login registry.replicate.ai --username user --password-stdin`, with API token passed as stdin.

This might be better in a couple of ways:
- Credentials are stored securely in system keychain instead of plain text in our config file
- Much simpler and eliminates the need for a separate binary

The downside I can think of is that it is not easy to check if you are already logged in, in case we want to do that in Cog for some reason. (E.g. to fail fast when pushing when not logged in.) From some browsing of Stackoverflow, the simplest/fastest way might be to try pulling an empty image on the registry.

andreasjansson:
@bfirsh That's a great idea, let's do that!

andreasjansson:
superseded by https://github.com/replicate/cog/pull/184

PR #176, opened 2021-07-30
Add --progress={auto,plain,tty} option

andreasjansson:
Adds the same ability that Docker Compose and Docker have to control
build output. --progress=plain replaces buildkit's cursor movement
with a linear build log, making it useful on dumb
terminals (e.g. the Emacs Shell).

Builds on https://github.com/replicate/cog/pull/155


bfirsh:
We pass through environment, so is the problem that buildkit doesn't check for terminal capabilities / `TERM=dumb`? Could we do that check inside Cog and pass that through to Docker as a `--progress` option, instead of adding an extra option?

andreasjansson:
BuildKit does check for term capabilities, but Emacs shell (and probably others) don't report it correctly. I assume that's the reason Docker and Docker Compose chose to add the --progress flag.

bfirsh:
Gotchya. But Emacs shell sets `TERM=dumb`, right? Maybe we could detect that and automatically set `--progress=plain`, to make up for Docker's deficiency. (We should also submit a patch to Docker to fix this...)

We should probably also have this option, but I'm thinking this might actually improve this for you, so you don't need to remember to pass the option!

andreasjansson:
Superseded by https://github.com/replicate/cog/pull/187

Issue #181, opened 2021-07-30
Don't be tied hard to new Tensorflow/PyTorch versions


bfirsh:
I just have `tensorflow==2.5.0` in my `cog.yaml` and this shouldn't fail hard:

```
ⅹ Failed to generate Dockerfile: No matching tensorflow CPU package for version 2.5.0
```

It doesn't need to do anything for CPU packages (I think?) and even if it did, it should be a warning not a hard failure.

bfirsh:
This has been fixed for PyTorch in #229, but still an issue for Tensorflow I believe.

bfirsh:
It might be worth stepping back and think how we design this from a user's point of view. There are two cases where a user might be selecting a PyTorch version which is different to what Cog allows you do to:

1. They know what they are doing, and selecting a version that Cog doesn't support. Either Cog is doing something wrong, or they want to use a PyTorch version which isn't supported by Cog, or they're doing something else custom.
2. They are doing something wrong.

In the (1) case, they are doing something very intentional. In (2), they are not. The problem with displaying warnings, like we have done in #229 and #269 is that user (1) is fine, but user (2) might not see the warnings and then something is broken for some reason.

Perhaps Cog should fail hard when it fails to resolve versions, and then have some way that a user explicitly says "I know what I'm doing" and then it doesn't fail hard. Some `cog.yaml` option to disable magic version resolving, or something.

Either way, this deserves a bit more thought.

PR #183, opened 2021-07-31
Add Tensorflow 2.5.0 and 2.6.0

bfirsh:
This doesn't work, as per tests. I think it's because there's a missing base image, but I didn't have time to figure it out.

I've disabled Tensorflow package picking for CPU in the meantime, because it's all just `tensorflow`: https://github.com/replicate/cog/commit/c02b0a3791aa7a0f71af10c311e84c6770c4e16c

Also: PyTorch scraper is broken, it seems, so that's not updated here.

Related: #181 #204


bfirsh:
Tried regenerating but still doesn't work... I wonder what that error is...

bfirsh:
Superseded by #329.

PR #184, opened 2021-08-03
cog login to registry

andreasjansson:
The cog login command fetches a login token from a remote and
uses Docker's default credential helper to save that token.


bfirsh:
A few things:

- For simplicity and consistency, I wonder if we should just call `docker login r8.im --username ... --password-stdin`? Every other interaction with Docker is done through the CLI now, and I can imagine future incompatibility problems if we interact with Docker config directly.
- I was imagining the semantics of `cog login` to be "log me in to whatever registry I have specified in `cog.yaml`". In other words, it would shell out to `docker login` if the registry was a not `r8.im`. With this design, the user needs to be aware of the difference between a "Cog" registry and a "normal" registry, which I don't think they need to be if we shell out to `docker login`.
- User should initiate opening a browser by hitting a key, otherwise they are thrown into their web browser without any context.

I can fix this stuff if you want -- will check in after dinner :)

bfirsh:
As talked about on call with @andreasjansson:

`docker login` does special call to registry to check you're logged in, but this is not implemented on GCR. @andreasjansson tried to short-circuit that response, and Andreas gave up on that. It's much easier to do it this way. The way in currently works is a bodge until we figure out what the correct response is.

PR #190, opened 2021-08-11
Implement expected file methods on Cog's queue logger

andreasjansson:
Since we're overwriting sys.std{out,err} with our custom logger, it
needs to implement the file methods present on the overwritten objects.


bfirsh:
@andreasjansson test failures look real :(

andreasjansson:
@bfirsh fixed

PR #194, opened 2021-08-12
Use buildkit caching for pip and apt

bfirsh:
This dramatically speeds up `pip install` when adding/changing pip packages, because tensorflow/torch are so goddamn huge.


andreasjansson:
Excellent!

PR #195, opened 2021-08-12
Fix cog login when creds is not set in docker's config.json

andreasjansson:


andreasjansson:
> I'm not sure I understand what the problem is. What was going on here?

You can't set the `Auth` field directly in `AuthConfigs`. Instead you have to set `Username` and `Password` and the `Save()` method does the concatenation and base64 encoding.

PR #211, opened 2021-08-20
Set all labels after build

bfirsh:
A quote in a `pre_install` step was being escaped incorrectly.
Unsurprisingly -- we were putting a shell script inside an argument
inside a string, inside a JSON document, inside a string, inside a
Dockerfile instruction.

By doing this in the post-build labelling step this works properly
(shrug). We should be really be using proper APIs as much as we can
or we're going to be endlessly battling typing/escaping problems
like this.

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -1,9 +1,7 @@
 package dockerfile
 
 import (
-	"bytes"
 	_ "embed"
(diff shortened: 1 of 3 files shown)


bfirsh:
This should really have a test.

andreasjansson (approved):
Nice! I ran into this as well.

PR #215, opened 2021-08-24
Add hidden --registry flag to cog login

andreasjansson:
This is needed when developing against a local registry


bfirsh:
Super. `docker login` uses first argument for this. Not a problem because this is hidden, but just as a note to ourselves in case we want to map `cog login` closer to `docker login`.

PR #219, opened 2021-09-02
Add `run` build option to `cog.yaml`

### LINKED ISSUE #159: Add install commands to `cog.yaml`
This is half done with `pre_install`, but this needs some more work:

- Decide on a name. (It's "pre" what?) Now the section is `build`, perhaps this makes sense to be simply `run`.
- Document.
- Make more robust. If there is a new line, it breaks. It should probably be written out to a script, then executed, or something like that. That way we aren't concatenating strings within bash syntax within Dockerfile syntax, which is a recipe for brittleness.
- Sometimes these commands modify the source directory (like chromagan downloading CGG16 weights). Perhaps these commands should be run with the source directory mounted as a volume, somehow, when working locally.

bfirsh:
The existing `pre_install` was intentionally left undocumented
because the name was not ideal. (What is it "pre"? What's
"install"?)

"run" seems apt, because it maps directly to the Dockerfile
instruction and it's "running" these things, not "installing" them.

Known issues:
- This does not work with multiple lines.
- There should be some method of getting files in that works
  efficiently even when building base images. Maybe a run command
  could say what files it needs copied?

I will create separate issues for these when this is merged.

Closes #159

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -58,6 +58,21 @@ build:
   python_version: "3.8.1"
 ```
 
+### `run`
+
+A list of commands to run to run in the environment to set it up. They are run last, after your system packages and Python packages have been installed. If you've used Docker, it's just like a `RUN` instruction in your `Dockerfile`.
+
+Your code is _not_ available to commands in `run`. This is so we can build your image efficiently when running locally.
+
+For example:
+
+```yaml
+build:
+  run:
+    - curl -L https://github.com/cowsay-org/cowsay/archive/refs/tags/v3.7.0.tar.gz | tar -xzf -
+    - cd cowsay-3.7.0 && make install
+```
+
 ### `system_packages`
 
 A list of Ubuntu APT packages to install. For example:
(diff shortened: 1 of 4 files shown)


docs/yaml.md
```diff
@@ -58,6 +58,21 @@ build:
   python_version: "3.8.1"
 ```
 
+### `run`
+
+A list of commands to run to run in the environment to set it up. They are run last, after your system packages and Python packages have been installed. If you've used Docker, it's just like a `RUN` instruction in your `Dockerfile`.
+
+Note that your code is _not_ available to commands. This is so we can build your image efficiently when running locally.
```
andreasjansson, line 65: I was confused by the phrasing "to commands". Perhaps "to commands in the `run` clause"

andreasjansson (approved):
Nice! Yes "pre_install" doesn't make sense now that we don't have "post_install". "install" in that case meant installing the code.

docs/yaml.md
bfirsh: Clumsy "note that" too, reading again.

bfirsh:
Makes sense. In case we need to mention it anywhere, perhaps to "copy" code is better nomenclature, following the same logic in this change.

bfirsh:
Known issues:
https://github.com/replicate/cog/issues/222
https://github.com/replicate/cog/issues/223

Issue #220, opened 2021-09-02
Fix missing version errors


bfirsh:
The error messages when PyTorch and Tensorflow versions are not supported are mighty confusing:

<img width="440" alt="Screen Shot 2021-09-02 at 15 40 42" src="https://user-images.githubusercontent.com/40906/131925201-4cf80bfb-33ed-4c81-91dc-0a9193e57b77.png">

This should be something like:

```
Cog does not currently support PyTorch version 1.2.3.

It supports versions:
1.2.0
1.4.0
1.5.0
1.5.1
1.6.0
1.7.0
1.7.1
1.8.0

You might want to try upgrading Cog, in case it supports more versions: https://github.com/replicate/cog
```

And when PyTorch/CUDA versions are mismatched, it should tell you what CUDA versions are supported. Ben has a WIP branch for this.

markphelps:
This issue has been inactive for over a year, so I'm closing it as stale. The referenced WIP branch was never linked and there hasn't been any follow-up. If this is still a problem in the current version of Cog, please feel free to reopen or create a new issue.

PR #221, opened 2021-09-02
Allow installing any PyTorch CPU version

bfirsh:
If it's not in the compatibility matrix, we don't need to stop users from installing it. It's not like GPU where we need to know the correct CUDA version.

Historically, we needed this because Cog has to support CPU and GPU _simultaneously_. But we don't do that any longer, so just let users install whatever the heck packages they like.

See also #220


andreasjansson:
Great!

andreasjansson (approved):

Issue #222, opened 2021-09-03
`run` should work with multiple lines


bfirsh:
Currently `run` commands in the `build` section of can only be a single line. They should support multiple lines.

The sensible way of supporting this would probably be to write out the build commands to a shell script and then run that. E.g. write out string to `build/0`, `build/1`, etc, then output this in the `Dockerfile`:

```
COPY build /tmp/build
RUN /bin/sh /tmp/build/0
RUN /bin/sh /tmp/build/1
```

bfirsh:
Oh, we can just do this: https://www.docker.com/blog/introduction-to-heredocs-in-dockerfiles/

With a suitable unique heredoc string.

anish-sahoo:
Closing due to inactivity. Multi-line run commands can be expressed by splitting them into separate YAML list items or by invoking a repo script from a single-line run command.

PR #224, opened 2021-09-03
Better build temp directory

bfirsh:
Use a single, randomly generated tempdir when building. This
means more than one build can run concurrently without colliding,
and neatens up the cleanup process.

This was a WIP implementation of #222 that I abandoned, but this
is useful separately.

### DIFF
--- a/pkg/dockerfile/generator_test.go
+++ b/pkg/dockerfile/generator_test.go
@@ -10,9 +10,9 @@ import (
 	"github.com/replicate/cog/pkg/config"
 )
 
-func testInstallCog(generatedPaths []string) string {
-	return `COPY .cog/tmp/cog-0.0.1.dev-py3-none-any.whl /tmp/cog-0.0.1.dev-py3-none-any.whl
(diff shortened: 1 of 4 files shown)


pkg/dockerfile/generator_test.go
```diff
@@ -10,9 +10,9 @@ import (
 	"github.com/replicate/cog/pkg/config"
 )
 
-func testInstallCog(generatedPaths []string) string {
-	return `COPY .cog/tmp/cog-0.0.1.dev-py3-none-any.whl /tmp/cog-0.0.1.dev-py3-none-any.whl
-RUN --mount=type=cache,target=/root/.cache/pip pip install /tmp/cog-0.0.1.dev-py3-none-any.whl`
+func testInstallCog(relativeTmpDir string) string {
```
bfirsh, line 13: I'm not sure I like these tests that contain generated values. I almost got screwed by a false positive where `relativeTmpDir` was the empty string. We should be injecting this value into the generator.

andreasjansson (approved):
Great!

PR #225, opened 2021-09-03
More sensible pytest output

bfirsh:
1) -vv displays full diff output so you can see what failure is
2) Enabled output capturing again, so output isn't a big mess
   and failure is associated with output.

Not sure why it was this way to start with.

### DIFF
--- a/test-integration/Makefile
+++ b/test-integration/Makefile
@@ -1,3 +1,3 @@
 .PHONY: test
 test:
-	pytest -s --tb=short
+	pytest -vv
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ test-integration: install
 
 .PHONY: test-python
 test-python:
-	cd python/ && pytest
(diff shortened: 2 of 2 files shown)


test-integration/Makefile
```diff
@@ -1,3 +1,3 @@
 .PHONY: test
 test:
-	pytest -s --tb=short
+	pytest -vv
```
andreasjansson, line 3: The purpose of `--tb=short` was so pytest didn't spit out thousands of irrelevant lines of junk, and `-s` was there so I could drop to ipdb if I needed to debug a test. I might bring those back if I run into that again.

andreasjansson (approved):

test-integration/Makefile
bfirsh: Perhaps it could be a separate command, or you could just add it if you need it. No output capturing makes debugging without ipdb impossible. ;)

PR #240, opened 2021-09-17
Document not downloading weights in setup()

bfirsh:


zeke (approved):
immutability ftw 🙌🏼

PR #243, opened 2021-09-17
use sudo to install

zeke:
This PR updates the installation instructions in the Getting Started guide to use `sudo` to download and chmod the cog binary. This presumably wasn't working because the latest version of homebrew doesn't make `/usr/local/bin` writeable by my user.

🍐 @bfirsh


bfirsh (approved):
🌞

bfirsh:
This has the nice side-effect of it working by default on Linux, too.

Issue #101, opened 2021-05-31
Require users to set `python_version`


bfirsh:
This will stop models from breaking when we update our default.

zeke:
To clarify, are you suggesting that `python_version` should be a required property in `cog.yaml`?

I think you also mentioned something about wanting to rename this property to `python`, right? If so, it might be good to roll up these two changes together.

bfirsh:
Yes, that's right. Not sure if renaming is the right thing to do, but if it is, that sounds good.

Some of these issues are hastily written with a bunch of assumed knowledge -- sorry. Feel free to edit the main issue if they need clarification. Sometimes I treat the main issue a bit like a wiki, and add a note to the issue to say it was written by / updated by multiple people: https://github.com/replicate/cog/issues/29

zeke:
> Feel free to edit the main issue if they need clarification

Updated the issue title. 👍🏼

mcam10:
I think we can close this issue out?

PR #249, opened 2021-09-19
Add check_python to make targets (#101)

amarjandu:
#101 

```
gh0st-4:cog amar$ pyenv local 3.7.8
gh0st-4:cog amar$ make check_python
Python 3.7.8

Use Python 3.8

make: *** [check_python] Error 1
gh0st-4:cog amar$ pyenv local 3.8.2
gh0st-4:cog amar$ make check_python
gh0st-4:cog amar$
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 SHELL := /bin/bash
 
+include common.mk
 RELEASE_DIR := release
 GOOS := $(shell go env GOOS)
 GOARCH := $(shell go env GOARCH)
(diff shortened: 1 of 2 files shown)


zeke (commented):
I just gave this a try locally and it works for me.

However, I'm not sure that this change is resolving #101. I'll defer to @bfirsh to chime in on that.

zeke:
@amarjandu were you using `3.7` and encountering issues?

amarjandu:
It seemed like the issue was requesting that python was pinned to a specific version. I could have interpreted the issue incorrectly. @bfirsh is #101 to pin the version of python that is used to development or that is within the `yaml` files?

@zeke I have not encountered issues with 3.7, but I only installed it so I could test this PR, I use the python/go version that was specified in `.github/workflows/ci.yaml`.

zeke:
Yeah #101 is about requiring that the version is specified in `cog.yaml`. I've updated that issue for clarity. Sorry for the confusion there.

As far as I'm aware, the dev environment doesn't currently require a specific version of Python, so I'm not sure this PR is needed.

PR #260, opened 2021-09-24
update install commands in README

zeke:
Like https://github.com/replicate/cog/pull/243, but for the README this time.


bfirsh:
Ah, whoops! Nice.

PR #264, opened 2021-09-28
Add prediction timeouts to redis worker

bfirsh:
Timeouts are passed as seconds to the somewhat unwieldy list of
arguments. It is optional, so it remains backwards compatible with
models.

I have made a little helper to use Python's argument handling to parse
the argument string and set types. This should really be kwargs, or
JSON, or something.

One thought about this: timeouts using signals might happen anywhere and
might leave Predictors in unpredictable (lol) states. We might have to
see whether this is a problem in reality. If so, then we might want to
just reboot the whole worker on timeout.

@andreasjansson Can you think of any reasons off the top of your head why GPUs/Tensorflow/Torch might get left in weird states?

An additional caveat of this: if a model has a bare `except:` in it, the timeout won't work, because the signal is raised as an exception. But this won't be a problem in Replicate once we have an additional timeout on Replicate's side that kills naughty pods that aren't timing out.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -24,8 +24,11 @@ from ..predictor import Predictor, run_prediction, load_predictor
 class timeout:
     """A context manager that times out after a given number of seconds."""
 
-    def __init__(self, seconds=1, error_message="Prediction timed out"):
-        self.seconds = seconds
+    def __init__(self, seconds, elapsed=None, error_message="Prediction timed out"):
+        if elapsed is None or seconds is None:
+            self.seconds = seconds
+        else:
+            self.seconds = seconds - int(elapsed)
         self.error_message = error_message
 
     def handle_timeout(self, signum, frame):
@@ -33,8 +36,11 @@ class timeout:
 
     def __enter__(self):
         if self.seconds is not None:
-            signal.signal(signal.SIGALRM, self.handle_timeout)
-            signal.alarm(self.seconds)
+            if self.seconds <= 0:
+                self.handle_timeout(None, None)
+            else:
+                signal.signal(signal.SIGALRM, self.handle_timeout)
+                signal.alarm(self.seconds)
 
     def __exit__(self, type, value, traceback):
         if self.seconds is not None:
@@ -209,6 +215,7 @@ class RedisQueueWorker:
             self.push_error(response_queue, e)
             return
 
+        start_time = time.time()
         with self.capture_log(self.STAGE_RUN, prediction_id), timeout(
             seconds=self.predict_timeout
         ):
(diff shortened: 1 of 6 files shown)


python/cog/server/redis_queue.py
```diff
@@ -186,15 +209,19 @@ def handle_message(self, response_queue, message, cleanup_functions):
...
+        ):
             return_value = self.predictor.predict(**inputs)
         if isinstance(return_value, types.GeneratorType):
             last_result = None
 
             while True:
                 # we consume iterator manually to capture log
                 try:
-                    with self.capture_log(self.STAGE_RUN, prediction_id):
+                    with self.capture_log(self.STAGE_RUN, prediction_id), timeout(
```
andreasjansson, line 222: Doesn't this mean that timeouts are applied per-yield? So the whole prediction can take `n_yields * timeout` seconds?

andreasjansson (commented):
Unless I'm misreading it looks like yield timeouts are per-yield as opposed to per-prediction?

test-integration/test_integration/test_redis_queue.py
```diff
@@ -342,6 +342,89 @@ def test_queue_worker_timeout(docker_image, redis_port, request):
...
+                    }
+                ),
+            },
+        )
+
+        response = json.loads(redis_client.brpop(response_queue_name, timeout=10)[1])
+        assert response == {"value": "yield 0", "status": "processing"}
+
+        response = json.loads(redis_client.brpop(response_queue_name, timeout=10)[1])
+        assert response == {"status": "failed", "error": "Prediction timed out"}
```
bfirsh, line 425: This seems flakey -- it's depending on fraction of a second timing, if I'm reading this right. If the execution of `predict()` is interrupted for ~0.1s, then it'll fail.

Perhaps a more robust way to do this would be to sleep for 0.1, and keep on yielding and ensure at some point within 20 or so iterations it times out.

python/cog/server/redis_queue.py
bfirsh: Correct, thanks for the fix!

test-integration/test_integration/test_redis_queue.py
andreasjansson: I was thinking 300ms is a lot of time, but if it turns out to be flaky we can always revisit. I can put a TODO in there

bfirsh: I think the reason this intuitively feels flakey is this depends on various I/O operations (e.g. pushing result to queue, posting logs, etc), which could block on the order of 100ms. If it were just a single thread doing CPU stuff, then 300ms does feel like a long time.

Happy to leave this as a todo though.

PR #269, opened 2021-09-28
Relax torchvision version check

bfirsh:
Let the user manually specify things like `torchvision==0.10.0+cu111`.
Or, shoot themselves in the foot. But we told them, right?

Related to #181

### DIFF
--- a/pkg/config/compatibility.go
+++ b/pkg/config/compatibility.go
@@ -327,7 +327,9 @@ func torchvisionGPUPackage(ver string, cuda string) (name string, cpuVersion str
 		}
 	}
 	if latest == nil {
-		return "", "", "", fmt.Errorf("No torchvision GPU package for version %s that's lower or equal to CUDA %s", ver, cuda)
+		// TODO: can we suggest a CUDA version known to be compatible?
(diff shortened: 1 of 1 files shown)


pkg/config/compatibility.go
```diff
@@ -327,7 +327,9 @@ func torchvisionGPUPackage(ver string, cuda string) (name string, cpuVersion str
 		}
 	}
 	if latest == nil {
-		return "", "", "", fmt.Errorf("No torchvision GPU package for version %s that's lower or equal to CUDA %s", ver, cuda)
+		// TODO: can we suggest a CUDA version known to be compatible?
+		console.Warnf("Cog doesn't know if CUDA %s is compatible with torchvision %s. This might cause CUDA problems.", cuda, ver)
```
zeke, line 331: style question: could this alternatively be written as `f"Cog doesn't know if CUDA {cuda} is compatible with torchvision {ver}. This might cause CUDA problems."` ?

bfirsh: That's Python, unfortunately. Golang doesn't have anything nice like that as far as I know.

andreasjansson (approved):

pkg/config/compatibility.go
zeke: Oh ha! Didn't even notice. 🙈

zeke (approved):

PR #271, opened 2021-09-29
use curl instead of wget to download weights

zeke:
`wget` is not installed by default on macOS. It's easy to `brew install wget`, but then we're assuming the user has homebrew, and would have to add some macOS-specific clause in the docs.

This PR updates the getting started to use `curl` instead of `wget`, as most users will have this command in their environment.

```
 -O, --remote-name   Write output to a file named as the remote file
```


bfirsh:
🥌 🧹

PR #277, opened 2021-10-01
Use r8.im in docs

bfirsh:
Makes this make marginally more sense as a way to try out Replicate. I think JSON output is still broken though. :(


bfirsh:
Opening PR for visibility, merging to get this fixed quick.

PR #278, opened 2021-10-02
update cog.yaml reference doc

zeke:
This PR makes a few minor updates to the `cog.yaml` reference documentation.

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -17,7 +17,7 @@ predict: "predict.py:JazzSoloComposerPredictor"
 
 ## `build`
 
-This how to build the Docker image your model runs in. It contains various options within it:
+This stanza describes how to build the Docker image your model runs in. It contains various options within it:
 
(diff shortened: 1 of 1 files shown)


docs/yaml.md
```diff
@@ -88,14 +88,14 @@ build:
...
 
-If you don't provide this, a name will be generated from the directory name.
-
 For example:
 
 ```yaml
 image: "registry.hooli.corp/jazz-solo-model"
 ```
 
+If you don't provide this, a name will be generated from the directory name.
```
zeke, line 97: I moved the examples code blocks up above these caveat/gotcha blurbs so the examples won't be misunderstood as being related to those caveats. Does that make sense?

```
NAME
DESCRIPTION
EXAMPLE
CAVEATS
```

bfirsh: Great idea.

bfirsh:
Wonderful!

Issue #157, opened 2021-07-22
Design how Python requirements are specified


bfirsh:
We should decide between `python_requirements` and `python_packages` and make it work properly.

There is some background here that is not written down anywhere that needs writing down. tl;dr `python_requirements` doesn't work properly, `python_packages` is not ideal in various ways.

zeke:
> The main downside of this is that it is not clear to the user that Cog is rewriting versions.

Is Cog NOT rewriting versions when you use `python_packages`?

bfirsh:
It is. And come to think of it this is a downside of both (1) and (2).

Torch/Tensorflow packages need to be compiled for a particular version of CUDA, and in Torch's case, those packages are not on PyPi but on their custom index.

For example, if a user specifies `torch==1.10.0` in `python_packages`, then Cog detects the best CUDA version to use and rewrites the install to be something like `torch==1.10.0+cu113 -f https://download.pytorch.org/whl/cu113/torch_stable.html`.

zeke:
> users normally already have some way of defining their dependencies (pip, Poetry, Conda, etc) and are unwilling to switch to a different method. [...] read dependencies from wherever the user has already defined them.

☝🏼 I think this makes sense.

- Keep `python_requirements`
- Remove `python_packages`
- Use docs and logging to make it clear that torch and tensorflow versions are being modified.

zeke:
Speaking of the myriad ways to define dependencies in Pythonland, I recently took the [2021 Python Developers survey](https://surveys.jetbrains.com/s3/c1-python-developers-survey-2021) and this was one of the questions:

<img width="612" alt="Screen Shot 2021-11-08 at 6 31 38 PM" src="https://user-images.githubusercontent.com/2289/142293711-2fc2e075-b0aa-436a-8caa-f316a24436cf.png">


I checked out the [results from the 2020 survey](https://www.jetbrains.com/lp/python-developers-survey-2020/) thinking it might help us know which formats to prioritize, but it looks like they didn't ask that question last year.

bfirsh:
A bit of interesting data from #325: we have some undocumented options `python_extra_index_urls` and `python_find_links`, but looks like nobody is using these for any public models:

https://github.com/search?q=python_find_links+filename%3Acog.yaml&type=code
https://github.com/search?q=python_extra_index_urls+filename%3Acog.yaml&type=code

zeke:
Here are some rough stats from GitHub on the popularity of different dependency formats:

Filename | Count
[...]

PR #279, opened 2021-10-02
cog init

zeke:
This PR introduces a new CLI command `cog init` that can be used to bootstrap an existing project with starter `cog.yaml` and `predict.py` files.

```
$ cog init
Created /Users/z/git/replicate/cog/cog.yaml
Created /Users/z/git/replicate/cog/predict.py
```

### To Do

- [x] Make sure I'm Doing It Right™, Golang-wise
- [x] Do not overwrite files if they already exist
- [x] Refine the `cog.yaml` and `predict.py` templates to contain the right starter content
- [x] Write docs
- [x] Write tests

### Stretch goals

- [ ] Derive `python_packages` somehow ([pipreqs](https://pypi.org/project/pipreqs/)? look for `requirements.txt`?)
- [ ] Add example `script/download-weights`?


zeke:
This is ready for a first look! 👀

bfirsh:
This looks fantastic!

> Derive python_packages somehow

We probably shouldn't put effort into this until we've decided on #157.

zeke:
This now has tests and docs! It's ready for another look.

pkg/cli/init-templates/cog.yaml
```diff
@@ -0,0 +1,28 @@
+# Configuration for Cog ⚙️
+# Reference: https://github.com/replicate/cog/blob/main/docs/yaml.md
+
+build:
+  # set to true if your model requires a GPU
+  gpu: false
+
+  # a list of ubuntu apt packages to install
+  system_packages:
+    - "libgl1-mesa-glx"
```
andreasjansson, line 10: Might want some indication that these are just example packages. Maybe they should be commented out by default?

pkg/cli/init-templates/cog.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - "libglib2.0-0"
+
+  # python version in the form '3.8' or '3.8.12'
```
andreasjansson, line 13: Do patch versions work now? They didn't use to but perhaps @bfirsh added that.

pkg/cli/init-templates/cog.yaml
```diff
@@ -0,0 +1,28 @@
...
+  python_version: "3.8"
+
+  # a list of packages in the format <package-name>==<version>
+  python_packages:
+    - "numpy==1.19.4"
```
andreasjansson, line 18: Same here, perhaps comment them out?

pkg/cli/init-templates/predict.py
```diff
@@ -0,0 +1,18 @@
...
+	
+import cog
+# import torch
+
+class Predictor(cog.Predictor):
+    def setup(self):
+      """Load the model into memory to make running multiple predictions efficient"""
+      # self.model = torch.load("./weights.pth")
+
+    @cog.input("image", type=cog.Path, help="Grayscale input image")
```
andreasjansson, line 12: Maybe we could have a comment pointing to the documentation for the predict() function here, to make it super easy to read about input types, etc.?

andreasjansson (approved):
Fantastic! This lowers the barrier to start using Cog by a lot!

pkg/cli/init-templates/cog.yaml
bfirsh: I think they always have worked? It just passes down to pyenv which supports either. The thing I added was just using Docker's `python` image, which also supports either.

pkg/cli/init-templates/predict.py
zeke: There's a comment at the top of the file:

```yaml
# Prediction interface for Cog ⚙️
# Reference: https://github.com/replicate/cog/blob/main/docs/python.md
```

pkg/cli/init-templates/cog.yaml
zeke: Yep we waffled on this. You tipped the scales. ⚖️

zeke: 51f59536c135b97daa87676631ebe4b55e911f8e

pkg/cli/init-templates/cog.yaml
zeke: Yep. Done. 51f59536c135b97daa87676631ebe4b55e911f8e

pkg/cli/init-templates/cog.yaml
zeke: I just did a test `cog build` with `python_version: "3.8.11"` and it appears to work:

```
 => [internal] load metadata for docker.io/library/python:3.8.11                                                                        1.1s
 => [stage-0 1/8] FROM docker.io/library/python:3.8.11@sha256:a0a734233420b17d9ab37125afc9d8217b75db153d55854ac6683e639f00a8e8         13.4s
```

zeke:
@bfirsh I made a few minor updates based on @andreasjansson's feedback and updated the output a bit. Here's what it looks like now:

```

✅ Created /Users/z/git/replicate/cog/cog.yaml
✅ Created /Users/z/git/replicate/cog/predict.py

Done! For next steps, check out the docs at https://github.com/replicate/cog/tree/main/docs
```

This is ready for another look before we ship it.

bfirsh:
Looks like there are merge conflicts.

I wonder if this would be a better URL for docs? Actually describes what each thing is. https://github.com/replicate/cog#next-steps

Could even change the heading to "Documentation" so the anchor becomes `#documentation`.

bfirsh (approved):
Minor thing anyway. Looks good otherwise!

zeke:
> I wonder if this would be a better URL for docs?

We discussed this in meatspace. Gonna ship with the docs URL as-is.

Issue #280, opened 2021-10-02
No CUDA runtime is found / Found no NVIDIA driver on your system.


bwhitman:
When running a GPU cog: 

```
$ sudo cog predict r8.im/allenhung1025/looptest@sha256:f5cd715e99046e0513fe2b4034e8f7d8c102525b02f49efb52b05f46fcb9ea83
Starting Docker image r8.im/allenhung1025/looptest@sha256:f5cd715e99046e0513fe2b4034e8f7d8c102525b02f49efb52b05f46fcb9ea83 and running setup()...
No CUDA runtime is found, using CUDA_HOME='/usr/local/cuda'
Traceback (most recent call last):
...
AssertionError: 
Found no NVIDIA driver on your system. Please check that you
have an NVIDIA GPU and installed a driver from
http://www.nvidia.com/Download/index.aspx
ⅹ Failed to get container status: exit status 1
```

But CUDA is fine:
```
$ cd /usr/local/cuda/samples/1_Utilities/deviceQuery
$ sudo make
$ ./deviceQuery
Detected 1 CUDA Capable device(s)

Device 0: "NVIDIA GeForce RTX 2080 Ti"
  CUDA Driver Version / Runtime Version          11.4 / 11.2
  CUDA Capability Major/Minor version number:    7.5
  Total amount of global memory:                 11016 MBytes (11551440896 bytes)
  (68) Multiprocessors, ( 64) CUDA Cores/MP:     4352 CUDA Cores
...
Result = PASS
```

andreasjansson:
Hi @bwhitman! 👋 

Does it work if you run the model with Docker directly?

```
$ docker run -d -p 5000:5000 --gpus all r8.im/allenhung1025/looptest@sha256:f5cd715e99046e0513fe2b4034e8f7d8c102525b02f49efb52b05f46fcb9ea83

$ curl http://localhost:5000/predict -X POST -F seed=-1
```

bwhitman:
Hm, no: 

```$ docker run -d -p 5000:5000 --gpus all r8.im/allenhung1025/looptest@sha256:f5cd715e99046e0513fe2b4034e8f7d8c102525b02f49efb52b05f46fcb9ea83
f189dd7fdffa728fc77f7c8976a3b9e4e4619cb670ed411774689439bd9e302c
docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].```

I'll dig in, doesn't seem like a cog issue

bwhitman:
[...]

PR #285, opened 2021-10-05
Make remote models use GPUs in cog predict, if model requires GPU

andreasjansson:
Fixes https://github.com/replicate/cog/issues/280

### DIFF
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -77,9 +77,15 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 		// Use existing image
 		imageName = args[0]
 
-		console.Infof("Pulling image: %s", imageName)
-		if err := docker.Pull(imageName); err != nil {
-			return fmt.Errorf("Failed to pull %s: %w", imageName, err)
+		exists, err := docker.ImageExists(imageName)
+		if err != nil {
+			return fmt.Errorf("Failed to determine if %s exists: %w", imageName, err)
+		}
+		if !exists {
+			console.Infof("Pulling image: %s", imageName)
+			if err := docker.Pull(imageName); err != nil {
+				return fmt.Errorf("Failed to pull %s: %w", imageName, err)
+			}
 		}
 		conf, err := image.GetConfig(imageName)
 		if err != nil {
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"fmt"
 	"io"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/image/config.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func GetConfig(imageName string) (*config.Config, error) {
+	image, err := docker.ImageInspect(imageName)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to inspect %s: %w", imageName, err)
+	}
+	configString := image.Config.Labels["org.cogmodel.config"]
+	if configString == "" {
+		return nil, fmt.Errorf("Image %s does not appear to be a Cog model", imageName)
```
bfirsh, line 18: I wonder if it should also check Cog version here for future compatibility?

"You are running Cog version X but this image was produced by Cog version Y. You need to upgrade Cog to run this image: ..."

It might be annoying for it to happen for every patch version while we're in `0.0.X`, so maybe it triggers at `0.1.0` and we can figure out a better versioning scheme then.

pkg/docker/image_inspect.go
```diff
@@ -0,0 +1,29 @@
...
+	"os"
+	"os/exec"
+
+	"github.com/docker/docker/api/types"
+)
+
+func ImageInspect(id string) (*types.ImageInspect, error) {
+	cmd := exec.Command("docker", "image", "inspect", id)
+	cmd.Env = os.Environ()
+
```
bfirsh, line 15:
```suggestion
console.Debug("$ " + strings.Join(cmd.Args, " "))
```

(github doesn't let me add the tab)

bfirsh (approved):
Lovely! Do you think this needs any tests? Maybe for some subset?

pkg/image/config.go
andreasjansson: I agree, but perhaps that should be a different PR. I'll open an issue for now so we don't forget.

pkg/docker/image_inspect.go
andreasjansson: Pushed with tab :)

andreasjansson:
> Lovely! Do you think this needs any tests? Maybe for some subset?

The non-GPU case is covered by `test_predict.py:test_predict_with_existing_image`. Once we can build GPU images on CPU we can add a GPU test as well.

andreasjansson:
The test failure is legit, fixing.

pkg/cli/predict.go
```diff
@@ -76,7 +77,17 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 		// Use existing image
 		imageName = args[0]
 
-		// TODO: check metadata for GPUs
+		console.Infof("Pulling image: %s", imageName)
+		if err := docker.Pull(imageName); err != nil {
+			return fmt.Errorf("Failed to pull %s: %w", imageName, err)
```
zeke, line 82: Hadn't seen `%w` before. TIL: https://go.dev/blog/go1.13-errors

Screen Shot 2021-10-05 at 9 39 50 AM

PR #298, opened 2021-10-11
Fix cog predict on remote image

andreasjansson:
The docker inspect step throws an error if the image doesn't exist,
which causes Cog to fail.


bfirsh:
Looks good to me anyway, if we add that check back in, maybe with a comment to explain why the heck there are two checks... 🙄

PR #317, opened 2021-11-01
document r8.im registry in YAML reference

zeke:
This PR updates the cog.yaml reference page to mention the `r8.im` registry.


bfirsh:
Nice! I wonder whether this could be the only one mentioned for simplicity...

zeke:
You mean instead of the existing hoolio thing?

bfirsh:
Yep. "r8.im is Replicate's registry, but this can be any Docker registry." or somesuch. No big deal though.

PR #318, opened 2021-11-01
document prerequisites

zeke:
This PR updates the two getting started guides with a list of prerequisites, so folks know exactly what's required before diving it. It also moves a bit of the existing content around into a more logical order

### DIFF
--- a/docs/getting-started-own-model.md
+++ b/docs/getting-started-own-model.md
@@ -2,7 +2,20 @@
 
 This guide will show you how to put your own machine learning model in a Docker image using Cog. If you haven't got a model to try out, you'll want to follow the [main getting started guide](getting-started.md).
 
+## Prerequisites
+
+- **A trained model**. If you don't already have your own trained model, you can use one from [https://github.com/replicate/cog-examples](replicate/cog-examples).
+- **macOS or Linux**. Cog works on macOS and Linux, but does not currently support Windows.
+- **Docker**. Cog uses Docker to create a container for your model. You'll need to [https://docs.docker.com/get-docker/](install Docker) before you can run Cog.
+
 ## Initialization
+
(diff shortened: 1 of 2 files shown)


docs/getting-started-own-model.md
```diff
@@ -2,8 +2,21 @@
 
 This guide will show you how to put your own machine learning model in a Docker image using Cog. If you haven't got a model to try out, you'll want to follow the [main getting started guide](getting-started.md).
 
+## Prerequisites
+
+- **A trained model**. If you don't already have your own trained model, you can use one from [https://github.com/replicate/cog-examples](replicate/cog-examples).
```
bfirsh, line 7: Might be worth a nod to "Get started with an example model" in `getting-started.md` here?

docs/getting-started.md
```diff
@@ -2,11 +2,21 @@
 
 This guide will walk you through what you can do with Cog by using an example model.
 
+## Prerequisites
+
+- **A trained model**. If you don't already have your own trained model, you can use one from [https://github.com/replicate/cog-examples](replicate/cog-examples).
```
bfirsh, line 7: Not for this guide -- it explains where to get a trained model.

bfirsh (changes requested):
Super! One little change.

docs/getting-started-own-model.md
zeke: Gotcha. Removed that line from the prereqs because it's a bit redundant given the context.

docs/getting-started.md
zeke: Gotcha. Removed.

zeke:
Thanks for the feedback @bfirsh. This is ready for another look.

bfirsh (approved):
nice!

PR #319, opened 2021-11-01
use r8.im as the only `image` example

zeke:
A tiny followup to https://github.com/replicate/cog/pull/317#issuecomment-956577635


bfirsh:
Nice!

Issue #34, opened 2021-04-22
Throw error if invalid keys exist in cog.yaml


andreasjansson:
Currently invalid keys are silently ignored.

For example, a `cog.yaml` that contains this:

```
buildd:
  python_version: "3.8"
  python_packages:
    - "torch=1.8.0"
```

Will silently not install `torch`, instead of complaining that the key `buildd` doesn't exist.

## Future

We may also want to validate the values at some point. For example, ensuring `python_version` is a string and matches a given pattern. (If you omit the quotes, `3.1` and `3.10` are indistinguishable! Thanks YAML!)

(Edited by @andreasjansson and @bfirsh.)

bfirsh:
Closed by #325.

PR #321, opened 2021-11-03
Add cog.yaml validator to package

imshashank:
## Description:

This PR adds code to validate the yaml package against the "json schema".

### Binding Data
To bind the data in the package I used `https://github.com/jteeuwen/go-bindata`.
The JSON schema is in config/data/
binddata.go is auto-generated using the command: `go-bindata data/`

### Validator
In Validator, the yaml is converted into JSON and then validated against the JSON schema. 

### What is validated?
At this time the json schema validates for only two fields:
`build` & `build.python_version`

New fields can be added by simply updating the schema.

P.S. Sent a different PR for adding the "validate" command.


pkg/config/bindata.go
```diff
@@ -0,0 +1,105 @@
...
...68\x1d\x69\xce\xfb\x39\x3c\xac\xc1\xde\x76\x0e\x8d\xd5\x54\x83\x8d\x67\x4d\x90\x4a\xc9\x50\x9a\x4e\xa3\x17\xaa\xab\x25\xd9\xe0\x70\x0c\x81\x71\x31\x3e\xd9\xe0\xdb\x0a\x00\x0e\xe1\x04\x92\x30\x40\x2f\x1c\x03\x40\xd2\xee\xa5\xb2\xe6\xfb\x8e\x9c\x1f\xbf\xfd\x26\xbd\x4d\xc6\xfc\x71\x35\x9d\x0f\xc1\xd9\xd1\xcf\x8e\x1d\xe9\x93\xf3\xe8\x78\x12\xb4\xce\xb6\xe4\x84\x43\xf3\xc3\x6a\xb1\xe9\x84\xf4\x45\xf6\xac\xce\x06\xd1\x7a\x92\x2c\x90\x1c\xe2\x13\xcd\xfb\x8a\xc6\x75\x46\x14\x83\x64\x91\xa4\x28\xf3\x5b\x61\x48\x6d\xcf\x81\x48\x45\x73\x24\xd1\xd2\x5d\x67\x3c\xd8\xa4\xf8\x24\xfd\xdd\x11\xc5\xc6\x63\xa7\x1c\xdb\xce\xc3\xb6\x03\xa7\x27\x96\x8a\x0d\x58\x36\xf1\x18\x73\x2a\x21\x78\x41\x26\xa6\xf3\x7f\x1c\x13\x90\x71\xdb\xa1\xe0\x02\xc9\x92\xcd\x45\xd5\x02\xa7\xe5\xf6\xb3\xe1\xfe\x0a\x2e\x0e\x9c\x59\xac\x63\x83\x09\xab\x17\xc7\xa6\x3c\xcb\xc5\x68\xe7\x26\xe7\x8c\x97\x39\x13\x1a\x36\xd6\xe1\xf5\x4d\x7a\x7b\x05\xeb\xd0\x2a\xc9\xab\xf0\x37\x7d\x7b\x85\xc9\xcb\x16\xb8\x0b\xee\xfd\x05\xe8\xfc\xf8\x8a\x22\xdb\x89\x5b\x32\x4f\x2c\xb2\x0b\x99\x19\x21\xe0\xe1\x82\xd6\xf3\x05\x50\x5a\x73\x3f\xb1\xaa\xef\x62\x02\xe2\x3a\x3a\x3d\xb5\xe1\x11\xff\x4b\x79\xfc\x13\x00\x00\xff\xff\x50\x29\xe5\xed\x51\x05\x00\x00")
```
andreasjansson: You might want to look into the embed package (https://pkg.go.dev/embed). There's examples of using that in pkg/cli/init.go, for example.

pkg/config/data/config_schema_v1.0.json
```diff
@@ -0,0 +1,49 @@
...
+  "default": {},
+  "examples": [
+    {
+      "build": {
+        "python_version": "3.8"
+      }
+    }
+  ],
+  "required": [
+    "build"
```
andreasjansson, line 16: I'm not sure build should be required. There are defaults  `python_version`, `cuda`, etc.

pkg/config/data/config_schema_v1.0.json
```diff
@@ -0,0 +1,49 @@
...
+          "type": "string",
+          "title": "The python_version schema",
+          "description": "The minor (3.8) or patch (3.8.1) version of Python to use.",
+          "default": "",
+          "examples": [
+            "3.8"
+          ]
+        }
+      },
+      "additionalProperties": true
```
andreasjansson, line 45: Is there a reason to leave `additionalProperties` as `true`? It'd be nice to protect users who accidentally mistype or misindent fields, and error loudly if they do.

pkg/config/validator.go
```diff
@@ -0,0 +1,151 @@
...
+	return err
+}
+
+const (
+	jsonschemaOneOf = "number_one_of"
+	jsonschemaAnyOf = "number_any_of"
+)
+
+func getDescription(err validationError) string {
+	switch err.parent.Type() {
```
andreasjansson, line 73: Some documentation would be nice here. Does this collapse a list of errors into a single error?

imshashank: While validating the schema there can be a lot of errors and it was confusing on what to return. One approach was to return a list of errors.

 I was trying to find a standard way and found this in docker project: 
https://github.com/farfeduc/docker-ce/blob/74c07ebc84244a5df3c25786e477500b446e9704/components/cli/cli/compose/schema/schema.go#L79

This tries to find the most specific error based on the number of times it was repeated.
Also, it tries to find the parent error.

pkg/config/data/config_schema_v1.0.json
imshashank: So build is made "required" to avoid this issue: https://github.com/replicate/cog/issues/34
If someone misspells "build" keyword then cog assumed no config is present and was using default configs. 

Making it required avoids this issue.

PR #323, opened 2021-11-03
Add golangci config to enable golint and other useful checks

andreasjansson:
Also fix a few issues picked up by golint. We should add more of these checks over time.


andreasjansson:
I'm merging this to unblock @imshashank

.golangci.yaml
```diff
@@ -0,0 +1,187 @@
...
+  skip-files:
+
+linters:
+  disable-all: true
+  enable:
+  - deadcode
+  - errcheck
+  - exportloopref
+  - gocritic
+  # - gofumpt
```
zeke, line 27: > gofumpt

Love it.

PR #325, opened 2021-11-06
Throw error if invalid keys exist in cog.yaml

imshashank:
Fix for: 
## Throw error if invalid keys exist in cog.yaml #34

# Approaches:

## 1) Use hard-coded key pair to test the yaml file
Create a list of key pairs or a struct to validate the cog.yaml file

Pros:
- Simple & straight forward approach
- Easy to test and validate it works

Cons:
- Hard to implement complex logic
- Needs more code changes
- Need to learn `go` and have knowledge of internal systems to make changes
- Hard to keep track & validate the different versions of cog.yaml file

## 2) Use a JSON Schema to validate

This approach uses a JSON Schema to validate the `cog.yaml` file.

Pros:
- Uses Industry standard approach to validate the yaml file
- Most cases will need a simple change to json schmema
- No need to learn `go` to understand various fields in cog.yaml
- The json-schema serves as the documentation.

Cons:
- Dependencies on external libraries and validation mechanism
- Would be hard to understand for people not familiar with jsonSchemas
- Required editing a large json file to make changes

# Approach Selected: "Using JSONSchema"

Decided to implement the `jsonschema` approach based on the pros & cons and especially keeping the end-user in mind and the "open-source" nature of the project.

The approach will not enforce new rules but only validates that the keys present in the yaml are defined in the schema.

### How to generate the initial JSON schema
- Create a yaml file containing all the fields present in the cog.yaml
- Converted the yaml to JSON using this tool: https://onlineyamltools.com/convert-yaml-to-json
- Then generate the jsonschema using this tool: https://www.jsonschema.net/home

The schema required few changes:
- Remove the `required` field
- Allow both number and string for python_version field

## Validate command
The PR also includes a `validate` command to verify any cog.yaml in the current directory.

### Usage:

```
cog validate
```

### Empty cog.yaml

No error will be thrown and `cog` will use the default schema generated by the below code:

```
func DefaultConfig() *Config {
	return &Config{
		Build: &Build{
			GPU:           false,
			PythonVersion: "3.8",
		},
	}
}

```
These are all valid `cog.yaml` files:
- No File
- Empty file
- `build:`

### Bad yaml file example & errors

- 
cog.yaml:
``` 
buildd:
```

Output:
`ⅹ (root) Additional property builds is not allowed`

-  
cog.yaml:
```
buildd:
  python_version: "3.8"
  python_packages:
    - "torch=1.8.0"
```

Output:
`ⅹ (root) Additional property buildd is not allowed`

-  
cog.yaml:
```
build:
  python_versions: "3.8"
  python_packages:
    - "torch=1.8.0"
```
Output:

`ⅹ build Additional property python_versions is not allowed`
..........................................................................................

Signed-off-by: Shashank Agarwal 2 <shashankmailster@gmail.com>


pkg/cli/validate.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		return err
+	}
+	err = config.ValidateAndCompleteConfig()
+	if err != nil {
+		return err
+	}
+	console.Output("Valid cog.yaml file")
+	return nil
+}
```
bfirsh, line 38: Adding a top-level command is something that'll need a bit of thought. For the sake of keeping this PR simple, let's not include `cog validate`.

pkg/config/config.go
```diff
@@ -56,7 +56,12 @@ func FromYAML(contents []byte) (*Config, error) {
 		return nil, fmt.Errorf("Failed to parse config yaml: %w", err)
 	}
 	// Everything assumes Build is not nil
-	if config.Build == nil {
+	if string(contents) != "" && config.Build != nil {
+		err := Validate(string(contents), "")
+		if err != nil {
+			return nil, err
+		}
+	} else {
```
bfirsh, line 64: Should this be inside this conditional? Wouldn't it make sense to always validate the config, even when `build` is not set?

pkg/config/data/config_schema_v1.0.json
```diff
@@ -0,0 +1,252 @@
...
+        ],
+        "run": [
+          "curl -L https://github.com/cowsay-org/cowsay/archive/refs/tags/v3.7.0.tar.gz | tar -xzf -",
+          "cd cowsay-3.7.0 && make install"
+        ]
+      },
+      "image": "registry.hooli.corp/jazz-solo-model",
+      "predict": "predict.py:JazzSoloComposerPredictor"
+    }
+  ],
```
bfirsh, line 33: It'd be great to remove all the superfluous stuff here. Anything that's not actually needed to validate the config.

pkg/config/data/config_schema_v1.0.json
```diff
@@ -0,0 +1,252 @@
...
+          "default": "",
+          "examples": [
+            "10.1"
+          ]
+        },
+        "gpu": {
+          "$id": "#/properties/build/properties/gpu",
+          "type": "boolean",
+          "title": "The gpu schema",
+          "description": "An explanation about the purpose of this instance.",
```
bfirsh, line 82: This too.

pkg/config/validator.go
```diff
@@ -0,0 +1,158 @@
...
+	case jsonschemaOneOf, jsonschemaAnyOf:
+		if err.child == nil {
+			return err.parent.Description()
+		}
+		return err.child.Description()
+	}
+	return err.parent.Description()
+}
+
+func humanReadableType(definition string) string {
```
bfirsh, line 94: This stuff is from Docker, isn't it? This should probably be acknowledged at least. It might be a license violation. Maybe there is a way of importing it from Docker instead of copy & pasting.

python/setup.py
```diff
@@ -1,6 +1,6 @@
 import setuptools
 
-with open("../README.md", "r", encoding="utf-8") as fh:
+with open("../README.md", "r") as fh:
```
bfirsh, line 3: Why was this needed? I think this means it won't work with utf-8 characters.

bfirsh (changes requested):
Thanks for the more detailed PR, @imshashank!

I've added some comments inline.

Screen Shot 2021-11-08 at 15 28 05

It'd also be good to get a more descriptive error. E.g. this should probably read something like:

```
❌ There is a problem in your cog.yaml file. The option "booild" is 
not recognised.

To see what options you can use, take a look at the docs: 
https://github.com/replicate/cog/blob/main/docs/yaml.md

You might also need to upgrade Cog, if this option was added in a 
later version of Cog.
```

We're happy to take over this pull request if you'd like. Let us know.

pkg/cli/validate.go
imshashank: Sure, will remove the `validate` command code. This command makes it quite easy to test the functionality and is backward compatible.

pkg/config/config.go
imshashank: This was added to make sure the `validate` functionality is backward compatible.  Right now `cog.yaml` is not required and also build is not a required keyword. I think this is actually nice and let's user avoid writing a cog.yaml file if they don't want to. This can definitely be changes in future and restrictions can be added.

Let me know what do you think.

pkg/config/data/config_schema_v1.0.json
imshashank: These are just example strings that were generated automatically in the json schema. This is actually useful for users if they want to see examples of how certain variables need to be declared.
This was directly taken from the cog.md documentation.

pkg/config/data/config_schema_v1.0.json
imshashank: Will remove all these strings. Good catch.

pkg/config/validator.go
imshashank: Yes some of this code is from this file: https://github.com/docker/docker-ce/blob/f76280404059080d79fcda620caf8cef5a4a22f7/components/cli/cli/compose/schema/schema.go
The code is available through Apache v2 license: https://github.com/docker/docker-ce/blob/master/LICENSE

I couldn't find a way to import this, if you have any ideas please let me know.

If there is a concern, I can rewrite this code from scratch. Please let me know

python/setup.py
imshashank: I can undo this, it was just causing problems for me while compiling on my OSX laptop.

imshashank:
Please let me know if this is okay:

```
ⅹ There is a problem in your cog.yaml file.
ⅹ (root) Additional property booild is not allowed.
ⅹ
ⅹ To see what options you can use, take a look at the docs:
ⅹ https://github.com/replicate/cog/blob/main/docs/yaml.md
ⅹ
ⅹ You might also need to upgrade Cog, if this option was added in a
ⅹ later version of Cog.
exit status 1
```

This string `"Additional property ${property} is not allowed."` is a standard error, would recommend against changing it. 
Waiting for reply on comments and will then send an update.

bfirsh:
"Additional property not allowed" makes a lot less sense for a user, though. And "(root)" is out of context and looks scary.

Is it possible to customize these error messages, I wonder? This might be an argument for us to do something simpler.

I remember this from implementing this in the Python version of Compose. The initial version was simple, but we had to put in a lot of legwork to make the error messages less robotic and understandable.

pkg/config/config.go
bfirsh: Gotchya. Sounds sensible to keep current behavior.

pkg/config/validator.go
bfirsh: If you're just using public methods (starting with a capital letter) you should in theory just be able to import the package.

If not, a nod to where the code comes from at the top of the file with a link to the license seems sensible.

bfirsh:
Either way though this is not a big deal -- we can fix the error messages later. Some error is a lot better than silently failing. :)

imshashank:
Removed `(root)`, it was not very helpful. I think there were ways to provide a template to the error messages but that might require some configuration within the jsonschema package. So just removed the `root` etc for now. The user will get the simple error now.

Also removed the `validate` command. Also removed everything not required in the JSON schema itself. Please have a look at the updated PR.

bfirsh:
Fantastic, thanks @imshashank.

This is actually missing the options `python_extra_index_urls`, `python_find_links`, and `cudnn`. See here: https://github.com/replicate/cog/blob/affa6d65ce5938b673b42ad3f98a5fee79cd23c1/pkg/config/config.go#L19-L32

I did a search on GitHub though and it looks like nobody is using these for any public models:

https://github.com/search?q=python_find_links+filename%3Acog.yaml&type=code
https://github.com/search?q=python_extra_index_urls+filename%3Acog.yaml&type=code
https://github.com/search?q=cudnn+filename%3Acog.yaml&type=code

The first two depend on #157. Given that nobody is probably using these undocumented features, perhaps we just leave them out for now and either come with a design or document them in as a piece of follow-up work.

Thoughts @andreasjansson @zeke?

zeke:
> Given that nobody is probably using these undocumented features, perhaps we just leave them out for now

👍🏼

imshashank:
@bfirsh I can create an issue and mark it as backlog. Is there a place where you track backlog items?
Or I can add them to the JSON schema but I think it's better to leave them out of this PR as there are no tests, docs, etc. for these values. Please let me know

zeke:
Adding an issue sounds great. Maybe with a link to this PR for context. 👍🏼

bfirsh (approved):
This sounds good to me. I think the Python options are covered by #157, but we could probably do with an issue to remind us about `cudnn`.

Thanks so much for getting this to the finish line, @imshashank. :)

Issue #204, opened 2021-08-13
Ensure we support latest versions of tensorflow and torch


bfirsh:
See also #183 #202 #390

(edited by @zeke)

bfirsh:
Fixed by #329.

bfirsh:
Actually I'm going to leave this open, to remind us to run this before our initial release.

PR #228, opened 2021-09-03
Fix PyTorch version scraper & update PyTorch versions

bfirsh:
Did not update Tensorflow because it is broken for some reason.

See also #183 #204

Fixes #202

PR #329, opened 2021-11-10
Add support for CUDA 11.2 by upgrading Torch/Tensorflow/CUDA compatibility maps

andreasjansson:
Also includes fixes for torch version scraper, since the torch json was changed.


bfirsh:
See also #183 and #228. Did you manage to fix the issues mentioned in those PRs, causing the tests to fail?

Related to #181 
Closes #202 
Closes #204

andreasjansson:
> See also #183 and #228. Did you manage to fix the issues mentioned in those PRs, causing the tests to fail?

Yes, the issue there was that TF names the versions "11.2" and the CUDA base images use the full "11.2.0" notion. Using `version.Equal` instead of plain `==` fixes that.

pkg/dockerfile/generator_test.go
```diff
@@ -92,7 +92,7 @@ predict: predict.py:Predictor
 	require.NoError(t, err)
 
 	expected := `# syntax = docker/dockerfile:1.2
-FROM nvidia/cuda:11.0-cudnn8-devel-ubuntu16.04
+FROM nvidia/cuda:11.2.0-cudnn8-devel-ubuntu20.04
```
bfirsh, line 95: I would expect the default Ubuntu and CUDA version to remain stable. What's going on here? Is it non-deterministically picking the Ubuntu version somehow?

bfirsh (approved):
Happy to get this in as a bodge, but perhaps we need to open another issue about default Ubuntu versions not remaining stable.

pkg/dockerfile/generator_test.go
andreasjansson: If you set `gpu: true` without specifying `cuda:` or including any python packages that pick a CUDA version for you, it defaults to the latest CUDA version. NVIDIA's base image uses Ubuntu 20.04 for CUDA 11.2.

bfirsh: Gotchya. I have created an issue here: https://github.com/replicate/cog/issues/330

PR #331, opened 2021-11-12
Capture _all_ stdout/stderr in Redis worker

andreasjansson:
We previously missed output printed from C modules, notably TensorFlow's GPU logs that are really helpful for debugging CUDA issues.

This PR uses os.dup2() to tee stdout/stderr to a subprocess that sends log messages to the log queue. It's fiddly low-level stuff so I've added lots of comments inline.

### DIFF
--- a/python/cog/server/redis_log_capture.py
+++ b/python/cog/server/redis_log_capture.py
@@ -1,0 +1,161 @@
+import time
+import json
+import sys
+import uuid
+import multiprocessing
+import os
+import contextlib
+
+import redis
+
+
+@contextlib.contextmanager
+def capture_log(redis_host, redis_port, redis_db, log_queue, stage, prediction_id):
+    """
+    Send each log line to a redis RPUSH queue in addition to an
+    existing output stream.
+    """
+
+    # if we don't have a log queue, this is a no-op
+    if log_queue is None:
+        yield
+
+    # if we have a log queue, tee stdout and stderr to redis via a logging subprocess
+    else:
+        outs = {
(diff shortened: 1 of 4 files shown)


python/cog/server/redis_log_capture.py
```diff
@@ -0,0 +1,161 @@
...
+                stage=stage,
+                pipe_reader=pipe_readers[out_name],
+                old_out_fd=old_fds[out_name],
+                done_token=done_token,
+            )
+            procs[out_name].daemon = True
+            procs[out_name].start()
+
+        try:
+            # do the work
```
zeke, line 67: :)

zeke (approved):
I love the meticulous commenting. I tried my best but I couldn't find a single typo.

test-integration/test_integration/conftest.py
```diff
@@ -72,8 +77,14 @@ def setup(self):
     @cog.input("path", type=Path)
     @cog.input("output_file", type=bool, default=False)
     def predict(self, text, path, output_file):
+        logging.warn("writing log message")
+        time.sleep(.1)
```
zeke, line 81: I know this only amounts to an extra 0.4 seconds in test duration and doesn't really matter, but is there another way to do this that would run faster? Does python have something like Node's `process.nextTick()`?

andreasjansson: I'm not sure how to make that work since there are two processes reading from stderr and stdout. Without explicit synchronization it would be hard to have them arrive in order without a small timeout.

Issue #295, opened 2021-10-07
Output printed at import time causes type signature parsing to fail


andreasjansson:
If any of the imports write to stdout, the type signature parser fails and cog dies with a cryptic error.

For example:
```
$ cog predict --debug

[...]
[+] Building 1385.8s (16/16) FINISHED
[...]

Adding labels to image...
$ docker run --rm --shm-size 8G r8.im/user/model python -m cog.command.type_signature
Some debug output
{
  "inputs": [
    {
      [...]
    }
  ]
}


ⅹ Failed to get type signature: invalid character 'S' looking for beginning of value
```

josephcatrambone-crucible:
I think I'm stumbling into this issue again, but the deeper problem appears to be that nVidia is throwing a deprecation message out which is interfering with parsing.  I'd wager this can be alleviated by updating the image, but not being able to rebuild images makes me a little worried.

```
 => => preparing build cache for export
Adding labels to image...

==========
== CUDA ==
==========

CUDA Version 11.6.0

Container image Copyright (c) 2016-2022, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.

*************************
** DEPRECATION NOTICE! **
*************************
THIS IMAGE IS DEPRECATED and is scheduled for DELETION.
    https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/support-policy.md

{
  "openapi": "3.0.2",
  "info": {
    "title": "Cog",
    "version": "0.1.0"
...
  }
}


ⅹ Failed to get type signature: invalid character '=' looking for beginning of value
```

zeke:
Thanks for following up @josephcatrambone-crucible 👍🏼

PR #332, opened 2021-11-17
Suppress output before parsing type signature

andreasjansson:
Fixes https://github.com/replicate/cog/issues/295


zeke (approved):

zeke:
Thanks @andreasjansson 🍰

I just released [v0.0.18](https://github.com/replicate/cog/releases/tag/v0.0.18) which includes this fix.

cc @dribnet

bfirsh:
@andreasjansson Shouldn't this redirect stdout to stderr instead of suppressing it completely? I'm running into an issue atm where errors aren't printed correctly.

Either way, stderr should still be output. Only stdout is parsed for the type signature, I think.

PR #340, opened 2021-11-19
Mock time instead of sleeping in tests

bfirsh:
This is hard to do without in integration tests, but we shouldn't have to do this in unit tests. 1 second aggregated over number of times we run unit tests, etc...


andreasjansson:
Nice, thanks for cleaning that up! 👍

andreasjansson (approved):

zeke:
Time, we mock thee!

bfirsh:
download-1
download

PR #341, opened 2021-11-20
Add full example of cog.yaml to getting started

bfirsh:
I was finding it hard to find an example of `cog.yaml` and `predict.py`
the other day when showing someone Replicate. This is the documentation
that is linked to from replicate.ai/docs.

This also makes it a bit clearer what is going on, I think.


zeke (approved):
👌🏼

Issue #238, opened 2021-09-14
document workflow for automatic git sign-off to reduce contribution friction


zeke:
The [contributing guidelines](https://github.com/replicate/cog/blob/main/CONTRIBUTING.md#signing-your-work) for this project require that every commit be "signed off" with an extended commit message:

> Each commit you contribute to Cog must be signed off. It certifies that you wrote the patch, or have the right to contribute it. It is called the Developer Certificate of Origin and was originally developed for the Linux kernel.

After submitting a few PRs to this repo, I've found that requirement to be a slight hurdle to getting things done. To reduce friction for future contributors, I think we should amend the contributing docs to describe ways that this sign-off process can be automatic.

Here's a process that is working for me locally: https://github.com/zalando/zalando.github.io/blob/9d7cbdf1e5e5b27838899ef685838460cdb69b6a/_docs/releasing/contributions.md#automatic-git-sign-off

PR #342, opened 2021-11-23
document how to automatically sign off commits

zeke:
This PR updates the CONTRIBUTING docs with info about how (and why) to use a git hook for sign-off.

Resolves https://github.com/replicate/cog/issues/238

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 
 ### Signing your work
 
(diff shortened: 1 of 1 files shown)


bfirsh:
Super.

For reference I put this in my `cog/.git/config`:

```
[commit]
    template = /Users/ben/.gitsignoffmessage
```

Then `~/.gitsignoffmessage` is:

```


```

I think this doesn't work with `git commit -m ...` though, so the hook method is probably better.

zeke:
> I think this doesn't work with `git commit -m ...`

When does it work?

bfirsh:
When you just type `git commit` and it uses `$EDITOR` to create the commit message.

Issue #327, opened 2021-11-09
Strawman: FastAPI model server


bfirsh:
This is potential design that could hit three interconnected birds with one stone:
- 🐣 #193 
- 🐥 #205 
- 🐔 #259 

Those tickets contain user data and requirements. Please read them for context.

This is a strawman. It is most likely wrong. Please attack with your sharpest pitchforks!

## Model definition

The model API is specified using Pydantic types in the predict function.

```python
import cog
from pathlib import Path
from pydantic import Field
import torch


class Predictor(cog.Predictor):
    def setup(self):
        self.net = torch.load("weights.pth")

    def predict(
        self,
        input: Path = Field(..., title="Image to enlarge"),
        scale: float = Field(1.5, description="Factor to scale image by"),
    ) -> Path:
        """Run a single prediction on the model"""
        # ... pre-processing ...
        output = self.net(input)
        # ... post-processing ...
        return output
```

You could imagine some custom types like `cog.Field` if we need custom behavior, or don't want the extra import.

This predictor serves:

### `GET /predict`

Takes a JSON object in request body:

```json
{
  "input": "https://bucket.s3.amazon.com/...", # or "data:foo/bar;base64,..."
  "scale": 1.7,
}
```

Returns:

```
{
  "output": "https://bucket.s3.amazon.com/..."
}
```

Everything is JSON. Files are represented as URLs -- HTTP, cloud storage, data: URLs, or perhaps even file:// for mounted storage. This is for simplicity, so it can be represented sensibly in OpenAPI, and so more complex data structures can be represented.

If a more complex Pydantic object was defined as the return value, `output` would be a JSON object. The root object can also include other information about the run -- some prior art in #259 might be informative about what we want to include (metrics, etc).

Authentication and configuration for cloud storage needs some thought. It could specify a prefix for storing output files in the request.

### `GET /docs`

A Swagger UI, generated automatically by FastAPI.

### `GET /openapi.json`

An OpenAPI definition, generated automatically by FastAPI. This is used as the type definition. It contains a complete documentation of the HTTP API and entries like this to define the types for input/output:

```
{
  "components": {
    "schemas": {
      "Input": {
        "title": "Input",
        "required": [
          "input"
        ],
        "type": "object",
        "properties": {
          "input": {
            "title": "Image to enlarge",
            "type": "string",
            "format": "path"
          },
          "scale": {
            "title": "Scale",
            "type": "number",
            "description": "Factor to scale image by",
            "default": 1.5
          }
[...]

Issue #266, opened 2021-09-28
Cog-native way of handling temp folders


andreasjansson:
Many models depend on temporary files and folders, and it's currently up to the user to clear them all up at the end of a prediction.

There should be some way in Cog to create temporary files and folders that are scoped to a single prediction and cleaned up automatically by Cog at the end of the prediction.

bfirsh:
I wonder if there's something clever we can do with file handles, in a way that's also not confusing and memory inefficient...

andreasjansson:
I think we need a more general solution, since many of these temp files are generated by things far away from predict.py. For example, a common pattern is for an audio model to output a .wav file, and then shell out to ffmpeg to convert it to mp3.

qiao-lang1997:
as

markphelps:
Closing as stale because there has been no activity on this issue for over 4 years. If this feature is still needed, please open a new issue or reopen with updated context.

Issue #259, opened 2021-09-23
Review model API design


bfirsh:
The current model API was hastily written, and we have now learned many things that could be incorporated into its design.

## User data

- Replicate! We currently use an undocumented Redis queue consumer.
- An industry user wants a gRPC API with support for non-blocking requests for long-running models.
- An industry user wants an AMQP RPC API. It is unclear whether the model itself needs to interact over this API, or whether a sidecar 

### Essential

- The primary API for Cog should be the same thing we use for Replicate.
- Make predictions as standard blocking REST API. JSON over HTTP.
- Make it possible to use with a queuing system. Either built-in, as a sidecar at the container-level, or as an extension at the Python-level.
- Pass files as either URLs or base64 encoding, depending on if efficiency is important. (Interdependent with #205.)

### Future, to design in context

- gRPC interface. This is harder to use, so shouldn't be the primary interface, but it seems to be becoming widely used for ML.
- Other serving platforms, like AI Platform or Kserve.
- Non-blocking API with requests
- #230 

## Prior art

We don't need to reinvent the wheel.

### Off the shelf

- https://www.tensorflow.org/tfx/serving/api_rest
- Kserve's model API: https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md
- Kserve has a sidecar queue proxy: https://github.com/kserve/website/blob/main/docs/modelserving/control_plane.md
- Sagemaker? (Can anyone find the actual API documentation?)
- https://github.com/triton-inference-server/server/blob/main/docs/architecture.md#models-and-schedulers
- https://pytorch.org/serve/rest_api.html

### Real world

- How is it done at spotify @andreasjansson?
- Might be some descriptions of how people do serving here: https://huyenchip.com/mlops/

## Areas for discussion

- What are the trade-offs between an HTTP based API and a queue based API?
- How do we support queuing systems?

## Next steps
- [ ] Use default Cog server for Replicate

## See also
- #193
- #205 
- #212

(Maintainers -- please edit this and consider it a wiki! Edited by @bfirsh, ...)

bfirsh:
Hmm, I wonder whether we can consider this superseded by #443?

Issue #205, opened 2021-08-14
Design type system & signature


bfirsh:
The type system and signature needs some thought.

## User data

- An industry user wants a way of documenting the inputs and outputs of their models so they can figure out how they fit together. Their models use a combination of video, image, and text content with additional metadata that could be represented as JSON.
- We, Replicate, need a type signature to generate forms on Replicate. Specific problems we have encountered:
  - ints and bools are strings
  - There is no image type
  - There are no output types

## Requirements

- We need to define input and output types for a prediction function. It should have some kind of schema (what we currently have at `GET /type-signature`).
- Standard types and custom types (e.g. images, or complex objects) need to be represented.
- Multiple or complex outputs need to be represented somehow. #212 
- It needs to be able to represent files.
- It should support standard REST JSON + HTTP, plus other representations and transports (e.g. gRPC, AMQP)
- Ideally, we want a standard off-the-shelf system (e.g. OpenAPI).

## Potential implementations

We could use OpenAPI / JSON Schema, which would handily generate a self-documenting API, instead of our unnecessarily-invented-here `/type-signature` endpoint. This might happen automatically if we make use of FastAPI in #193. (Hello @zeke!)

A potential issue with this is mixing in files to input/output. There are a few potential solutions: gRPC, multipart forms, references to external files (but how do output files work?). Maybe there is prior art here we can learn from.

## See also

- #193 
- #259 (blocked on / closely related)

andreasjansson:
Another reason for the type signatures to live in Python is that it makes the Cog Python package a standalone thing that can be used and integrated in other places, without using the Cog command line or cog.yaml. For example, if you have a custom Docker image and still want to push to Replicate (https://github.com/replicate/cog/issues/290), or if you want to import someones Predictor class at the Python layer.

PR #343, opened 2021-11-24
Design exploration for new Cog schema and API

bfirsh:
49b5af0a584b49f6a84f46a1548c19cc-superJumbo

An exploration of #193, #205, and #259, based on a design we have sketched out in private (soon to be on GitHub). Also incorporated in this design, not represented on GitHub, is how we can add a public queue-based API for Cog.

This is here for visibility and so @zeke and I can collaborate on this. Not ready for reviewing yet.

A few notes:
- The README is an exploration of how we might simplify and explain the value proposition given the new design. It is not the right design.
- I have started converting the tests to explore what we need to implement. Most of these tests can be removed -- they are merely testing Pydantic/FastAPI behavior.
- The main missing part of the HTTP server is handling files. URLs need converting to files. #266 might want to be considered as part of this too.

# See also 
- #327


bfirsh:
Rebased on top of #369.

bfirsh:
Notes from conversation with @andreasjansson about integrating this with Replicate's Redis queue:

Serving mostly blinding passes through JSON, so we can mostly re-use the existing system. The only thing that serving modifies is the `status` field, for statuses like `queued` etc.

## Files

Inputs will be signed URLs from web.

Outputs will be posted to a URL web that is passed as part of request, which will return a URL that will be embedded in output object. E.g.:

```
{
    "input": ...
    "file_upload_url": ...
}
```

All file handling will be removed from serving, and serving will blindly pass through these JSON objects.

## Logs

We want to get rid of the separate logging queue because it doesn't work very well. We will make the prediction object returned by Cog consistent with what is current returned by serving. Both HTTP and redis will use this object for prediction output/status:

```
{
    "status": ...
    "output": ...
    "logs": ...
}
```

Serving will blindly pass this through and not have separate queue for logging.

bfirsh:
Replaced by #378.

PR #346, opened 2021-11-26
Configure VSCode for py.test

bfirsh:
All the testing integration works now, including handy little "run this test" buttons on tests:

Screen Shot 2021-11-25 at 20 19 20


zeke (approved):
👌🏼

Issue #320, opened 2021-11-02
Copy ~/.cache/x into Docker


yeldarby:
Every time I do `cog predict`, the `huggingface` package is downloading the 500MB GPT2 pretrained weights into `~/.cache/huggingface` which is quite slow & bandwidth intensive.

I can't figure out how to bundle these into the Docker.

So far I've tried
* Using `run` in `cog.yaml` (doesn't work because it doesn't have access to the file yet)
* Copying `~/.cache/huggingface` into my working dir and having my `predict.py` copy the files at the top (which doesn't work for unknown reasons; possibly everything outside of `/src` is read-only) `os.system("cp -r /src/cache/huggingface /root/.cache/huggingface")`

zeke:
Hi @yeldarby. Thanks for opening this issue and following up with the fix you found.

It looks like this issue is specific to the `huggingface` package, so I'm not sure what generic changes we could make here to prevent other users of Cog from running into this issue. Do you have any suggestions on how we might improve behavior or documentation around this in a generalizable way?

yeldarby:
It's not just `huggingface`; other libraries like `openai/clip` and `torch` also store data in `~/.cache`.

Ideally make a way to mount volumes in custom locations where data will persist between runs. Or being able to add `COPY` lines to the `Dockerfile` to move files into the proper location on the container's filesystem.

hangtwenty:
### I want this!

As a user I want something like this too. In Docker terms I'm leaning more towards `ENV` than `COPY` as a way to address it.

## Brainstorming

### Security note

**It's very important** that the host's `~/.cache` does not get automatically pulled in, because a user could have private/sensitive/unrelated data in there. (Aside from bloat.)

### `$XDG_CACHE_HOME`

[...]

Issue #291, opened 2021-10-07
Environment variables in cog.yaml


andreasjansson:
Cog should have an equivalent to the Dockerfile `ENV` stanza or the Docker Compose `environment` option.

Like this:

```yaml
build:
  environment:
    MY_VAR: my_value
```

hangtwenty:
I implemented a rough draft of this feature. I still need to test it though.

wolfgangmeyers:
I literally need this to fix a problem in rembg.

Majdoddin:
I also need it to set some CUDA environment variables.
Meanwhile, this is a workaround: 
Put this Dockerfile in some empty subfolder, say `docker_env/` in your project:
```
FROM your_model
ENV var1=val1
ENV var2=val2
```
now, after building your model the usual way, 
```
$ cog  build -t your_model
$ cog login
$ cd docker_env
$ docker build -t your_model_env_vars .
$ docker tag model_env_vars r8.im/username/model_name:latest
$ docker push r8.im/username/model_name:latest
```
And that is all. 
If you inspect it:
```
docker inspect your_model_env_vars
```
you can find your vars at the end of `"Env": ` part.

PR #361, opened 2021-12-15
Cog environment variables

hangtwenty:
Implements #291.

This makes it so you can set environment variables in the Docker image that cog generates. In Dockerfile parlance, it's support for `ENV`.

This enables adding to cog.yaml:

```yaml
environment_variables:
    - FOOBAR=baz
    - SOME_LIBRARY_CACHE=/src/some_library_cache
```

And those become `ENV` directives in the Dockerfile.

----

It also applies a thoughtful default for [`$XDG_CACHE_HOME`](https://github.com/replicate/cog/issues/320#issuecomment-989545205) (see #320 for context). It's inside of `/src` so that it's preserved between runs.


pkg/dockerfile/generator.go
```diff
@@ -131,9 +132,53 @@ func (g *Generator) baseImage() (string, error) {
...
+
+	if _, ok := envVarMap["XDG_CACHE_HOME"]; !ok {
+		// Cog sets a default value for $XDG_CACHE_HOME. Why:
+		// Cog mounts the project directory so anything in there (in /src in the
+		// image) gets retained between runs. $XDG_CACHE_HOME is used by various
+		// libraries including popular ML libraries; so, setting it to a subdirectory
+		// within /src results in retaining the cache between runs. Ultimately,
+		// everything in /src gets "baked in" to the cog image. So this is great
+		// for caching pretrained models and so on.
+		// For more context see: https://github.com/replicate/cog/issues/320
```
hangtwenty, line 165: This comment is kinda verbose. I'd like to clean it up.

bfirsh:
Great. Haven't reviewed this properly, but for starters: this should be `environment` for consistency with Docker Compose.

(Prepare for some noodling because this is my wheelhouse. 😁)

It's also worth noting that Docker Compose's `environment` is a _runtime_ feature rather than a _build_ feature. It is options passed to `docker run` rather than `ENV` statements in `Dockerfile`. It has the ability to read environment variables from the environment at runtime, etc.

I haven't thought this out in depth, but perhaps there is an alternate design here where environment variables feel like they're set at runtime. (But we do also want to set environment variables at build time to be baked into the image.)

It might be worth considering a future where we want to read environment variables from the environment. I wonder how that would work, and whether it semantically makes sense for them to live under `build.environment`. Where should options live in `cog.yaml` if they're used both at build time and run time?

Maybe this is a distinction that Cog doesn't need to make, because it is always in control of the image it is running, so perhaps `environment` makes sense as part of the build step. We can also move `environment` to somewhere else in the future if we change the semantics.

hangtwenty:
@bfirsh yeah to be sure, the `environment_variables` name was verbatim from #291. I wasn't sure why `environment_variables` except that it might be to avoid confusion with the semantics of `environment` in Docker Compose.

hangtwenty:
@bfirsh I grok what you mean about a design that considers runtime variables as well. This PR is meant to implement `ENV` directives for the generated `Dockerfile`s, while leaving it open for future features.

hangtwenty:
In this first cut, `environment_variables` is under `build`.

I figure that the future features for runtime environment variables would live under `predict` rather than `build`. I say "under" because `predict` is currently top-level

(Notable precedent in `cog`'s current design: there's a `run` key under `build`, with semantics similar to `RUN` in Dockerfiles; i.e. it's run during build. So maybe instead of `build.environment_variables`, it's just `build.env`. `build.env`, `build.run` ... It makes sense to me anyways.)

bfirsh:
To be clear — I might be wrong! Was just noodling on some possible design questions. environment in build is likely the right design…

PR #362, opened 2021-12-15
Support predict modules in subdirectories

### LINKED ISSUE #164: Support importing model from subdirectories
This works:

```yaml
model: "predict.py:MyModel"
```

But this doesn't:

```yaml:
model: "mymodel/predict.py:MyModel"
```

It should.

andreasjansson:
Closes #164

### DIFF
--- a/test-integration/test_integration/test_predict.py
+++ b/test-integration/test_integration/test_predict.py
@@ -97,3 +97,44 @@ def test_predict_with_remote_image(tmpdir_factory):
     # lots of docker pull logs are written to stdout before writing the actual output
     # TODO: clean up docker output so cog predict is always clean
     assert out.strip().endswith("hello world")
+
+
+def test_predict_in_subdirectory_with_imports(tmpdir_factory):
+    tmpdir = tmpdir_factory.mktemp("project")
+    subdir = tmpdir.mkdir("my-subdir")
+    with open(subdir / "predict.py", "w") as f:
+        f.write(
+            """
+import cog
+
+from mylib import concat
+
+class Predictor(cog.Predictor):
+    def setup(self):
+        pass
+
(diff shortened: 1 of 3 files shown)


bfirsh:
Smashing.

To encode a verbal conversation: at some point, there is probably some prior art here we can build upon, instead of reinventing the wheel. For example, a few places in the Python ecosystem (gunicorn, etc) use the `module.name:function` form. That's not ideal for various reasons, but maybe best to be consistent and not have to maintain our own lookup system. This feels like a thing that will unravel with edge cases and import path fun.

test-integration/test_integration/test_predict.py
```diff
@@ -97,3 +97,44 @@ def test_predict_with_remote_image(tmpdir_factory):
     # lots of docker pull logs are written to stdout before writing the actual output
     # TODO: clean up docker output so cog predict is always clean
     assert out.strip().endswith("hello world")
+
+
+def test_predict_in_subdirectory_with_imports(tmpdir_factory):
+    tmpdir = tmpdir_factory.mktemp("project")
+    subdir = tmpdir.mkdir("my-subdir")
+    with open(subdir / "predict.py", "w") as f:
```
bfirsh, line 105: It'd be nice to just use fixtures on disk. Less fiddly to write, IDEs work, etc.

bfirsh (approved):

bfirsh:
I was going to merge but I see you've pushed a new commit. Feel free to merge once they pass!

test-integration/test_integration/test_predict.py
zeke: We did "fixtures" on disk for `cog init`

zeke:
Nice one, @andreasjansson 🙌🏼

PR #366, opened 2021-12-27
recognize all contributors

zeke:
This PR sets up [All Contributors](https://allcontributors.org/) on the project so we can recognize contributions in all their various forms, not just commits.

- Added the @allcontributors bot which we can use to add contributors directly from Issue and PR comments.
- Added docs on how and we acknowledge contributions.
- Added the folks from GitHub's [Contributors](https://github.com/replicate/cog/graphs/contributors) page. There are more folks not on that list that we can add manually after this lands.
- There's a CLI for adding contributors but we probably need it since we have the GitHub bot. We can add it later if needed.

---

Here's what it look like in the rendered README:

Screen Shot 2021-12-27 at 1 00 20 PM

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1,0 +1,96 @@
+{
+  "projectName": "cog",
+  "projectOwner": "replicate",
+  "repoType": "github",
+  "repoHost": "https://github.com",
+  "files": [
+    "README.md"
+  ],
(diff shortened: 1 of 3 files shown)


bfirsh:
Wonderful!

bfirsh:
@allcontributors please add @zeke for tool

PR #369, opened 2021-12-28
Don't require setup() and simplify tests

bfirsh:
A few related changes here. Notably I have simplified all the tests by not testing `setup()` in each test. It muddled what each test was actually testing.

It's a single code path that calls setup in the HTTP server, so this should fully test it. If that code path broke it would be more helpful to have one failed test saying "setup doesn't work" than a few dozen broken tests.

@andreasjansson Was there reasoning behind `setup()` being tested in each test that I'm not aware of?

Commits are significant.


zeke (approved):
🐕 wow such tests

andreasjansson:
> @andreasjansson Was there reasoning behind setup() being tested in each test that I'm not aware of?

That test suite mostly grew out of copy-paste, so no particular reason. It makes sense to test setup separately :+1:

Issue #286, opened 2021-10-05
Display warning or throw error if user tries to run remote model with outdated local cog


andreasjansson:
We should either display a warning or throw an error if a user tries to `cog predict` a remote model that has been built with a newer version of Cog that is incompatible with the local version.

We currently don't have a definition of "incompatible", so we could just use minor version.

markphelps:
This issue has been inactive for over 4.5 years with no activity or progress. I'm closing it as stale. If version compatibility between local Cog and remote models is still a concern with current versions of Cog, please open a new issue with updated context.

Issue #114, opened 2021-06-04
Use separate `required` option for `@cog.input()`


bfirsh:
The double behavior of `default` is not obvious.

This might want to be `optional=True` instead, so that inputs are required by default. If the inputs were not required by default, then users normally would make them required. If required inputs are not marked as required, this will cause breakage.

zeke:
Closing, as `@cog.input` has been removed as of v0.1.0-alpha.

PR #378, opened 2022-01-05
Use Pydantic for type annotations and FastAPI/OpenAPI for HTTP API

### LINKED ISSUE #58: Default values are sometimes the string "None"

### LINKED ISSUE #113: Define path types without having to import pathlib
This is clumsy to explain in getting started, and means we have to have extra imports in all the model definition docs. Also a thing users will stumble on.

### LINKED ISSUE #193: Revisit design of Python type annotations
`@cog.input()` is separate from the actual type signature and involves referencing arguments with strings. We are also doing a bunch of inventing-here with type checking and API generation.

Can we do something more elegant with Python types/Pydantic/FastAPI?

## Blockers

- Probably depends on #259 to ensure we can output correct APIs.
- Closely intwined with #205 -- they might be one and the same thing if we use something like FastAPI.

## Proposed design

Pydantic and [FastAPI](https://fastapi.tiangolo.com/tutorial/query-params-str-validations/) have a pattern where the default value can be an object with more details about the object.

For example, to translate an example from [our docs](https://github.com/replicate/cog/blob/main/docs/getting-started-own-model.md):

```python
import cog
from pathlib import Path
import torch


class Predictor(cog.Predictor):
    def setup(self):
        self.net = torch.load("weights.pth")

    def predict(
        self,
        input: Path = cog.Input(description="Image to enlarge"),
        scale: float = cog.Input(default=1.5, description="Factor to scale image by"),
    ) -> Path:
        # ... pre-processing ...
        output = self.net(input)
        # ... post-processing ...
        return output

```

`cog.Input` is the same as `pydantic.Field`. We could just use that.

The beauty of incorporating Python types and FastAPI is:
- IDE type completion / checking works
- We can automatically generate a validating API and an OpenAPI schema for the model (which has a number of additional benefits: automatic docs, automatic forms, automatic client libraries, etc)

### LINKED ISSUE #205: Design type system & signature

### LINKED ISSUE #212: Output types and multiple outputs
We support multiple inputs and types for them. We should do the same for outputs.

Probably depends on #205

### LINKED ISSUE #246: Cog requires inputs for predict function
There's no reason that it should. This should work:

```
class Predictor(cog.Predictor):
    def setup(self):
        # ...

    def predict(self):
        return "hello"
```

It at least should not throw a confusing error message about "`_inputs` not set" or something.

### LINKED ISSUE #259: Review model API design

### LINKED ISSUE #262: Integration tests should use Docker networking instead of relying on host's networking
The redis integration tests connect to redis by going out of Docker, to the host machine, then back into Docker via the exposed port, using the IP address of the host machine.

This is fragile because it might break depending on the network setup of the host machine. A more reliable way would be to use Docker networks (and just connect to `redis:6379` inside the model container).

### LINKED ISSUE #304: Everything in predict method but `name` needs to be a named argument
When attempting to specify `type` and `help` without names:

```py
@cog.input("seed", str, "A random seed", min=0, max=1000, default=None)
```

The following error is produced:

```
E   TypeError: input() got multiple values for argument 'default'
```

### LINKED ISSUE #306: String max_length parameter on @cog.input
We don't have validation on string length in @cog.input. A `max_length` field could provide that.

Additionally it could be used when building UIs on top of Cog, e.g. Replicate.ai, to determine if the input element should be a single line input or multi-line textbox.

### LINKED ISSUE #327: Strawman: FastAPI model server

bfirsh:
The exploration in #343 was sufficiently explored and getting very messy, so I thought I would create a new pull request for the real implementation.

This is an implementation of #193, #205, and #259 using Pydantic for the type system, FastAPI for the model server, and OpenAPI for the schema.

Models now look a bit like this:

```python
from cog import Predictor, Input, Path
import torch

class ColorizationPredictor(Predictor):
    def setup(self):
        self.model = torch.load("./weights.pth")

    def predict(self,
          image: Path = Input(title="Grayscale input image")
    ) -> Path:
        processed_input = preprocess(image)
        output = self.model(processed_input)
        return postprocess(output)
```

The HTTP API is:

- `POST /predictions`: Run a prediction.
- `GET /openapi.json`: OpenAPI schema for HTTP API, including input and output JSON Schema types generated from the `predict()` type signature.

Files are passed as URLs. Currently only `data:` and `http:` URLs are supported. It is intended that `file:`, `http:`, `s3:`, and `gs:` URLs will be supported. For Replicate, we will pass files using HTTP directly from web. Output files will be POSTed directly back to web.

The Redis queue worker remains in place. It now uses the same request and response objects as the HTTP API. The tests should explain its behavior. Logging still uses the old logging API, but is intended to add this to the response object like the Replicate serving API.

Commits are significant and this should be reviewed commit by commit. This is an intentionally large and long-running branch -- `main` is currently used as the documentation for Replicate.

## Todo
- [x] `cog predict` with new API
- [x] Input ordering?
- [x] Ensure HTTP server is single threaded.
- [x] HTTP URLs for input and output
- [x] Add `choices` option to `Input` to generate an `Enum`.
- [x] `cog init` template
- [x] Update getting started docs
- [x] Update reference docs
- [x] Rename `Predictor` to `BasePredictor` so we can do `from cog import BasePredictor, Input, Path` without colliding with a `Predictor` subclass.
- [x] Better errors if you get output type wrong.
- [x] ~Reliably set Cog version. In development, the version is set to `0.0.1` which is pretty useless. It should probably be hardcoded somewhere, and the release process needs updating. We need this to be able to switch between old and new Cog API in Replicate.~ We're just going to support version `dev` everywhere which means "latest".
- [x] Update examples https://github.com/replicate/cog-examples/pull/8
- [x] Update the OpenAPI Outputs to be a consistent place. (Simple outputs currently end up in Request object schema)
- [x] Test returning object as output
- [x] Test arbitrary JSON objects in output
- [ ] Is `go.sum` up to date?

## Future (Pydantic + OpenAPI)
- [ ] Tell users to upgrade if they use `@cog.input()` with new Cog.
- [ ] Document nouns to make sure we're all on the same page. "Input"/"Inputs"?
- [ ] Document how to migrate `@input()` to Pydantic type annotations.
- [ ] Go through and fix any TODOs
- [ ] Make input validation errors work for `cog predict` in a simple way until we lock down HTTP API
- [ ] Make input errors work for Redis API & Replicate
- [ ] Integrate with Replicate - https://github.com/replicate/replicate-web/pull/907
- [ ] Force users to pick an output type -- throw an error if it's not set. (Couldn't just be `Any`, but they need to be explicit!)
- [ ] Clean up temporary files -- both input and output
- [ ] Do we want to make `cog predict` with an existing image backwards compatible? ('cos there are old models on Replicate)
- [ ] What happens if you use `pathlib.Path` as an input type? Does it work? Should we throw an error? (This might happen if you import wrong package / collide imports)
- [ ] What happens if you return `pathlib.Path` from a predict function? Does this work? Should it throw an error?
- [ ] Should we have a fixed set of supported inputs/outputs somehow?
- [ ] Do we want to force users to put extra properties in `additionalProperties`? Should we even allow extra properties to `cog.Input()`?
- [ ] Merge examples PR https://github.com/replicate/cog-examples/pull/8

## Far-flung future
- Vendor Pydantic and FastAPI
- Add back support for PyTorch and Tensorflow tensors as output. This was removed because it was fiddly to get working with Pydantic. You now need to convert them to numpy arrays.
- Add back timing information to server.
- Better error if you forget an output type
    ```
    pydantic.error_wrappers.ValidationError: 1 validation error for Prediction
    output
      unexpected value; permitted: None (type=value_error.const; given=hello foo; permitted=(None,))
    ```
- Multiple file outputs for `cog predict`. It currently just outputs the plain JSON if there isn't a single file output.
- Logging, as described in comment in #343
- Review request/response objects and make sure we're happy with them.
     - This might be an opportunity to rename the statuses and ensure they're consistent ("success" vs "failed").
     - Do we want the input type to have a second `input` level so we can have top-level request options?
- Do we need separate URLs for input/output schema, or can they be reliably fetched from OpenAPI?
- [ ] Document HTTP API
- [ ] Do we want to version the prediction API, or let the client handle it? https://github.com/replicate/cog/issues/94
- [ ] Test arbitrary objects in input
- [ ] What if people run new models from a directory of code with old Cog? What happens if we do that? Do we care? #286 
- [ ] What if people run new models from an image with old Cog? Maybe we throw a simple error from `/predict`?

## Based on (& blocked by)

- #383
- #385 

## Refs
Closes #58
Closes #113 
Closes #193
Closes #205
Closes #212 
Closes #246
Closes #259
Closes #262
Closes #263
Closes #304 
Closes #306 
Closes #327
Closes #328


README.md
```diff
@@ -1,16 +1,14 @@
 # Cog: Containers for machine learning
 
-Use Docker for machine learning, without all the pain.
+Put your machine learning model in a standard, production-ready Docker container without having to know how Docker works.
```
zeke, line 3: ❤️

README.md
```diff
@@ -87,10 +63,34 @@ $ cog build -t my-colorization-model
 
 $ docker run -d -p 5000:5000 --gpus all my-colorization-model
 
-$ curl http://localhost:5000/predict -X POST -F input=@image.png
+$ curl http://localhost:5000/predictions -X POST \
+   --data '{"input": "https://.../input.jpg"}'
 ```
 
-That's it! Your model will now run forever in this reproducible Docker environment.
+<!-- In development, you can also run arbitrary commands inside the Docker environment:
```
zeke, line 70: Why is this commented out?

README.md
```diff
@@ -87,10 +63,34 @@ $ cog build -t my-colorization-model
...
+```
+$ cog run jupyter notebook
+```
+-->
+
+## Deploying models to production
+
+Cog does a number of things out of the box to help you deploy models to production:
+
+- **Standard interface.** Put models inside Cog containers, and they'll run anywhere that runs Cog containers.
```
zeke, line 89: `they'll run anywhere that runs Cog containers.` --- Should this be **Docker** instead of **Cog**

bfirsh: I mean, technically correct, but yes that is a more useful statement haha.

README.md
bfirsh: Just tinkering with this. Perhaps it's more clear what Cog is for (packaging models into Docker containers) when there isn't the diversion into development environments.

docs/getting-started-own-model.md
```diff
@@ -88,16 +88,25 @@ class Predictor(cog.Predictor):
...
+- `int`: an integer
+- `float`: a floating point number
+- `bool`: a boolean
+- `cog.File`: a file-like object representing a file
+- `cog.Path`: a path to a file on disk
+
+You can provide more information about the input with the `Input()` function, as shown above. It takes these basic arguments:
+
+- `description`: A description of what to pass to this input for users of the model
 - `default`: A default value to set the input to. If this argument is not passed, the input is required. If it is explicitly set to `None`, the input is optional.
```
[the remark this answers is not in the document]
bfirsh: There could be an argument for a more explicit, easier to understand design here. https://github.com/replicate/cog/issues/114 Not sure what is best.

bfirsh:
I have rebased this on main to get the vscode editor goodness.

bfirsh:
@zeke Looks like the docstrings do indeed end up in the schema, as the tests have told us. 😂

zeke:
> docstrings do indeed end up in the schema

Gotcha. I'll take a look.

zeke:
🍏

bfirsh:
This is an intentionally large and long-running branch because `main` is used as the documentation for Cog.

I wonder if we could get this work into `main`, but not update the documentation? Or maybe in a separate long-running branch on this repository that we can all easily open pull requests against?

zeke:
> I wonder if we could get this work into main, but not update the documentation? Or maybe in a separate long-running branch on this repository that we can all easily open pull requests against?

I'd be in favor of the latter: long-running branch on this repo (not a fork, and with a cool name)

1. Create a new branch from `main` called `new-cog` or `year-of-the-tiger`. Can we use emojis in branch names? 🤔 
1. Point this PR at `new-cog`
1. Merge this PR
1. Open followup PRs against `new-cog`

bfirsh:
👍

To summarize an IRL conversion: we chose the extremely cool name "future" for the branch.

I have cleaned up the commits a little (squashed some "fix tests" commits etc), updated go.mod, and based this PR on our new future branch. We're ready to roll, I think!

Next steps:
- [ ] Create a new Mega Pull Request for the "future" branch. This branch will be an initial version of Cog that we can use to start testing it with Replicate internally.
- [ ] Migrate "Future (Pydantic + OpenAPI)" tasks to that pull request.
- [ ] Create issues for the "Far flung future" and add them to the "next" in the project to remind us to do them before we ship new cog.

zeke (approved):
The `future` looks bright. 😎 

Ship it (to a feature branch)! 🚢

bfirsh:
In case it's not clear from the previous conversation: we decided to defer the design of the HTTP API to the future. The design in the body of this pull request is merely a strawman for a future design that we will lock down.

bfirsh:
I have added a few tickets for the far flung future that can be seen in the mentions above. They're in the "next" status in our project to remind us to look at them before we ship new Cog.

PR #379, opened 2022-01-07
New Cog bits n bobs

bfirsh:
Lifting some things out of #378 to avoid the curse of a giant pull request. Feel free to merge at any point, and I will continue to add things to this if they are independent of that work.


zeke (approved):
🧹

PR #385, opened 2022-01-12
Run redis integration tests in Docker network

### LINKED ISSUE #262: Integration tests should use Docker networking instead of relying on host's networking

bfirsh:
The queue_controller server has been replaced by a separate container
that can run inside the network.

Closes #262
Closes #263


test-integration/test_integration/util.py
```diff
@@ -54,33 +57,23 @@ def docker_run(
...
-        "network",
-        "inspect",
-        "bridge",
-        "--format='{{ json (index .IPAM.Config 0).Gateway }}'",
-    ]
-    out, _ = subprocess.Popen(cmd, stdout=subprocess.PIPE).communicate()
-
-    return out.decode().strip().replace('"', "").replace("'", "")
+        pass
+        # subprocess.Popen(["docker", "rm", "--force", name]).wait()
```
zeke, line 79: did you mean to live this in here?

test-integration/test_integration/conftest.py
```diff
@@ -1,25 +1,76 @@
...
+
+
+@pytest.fixture
+def upload_server_image():
+    """
+    Build the upload server once for the test run. The image doesn't change.
+    """
+    subprocess.run(
+        ["docker", "build", "-t", "cog-test-upload-server", "."],
+        cwd=Path(__file__).parent.parent / "upload_server",
```
zeke, line 55: TIL: `Path(__file__).parent.parent` -- I like how python uses dot notation for imports and paths.

zeke (approved):
Looks good!

test-integration/test_integration/util.py
bfirsh: No -- good catch. Thank you!

bfirsh:
@andreasjansson would appreciate some of your lovely eyeballs on this too, because you did previous implementation I think. A quick check just to make sure it makes conceptual sense with your knowledge of serving at least!

bfirsh:
It's also worth noting that GitHub is hiding the bulk of this pull request behind an easy-to-miss "Load Diff" button because it's too large (sigh).

andreasjansson (approved):
Amazing, I love this. Better in every way, feels really good to get rid of that janky multiprocessing queue code.

bfirsh:
Thanks for the 3rd and 4th eyes 🙌

Issue #196, opened 2021-08-12
Define what versions of Python are supported


bfirsh:
It should be documented in `cog.yaml` docs what versions are supported and warn if you're using a version that isn't supported. Also test this in CI.

bfirsh:
I think this got closed by some overly eager automation in #415.

We're still missing the actual check that validates `python_version`. This could be a separate ticket, I suppose.

PR #387, opened 2022-01-12
Run Python tests on multiple Python versions

bfirsh:
This would have caught a bug we found in new Cog on Python 3.7.

Ref #196. This is implicitly defining some of the Python versions we
support, but they are not enforced in Cog yet.

/cc @preeth1


preeth1 (approved):

zeke:
Build Matrices muahaaahaa. All the pythons. 

✨👐🏼✨ 
🐍🐍🐍🐍

PR #388, opened 2022-01-12
Install wheel with pyenv Python

bfirsh:
Not having wheel installed makes installing some packages very slow. We
noticed dlib particularly.

Docker's Python base image already installs the wheel package, but it is
not installed when we install Python ourselves with pyenv.


bfirsh:
This doesn't seem to actually fix the problem we were seeing, unfortunately. dlib is still being built from scratch.

Screen Shot 2022-01-12 at 12 13 13

That's because dlib doesn't have any wheels: https://pypi.org/project/dlib/#files

bfirsh:
It seems like wheel still should be installed though, because `pip install` was throwing warnings, so this is still a useful PR!

preeth1 (approved):

zeke:
I wonder if this change would also make this problem go away:

https://github.com/replicate/cog/blob/d15a16d818035a8b60e9e8c5967326d928a4edc7/CONTRIBUTING.md?plain=1#L145-L147

bfirsh:
Unfortunately not -- that's on the host machine when running the development environment, but this change installs the wheel package _inside_ Cog models.

zeke:
Ah gotcha

PR #391, opened 2022-01-13
update VS Code settings to format Python files on save

zeke:
Does what it says on the tin. 🥫


bfirsh (approved):

bfirsh:
Great! I have added a few more settings from my user config, so right back at you for review. :D

Lalmia11 (commented):
B612_20230307_025817_677

Issue #386, opened 2022-01-12
Switch to pyproject.toml for Python package


bfirsh:
This is the Correct Python Packaging way of doing things now.

mattt:
Resolved by https://github.com/replicate/cog/pull/843

PR #396, opened 2022-01-19
Add minimal pyproject.toml

clbarnes:
See #386.

Does not replace setup.py (unless an alternative build backend like
poetry or flit is used), but just lets installers like pip know what is
required and how to run setup.py.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,0 +1,4 @@
+[build-system]
+requires = ["setuptools", "wheel"]
+build-backend = "setuptools.build_meta"
+


zeke:
Thanks for the PR, @clbarnes!

I'll defer to resident python experts @bfirsh and @andreasjansson to review this.

hangtwenty:
I tried `poetry install` locally and got an error, but it was a pretty quick fix!

<details><summary>Expand for details of the error messages</summary>

I tried the `poetry` compatibility locally and got this error:

```python
❯ cd <...>/cog_repo/python
❯ poetry install

  PyProjectException

  [tool.poetry] section not found in <...>/cog_repo/python/pyproject.toml

  at ~/.poetry/lib/poetry/_vendor/py3.9/poetry/core/pyproject/toml.py:56 in poetry_config
      52│     def poetry_config(self):  # type: () -> Optional[TOMLDocument]
      53│         if self._poetry_config is None:
      54│             self._poetry_config = self.data.get("tool", {}).get("poetry")
      55│             if self._poetry_config is None:
    → 56│                 raise PyProjectException(
      57│                     "[tool.poetry] section not found in {}".format(self._file)
      58│                 )
      59│         return self._poetry_config
```

I added `[tools.poetry]` with just name but then got:

```python
❯ poetry install

  RuntimeError

  The Poetry configuration is invalid:
    - 'version' is a required property
    - 'description' is a required property


  at ~/.poetry/lib/poetry/_vendor/py3.9/poetry/core/factory.py:43 in create_poetry
       39│             message = ""
       40│             for error in check_result["errors"]:
       41│                 message += "  - {}\n".format(error)
       42│
    →  43│             raise RuntimeError("The Poetry configuration is invalid:\n" + message)
       44│
       45│         # Load package
       46│         name = local_config["name"]
       47│         version = local_config["version"]
```

So I added `version`/`description` and made `setup.py` read the `version` from `pyproject.toml`. After that, `poetry install` is happy.

Then tested `flit install` and got:

```
❯ flit install
Neither [project] nor [tool.flit.metadata] found in pyproject.toml
```

and updated accordingly. `flit install` still throws an error but it's not due to pyproject.toml so I'm going to leave that alone. Glad `poetry install` is happy now.
</details>

clbarnes:
Maybe I'm misunderstanding poetry and flit, but aren't they alternative build backends rather than frontends? The point of specifying the `build-backend` in `pyproject.toml` is to tell installers/ build frontends (i.e. pip) to use setuptools and setup.py to build it, not some other build backend. Both of those errors are the result of trying to use it like it's a poetry-built or flit-built project (which is why it's looking for metadata specific to those build tools), where it's not.

I think it's best to avoid setup.py reading pyproject.toml, as build frontends/ installation tools (i.e. pip; as opposed to build backends, i.e. flit/poetry) read pyproject.toml to learn that they should read setup.py in order to install it.

hangtwenty:
@clbarnes I agree that it's crufty, I was surprised that `poetry` (`1.1.12`) still requires `[tools.poetry]` even when `version`/`description` are specified in the generic part of `pyproject.toml`. So without that, `poetry` cannot be used as a development tool for managing a virtualenv for `cog`.

> The point of specifying the `build-backend` in `pyproject.toml` is to tell installers/ build frontends (i.e. pip) to use setuptools and setup.py to build it, not some other build backend.

Correct, and in your branch the `pyproject.toml` is pointing to the `setup.py`, and in my branch it is as well.

Subtle point: adding the `[tools.poetry]` causes no issues, and it is ignored when someone's not using `poetry`. So it's crufty in the `pyproject.toml`, but cleaner behavior IMHO

hangtwenty:
@clbarnes Does my PR make a little more sense now?

Honestly no worries if don't want to merge it! I poked around because I was curious, there's nothing wrong with the minimal `pyproject.toml` for the sake of `build-backend = "setuptools.build_meta"`

clbarnes:
Interesting. My understanding was that Poetry and Flit are not generic developer-side tools to be applied to any project you work on (the way that, say, pyenv or virtualenvwrapper are). They are opinionated tools which enforce a particular development workflow on particular projects configured to use them (as does setuptools, it's just a workflow which has been round a lot longer). Don't get me wrong, they're great tools! But I think any given project's compatibility being mutually exclusive between setuptools/poetry/flit is a feature rather than a bug; one probably shouldn't use setuptools to manage a poetry-based project and vice versa, and the fact that you *can* write a frankenstein pyproject.toml which happens to support several build backends for a particular subset of their configurations is more likely to lead to confusion than anything else.

hangtwenty:
@clbarnes I see where you're coming from. I get the contrast you're drawing between `{setuptools, flit}` vs. `{virtualenv, pyenv}`, but I think `{poetry, Pipenv}` (but especially `poetry`) can be used as a personal development tool. Sometimes I use `poetry` even when a project doesn't require it. (In those cases I leave `pyproject.toml` untracked if teammates don't want to switch to `poetry`).

My branch has the same result as yours, with just the additional _option_ of using `poetry install` as a one-command way to install into a virtualenv. But it's not necessary!

evilstreak:
We ended up replacing `setup.py` with `pyproject.toml` in #843.

PR #402, opened 2022-01-26
cog: add support for "environment" to set environment variables in the generated Dockerfile

hangtwenty:
ℹ️  Docs preview: https://github.com/hangtwenty/cog/blob/cog-environment-variables/docs/yaml.md#environment

[](https://github.com/hangtwenty/cog/blob/cog-environment-variables/docs/yaml.md#environment)

----

### Change 1/2: support for `environment` (`ENV`)

Closes issue #291, "Environment variables in cog.yaml." In Dockerfile parlance, it's support for `ENV`. This supersedes my the previous PR, #361.

This enables adding `environment` to `build` in `cog.yaml`:

```yaml
build:
  environment:
    - EXAMPLE=/src/example
    - DEBUG=1
```

And those become `ENV` directives in the Dockerfile.

### Change 2/2: A single default environment variable to configure caching for PyTorch and some other libraries (`XDG_CACHE_HOME`)

One of the libraries used most often with `cog` is PyTorch. You don't have to do configure anything for it, and now the caching should "just work" in most cases. Likewise for other popular libraries.

<details> <summary>Variables you DON'T need to set </summary>

You don't need to set any of the following because the sane default for `$XDG_CACHE_HOME` will take care of it. (Because it's part of a standard, several libraries use it for their defaults.)

* pytorch: You **don't** need to set `PYTORCH_TRANSFORMERS_CACHE` — [docs](https://pytorch.org/docs/stable/hub.html#:~:text=XDG_CACHE_HOME)
* Hugging Face: You **don't** need to set `TRANSFORMERS_CACHE` — [docs](https://huggingface.co/transformers/v4.0.1/installation.html#caching-models)
    * You _shouldn't_ have to set `PYTORCH_PRETRAINED_BERT_CACHE` — [docs](https://huggingface.co/transformers/v1.0.0/model_doc/overview.html#:~:text=PYTORCH_PRETRAINED_BERT_CACHE)
* pip: no variables needed! Before this PR, `cog` already configured the pip/apt-get cache directories

#### You don't need  to set `TORCH_HOME`

If `TORCH_HOME` is set, that'll take precedence over `XDG_CACHE_HOME` — see [pytorch docs](https://pytorch.org/docs/stable/hub.html#:~:text=XDG_CACHE_HOME), [huggingface docs](https://huggingface.co/transformers/v1.0.0/model_doc/overview.html#:~:text=PYTORCH_PRETRAINED_BERT_CACHE). So, if you set `TORCH_HOME` to something else, just be sure to set it within `/src/`, like `/src/.cache-torch`. That way, it will get cached across runs.

 </details>


pkg/config/config.go
```diff
@@ -15,6 +15,7 @@ import (
 // TODO(andreas): custom cpu/gpu installs
 // TODO(andreas): validate python_requirements
 // TODO(andreas): suggest valid torchvision versions (e.g. if the user wants to use 0.8.0, suggest 0.8.1)
+// (Possibly useful reference code: https://github.com/python-poetry/poetry/blob/master/src/poetry/puzzle/solver.py)
```
hangtwenty, line 18: - [ ] Need to remove this note, it was not for this branch.

pkg/dockerfile/generator.go
```diff
@@ -131,9 +132,53 @@ func (g *Generator) baseImage() (string, error) {
...
-ENV PYTHONUNBUFFERED=1
-ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/x86_64-linux-gnu:/usr/local/nvidia/lib64:/usr/local/nvidia/bin`
+	// If given, variables should be a list of strings, formatted like `KEY=VALUE`.
+	environmentVariables := g.Config.Build.BuildEnv
+
+	// Regex for valid environment variable names
+	regexpVariableName := regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
+
+	// Parse them into a list of strings. We also preserve the order.
+	// TODO(hangtwenty): Clean up by using an ordered map (maybe from a library)
```
hangtwenty, line 142: - [ ] `TODO(hangtwenty): Clean up`

pkg/dockerfile/generator.go
```diff
@@ -131,9 +132,53 @@ func (g *Generator) baseImage() (string, error) {
...
+	}
+
+	var envVarLines []string
+	for _, envVarName := range envVarKeys {
+		v := envVarMap[envVarName]
+		envVarLines = append(envVarLines, fmt.Sprintf("ENV %s=%s", envVarName, v))
+	}
+
+	return `ENV DEBIAN_FRONTEND=noninteractive \
+	PYTHONUNBUFFERED=1 \
```
hangtwenty, line 179: - [ ] I should remove the `\` changes here; I think it would cause a layer to rebuild, even for projects that do not adopt `build.env`. Instead, lead the existing `ENV DEBIAN_FRONTEND=...`/`ENV PYTHONUNBUFFERED=`... as separate lines. (In other words, just avoiding unnecessarily invalidating cached layers for models re-built after this change.)

pkg/dockerfile/generator.go
```diff
@@ -267,3 +312,13 @@ func filterEmpty(list []string) []string {
 	}
 	return filtered
 }
+
+// TODO(hangtwenty): De-duplicate this - the same function is in config.go
```
hangtwenty, line 316: - [ ] `// TODO(hangtwenty): De-duplicate this - the same function is in config.go`

pkg/dockerfile/generator.go
hangtwenty: On second thought, maybe it's fine? (LMK)

pkg/config/config.go
hangtwenty: I removed this comment, not sure why it is showing

pkg/config/config.go
```diff
@@ -24,6 +24,7 @@ type Build struct {
 	PythonFindLinks      []string `json:"python_find_links,omitempty" yaml:"python_find_links"`
 	PythonPackages       []string `json:"python_packages,omitempty" yaml:"python_packages"`
 	Run                  []string `json:"run,omitempty" yaml:"run"`
+	BuildEnv             []string `json:"env,omitempty" yaml:"env"`
```
bfirsh, line 27: This should be `environment` for consistency with Docker Compose.

It also probably doesn't need the `Build` prefix because it's already in a struct called `Build`. So maybe `Environment` if we're also renaming to "environment"?

pkg/config/config.go
```diff
@@ -206,7 +207,7 @@ func (c *Config) pythonPackageForArch(pkg string, goos string, goarch string) (a
 func (c *Config) validateAndCompleteCUDA() error {
 	if c.Build.CUDA != "" && c.Build.CuDNN != "" {
 		compatibleCuDNNs := compatibleCuDNNsForCUDA(c.Build.CUDA)
-		if !sliceContains(compatibleCuDNNs, c.Build.CuDNN) {
+		if !slices.ContainsString(compatibleCuDNNs, c.Build.CuDNN) {
```
bfirsh, line 210: Nice!

pkg/dockerfile/generator.go
```diff
@@ -131,9 +133,62 @@ func (g *Generator) baseImage() (string, error) {
...
+	if len(environmentVariables) > 0 {
+		envVarMap = make(map[string]string)
+		for _, v := range environmentVariables {
+			parts := strings.SplitN(v, "=", 2)
+
+			if ok := regexpVariableName.MatchString(parts[0]) && len(parts) == 2; ok {
+				envVarMap[parts[0]] = parts[1]
+				envVarKeys = append(envVarKeys, parts[0])
+			} else {
+				fmt.Printf("Ignoring invalid environment variable: %s\n", v)
```
bfirsh, line 167: I wonder if this should be a hard error?

I wonder if we can just defer to Docker for this validation, and just put blindly put in the string into `ENV`?

pkg/dockerfile/generator_test.go
```diff
@@ -68,6 +68,116 @@ FROM python:3.8
...
+	expected := `# syntax = docker/dockerfile:1.2
+FROM python:3.8
+ENV DEBIAN_FRONTEND=noninteractive
+ENV PYTHONUNBUFFERED=1
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/x86_64-linux-gnu:/usr/local/nvidia/lib64:/usr/local/nvidia/bin
+ENV TRANSFORMERS_CACHE=/src/custom_huggingface_transformers_cache
+ENV FOOBAR=foobar
+ENV XDG_CACHE_HOME=/src/$FOOBAR
+ENV RETICULATING=splines
+ENV empty=
```
bfirsh, line 180: I wonder if we could isolate testing the environment stuff with a separate function that is unit tested? If we wanted to change another part of the Dockerfile generation, we now have to update 3 additional tests.

bfirsh (changes requested):
Amazing, thanks @hangtwenty! A few comments inline.

pkg/dockerfile/generator_test.go
hangtwenty: Good call, I'll give it a refactor.

pkg/dockerfile/generator.go
hangtwenty: I'm OK with either of those options as well. I'll try them out and compare the experience

pkg/config/config.go
hangtwenty: <details><summary> For other reviewers /  posterity, here's the discussion about this from the previous PR</summary>

image

(From PR #361 )

</details>

hangtwenty: At first I felt differently about this, because `environment` in Docker Compose has its own semantics, but I've come around to agreeing that `build.environment` makes the most sense!

Why I agree now:

* `environment` looks plenty clear in `cog.yaml`; it's under `build`
*  although `env` might resemble `ENV` to me, I think it's because of handwriting lots of Dockerfiles.. but really, `env` is not `ENV`. If someone's experienced with Python (and/or Docker Compose) but not Dockerfiles, `env` might resemble/remind of `virtualenv` or `pyenv`, and that would be confusing.

pkg/dockerfile/generator.go
hangtwenty: Made it hard-error.

pkg/dockerfile/generator_test.go
hangtwenty: Refactored!

pkg/dockerfile/generator_test.go
```diff
@@ -46,29 +50,54 @@ RUN curl https://pyenv.run | bash && \
...
+	conf, err := config.FromYAML([]byte(cogYaml))
+	require.NoError(t, err)
+	require.NoError(t, conf.ValidateAndCompleteConfig())
+	return conf
+}
+
+//===============================================================================
+// tests: build.gpu
+//-------------------------------------------------------------------------------
+
```
hangtwenty, line 87: ℹ️  This diff is hard to read in "unified" diff, I'd def recommend using side-by-side or viewing the file.

pkg/config/config.go
hangtwenty: [It's resolved, but leaving it open for now for easier review.]

hangtwenty: @bfirsh I also revised the Pull Request description so that the usage example is right (`build.environment` not `build.env`)

hangtwenty:
Updated the PR description, so that it can serve as docs for advanced usage.

pkg/dockerfile/generator.go
```diff
@@ -130,10 +136,61 @@ func (g *Generator) baseImage() (string, error) {
...
+	return preamble, err
+}
+
+func (g *Generator) environmentVariables() ([]string, error) {
+	// build.environment: list of strings formatted like `KEY=<VALUE>`
+	environmentVariables := g.Config.Build.Environment
+
+	regexpVariableName := regexp.MustCompile("^[A-Za-z_][A-Za-z0-9_]*$")
+
+	var envVarKeys []string
```
bfirsh, line 158: What's the intention behind this? So it's ordered correctly?

bfirsh (approved):
Super, thanks @hangtwenty for the fixes! Docs for this would be good too.

pkg/dockerfile/generator.go
hangtwenty: That's right! (Though maybe there's a more idiomatic way to implement it in Go?) 

The order can be significant since environment variables can refer to each other. (As normal with Docker/Dockerfiles.)

zeke:
Nice! It looks like this is ready to go implementation-wise and just needs some documentation.

@hangtwenty is that something you can work on?

hangtwenty:
@zeke Sure, I'll make some time this week.

docs/yaml.md
```diff
@@ -27,6 +27,92 @@ This stanza describes how to build the Docker image your model runs in. It conta
 
 Cog automatically picks the correct version of CUDA to install, but this lets you override it for whatever reason.
 
+### `environment`
```
hangtwenty, line 30: "View File" for easier reviewing:

https://github.com/hangtwenty/cog/blob/cog-environment-variables/docs/yaml.md#environment

zeke (approved):
Love the thorough docs! Thanks for the quick turnaround on that. I left some comments inline.

docs/yaml.md
```diff
@@ -27,6 +27,79 @@ This stanza describes how to build the Docker image your model runs in. It conta
...
+An optional list of environment variables. They will be set in the Dockerfile using [`ENV` instructions](https://docs.docker.com/engine/reference/builder/#env).
+These are set in the Docker container, so your `predict.py` and imported libraries will be able to use them.
+
+For example:
+
+```yaml
+build:
+  environment:
+    - SOME_DIR=/src/example
+    - ANOTHER_DIR=$SOME_DIR/weights
```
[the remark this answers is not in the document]
hangtwenty: Good question — yes order matters. I'll clarify that.

Regarding a realistic example, I'm with you, it's a little tough in this case because I don't want to encourage people to hardcode unnecessary variables, and `XDG_CACHE_HOME` takes care of many realistic use-cases for both PyTorch and Huggingface. I thought about using other realistic examples, but I don't want it to seem like it's endorsing a specific library, since that could get stale.

But I'll try!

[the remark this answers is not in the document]
hangtwenty: OK now it's a more realistic example, while still avoiding encouraging unnecessary vars.

image

docs/yaml.md
```diff
@@ -27,6 +27,79 @@ This stanza describes how to build the Docker image your model runs in. It conta
...
+```
+
+That example would set `$SOME_DIR` to the string `/src/example` and `$ANOTHER_DIR` to `/src/example/weights`.  `DEBUG` would be set to an empty string.
+
+<details>
+<summary>Telling libraries where to cache things</summary>
+
+Cog already re-uses `/src/` across invocations; so, if we tell libraries to cache inside of `/src/`, the cached files will be persisted across invocations.
+
+Caching between runs will "just work" for some libraries, including PyTorch.
```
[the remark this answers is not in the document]
hangtwenty: Great feedback, thank you.

hangtwenty:
image

Oh, I forgot to do the DCO thing. What should I do about that?

hangtwenty:
Hmm, it's getting an ❌  for the DCO check, but I just did `git commit --amend -s` and re-pushed, and it did not resolve it. What do I do about that?

zeke:
> DCO check

I approved it manually. 🍏

zeke:
Great work, @hangtwenty!

I'll defer to @bfirsh and @andreasjansson to make a call on whether this is ready to ship.

bfirsh:
@hangtwenty Thanks for all your help getting this together. I think what is making this hard to review is there are some unrelated changes in this PR that need extra review -- the extra cache environment variables, all the extra docs about how to do caching, etc.

Perhaps we could pull out just the change that adds `environment` to `cog.yaml` and have some really minimal documentation for just that thing, then we could have separate PRs for the other changes.

No worries if you haven't got time -- we want to add this too so maybe we'll have the chance to slim this down.

hangtwenty:
> No worries if you haven't got time -- we want to add this too so maybe we'll have the chance to slim this down.

Feel free, of course! Just confirmed "allow edits by maintainers" is enabled.

bfirsh:
@hangtwenty sorry this got stuck. We'd still love to include this if you ever find the time again for a slimmed down PR. Hope you're well. :)

waffletower:
😢

Majdoddin:
Meanwhile a workaround:
https://github.com/replicate/cog/issues/291#issuecomment-2653025598

Issue #171, opened 2021-07-28
Decide on and define nouns


bfirsh:
They're a bit fuzzy currently. E.g. "image" and "model"

Let's add something to the contributing guide to clarify concepts and terminology in the project.

See also #428 

edited by @zeke

PR #407, opened 2022-02-08
Use Pydantic for type annotations

### LINKED ISSUE #58: Default values are sometimes the string "None"

### LINKED ISSUE #113: Define path types without having to import pathlib

### LINKED ISSUE #193: Revisit design of Python type annotations

### LINKED ISSUE #205: Design type system & signature

### LINKED ISSUE #212: Output types and multiple outputs

### LINKED ISSUE #246: Cog requires inputs for predict function

### LINKED ISSUE #304: Everything in predict method but `name` needs to be a named argument

### LINKED ISSUE #306: String max_length parameter on @cog.input

bfirsh:
This is an implementation of #193 and #205 to use Pydantic as the type annotations for model predictors and FastAPI for serving the model via HTTP.

Models now look a bit like this:

```python
from cog import BasePredictor, Input, Path
import torch

class ColorizationPredictor(BasePredictor):
    def setup(self):
        self.model = torch.load("./weights.pth")

    def predict(self,
          image: Path = Input(title="Grayscale input image")
    ) -> Path:
        processed_input = preprocess(image)
        output = self.model(processed_input)
        return postprocess(output)
```

This is a follow-up of #378, where the initial working version was made.

At some point, we want to lock down and define the HTTP API, but we are deferring that to a future piece of work. The Redis queue worker remains unchanged and supports the old API for Replicate to run these models.

The branch has now been moved to this repository so we can collaborate together by opening pull requests against it. We aren't putting this work in `main` because the documentation for the current version of Cog is there.

## Upgrading from Cog 0.0 to 0.1

[Note: This documentation should be copied to the release notes when we merge this.]

In Cog versions up to 0.0.20 you described inputs using annotations on your `predict` method. For example:

```py
@cog.input("image", type=Path, help="Image to enlarge")
@cog.input("scale", type=float, default=1.5, help="Factor to scale image by")
def predict(self, image, scale):
    ...
```

From Cog 0.1.0 onwards, we've started using [Pydantic](https://pydantic-docs.helpmanual.io/) to define input and output types. Rather than describing inputs using annotations, you now describe them with type hinting. Here's how you'd define the same inputs now:

```py
def predict(self,
    image: Path = Input(description="Image to enlarge"),
    scale: float = Input(description="Factor to scale image by", default=1.5)
) -> Path:
    ...
```

The parameters that `Input()` takes are pretty similar to those `@cog.input()` used to take. Here are the differences:

- It no longer takes a `type` parameter; use a type hint instead.
- The `help` parameter has been renamed to `description`.
- The `options` parameter has been renamed to `choice`.
- The `min` option has been replaced by two options:
    - `ge`: greater than or equal to (direct replacement)
    - `gt`: greater than (a new alternative)
- The `max` option has been replaced by two options:
    - `le`: less than or equal to (direct replacement)
    - `lt`: less than (a new alternative)

The other major difference is that you now need to define the output type of your model. That's the `-> Path` bit in the example above. That might be a simple type like `str`, `float` or `bool`. If you need to handle multiple outputs, check out the [new documentation for complex output objects](…). If you only have a single output, but it can be of different types depending on the run, you can use `typing.Any`:

```py
from typing import Any

def predict(self) -> Any:
    ...
```

## Todo

- [x] Disallow users from passing extra options to `cog.Input()`. #408
- [x] Make input validation errors work for `cog predict` in a simple way until we lock down HTTP API. #420 
- [x] Force users to pick an output type and throw an error if it isn't set. This is because the validation error when it isn't set is really weird, but this also seems sensible anyway. (Couldn't just be `Any`, but they need to be explicit!) #421 
- [x] Document how to migrate `@input()` to Pydantic type annotations.
- [x] Make input errors work for Redis API & Replicate #434 
  - If there is a validation error, `push_error()` in the redis worker should be called with an appropriate human-readable error.
  - This call all be done in Cog-land by writing an integration test for the redis worker.
- [x] Clean up temporary files -- both input and output: https://github.com/replicate/cog/pull/449
- [x] https://github.com/replicate/cog/pull/429
- [x] https://github.com/replicate/cog/issues/445
- [x] The name of the output type on a model Ben has locally is wrong. Needs investigation. #454 
- [x] Merge examples PR https://github.com/replicate/cog-examples/pull/8
- [x] Write release notes: https://gist.github.com/zeke/0ff121335e8c331e4343683696277ec5

## Future
- [x] #171
- [x] https://github.com/replicate/cog/issues/428
- [x] #456
- [x] What happens if you use `pathlib.Path` as an input type? Does it work? Should we throw an error? (This might happen if you import wrong package / collide imports) [It breaks! #429 prevents it being used as an input type]. See also https://github.com/replicate/cog/issues/444
- [ ] What happens if you return `pathlib.Path` from a predict function? Does this work? Should it throw an error? #456
- [ ] ~Do we want to make `cog predict` with an existing image backwards compatible? ('cos there are old models on Replicate)~ With a shim, this will just work.
- [x] https://github.com/replicate/cog/issues/427
- [x] #457
- [x] https://github.com/replicate/cog/issues/446
- [x] #459

~(These should be migrated to issues.)~ Done!

## Refs
Closes https://github.com/replicate/cog/issues/58
Closes https://github.com/replicate/cog/issues/113
Closes https://github.com/replicate/cog/issues/193
Closes https://github.com/replicate/cog/issues/205
Closes https://github.com/replicate/cog/issues/212
Closes https://github.com/replicate/cog/issues/246
Closes https://github.com/replicate/cog/issues/304
Closes https://github.com/replicate/cog/issues/306
Closes https://github.com/replicate/cog/pull/328


zeke (approved):
👏🏼 PY 👏🏼 DAN 👏🏼 TIC!

:shipit:

bfirsh:
All rebased and merge-conflict-resolved.

PR #408, opened 2022-02-08
Remove non-essential arguments to `cog.Input()`

bfirsh:
Hey look I'm opening a pull request against the future branch!

These all don't seem relevant (and in some cases undocumented?).

I have also removed support for `kwargs`, which just shows up as extra
fields in the schema. What a lovely foot gun that is. They aren't even
put behind `additionalProperties`, or something like that.

Adding support for things is much easier to take it away, so it seems to
be conservative here.


zeke (approved):
Nice cleanup!

PR #414, opened 2022-02-08
Upgrade numpy to 1.21.5

evilstreak:
Installing 1.21.1 failed locally because the binary is only available for older versions of Python.

Although the latest version of numpy is 1.22.2, the latest version that includes support for Python 3.7 is 1.21.5.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 fastapi==0.70.1
-numpy==1.21.1
+numpy==1.21.5
 pillow==9.0.0
 pydantic==1.8.2
 pytest==6.2.4


zeke (approved):
Looks good to me. 👍🏼 

On the topic of Python versions, I think I heard @andreasjansson mentioning to @dashstander yesterday that Cog support Python versions 3.6 and greater. I noticed recently that 3.6 is EOL as of two months ago. We should probably consider dropping support for 3.6 and documenting what Python versions we support.

bfirsh:
We support >= 3.7, which is a requirement for new Cog because Pydantic/FastAPI don't support 3.6. https://github.com/replicate/cog/blob/7c5d4b86c2f3f173495e7d9c35de627b8bd5ca77/.github/workflows/ci.yaml#L53

One thing we don't do is validate the Python version in `cog.yaml`: https://github.com/replicate/cog/issues/196

bfirsh:
Thanks for the fix!

PR #415, opened 2022-02-08
Document supported Python versions and display prerequisites in the README

zeke:
This PR:

- documents the Python versions we support
- adds a Prerequisites section to the README
- adds a TOC to the top of the README so you can see its contents at a glance.

Partially fixes https://github.com/replicate/cog/issues/196

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,20 @@
 
 Use Docker for machine learning, without all the pain.
 
-Cog gives you a consistent environment to run your model in – for developing on your laptop, training on GPU machines, and for other people working on the model. Then, when the model is trained and you want to share or deploy it, you can bake the model into a Docker image that serves a standard HTTP API.
+- [What is Cog?](#what-is-cog)
+- [Develop and train in a consistent environment](#develop-and-train-in-a-consistent-environment)
+- [Put a trained model in a Docker image](#put-a-trained-model-in-a-docker-image)
+- [Why are we building this?](#why-are-we-building-this)
+- [Prerequisites](#prerequisites)
+- [Install](#install)
+- [Upgrade](#upgrade)
+- [Next steps](#next-steps)
+- [Need help?](#need-help)
+- [Contributors ✨](#contributors-%E2%9C%A8)
+
+## What is Cog?
+
+Cog is an open-source command-line tool that gives you a consistent environment to run your model in – for developing on your laptop, training on GPU machines, and for other people working on the model. Once you've trained your model and you want to share or deploy it, you can bake the model into a Docker image that serves a standard HTTP API and can be deployed anywhere.
 
 Cog does a few handy things beyond normal Docker:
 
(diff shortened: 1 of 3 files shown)


bfirsh:
The version of Python on the _host_ doesn't matter. The Python versions we support are in the Docker container.

In other words, the thing that needs documenting is the available options for `python_version` in `cog.yaml`.

zeke:
Yeah good point. It's not a prerequisite for using Cog per se, more of a prerequisite for your model to actually work. I moved this around and reworded it in https://github.com/replicate/cog/pull/415#:~:text=clarify%20supported%20Python%20versions -- let me know what you think.

README.md
```diff
@@ -104,6 +117,11 @@ We realized that, in addition to Spotify, other companies were also using Docker
...
 Hit us up if you're interested in using it or want to collaborate with us. [We're on Discord](https://discord.gg/QmzJApGjyE) or email us at [team@replicate.com](mailto:team@replicate.com).
 
+## Prerequisites
+
+- **macOS or Linux**. Cog works on macOS and Linux, but does not currently support Windows.
+- **Docker**. Cog uses Docker to create a container for your model. You'll need to [install Docker](https://docs.docker.com/get-docker/) before you can run Cog.
+
 ## Install
 
 First, [install Docker if you haven't already](https://docs.docker.com/get-docker/). Then, run this in a terminal:
```
bfirsh, line 127: This is duplicated now.

README.md
```diff
@@ -2,7 +2,20 @@
 
 Use Docker for machine learning, without all the pain.
 
-Cog gives you a consistent environment to run your model in – for developing on your laptop, training on GPU machines, and for other people working on the model. Then, when the model is trained and you want to share or deploy it, you can bake the model into a Docker image that serves a standard HTTP API.
+- [What is Cog?](#what-is-cog)
```
bfirsh, line 5: This top bit is intended to be a bit like a landing page for the project, not documentation. "Develop and train in a consistent environment", "Put a trained model in a Docker image", and "Why are we building this?" are all things you might imagine being on the home page of the Cog website, were it to exist.

I wonder whether the table of contents should live under those bits, so we aren't burying it? Maybe we could change that if/when we have a website.

bfirsh (commented):
Great. Couple of comments!

README.md
zeke: I get that, but I think having a TOC up top is useful because you can see at a glance what's covered. I've always found it a little weird that the installation and usage docs are buried so deep in the page. Maybe it would look a little tighter if all the landing-page-ish content was tucked under the one "What is Cog?" TOC link?

- [What is Cog?](#what-is-cog)
- [Prerequisites](#prerequisites)
- [Install](#install)
- [Upgrade](#upgrade)
- [Next steps](#next-steps)
- [Need help?](#need-help)
- [Contributors ✨](#contributors-%E2%9C%A8)

bfirsh: Yeah, that's better. It still feels like it needs more above the fold though. We can always tweak later -- this is being rewritten in future anyway.

README.md
zeke: Yeah that's intentional. I put it in the README and the two getting started guides, so it'll be seen by people regardless of how they end up on the repo. 🤷🏼‍♂️

PR #420, opened 2022-02-10
Better error message when providing invalid inputs

evilstreak:
The new Cog internals return a 422 instead of a 400 when the input is invalid. Rather than dumping that out as a cryptic "returned a 422" message, pull out the individual validation issues from the response and present them to the user.

Adding the hint about how to provide inputs on the command line will hopefully help first time users who are running a bare `cog predict`.

## Before

Screenshot 2022-02-10 at 13 38 12

## After

Screenshot 2022-02-10 at 13 35 09

### DIFF
--- a/pkg/predict/predictor.go
+++ b/pkg/predict/predictor.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"math/rand"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
@@ -27,6 +28,14 @@ type Response struct {
 	Status status       `json:"status"`
 	Output *interface{} `json:"output"`
 	Error  string       `json:"error"`
+}
+
+type ValidationErrorResponse struct {
+	Detail []struct {
+		Location []string `json:"loc"`
+		Message  string   `json:"msg"`
+		Type     string   `json:"type"`
+	} `json:"detail"`
 }
 
 type Predictor struct {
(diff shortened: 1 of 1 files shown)


pkg/predict/predictor.go
```diff
@@ -131,18 +132,45 @@ func (p *Predictor) Predict(inputs Inputs) (*Response, error) {
...
+
+    cog predict -i blur=3.5
+
+If your input is a local file, you need to prefix the path with @ to tell Cog to read the file contents. For example:
+
+    cog predict -i path=@image.jpg`,
+			strings.Join(errorMessages, "\n"),
+		)
+
+		return nil, fmt.Errorf(validationErrorMessage)
```
bfirsh, line 173: `fmt.Errorf()` does formatting like `fmt.Sprintf()`, so this could be simplified a little by doing the formatting in `fmt.Errorf()`.

pkg/predict/predictor.go
```diff
@@ -131,18 +132,45 @@ func (p *Predictor) Predict(inputs Inputs) (*Response, error) {
...
-	// TODO
-	if resp.StatusCode == http.StatusBadRequest {
+	if resp.StatusCode == http.StatusUnprocessableEntity {
 		body := struct {
-			Message string `json:"message"`
+			Detail []struct {
+				Location []string `json:"loc"`
+				Message  string   `json:"msg"`
+				Type     string   `json:"type"`
+			} `json:"detail"`
```
bfirsh, line 141: This struct could be defined at the top level so it's reuseable, like `Request` and `Response`. This isn't important nor essential, but Go code tends to do that except for trivial types, because as soon as you want to define a function that needs to take/return this type, then it needs to be defined somewhere. It also will then show up in compile errors and your editor as `ErrorResponse` or whatever, rather than some cryptic unnamed type.

Also, another consideration -- I expect we'll use reuse struct for other error responses. It's a generic structure for FastAPI errors, I think.

Anyway -- small, non-important thing, but reviewing this intentionally verbosely for information transfer. 😄

bfirsh (approved):
Looks great! A couple of nits.

PR #421, opened 2022-02-10
Display better error when output type is undefined

evilstreak:
If an output type isn't specified, Pydantic expects the model to return nothing. When the model returns an output and it's compared to `None`, an error is logged and a 500 error is returned.

This change explicitly raises an error when the output type is undefined and raises an error message, rather than defaulting to `None`. The error message provides explanation of how to fix the issue more clearly than the original behaviour.

## Before

Screenshot 2022-02-10 at 14 29 03

## After

Screenshot 2022-02-10 at 14 48 12

### DIFF
--- a/python/tests/server/test_http_input.py
+++ b/python/tests/server/test_http_input.py
@@ -191,7 +191,7 @@ def test_extranous_input_keys():
         text: str
 
     class Predictor(BasePredictor):
-        def predict(self, input: Input):
+        def predict(self, input: Input) -> str:
             return input.text
 
     client = make_client(Predictor())
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -133,7 +133,20 @@ def get_output_type(predictor: BasePredictor):
 
     signature = inspect.signature(predictor.predict)
     if signature.return_annotation is inspect.Signature.empty:
-        OutputType = Literal[None]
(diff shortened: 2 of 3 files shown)


python/tests/server/test_http_input.py
```diff
@@ -191,7 +192,7 @@ class Input(BaseModel):
         text: str
 
     class Predictor(BasePredictor):
-        def predict(self, input: Input):
+        def predict(self, input: Input) -> Any:
```
bfirsh, line 195: In a real predictor, I suppose this would actually be `str`, but in this case the test isn't testing the output types so it doesn't matter.

bfirsh (approved):
Fantastic. Love the clear error message.

I wonder if users know they need to do `from typing import Any`. Perhaps that needs to be part of the error message.

python/tests/server/test_http_input.py
evilstreak: I've updated both tests to use `str` because it's a slightly better representation of real world usage, and it avoids needing the import in both tests.

evilstreak:
I added `from typing import Any` to the error message. Thanks!

zeke (approved):
👌🏼

zeke:
@allcontributors please add @evilstreak for code and tests

PR #426, opened 2022-02-15
Rename annotation to InputType for clarity

evilstreak:
The source of this information is the annotation, but the thing we're actually using this for is the type of the input. Name it `InputType` to make that clearer, and for consistency with the variable `OutputType` used for type checking outputs.

Refactored whilst working on limiting input types.


zeke (approved):
🍰

PR #429, opened 2022-02-16
Strictly limit the allowed model input types

evilstreak:
Allow only a short list of types for model inputs. We can extend this list in the future much more easily than restricting it.

Add a test to assert an error is raised when starting the server. Remove the test for extraneous keys, because the server won't start any more with that predictor definition.

There's a minor weirdness for the Cog types: within Cog, `Path` and `File` are available in the module `cog.types`, but when Cog is imported into another project they're in the module `cog`. We've written some special case code to handle that, but don't know enough about Python packaging to understand why there's a difference, and whether there's a better way to handle this.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -13,7 +13,10 @@ from typing_extensions import Literal, get_origin, get_args
 import yaml
 
 from .errors import ConfigDoesNotExist, PredictorNotSet
-from .types import Input
+from .types import Input, Path as CogPath, File as CogFile
+
+
+ALLOWED_INPUT_TYPES = [str, int, float, bool, CogFile, CogPath]
 
 
 class BasePredictor(ABC):
@@ -92,9 +95,14 @@ def get_input_type(predictor: BasePredictor):
     for name, parameter in signature.parameters.items():
         InputType = parameter.annotation
 
-        if not InputType:
-            # TODO: perhaps should throw error if there are arguments not annotated?
-            continue
+        if InputType is inspect.Signature.empty:
+            raise TypeError(
+                f"No input type provided for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}."
+            )
+        elif InputType not in ALLOWED_INPUT_TYPES:
+            raise TypeError(
+                f"Unsupported input type {human_readable_type_name(InputType)} for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}."
+            )
 
         # if no default is specified, create an empty, required input
         if parameter.default is inspect.Signature.empty:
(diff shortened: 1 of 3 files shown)


python/cog/predictor.py
```diff
@@ -92,9 +95,14 @@ class Input(BaseModel):
...
-        if not InputType:
-            # TODO: perhaps should throw error if there are arguments not annotated?
-            continue
+        if InputType == inspect.Signature.empty:
+            raise TypeError(
+                f"No input type provided for parameter `{name}`. Supported input types are: {get_readable_type_names(ALLOWED_INPUT_TYPES)}."
+            )
+        elif InputType not in ALLOWED_INPUT_TYPES:
+            raise TypeError(
+                f"Unsupported input type {get_type_name(annotation)} for parameter `{name}`. Supported input types are: {get_readable_type_names(ALLOWED_INPUT_TYPES)}."
```
zeke, line 104: Where does `annotation` come from?

evilstreak: Argh, thank you. That would've gone wrong in a rebase – `InputType` used to be called `annotation`.

The fact this passes CI rather suggests there's no test coverage here, so I might add one.

evilstreak: Added a test (reused the old one I'd previously removed, in fact) and fixed the error.

zeke: Nice!

python/cog/predictor.py
```diff
@@ -92,9 +95,14 @@ class Input(BaseModel):
     for name, parameter in signature.parameters.items():
         InputType = parameter.annotation
 
-        if not InputType:
-            # TODO: perhaps should throw error if there are arguments not annotated?
-            continue
+        if InputType == inspect.Signature.empty:
```
bfirsh, line 98:
```suggestion
        if InputType is inspect.Signature.empty:
```

Maybe?

bfirsh (approved):
Very nice!

Where was this set of supported types derived from? The previous types we supported, or what we support on Replicate, or something? I'm trying to think if there is other stuff we might want to support.

I wonder whether we want to support subclasses? E.g. `class MyWeirdString(str): ...`

Perhaps best to be conservative for now, and we can always add that support in the future.

python/cog/predictor.py
bfirsh: Actually yes, looks like it is this below.

evilstreak:
> Where was this set of supported types derived from?

I pulled the list from the docs on `future`: https://github.com/replicate/cog/blob/future/docs/python.md#inputkwargs

> I wonder whether we want to support subclasses? E.g. `class MyWeirdString(str): ...`
> 
> Perhaps best to be conservative for now, and we can always add that support in the future.

I'd err on the side of not building support for subclasses yet, partly because we could probably live without any version of this feature for shipping the `future` branch.

python/cog/predictor.py
```diff
@@ -159,3 +167,16 @@ class Output(BaseModel):
         __root__: OutputType
 
     return Output
+
+
+def get_type_name(t):
```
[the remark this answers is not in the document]
evilstreak: I've added a docstring, and also renamed the method to `human_readable_type_name`, which is hopefully more descriptive.

Your understanding is right. For something neither builtin nor Cog it'll use the full module (eg `collections.abc.Set`).

I'd appreciate an explainer on modules within a codebase vs modules from an import some time, so I can understand why it's `cog.types.Path` internally but `cog.Path` within models.

bfirsh:
All sounds good!

PR #431, opened 2022-02-17
support and document prereleases

zeke:
This updates the goreleaser config to support prereleases, and updates the CONTRIBUTING guide with instructions on how to create a prerelease.

https://goreleaser.com/customization/release/#github

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,6 +142,8 @@ Then visit [github.com/replicate/cog/actions](https://github.com/replicate/cog/a
 
 ### Publishing a prerelease
 
+Prereleases are a useful way to give testers a way to try out new versions of Cog without affecting the documented `latest` download URL which people normally use to install Cog.
+
 To publish a prerelease version, append a [SemVer prerelease identifer](https://semver.org/#spec-item-9) like `-alpha` or `-beta` to the git tag name. Goreleaser will detect this and mark it as a prerelease in GitHub Releases.
 
     git checkout some-prerelease-branch
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,6 +140,15 @@ To publish a new release, run the following in your local checkout of cog:
 
 Then visit [github.com/replicate/cog/actions](https://github.com/replicate/cog/actions) to monitor the release process.
 
+### Publishing a prerelease
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -140,6 +140,15 @@ To publish a new release, run the following in your local checkout of cog:
 
 Then visit [github.com/replicate/cog/actions](https://github.com/replicate/cog/actions) to monitor the release process.
 
+### Publishing a prerelease
+
+To publish a prerelease version, append a [SemVer prerelease identifer](https://semver.org/#spec-item-9) like `-alpha` or `-beta` to the git tag name. Goreleaser will detect this and mark it as a prerelease in GitHub Releases.
```
andreasjansson, line 145: What's the relationship between `-rc1`, `rc-2` above and `-alpha` and `-beta` here? Are they the same thing, like could you suffix a tag with dash-anything and it would be treated as a pre-release?

andreasjansson (approved):
Looks good to me! Maybe worth documenting in CONTRIBUTORS that prereleases do not affect the `latest` download URL which we use to install Cog?

CONTRIBUTING.md
zeke: > could you suffix a tag with dash-anything and it would be treated as a pre-release?

Yes!

From the SemVer spec:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92, 1.0.0-x-y-z.–.


I realize the `-rc1` comment is a bit confusing in contrast to the `alpha` thing mentioned in the docs, so I updated that here to be consistent: https://github.com/replicate/cog/pull/431/commits/84ddc3e847e2f86f4494ec841a466b6d4ebc3d02

zeke:
> Maybe worth documenting in CONTRIBUTORS that prereleases do not affect the latest download URL which we use to install Cog?

👍🏼

CONTRIBUTING.md
andreasjansson: 👍

Issue #198, opened 2021-08-12
Test `cog push`


bfirsh:
It's a bit of a gnarly thing to test. Maybe we want to mock `docker push` by mocking `exec`, or something? Or maybe we want to spin up a registry in the integration tests?

PR #447, opened 2022-02-28
Show a replicate.com link after pushing to r8.im

### LINKED ISSUE #210: `cog push` should give you URL to image / Replicate page after pushing

evilstreak:
Show a message after a successful push. If the push was to the Replicate registry then also show a link to the model page on replicate.com.

Screenshot 2022-02-28 at 17 51 48

(It's using `localhost:8100` as the replicate registry in development – in the wild that'd be matching against and replacing `r8.im`)

## The path not taken

I talked to @bfirsh about displaying a message from replicate.com rather than hardcoding this into Cog. Unfortunately, there's no existing call made to replicate.com after the `docker push` command has completed, and there's no way I know of to make `docker push` display more info by feeding it a suitable response from the server.

So, although it's much harder to update this when it's in Cog (because rolling out new code changes requires people to update their version), this seems like the right place to have it for now.

Fixes #210

### DIFF
--- a/pkg/cli/push.go
+++ b/pkg/cli/push.go
@@ -2,11 +2,13 @@ package cli
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/replicate/cog/pkg/config"
 	"github.com/replicate/cog/pkg/docker"
+	"github.com/replicate/cog/pkg/global"
 	"github.com/replicate/cog/pkg/image"
 	"github.com/replicate/cog/pkg/util/console"
 )
@@ -46,5 +48,14 @@ func push(cmd *cobra.Command, args []string) error {
 
 	console.Infof("\nPushing image '%s'...", imageName)
 
-	return docker.Push(imageName)
+	exitStatus := docker.Push(imageName)
+	if exitStatus == nil {
+		console.Infof("Image '%s' pushed", imageName)
+		replicatePrefix := fmt.Sprintf("%s/", global.ReplicateRegistryHost)
+		if strings.HasPrefix(imageName, replicatePrefix) {
+			replicatePage := fmt.Sprintf("https://%s", strings.Replace(imageName, global.ReplicateRegistryHost, global.ReplicateWebsiteHost, 1))
+			console.Infof("\nRun your model on Replicate:\n    %s", replicatePage)
+		}
+	}
+	return exitStatus
 }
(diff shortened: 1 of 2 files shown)


zeke (approved):
Looks good to me. Is there a way to write a test for this?

bfirsh:
"Run your model on Replicate:"?

We don't often use the word "predictions" because it sounds quite technical. And, a nit if we go with previous wording, it's usually "run predictions _on_", "not run predictions _with_".

evilstreak:
> "Run your model on Replicate:"?
> 
> We don't often use the word "predictions" because it sounds quite technical. And, a nit if we go with previous wording, it's usually "run predictions _on_", "not run predictions _with_".

I did think about you saying that as I wrote it! I decided that using "prediction" here would be OK, because it's in the context of someone pushing a model from Cog, and if they weren't pushing to replicate they'd be running a prediction using `cog predict`.

That said, I wasn't sure which way to go as I wrote it, so I'll swap it to "model".

evilstreak:
> Looks good to me. Is there a way to write a test for this?

I can't see an existing test for pushing. Which is to say, there's no prior art, but there's bound to be a reasonable way to test it. I can look into it tomorrow.

evilstreak:
To write a test for this we could:

1. Mock the `docker.Push()` call, so that it doesn't do anything and just returns a suitable exit status. This would be well suited to a unit test, though that's a style of test we don't have in this code base (yet!?)
2. Spin up a test registry locally that could receive the push. That would be well suited to an integration test, as we have now. It feels like that would be quite a lot of work.

The latter is more in keeping with the code base so far. The former feels like a more reasonable approach to me, because `docker.Push()` just shells out to Docker, and we're not really interested in testing whether Docker itself is working.

I think for mocking I'd also be inclined to refactor `cli.push()` to be smaller units, each of which could be tested in a more isolated fashion. For example, having the build part of the function reuse code that's tested in the integration tests for builds, and having the post-push-message extracted to a standalone function that could be unit tested. That'd reduce the complexity of the function that's hard to test, possibly to the point where we wouldn't think tests are necessary.

My instinct is to not take the time to add a test for this right now, but a decision on direction for the above would let us develop new features more appropriately.

evilstreak:
I've been combing through the open issues and found #198, which is about introducing tests for `cog push`!

bfirsh:
Wonderful!

bfirsh:
As discussed in standup, this is pretty trivial so probably fine to not have a test. ✨

PR #448, opened 2022-03-01
Remove `lt` and `gt` as `cog.Input` options

dashstander:
Due to [this pydantic bug](https://github.com/tiangolo/fastapi/issues/240), the `gt` and `lt` options for `cog.Input` do not work and cause the schema parsing to fail. This PR removes them and fixes the tests / documentation to not use them.

### DIFF
--- a/python/tests/server/test_http_input.py
+++ b/python/tests/server/test_http_input.py
@@ -216,7 +216,7 @@ def test_multiple_arguments():
 
 def test_gt_lt():
     class Predictor(BasePredictor):
-        def predict(self, num: float = Input(gt=3, lt=10.5)) -> float:
+        def predict(self, num: float = Input(ge=3.01, le=10.5)) -> float:
             return num
 
     client = make_client(Predictor())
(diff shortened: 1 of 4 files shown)


zeke (approved):
Ooh tricky. Nice fix.

Looks like the `ge` and `le` options are somehow NOT impacted that FastAPI bug https://github.com/uriyyo/fastapi-pagination/pull/148 🤷🏼

python/tests/server/test_http_input.py
```diff
@@ -216,18 +216,18 @@ def predict(
 
 def test_gt_lt():
     class Predictor(BasePredictor):
-        def predict(self, num: float = Input(gt=3, lt=10.5)) -> float:
+        def predict(self, num: float = Input(ge=3.01, le=10.5)) -> float:
```
zeke, line 219: This could still just be `3`, right?

bfirsh:
Another IRL conversation that @dashstander and I had briefly: `ge` and `le` are such gross non-intuitive options, so maybe this is another place where we break from the Pydantic standards and just call them `minimum` and `maximum`, consistent with OpenAPI.

Not essential for this PR if we want to do that -- we can fix in a follow-up one.

python/tests/server/test_http_input.py
dashstander: For some reason I thought it was important to maintain the test's original restrictions on having the value be greater-than-but-not-equal-to 3. But then after I had done all that I noticed that the actual test value that's passed through is 5 so.... yeah, it just remain 3. Whatever you wanna do

zeke:
@all-contributors please add @dashstander for bugs, code, and tests

PR #449, opened 2022-03-01
Clean up temporary input files

bfirsh:
This is a todo item in #407. This was much easier than expected. In short, we're adding a `cleanup()` method to the dynamically generated `Input` method that can be called to remove all of its files. It gets called after a prediction is run -- inside a try/finally so it is _always_ run.

Output files in the Redis queue were already cleaned up, it was just input files that were not. Output files are not cleaned up for the HTTP server, but it is not designed to be production-ready yet so this can be a piece of follow-up work.

@evilstreak I did a bit of exploratory work to see how complex this would be, and in the end the exploratory work just turned into the implementation. Sorry. 😂

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -215,7 +215,10 @@ class RedisQueueWorker:
         with self.capture_log(self.STAGE_RUN, prediction_id), timeout(
             seconds=self.predict_timeout
         ):
-            return_value = self.predictor.predict(**input_obj.dict())
+            try:
+                return_value = self.predictor.predict(**input_obj.dict())
+            finally:
+                input_obj.cleanup()
         if isinstance(return_value, types.GeneratorType):
             last_result = None
 
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -73,6 +73,22 @@ def load_predictor():
     return predictor_class()
 
 
+# Base class for inputs, constructed dynamically in get_input_type().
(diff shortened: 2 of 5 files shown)


python/cog/server/redis_queue.py
```diff
@@ -215,7 +215,10 @@ def handle_message(self, response_queue, message, cleanup_functions):
         with self.capture_log(self.STAGE_RUN, prediction_id), timeout(
             seconds=self.predict_timeout
         ):
-            return_value = self.predictor.predict(**input_obj.dict())
+            try:
+                return_value = self.predictor.predict(**input_obj.dict())
+            finally:
+                input_obj.cleanup()
```
bfirsh, line 221: This isn't actually tested. We don't have any unit tests so this is very hard to test. It's a very simple copy & paste of the HTTP implementation which is tested.

bfirsh, line 221: This isn't actually tested. We don't have any unit tests so this is very hard to test. It's a very simple copy & paste of the HTTP implementation which is tested.

evilstreak: This makes me wonder whether the code path for running a prediction could be abstracted in a way that both the HTTP and queue implementations could use it. Seems like a thing to try another time though!

evilstreak (approved):
Nice and simple!

python/cog/server/redis_queue.py
bfirsh: Yeah, that sounds like a sensible thing to do at some point.

PR #452, opened 2022-03-01
Document `cog.File()` and `cog.Path()`

zeke:
Does what it says on the tin.

Part of https://github.com/replicate/cog/pull/407

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -115,7 +115,7 @@ The `cog.File` object represents a _file handle_, and can be used for both input
 For models that return a `cog.File` object, the prediction output returned by Cog's built-in HTTP server will be a URL.
 
 ```python
-from cog import BasePredictor, Path
+from cog import BasePredictor, File, Path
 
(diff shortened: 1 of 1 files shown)


bfirsh:
A few notes:

1. It might worth mentioning that `File` and `Path` are how you get files in and out of models. That's kind of implied, but not explicit.
2. I wonder if it's worth wrapping the example with `class Predictor(BasePredictor):`? That might be confusing to someone not intimately familiar with Cog.
3. It's a bit confusing how the `File` example uses `Path` as an input. I reckon `File` will more commonly be used as _input_ than output. (Most libraries make it easy to read from a file handle, easy to write to disk, but most don't expose a file handle.)
4. I wonder whether we could have a more real example to illustrate what's going on? E.g.

    ```python
    from PIL import Image

        def predict(self, image: File):
             img = Image.open(image)
             ...
    ```

    (not sure if this actually works...)

zeke:
Good feedback! Made some updates. This is ready for another look.

Issue #453, opened 2022-03-01
`Input`/`Output` naming in OpenAPI schema is fragile


bfirsh:
We rely on `Input` and `Output` being in predictable places in the OpenAPI schema to determine input and output types. They should always be at `components.schemas.Input` and `components.schemas.Output`.

However, if there is another object in the schema called `Input` or `Output`, then Pydantic tries to cleverly dedupe them by prefixing them with the module name. For example, this is convoluted, but this will break because it will dedupe the `Input` object with the `Input` object defined internally in Cog:

```python
class Input(BaseModel):
  foo: string

class Output(BaseModel):
  input: Input

class Predictor(BasePredictor):
  def predict() -> Output:
    # ...
```

There are going to be less convoluted ways this will be a problem if we start having complex input objects. Either way, we should more robustly be defining `Input` and `Output` in a way that isn't dependent on the whims of Pydantic's automatic naming.

bfirsh:
An output type of `List[Output]` doesn't work for this reason.

bfirsh:
A quick fix would be a check to stop you from using objects named `Input` or `Output`.

anotherjesse:
Is this something we should test for during generation and warn the user?

bfirsh:
Yep, that's what I mean. Should be pretty simple.

PR #454, opened 2022-03-01
Only set custom Output object if it's not already called Output

bfirsh:
If you call your custom output object `Output`, then FastAPI cleverly tries to dedupe and called it `modulename_Output`. This isn't particularly helpful because the whole point is we're trying to make it consistently called `Output`.

This feels a bit fragile, and could break `Input` too. I've expressed the problem here for future fixing: https://github.com/replicate/cog/issues/453

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -178,11 +178,14 @@ For example:
     if get_origin(OutputType) is Generator:
         OutputType = get_args(OutputType)[0]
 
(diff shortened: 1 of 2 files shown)


zeke (approved):
🏁

evilstreak:
This change has broken `Any` as an output type. Reproduction of the issue here: https://github.com/replicate/cog/runs/5393017818?check_suite_focus=true#step:6:103

I'll add a proper test case and a fix.

Issue #435, opened 2022-02-22
document how to yield progressive outputs with pydantic Cog


zeke:
Most Cog models have a `predict()` method that returns a single value when the prediction is finished, but it's also possible for this function to `yield` output progressively as the function is running. This is particularly useful among image generation models.

Here's an example of a model that yields progressive output: https://github.com/dribnet/pixray/blob/f2a29ac5c355cb0c82a665a5edbd94751694e1a4/cogrun.py#L47-L50

Let's document how that works In the new [pydantic version of Cog](https://github.com/replicate/cog/releases/tag/v0.1.0-alpha).

PR #463, opened 2022-03-03
Make progressive output a list

### LINKED ISSUE #462: Design proposal: Progressive output as a list
## Background

In-progress output can be returned by using `yield` in a `predict()` function. When the function yields, the output is put on the Redis queue to return to the caller while the `predict()` function is still running.

This is useful for generative models like [pixray](https://replicate.com/pixray/text2image) which take a while to process, and users want to see the output as it is running.

The way this currently works is with a single output value that changes over time.

## Problem

Previous output gets discarded, and it is hard to track when output changes.

Due to this, displaying progressive output in Replicate is a big 'ol hack. If the output changes while the model is still running, we save that to the database and loosely keep track of what outputs get produced (with the `File.output_for_prediction` field). It’s hard to dedupe files (we have another hacky field `File.serving_url` for that) and we might miss output if we don’t poll in time (though this will be less of of a problem if we push with webhooks).

## Design

We could model progressive output as a list. Instead of each output being considered a point-in-time value, it could change over time. For example, consider these outputs changing over time in the previous design:

- `1.jpg`
- `2.jpg`
- `3.jpg`

We’d instead model it like this:

- `["1.jpg"]`
- `["1.jpg", "2.jpg"]`
- `["1.jpg", "2.jpg", "3.jpg"]`

In other words, the progressive output is considered part of the primary output, and part of the primary data model, rather than it being an effect of that data changing over time.

Some neat properties of this:

- It’s always consistent, in case data gets missed if we don’t poll fast enough or webhooks get missed
- It’s saved alongside the prediction, so wherever we display predictions the progressive output data is right there
- It kinda makes semantic sense. From a Python programmer’s point of view, a function that returns an iterator is returning a list that changes.
- For a backwards compatibility shim we're making for Replicate, we can keep track of the output changing. So instead of having to have the `serving_url` hack we can just be like `if len(new_output) > len(existing_output)`

[This is reminiscent of whether to send snapshots or incremental updates for logs](https://github.com/replicate/cog/issues/438). This would be inefficient if there are large numbers of outputs and we don’t care about the previous outputs.

Perhaps a difference with output is that progressive output is likely to be small amounts of data. And, it is primary output we care about, not just metadata.

## Implementation

- In the schema for Pydantic models, generators would be considered an `array`
- The schema for old models for backwards compatibility would *always* be `[{text: ..., file: ...}]`.
    - Most models will just return list of length 1.
    - We could be smart and figure out whether the function is a generator, but we’d then have to generate a different output schema for each model somehow. Currently, all old models have the same schema which makes the data migration really easy.
- Pydantic Cog models and the backwards compatibility shim would keep track of values that have been yielded and send a progressively growing list of outputs over the Redis API
- On Replicate, we’d just blindly store the output in `Prediction.output`. We would display outputs of list length 1 as if it was just a normal output. If the length of the list is >1, then we display a slider to skip through the outputs.

bfirsh:
An implementation of #462. This is a precursor to supporting progressive output for Pydantic Cog models on Replicate.

It's not documented, but it wasn't documented before, so I think we can defer that. #435

There is another bug fix in here with a little semantic change to the progressive output response. Details in the commit.

Closes #462

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from collections.abc import Generator
+from collections.abc import Iterator
 import enum
 import importlib
 import inspect
@@ -7,9 +7,10 @@ import os.path
 from pathlib import Path
 from pydantic import create_model, BaseModel
 from pydantic.fields import FieldInfo
+from typing import List
 
 # Added in Python 3.8. Can be from typing if we drop support for <3.8.
-from typing_extensions import Literal, get_origin, get_args
+from typing_extensions import get_origin, get_args
 import yaml
 
 from .errors import ConfigDoesNotExist, PredictorNotSet
@@ -174,11 +175,11 @@ For example:
     else:
         OutputType = signature.return_annotation
 
-    # The type that goes in the response is the type that is yielded
-    if get_origin(OutputType) is Generator:
-        OutputType = get_args(OutputType)[0]
+    # The type that goes in the response is a list of the yielded type
+    if get_origin(OutputType) is Iterator:
+        OutputType = List[get_args(OutputType)[0]]
 
-    if OutputType.__name__ != "Output":
+    if not hasattr(OutputType, "__name__") or OutputType.__name__ != "Output":
         # Wrap the type in a model called "Output" so it is a consistent name in the OpenAPI schema
         class Output(BaseModel):
             __root__: OutputType
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -65,15 +65,6 @@ def create_app(predictor: BasePredictor) -> FastAPI:
         if request:
             output_file_prefix = request.output_file_prefix
 
-        # loop over generator function to get the last result
-        if isinstance(output, types.GeneratorType):
-            last_result = None
-            for iteration in enumerate(output):
-                last_result = iteration
-                # TODO: clean up output files
-            # last result is a tuple with (index, value)
-            output = last_result[1]
-
         OutputType = get_output_type(predictor)
         Response = get_response_type(OutputType)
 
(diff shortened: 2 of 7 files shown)


python/cog/predictor.py
```diff
@@ -1,15 +1,16 @@
 from abc import ABC, abstractmethod
-from collections.abc import Generator
+from collections.abc import Iterator
```
zeke, line 2: Why the change from Generator to Iterator?

zeke (approved):
Love the branch name. Looks good to me. 🚀

python/cog/predictor.py
bfirsh: It's a subset of the functionality of `Generator` that is neater to write (`Iterator[str]` instead of `Generator[str, None, None]`).

See here: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

`Generator` is a subclass of `Iterator`.

python/cog/predictor.py
```diff
@@ -174,11 +175,11 @@ def predict(
...
 
-    # The type that goes in the response is the type that is yielded
-    if get_origin(OutputType) is Generator:
-        OutputType = get_args(OutputType)[0]
+    # The type that goes in the response is a list of the yielded type
+    if get_origin(OutputType) is Iterator:
+        OutputType = List[get_args(OutputType)[0]]
 
-    if OutputType.__name__ != "Output":
+    if not hasattr(OutputType, "__name__") or OutputType.__name__ != "Output":
```
bfirsh, line 182: This fixes a problem with `Any` that @evilstreak mentioned, I think?

evilstreak:
Yes, that’ll fix it!

bfirsh:
Pushed a minor fix for progressive files. It now calls `encode_json()` before adding to `output`, that way the filenames stay consistent and it doesn't upload unnecessary files to storage.

PR #467, opened 2022-03-07
Remove cog.Predictor

bfirsh:
We're doing a backwards incompatible change anyway, and this is actually
a neat bit of forced compatibility. If you try to use a old Cog model
with new Cog it will just fail because it can't import `Predictor`,
instead of failing somewhere deeper in the stack.

### DIFF
--- a/python/cog/__init__.py
+++ b/python/cog/__init__.py
@@ -3,14 +3,10 @@ from pydantic import BaseModel
 from .predictor import BasePredictor
 from .types import File, Input, Path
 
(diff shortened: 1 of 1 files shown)


zeke (approved):
This looks good to me.

As an alternative, is there a way to catch when someone tries to import `Predictor` and respond with a helpful error message? Maybe keep exporting `Predictor` but make it a function that says "hey you need to upgrade"?

bfirsh:
Yes! Feels like it could be a bit of follow-up work, if we find people are bumping into this?

PR #469, opened 2022-03-08
Add types to enums in the schema

bfirsh:
This is stopping `choices` work on `Input`.

This is what the pydantic docs tell us to do: https://pydantic-docs.helpmanual.io/usage/types/#enums-and-choices

We need to do multiple inheritance to make the types work. It's always
a string, but may as well use InputType in case we support other things
in the future.

Also included a related fix to add a type to Response.status.

h/t @dashstander for the report!


andreasjansson (approved):
LGTM!

PR #471, opened 2022-03-08
document how to yield progressive output

### LINKED ISSUE #435: document how to yield progressive outputs with pydantic Cog

zeke:
Resolves #435


bfirsh:
Fantastic!

PR #472, opened 2022-03-09
Fix yielding predictor cleanup bug

andreasjansson:
Inputs were being cleaned up prematurely for progressive predictors. This caused all progressive output models that take files as input to throw errors.

This fix moves input cleanup to the cleanup_functions that are run after the prediction has finished.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -211,14 +211,13 @@ class RedisQueueWorker:
             self.push_error(response_queue, e)
             return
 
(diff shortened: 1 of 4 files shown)


bfirsh:
Nice! Would it be hard to construct a test for this case? Unfortunately it'd have to be an integration test, so perhaps this is too much of an edge case to warrant an integration test. Maybe we can write some unit tests when we have a test harness for AMQP ;)

andreasjansson:
> Nice! Would it be hard to construct a test for this case?

Added test in latest commit.

Issue #265, opened 2021-09-28
Feature request: Add option to cli to output Dockerfile


ericallam:
As I was playing around with running cog I found myself curious about what actually went into the Docker image that `cog build` produces. It would be nice to be able to output the Dockerfile that `cog build` produces, as both a learning tool and the ability to "escape hatch" out of cog if necessary.

bfirsh:
cog debug dockerfile

Sssh... don't tell anyone. ;)

bfirsh:
Seriously though -- this should be documented somewhere. Thanks for the request. 😄

HitLuca:
I also wanted to take a look at the Dockerfile and found your code being what I needed. For me there wasn't any particular reason, I just wanted to see what cog does exactly :)

zeke:
Thanks for opening this issue, @ericallam. This is now documented by #474 

@all-contributors please add @ericallam for ideas.

Issue #433, opened 2022-02-18
CLI reference documentation


andreasjansson:
Cog has reference documentation for the Python API but no reference for the command line interface. The tutorials explain parts of the CLI, but it would be great to have a single page listing all of Cog's commands and their arguments.

markphelps:
This issue has been inactive for over 2.5 years, so I'm closing it as stale. The remaining CLI documentation items (the --debug flag, cog push output, and cog debug) are still welcome as contributions if anyone would like to open a PR.

PR #474, opened 2022-03-09
document `cog debug dockerfile`

zeke:
This PR updates the getting started guide with a note about how to view the dockerfile generated by Cog. A better long-term home for this information would be Cog's CLI docs, but those don't exist yet. See https://github.com/replicate/cog/issues/433

Resolves https://github.com/replicate/cog/issues/265

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -139,6 +139,12 @@ Building Docker image...
 Built resnet:latest
 ```
 
+Once you've built the image, you can optionally view the generated dockerfile to get a sense of what Cog is doing under the hood:
+
+```
+$ cog debug dockerfile
+```
+
 You can run this image with `cog predict` by passing the image name as an argument:
 
 ```


dashstander:
Looks good to me! And definitely nice to have until there are more comprehensive CLI docs.

dashstander (approved):
👍

bfirsh:
I wonder whether this might be a distraction mid-getting-started-flow? Or maybe it's helpful to understand what's going on? ("Wow look at all this complicated stuff I didn't have to do")

bfirsh:
Also this was intentionally under the `cog debug` namespace so we don't have to commit to maintaining it. (E.g. if we ever move off Dockerfiles.)

If we ever have some CLI documentation we should make it clear that there are no guarantees this will stick around.

zeke:
I've heard enough people asking how to do this on GitHub, Discord, and IRL that I think it's worth documenting, even if it does go away.

Issue #443, opened 2022-02-25
Queue worker in Cog


andreasjansson:
This is a follow-up from https://github.com/replicate/cog/issues/404, where I've tried to distill the discussion into separate issues and prioritize issues into required and future work.

## Required now

- [x] https://github.com/replicate/cog/issues/473
- [x] https://github.com/replicate/cog/issues/437
- [x] https://github.com/replicate/cog/issues/438
- [x] https://github.com/replicate/cog/issues/439

## Future
- [x] https://github.com/replicate/cog/issues/440
- [x] https://github.com/replicate/cog/issues/441
- [x] https://github.com/replicate/cog/issues/442
- [ ] Secrets management
- [ ] Handle streaming inputs
- [ ] Handle retries

sirupsen:
I'm commenting on the high-level plan here, because this feedback is high-level 👀 

I was sitting down with pen & paper thinking about how I'd design this from scratch... The big question I was asking myself: Why do the RPC in AMQP-land, rather than directly with the Cog HTTP server on _any_ of the predictors?

I don't think AMQP is a good RPC-interface to expose to users. It's too idiosyncratic. An HTTP-based API is simpler, and what we already do. HTTP should carry through from your quick local demo to production -- despite the backing mechanism changing with a CLI-flag (memory to Rabbit/Redis). 

To alleviate the barrier of entry of implementing an AMQP-based API, we'd likely get into the business of writing clients for all languages (Python, Go, Ruby, ...) and brokers (Redis, AMQP, Postgres, ..)... That's a serious amount of work. 👎🏻  It also means that we have to carefully consider the design of each broker. Of course we could make the AMQP the standard and turn that into everything else, but I think that's awkward too.

Enforcing the invariant that all communication with Cog goes through HTTP also gives us, and our users, far more leverage moving forward. They can move between Postgres, Rabbit, and Redis as the wind flies. I think we should put a hard line in the sand that users _never_ talk directly to brokers. Users always talk to the Cog HTTP server.

What I'm proposing:

![Untitled 18](https://user-images.githubusercontent.com/97400/157482002-1ac2bf81-076a-46b2-9db9-086ca052d927.png)

Now, you may find it awkward to use the HTTP server in the predictors as the API. However, it's a minimal amount of cycles to service API requests rather than predictions. I think it's simple to understand, and I think it has by far the most optionality down the line.

[...]

Issue #437, opened 2022-02-24
AMQP schema design


andreasjansson:
# Request body

- The message body is the same format as the HTTP API request body being implemented in #378. It contains the input and any other metadata.
- Message body encoded as JSON (`content_type` property set to `application/json`).
- The response queue is specified with the `reply_to` property on the message.
- Files are represented as URIs, prefixed with either `file://`, `http://`, `https://`, `s3://`, or `gs://`.
- Example request body:

```json
{
  "input": {
    "image": "s3://hooli-input-bucket/hotdog.jpg",
    "disentanglement_threshold": 5
  },
  "output_file_prefix": "s3://hooli-output-bucket/5245/"
}
```

# Response body

- The message body is the same format as the HTTP API response body. It contains the status, output, and potentially logs, metrics, and other metadata.
- Files are represented as URIs, like in the response. The model will upload them to `output_file_prefix` defined in the request.
- The response will be sent to the `reply_to` queue specified in the request.
- Message body encoded as JSON (`content_type` property set to `application/json`).
- Includes any message IDs present in the request (`message_id`, `correlation_id` properties).
- `status` can be one of: `success`, `processing`, `failed`, `cancelled`, `interrupted` (e.g. if a spot instance shut down the worker before completion)
- Depending on how we design observability features, the response body may include timing information and other performance metrics (setup time, prediction time, CPU time, memory usage, etc).
- Depending on how we design logging, the response body may include logs. For each log message, or some log message updating frequency, this message will be sent multiple times, with `status: "processing"`.
- For models that return in-progress output as as they are running, this message will be sent multiple times, with `status: "processing"` and the "output" key set to the output.
- Example request body:

```json
{
  "status": "success",
  "output": "s3://hooli-output-bucket/5245/output.jpg",
  "metrics": {
    "setup_time": <time>,
    "prediction_time": <time>
  },
  "logs": <array of strings, one per line>
}
```

## Schema

The AMQP API for a particular model is specified with a subset of the OpenAPI/JSON Schema used for the HTTP API.

It is available both as a `org.cogmodel.openapi_schema` label on the image, and at `/openapi.json` in the HTTP API.

The relevant subset looks like this for an example model that takes as inputs `image` and `disentanglement_threshold` and returns a string named `uri`:

```json
{
  "components": {
    "schemas": {
      "Input": {
        "title": "Input",
        "required": ["image", "disentanglement_threshold"],
        "type": "object",
        "properties": {
          "image": {
[...]

Issue #428, opened 2022-02-15
Pluralization of "input" and "output" in new Cog


zeke:
To pluralize, or not to pluralize? That is a question.

This issue seeks not to take a stance, but to document the state of naming input(s) and output(s) in the new [pydantic Cogsmos](https://github.com/replicate/cog/pull/407).

### In Python land (`predict.py`)

```py
from cog import Input, Output
```

It makes sense for these functions to be **singular**, because they're used one at a time:

```py
def predict(
  self,
  image: Path = Input(description="Input image"),
  blur: float = Input(description="Blur radius", default=5),
)
```

### In OpenAPI schema land

```js
components.schemas.Input
components.schemas.Output
```

`Input` or `Output` schemas can contain multiple properties, but the schemas themselves are currently singular to be consistent with all the other schemata like `Status`, `Request`, `Response`, `ValidationError` etc:

```
{
  "components": {
    "schemas": {
      "Input": {
        "type": "object",
        "title": "Input",
        "properties": {
          "username": {
            "type": "string",
            "title": "Username",
            "x-order": 0,
            "description": "GitHub username"
          },
          "accuracy": {
            "type": "integer",
            "title": "Accuracy",
            "x-order": 1,
            "description": "How accurate"
          }
        }
      },
      "Output": {
        "type": "string",
        "title": "Output"
      },
      "Status": {},
      "Request": {},
      "Response": {},
      "ValidationError": {},
      "HTTPValidationError": {}
      }
    }
  }
}
```

### In replicate.com

```py
prediction.inputs
prediction.output
```

Here is what I think is the rationale behind the current names:

`inputs` is expected to be a key-value object with one or more pairs, so it makes sense to be pluralized.
`output`, could be a string, or a complex object, or an array, so it's not clear that it should be pluralized. It might just be "one thing".

zeke:
We're sticking with the singular form ("input" and "output") for the Python Predictor API and the OpenAPI schema, as described above.

See https://github.com/replicate/cog/blob/dadab640914b15a27c818fdc152bfbb86c0b35b7/CONTRIBUTING.md#concepts

Resolved by #523

Issue #413, opened 2022-02-08
Finalize and document HTTP API


bfirsh:
The new FastAPI schema in #407 is not finalized or documented yet. We want this to be a public, stable API, so let's do a critique of it and document it.

Migrated from a todo in https://github.com/replicate/cog/pull/378.

bfirsh:
Fixed in #475.

PR #475, opened 2022-03-09
Design & document HTTP API

bfirsh:
This is a work-in-progress implementation of #413. I was considering creating an issue to discuss the design of the HTTP API, but I figure a pull request and some documentation might be the most productive medium for doing this.

The text of the documentation is incomplete and might be partial sentences. Mainly looking here for feedback on the API, then we can complete all the sentences.

I think it is mostly there, but a few design questions:

1. Do we like our request and response objects?
2. Do we like our status names? (Maybe `failure` instead of `failed`? Is there some prior art here we can build upon instead of inventing something new?)
3. #428
4. Is this compatible with #437 and #443?
5. File handling is a work in progress and I will create a separate piece of documentation about that.
6. Is this compatible with a potential future async version?
7. Do we want to version it? #94

### DIFF
--- a/docs/http.md
+++ b/docs/http.md
@@ -1,0 +1,61 @@
+# HTTP API
+
+When a Cog Docker image is run, it serves an HTTP API for making predictions.
+
+First, build your model:
+
+    cog build -t my-model
+
+Then, start the Docker container:
+
+    docker run -d -p 5000:5000 my-model
+
+    # If your model uses a GPU:
+    docker run -d -p 5000:5000 --gpus all my-model
+
+    # if you're on an M1 Mac:
+    docker run -d -p 5000:5000 --platform=linux/amd64 my-model
+
+Port 5000 is now serving the API:
+
+    curl http://localhost:5000
+
+To view the API documentation in browser for the model that is running, open [http://localhost:5000/docs](http://localhost:5000/docs).
+
+## `GET /openapi.json`
+
+The [OpenAPI](https://swagger.io/specification/) specification of the API, which is derived from the input and output types specified in your model's [Predictor](python.md) object.
+
+## `POST /predictions`
+
+Make a single prediction. The request body should be a JSON object with the following fields:
+
+- `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
+- `output_file_prefix`: A base URL to upload output files to. <!-- link to file handling documentation -->
+
+The response is a JSON object with the following fields:
+
+- `status`: Either `success` or `failed`.
+- `output`: The return value of the `predict()` function.
+- `error`: If `status` is `failed`, the error message.
+
+For example:
+
+    POST /predictions
+    {
+        "input": {
+            "image": "https://example.com/image.jpg",
+            "text": "Hello world!"
+        }
+    }
+
+Responds with:
+
+    {
+        "status": "success",
+        "output": "data:image/png;base64,..."
+    }
+
+Or, with curl:
+
+    curl -X POST -H "Content-Type: application/json" -d '{"input": {"image": "https://example.com/image.jpg", "text": "Hello world!"}}' http://localhost:5000/predictions
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ sudo chmod +x /usr/local/bin/cog
 - [Take a look at some examples of using Cog](https://github.com/replicate/cog-examples)
 - [`cog.yaml` reference](docs/yaml.md) to learn how to define your model's environment
 - [Prediction interface reference](docs/python.md) to learn how the `Predictor` interface works
+- [HTTP API reference](docs/http.md) to learn how to use the HTTP API that models serve
 
 ## Need help?
 


docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+    docker run -d -p 5000:5000 --init --gpus all my-model
+
+    # if you're on an M1 Mac:
+    docker run -d -p 5000:5000 --init --platform=linux/amd64 my-model
+
+Port 5000 is now serving the API:
+
+    curl http://localhost:5000
+
+To view the API documentation in browser for the model that is running, open [https://localhost:5000/docs](https://localhost:5000/docs).
```
zeke, line 23: Does `https` work here?

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+
+## `GET /openapi.json`
+
+The [OpenAPI](https://swagger.io/specification/) specification of the API.
```
zeke, line 27:
```suggestion
The [OpenAPI](https://swagger.io/specification/) specification of the API, which is derived from the input and output types specified in your model's [Predictor](python.md) object.
```

zeke (approved):
Great start! Left a few suggestions, and I will follow up with responses to your questions.

We get a `/docs` page for free too, right? But maybe we don't want to advertise that because it's ugly and probably doesn't work?

docs/http.md
```diff
@@ -0,0 +1,61 @@
+# HTTP API
```
sirupsen, line 1: Why not inline this in the README for now for maximum discoverability? I think fat READMEs are underrated, and I think this is high level enough to be there. I always Cmd-F for stuff

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+
+## `GET /openapi.json`
+
+The [OpenAPI](https://swagger.io/specification/) specification of the API.
+
+## `POST /predictions`
+
+Make a single prediction. The request body should be a JSON object with the following fields:
+
+- `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
```
sirupsen, line 33: Worth a sentence about the type-magic you accomplish for these signatures, it's clever and great, but bears repeating here without reading that whole doc

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+Responds with:
+
+    {
+        "status": "success",
+        "output": "data:image/png;base64,..."
+    }
+
+Or, with curl:
+
+    curl -X POST -H "Content-Type: application/json" -d '{"input": {"image": "https://example.com/image.jpg", "text": "Hello world!"}}' http://localhost:5000/predictions
```
sirupsen, line 61: Do we need to provide a timeout arg here? Often these are long and slow, no?

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+
+Then, start the Docker container:
+
+    docker run -d -p 5000:5000 --init my-model
+
+    # If your model uses a GPU:
+    docker run -d -p 5000:5000 --init --gpus all my-model
+
+    # if you're on an M1 Mac:
+    docker run -d -p 5000:5000 --init --platform=linux/amd64 my-model
```
sirupsen, line 17: Have you got this working with [M1's GPU yet](https://nod.ai/pytorch-m1-max-gpu/)?

sirupsen (approved):

docs/http.md
bfirsh: Hmm, yes. [I like long web pages too.](https://clig.dev/) We're thinking of turning this into some proper docs at some point too, so we'll have to think in context of that.

The downside of one long page is then search engines suffer. You can't google "Cog HTTP API" and get deep linked to the right place.

bfirsh:
@sirupsen @zeke Sorry I should have been clearer in the PR body: I'm mainly looking for feedback here on the actual API. The text of this documentation is incomplete (but thank you for the suggestions!).

@zeke I see answers to the questions are incoming. 😄

docs/http.md
bfirsh: Agreed. This needs expanding.

docs/http.md
bfirsh: Unfortunately I think it's very very hard to navigate through xhyve etc.

docs/http.md
bfirsh: Yes -- good idea. We have a timeout for the Redis worker, but not HTTP.

docs/http.md
bfirsh: No, no, it does not...

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+## `POST /predictions`
+
+Make a single prediction. The request body should be a JSON object with the following fields:
+
+- `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
+- `output_file_prefix`: A base URL to upload output files to. See [file handling documentation](...) for more details.
+
+The response is a JSON object with the following fields:
+
+- `status`: Either `success` or `failed`.
```
zeke, line 38: Per your question in the PR body: `failure` seems preferable here for consistency with `success`.

docs/http.md
bfirsh: I just realized Copilot wrote this for me! Not my fault! Honest!

docs/http.md
sirupsen: I'm hoping Apple invests in it.... would make the Ultra and whatever Mac Pro is to come more appealing

bfirsh: Yeah 100%.

It's starting to work in some form, but it only works if you run PyTorch and Tensorflow on the bare metal (and deal with dependency hell).

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+
+When a Cog Docker image is run, it serves an HTTP API for making predictions.
+
+First, build your model:
+
+    cog build -t my-model
+
+Then, start the Docker container:
+
+    docker run -d -p 5000:5000 --init my-model
```
zeke, line 11: Is `--init` needed here? And if so, why?

I read the [docker docs on `--init`](https://docs.docker.com/engine/reference/run/), but still don't really get what this is about.

> You can use the `--init` flag to indicate that an init process should be used as the PID 1 in the container. Specifying an init process ensures the usual responsibilities of an init system, such as reaping zombie processes, are performed inside the created container.
https://www.loom.com/share/41790390b7f1435c980bd5adddce7ff8

zeke: https://www.loom.com/share/41790390b7f1435c980bd5adddce7ff8

bfirsh:
For clarity for future reviewers: this is really for feedback on the API, not for feedback on the documentation. I have only half written the documentation. I'm particularly interested in the questions in the original post.

I opened this pull request instead of making a spec for the API because I figured it would be easier to comment on a PR than a GitHub issue.

This is also related to #443 where there are some ideas about how bi-directional HTTP might work. (Not necessarily for now, but could influence future design.)

docs/http.md
bfirsh: It's so signals work. Ctrl-C, etc.

bfirsh:
I wonder how an async API would work. It could be along the lines of:

```
POST /predictions?async=1
{
  "id": "332f6f8f39f1",
  "status": "processing",
}
```

```
GET /predictions/332f6f8f39f1
{
  "id": "332f6f8f39f1",
  "status": "success",
  "output": ...
}
```

sirupsen:
@bfirsh yeah I think something along those lines.

There's probably ~4 distinct cases to think through:

| HTTP Version | URL                      | Result                                                                                                                     |
|--------------|--------------------------|----------------------------------------------------------------------------------------------------------------------------|
| HTTP/1.1     | /predictions             | HTTP Uni-directional SSE Stream                                                                                            |
| HTTP/1.1     | /predictions?async=false | HTTP 201                                                                                                                   |
| HTTP/{2,3}   | /predictions             | [HTTP Bi-directional stream](https://developers.google.com/web/fundamentals/performance/http2#streams_messages_and_frames) |
| HTTP/{2,3}   | /predictions?async=false | HTTP 201

zeke:
> Do we like our request and response objects?

For reference, here's an example of what those objects look like today:

```json
{
  "Request": {
    "title": "Request",
    "type": "object",
    "properties": {
      "input": {
        "$ref": "#/components/schemas/Input"
      },
      "output_file_prefix": {
        "title": "Output File Prefix",
        "type": "string"
      }
    }
  },
  "Response": {
    "title": "Response",
    "required": ["status"],
    "type": "object",
    "properties": {
      "status": {
        "$ref": "#/components/schemas/Status"
      },
      "output": {
        "$ref": "#/components/schemas/Output"
      },
      "error": {
        "title": "Error",
        "type": "string"
      }
    },
    "description": "The status of a prediction."
  }  
}
```

Two things stand out to me:

1. `title` is redundant. Let's get rid of it. https://github.com/replicate/cog/issues/457
1. `description` is inconsistent. Here's a PR to fix that: https://github.com/replicate/cog/pull/498

evilstreak:
> Do we like our status names? (Maybe failure instead of failed? Is there some prior art here we can build upon instead of inventing something new?)

I've had a look, but can't find any sort of open standard here.

GitHub Actions use `success`, `failure` and `cancelled`, which works, but isn't exactly consistent. I think I'd prefer `succeeded`, `failed` and `cancelled` for consistency.

This random Microsoft status enum is very self-consistent: https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.jobstate?view=powershellsdk-7.0.0. I like the distinction between `…ed` statuses for a thing that's over and `…ing` statuses for something in progress. (One small nit: I'd probably prefer something like `Waiting` rather than `NotStarted`.

I think adopting `…ing` and `…ed` endings would make it easier to add consistent new statuses in the future. If we start with `success` and `failure` then we either end up inconsistent like GitHub Actions or we add a status like `cancellation` which feels super weird.

With `…ing` and `…ed`, if we want to add statuses for jobs that haven't started or are in progress, it's easy to find verbs: `queuing`, `waiting`, `running`.

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+        "input": {
+            "image": "https://example.com/image.jpg",
+            "text": "Hello world!"
+        }
+    }
+
+Responds with:
+
+    {
+        "status": "success",
```
evilstreak, line 55:
```suggestion
        "status": "succeeded",
```

docs/http.md
```diff
@@ -0,0 +1,61 @@
...
+## `POST /predictions`
+
+Make a single prediction. The request body should be a JSON object with the following fields:
+
+- `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
+- `output_file_prefix`: A base URL to upload output files to. See [file handling documentation](...) for more details.
+
+The response is a JSON object with the following fields:
+
+- `status`: Either `success` or `failed`.
```
evilstreak, line 38: As outlined in https://github.com/replicate/cog/pull/475#issuecomment-1070805035, I'd prefer sticking to "ed" and "ing" endings:

```suggestion
- `status`: Either `succeeded` or `failed`.
```

docs/http.md
bfirsh: Zeke says this is fixed in FastAPI now!

bfirsh:
Thanks for all your feedback here! I'm going to merge this as-is because it's a lot better than nothing, then turn the suggestions about the API into new tickets.

bfirsh:
👆 here are the tickets I made

PR #476, opened 2022-03-09
Support int choices

bfirsh:
This will probably work with Replicate, but at least the Cog side
definitely works so we can fix that later.


zeke (approved):
Looks good to me 👍🏼

andreasjansson (approved):
Nice!

PR #482, opened 2022-03-11
Handle return statements in yielding generator functions

andreasjansson:
Returning from generators is a bit of an anti-pattern, but it is
nonetheless supported in Python >=3.3.

We've already seen users return in yielding functions, so it's likely
more people will do that (even if we document that it's not allowed).

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -220,16 +220,24 @@ class RedisQueueWorker:
             return_value = self.predictor.predict(**input_obj.dict())
         if isinstance(return_value, types.GeneratorType):
             output = []
+            done = False
 
-            while True:
+            while not done:
                 # we consume iterator manually to capture log
                 try:
                     with self.capture_log(self.STAGE_RUN, prediction_id), timeout(
                         seconds=self.predict_timeout, elapsed=time.time() - start_time
                     ):
                         result = next(return_value)
-                except StopIteration:
-                    break
+                except StopIteration as e:
+                    # handle returns in generator functions. Returned values, if any, are
+                    # available in StopIteration.value
+                    if e.value is None:
+                        break
+                    else:
+                        result = e.value
+                        done = True
+
                 # Encode as JSON here instead of each time we push a result so the file URLs remain consistent
                 result = self.encode_json(result)
                 output.append(result)
--- a/test-integration/test_integration/fixtures/yielding-returning-project/cog.yaml
+++ b/test-integration/test_integration/fixtures/yielding-returning-project/cog.yaml
@@ -1,0 +1,3 @@
+predict: "predict.py:Predictor"
+build:
+  python_version: "3.8"
--- a/test-integration/test_integration/fixtures/yielding-returning-project/predict.py
+++ b/test-integration/test_integration/fixtures/yielding-returning-project/predict.py
@@ -1,0 +1,12 @@
+from typing import Iterator
+
+from cog import BasePredictor
+
+
+class Predictor(BasePredictor):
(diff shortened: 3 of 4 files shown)


python/cog/server/redis_queue.py
```diff
@@ -220,16 +220,24 @@ def handle_message(self, response_queue, message, cleanup_functions):
             return_value = self.predictor.predict(**input_obj.dict())
         if isinstance(return_value, types.GeneratorType):
             output = []
+            done = False
 
-            while True:
+            while not done:
```
zeke, line 225: > `while not done`

Looks very familiar. Has someone been reading our [new progressive output docs](https://github.com/replicate/cog/blob/12707a1609f9c390a4f87c94f08d7b2661e01abb/docs/python.md#progressive-output)? :)

zeke (approved):
Looks good to me.

What do you think about adding a note to the progressive output docs to clarify what will/won't happen when returning in a generator function, and what we recommend?

andreasjansson:
> What do you think about adding a note to the progressive output docs to clarify what will/won't happen when returning in a generator function, and what we recommend?

I wonder if it would be confusing if we started talking about returning vs yielding in the docs, since it's a very esoteric feature of Python. The examples in the docs are easy to both read and copy/paste. I doubt anyone will return in a generator unless by mistake, and this PR gracefully catches that mistake.

python/cog/server/redis_queue.py
```diff
@@ -220,16 +220,24 @@ def handle_message(self, response_queue, message, cleanup_functions):
...
-                except StopIteration:
-                    break
+                except StopIteration as e:
+                    # handle returns in generator functions. Returned values, if any, are
+                    # available in StopIteration.value
+                    if e.value is None:
+                        break
+                    else:
+                        result = e.value
+                        done = True
```
bfirsh, line 239: I wonder if there's a better way to structure this code.

Looking at some lines below, we actually send the final result twice because we can't tell if an iterator is finished unless we run the code between `yield` and the end of the function.

But, if there's a return value, we actually know for sure that the function is finished, so we can just send the `push_result(status="success")` and skip the `status="processing"` duplicate.

Maybe this would be better as calling a function that sends results under certain circumstances, or something.

Not a big deal, but might be worth adding as a comment.

bfirsh (changes requested):
Nice. This needs to be handled in `server/http.py` too.

I wonder if there could be problems if in a hypothetical future we wanted predictors to be used by something else? E.g. using a predictor right now is as simple as:

```
predictor = Predictor()
predictor.setup()
for prediction in predictor.predict():
  print(prediction)
```

Whereas now it's more complicated. 🤔

python/cog/server/redis_queue.py
andreasjansson: Yeah I was confused by the final push after the loop. What's the purpose of that?

andreasjansson:
> I wonder if there could be problems if in a hypothetical future we wanted predictors to be used by something else? E.g. using a predictor right now is as simple as:

You could say that currently the expected behavior of `return` within a generator is undefined, with the actual behavior being that we silently ignore the returned value. After this change we can still have the behavior be undefined (i.e. not documented), but we handle the returned value gracefully.

What if we just log a warning when we handle the `return`, something like

```
WARNING: The behavior of return is undefined when used inside a generator.
Please replace the return statement with yield
```

andreasjansson:
It's tricky to get right and annoying to maintain, with little value for the end user. We're better off just warning or throwing an error if someone tries to return from a generator predictor.

Closing this PR and opening https://github.com/replicate/cog/issues/483 to supersede it.

PR #485, opened 2022-03-12
Fix yielding predictor cleanup bug in HTTP server

andreasjansson:
Yielding predictors were deleting inputs before the prediction had
been run, resulting in `cog predict` being broken for progressive
output models that take files as input.

With this change, the cleanup doesn't happen until after the
prediction has completed.

Related to https://github.com/replicate/cog/pull/472

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -40,6 +40,7 @@ def create_app(predictor: BasePredictor) -> FastAPI:
     # response_model is purely for generating schema.
     # We generate Response again in the request so we can set file output paths correctly, etc.
     OutputType = get_output_type(predictor)
+    Response = get_response_type(OutputType)
 
     @app.post(
         "/predictions",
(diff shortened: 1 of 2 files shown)


python/cog/server/http.py
```diff
@@ -53,23 +54,15 @@ def predict(request: Request = Body(default=None)):
         """
         Run a single prediction on the model.
         """
-        if request is None or request.input is None:
-            output = predictor.predict()
-        else:
-            try:
+        has_input = request is not None and request.input is not None
```
bfirsh, line 57: Nit: this logic might be a bit neater if `request.input` was assigned to a variable `input`, which was by default `None`? (or a better name which isn't a global, haha)

bfirsh (approved):
Lovely! Good catch.

It feels like this logic should be consolidated in a way that both HTTP and Redis can share it, so we aren't fixing the same bugs twice all the time. I think @evilstreak mentioned this about a related problem recently.

python/cog/server/http.py
bfirsh: Nope no opinions, just an idea!

Issue #468, opened 2022-03-07
cog.yaml generated by `cog init` raises "must be a list"


andreasjansson:
Steps to reproduce:

1. `cog init`
2. `cog build`

Expected: The image is built
Actual: The following error is produced:

```
ⅹ There is a problem in your cog.yaml file. 
ⅹ must be a list.
ⅹ 
ⅹ To see what options you can use, take a look at the docs:
ⅹ https://github.com/replicate/cog/blob/main/docs/yaml.md
ⅹ 
ⅹ You might also need to upgrade Cog, if this option was added in a
ⅹ later version of Cog.
```

Removing the empty lists from cog.yaml fixes it, but that's not obvious from the error message.

Related to https://github.com/replicate/cog/issues/368
Also mentioned by https://github.com/replicate/cog/issues/424

Issue #368, opened 2021-12-28
Lists in cog.yaml should be allowed to be empty


bfirsh:
This cog.yaml:

```yaml
build:
  run:
  # - commented out
```

Throws this error:

```
ⅹ There is a problem in your cog.yaml file.
ⅹ must be a list.
ⅹ
ⅹ To see what options you can use, take a look at the docs:
ⅹ https://github.com/replicate/cog/blob/main/docs/yaml.md
ⅹ
ⅹ You might also need to upgrade Cog, if this option was added in a
ⅹ later version of Cog.
```

This should just evaluate to an empty list, so people can comment things out. (Also the error is broken.)

PR #493, opened 2022-03-15
fix invalid empty lists from `cog init`

zeke:
The `cog init` command generates a boilerplate `cog.yaml` file with commented-out stanzas. This default output includes empty lists, which are not actually valid YAML.

```yaml
# this is not valid because it's an empty list
foo:
  # - "bar"
  # - "baz"
```

This PR resolves that issue by commenting out the entire stanza, leaving it up to the user to selectively uncomment or remove the default config as needed.

Resolves https://github.com/replicate/cog/issues/468

### DIFF
--- a/test-integration/test_integration/test_build.py
+++ b/test-integration/test_integration/test_build.py
@@ -121,7 +121,7 @@ build:
 
 
 def test_build_with_cog_init_templates(tmpdir_factory, docker_image):
-    tmpdir = tmpdir_factory.mktemp("projectz")
+    tmpdir = tmpdir_factory.mktemp("project")
 
     subprocess.run(
         ["cog", "init"],
--- a/test-integration/test_integration/test_build.py
+++ b/test-integration/test_integration/test_build.py
@@ -29,8 +29,7 @@ def test_build_without_predictor(docker_image):
     assert "org.cogmodel.openapi_schema" not in labels
 
 
-def test_build_names_uses_image_option_in_cog_yaml(tmpdir_factory, docker_image):
-    tmpdir = tmpdir_factory.mktemp("project")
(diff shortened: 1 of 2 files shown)


andreasjansson (commented):
Should we add a test that runs cog init followed by cog build to make sure that `cog init` doesn't break in the future?

evilstreak (approved):

zeke:
> Should we add a test that runs cog init followed by cog build to make sure that `cog init` doesn't break in the future?

Good idea! Added a test in https://github.com/replicate/cog/pull/493/commits/8ee94cddd1755fff4dbfde8f5f3e94343b785c0f. This is ready for another look.

andreasjansson (approved):
Awesome!

test-integration/test_integration/test_build.py
```diff
@@ -118,3 +118,24 @@ def test_build_gpu_model_on_cpu(tmpdir_factory, docker_image):
         }
     }
     assert "org.cogmodel.openapi_schema" not in labels
+
+
+def test_build_with_cog_init_templates(tmpdir_factory, docker_image):
+    tmpdir = tmpdir_factory.mktemp("project")
```
andreasjansson, line 124: Nit: If you're only using one temp dir you can use the [`tmpdir`](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture) fixture.

zeke: Good catch. I updated all the other tests in that file to use `tmpdir` too: https://github.com/replicate/cog/pull/493/commits/415054b932a02637776b211c1611a4b884995cdf

bfirsh:
The problem I mentioned above is here: #368

PR #494, opened 2022-03-15
use `ge` and `le` options in predictor template

zeke:
The `lt` and `gt` options to the cog `Input` object are no longer supported: https://github.com/replicate/cog/pull/448

This PR updates the `predict.py` template generated by `cog init` to use the `ge` and `le` options instead.


andreasjansson (approved):
👍

evilstreak (approved):

PR #495, opened 2022-03-15
Fix int enums

andreasjansson:
Subclassing InputTypeEnum from `(int, enum.Enum)` caused validation
error in both the HTTP and Redis workers. Interestingly, this wasn't
caught by the existing `test_choices_int` test in test_http_input.py.

image


evilstreak (approved):
Nice find!

PR #498, opened 2022-03-16
make the description consistent in Request and Response schemas

zeke:
The multi-line `"""heredocs"""` we define in the Request and Response objects become the `description` values in the OpenAPI schema. This updates those descriptions for consistency.

Toward #413 and #475 

---

Screen Shot 2022-03-16 at 11 38 42 AM


evilstreak (approved):
I like it 👍

PR #503, opened 2022-03-21
document input and output types

zeke:
This PR updates the Python API reference with a new standalone section for `## Input and output types`. This makes it clear that this list of supported types applies not only to input, but output as well. This change also makes it possible to deep link to the list of supported types in the documentation.


evilstreak (approved):
Nice!

A few thoughts/questions about outputs as I read it (some of this isn't about the changes you've made, and none of this should stop us merging what's here):

- Is it worth pointing out that each of the object's properties must be one of the allowed output types?
 
- Do we support arbitrarily nested objects?

- Should we show an example of the JSON output generated by a simple output vs a complex output?

zeke:
Added some more clarification and example outputs based on @evilstreak's suggestion.

As for the question of supporting arbitrarily nested objects, I'm not sure. Punting on that for now.

PR #504, opened 2022-03-22
Update compatibility matrices

### LINKED ISSUE #204: Ensure we support latest versions of tensorflow and torch

bfirsh:
This adds support for Tensorflow 2.8 and PyTorch 1.11. The real change and the main review is of the JSON file in the last commit. Also some additional changes to the parser in here, which was the means to get that JSON file correct. Detail in the commit messages.

Closes #204


zeke:
Let's expand `tf` into `tensorflow...`

zeke (approved):
😅

PR #514, opened 2022-03-28
Fix output documentation

bfirsh:
`Output` is a thing the user defines, not a built-in Cog thing. I'm not sure the best hierarchy for this in the document, but this is clearer than it was before.

I have removed the API documentation because this is confusing for people trying to implement models for Replicate. Perhaps this should live in some separate API documentation?

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -11,7 +11,7 @@ Tip: Run [`cog init`](getting-started-own-model#initialization) to generate an a
   - [`Predictor.predict(**kwargs)`](#predictorpredictkwargs)
     - [Progressive output](#progressive-output)
 - [`Input(**kwargs)`](#inputkwargs)
-- [`Output(BaseModel)`](#outputbasemodel)
+- [Output](#output)
 - [Input and output types](#input-and-output-types)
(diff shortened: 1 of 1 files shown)


zeke (approved):
Okay with me.

docs/python.md
```diff
@@ -103,7 +103,7 @@ The `Input()` function takes these keyword arguments:
 
 Each parameter of the `predict()` method must be annotated with a type like `str`, `int`, `float`, `bool`, etc. See [Input and output types](#input-and-output-types) for the full list of supported types.
 
-## `Output(BaseModel)`
+## Output
```
zeke, line 106: I was on the fence about these headings when first writing this doc, but I like the simplicity of the change you've made here. Maybe we should we do this for `Input(**kwargs)` -> `Input` too?

bfirsh: I think it's fine how it is. This means "Output" as in, the concept of output. `Input(**kwargs)` is about the Python function `Input()`.

Maybe the structure of the document could clarify this. It's part explaining, part API reference.

PR #516, opened 2022-03-29
add `-p` to `cog run` to forward ports

### LINKED ISSUE #357: Add `-p` to `cog run` to forward ports
As mentioned in #123, this should do the right thing:

```
cog run -p 8080:8080 jupyter notebook --ip 0.0.0.0 --no-browser --port=8080 --allow-root
```

It should behave the same as `docker run`'s `-p`.

## Future

Not essential to do from day 1, but we might also benefit from a slight change in behavior. If you only pass one port to Docker (`-p 8080`), Docker will pick a random high port on the host machine. This is normally not what the user wants to do when running programs in development -- they want the port to be the same, known port, because the server inside will display something like `Running on http://localhost:8080` and that should work. And they're usually just running one thing so they won't collide.

So, perhaps `-p 8080` could expand to `-p 8080:8080`. We might need some way to get back the previous behavior, e.g. `-p 8080:` or `-p 8080:auto`.

zeke:
Resolves #357 

- [x] add example to help text for `cog run`
- [x] ~~add tests?~~ smoke tested with https://github.com/zeke/github-avatars/commit/b301f904d7c09748ec99ad5b2abebf81cee42ae1

### Notes

- Currently we shell out to the docker CLI, but ideally we'd call the Docker API directly.
- Allow `cog run` to pass flags arbitrarily instead of -p specifically? (No)
- Use something like `NextFreePort` to support `cog run notebook` without specifying a port?

🍐 @evilstreak

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"strconv"
 	"strings"
 
 	"github.com/replicate/cog/pkg/config"
@@ -8,6 +9,10 @@ import (
 	"github.com/replicate/cog/pkg/image"
 	"github.com/replicate/cog/pkg/util/console"
 	"github.com/spf13/cobra"
+)
+
+var (
+	runPorts []string
 )
 
 func newRunCommand() *cobra.Command {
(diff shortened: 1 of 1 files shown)


bfirsh:
> Use something like NextFreePort to support cog run notebook without specifying a port?

This would be ideal, but just FYI this is hard unless the container specifies what port it is exposing, which it is not in this case.

zeke:
@bfirsh and I paired on this today and got it into a working state.

This implementation differs a bit from the `docker run` behavior in that it supports **a single number** as input which is used for both the container and the host:

- 🍏 Yes: `cog run -p 8000 jupyter notebook ...` 
- 🍏 Yes: `cog run -p 8000 -p 8001 -p 8002 my_process ...` 
- 🍎 No: `cog run -p 8000:8000 ...`

This is ready for eyes.

bfirsh:
> This implementation differs a bit from the docker run behavior in that it supports a single number as input

... with the intention that we'll add support `x:y` and the various variations over time.

bfirsh:
Smashing!

PR #517, opened 2022-03-30
Add EXPOSE to Dockerfile

bfirsh:
It doesn't actually do anything, but a useful bit of metadata about the
container to explain what port is exposed.

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -100,8 +100,8 @@ func (g *Generator) GenerateBase() (string, error) {
 		pythonRequirements,
 		pipInstalls,
 		run,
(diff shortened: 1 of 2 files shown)


zeke (approved):
From https://docs.docker.com/engine/reference/builder/#expose

> The `EXPOSE` instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published

:shipit:

bfirsh:
A bit of nuance: this documentation is a lie if you have one-off commands or queue workers, but it's correct for the default case of running the HTTP server!

bfirsh:
Hrm, real failures I think.

PR #528, opened 2022-04-04
Make it clear what is incomplete

bfirsh:
Something that is kind-of missing here is that the Redis API is not just
undocumented, it's also not stable. I wonder whether we need to mention
that?


zeke:
> it's also not stable. I wonder whether we need to mention that?

Let's ship this as-is and cross that bridge when we get to it.

PR #532, opened 2022-04-04
Rename run.cog.cog_version label to run.cog.version

bfirsh:
Looking at them in `docker inspect` output the former just looks really
clumsy. Now that "cog" and "version" are next to each other, it's pretty
obvious that it refers to Cog.


zeke (approved):
👌🏼

PR #534, opened 2022-04-04
Only allow one of python_packages or python_requirements to be set

bfirsh:
A little bit of preparatory work for #157 I did last week. We want to use the same codepath for backwards compatibility for `python_packages`, and this'll be much simpler if only one of them is set.


zeke (commented):
This change looks good, but CI is failing because some of the fixtures in `generator_test.go` use both `python_requirements` and `python_packages`.

bfirsh:
Tests fixed!

Issue #496, opened 2022-03-16
Design: file handling


bfirsh:
[As was started to explore here](https://github.com/replicate/cog/pull/343#issuecomment-1005107133), we need a way to upload and download files over the [HTTP API](https://github.com/replicate/cog/pull/475) and the [queue-based API](https://github.com/replicate/cog/issues/443).

## Requirements
- Input files need to be sent to a prediction in the API request directly, from a separate HTTP location, or from cloud storage.
- Output files need to be sent back to the caller of the API in the response directly, by uploading to another HTTP endpoint, or by uploading to cloud storage.
- Multiple output files need to be stored for each prediction.
- This should work for both HTTP and queue-based APIs in the same way.

## Proposed design

Both input and output files are represented as URLs in the input and output objects.

Supported URL schemas:

- `data:` A data URL for passing files directly in the request/response. This is inefficient, but our only option for transferring binary blobs via JSON.
- `http:` An HTTP URL
- `gs:` Google Cloud Storage
- `s3:` Amazon S3

By default, data URLs are used. For example, a basic model that takes `image` as input and returns an altered image:

Request:
```
POST /predictions
{
  "input": {
    "image": "data:image/jpeg;base64,..."
  }
}
```
Response:
```
{
  "output": "data:image/jpeg;base64,..."
}
```

## Output file paths

For `http:`, `gs:`, and `s3:`, we need some way of hinting to the model of where to put output files.

Prediction requests have an `output_file_prefix` which defines the URL prefix to upload files to. The model can choose whatever arbitrary paths it wants to upload files within that prefix.

For example, an S3 version of the request above:

Request:
```
POST /predictions
{
  "input": {
    "image": "s3://bucket/prediction-input/25b743baec7cc6/input.jpg"
  },
  "output_file_prefix": "s3://bucket/prediction-output/25b743baec7cc6/"
}
```
Response:
```
{
  "output": "s3://bucket/predictions/25b743baec7cc6/output.jpg"
}
```

An alternate design could be to specify a base URL at runtime, then files are uploaded to some prediction ID or subpath for each prediction. For example, the model could be booted with `

## HTTP

HTTP is a little more nuanced than the cloud storage, because we need to design an API for the endpoint for uploading files.

We could make it behave like the cloud storage prefix, then files are `PUT` at subpaths of that prefix. For example, a request could look like this:

```
{
  "input": {
    "image": "https://..."
  },
[...]

PR #536, opened 2022-04-05
Future 0.3.0

### LINKED ISSUE #437: AMQP schema design

### LINKED ISSUE #520: Rename `success` status to `succeeded`
See @evilstreak's excellent analysis here: https://github.com/replicate/cog/pull/475#issuecomment-1070805035

This will require a minor bit of backwards compatibility in Replicate to check for both statuses. (Or `s.startswith("succe")`, said in an Italian accent, as @zeke and I joked).

### LINKED ISSUE #546: Document Redis API
- #437
- #536

evilstreak:
This PR includes several (mostly breaking) changes that implement some of the open issues about the API (both HTTP and queueing):

- Renames the `success` status to `succeeded` (closes #520)
- Implements the new response format for the queue API (closes #437)
- Adds the `logs` property to the queue response (see #547)
- [x] Documents the queueing API (see #549, closes #546)

We also need to ensure Replicate will work with the new version, so will need to release a new version of our serving system that can handle the new response format from the queuing API.

- [x] Update the Replicate serving system for 0.3.0 compatibility

This work does not include the new file handling design (see #496). We'll pick that up in a future piece of work.


bfirsh:
Looks great so far. I think `success` -> `succeeded` needs changing in the HTTP docs too.

zeke:
Yes yes. Looking good.

bfirsh:
All seems to work on a local build!

PR #538, opened 2022-04-05
Add support for numpy arrays as input/output

bfirsh:
This does not work, but here are the failing test cases demonstrating what should work.

Some wrangling of Pydantic validators and encoders is needed. We created a custom `encode_json()` method to make it easier to convert custom outputs to JSON. I wonder if we can get inputs to work without doing some custom wrapper like this.

Some guidance if anyone wants to grab this:
- We'd probably want to set `arbitrary_types_allowed` on the output object. (See Pydantic docs for it.)
- https://github.com/samuelcolvin/pydantic/issues/951 for ideas.

### DIFF
--- a/python/tests/server/fixtures/input_ndarray.py
+++ b/python/tests/server/fixtures/input_ndarray.py
@@ -1,0 +1,7 @@
+from cog import BasePredictor
+import numpy as np
+
+
+class Predictor(BasePredictor):
+    def predict(self, array: np.ndarray):
+        assert array == np.array([[1, 2], [3, 4]])
--- a/python/tests/server/test_http_input.py
+++ b/python/tests/server/test_http_input.py
@@ -252,6 +252,23 @@ def test_secret_str(client, match):
     assert resp.status_code == 422
 
 
+@uses_predictor("input_secret")
+def test_secret_str(client, match):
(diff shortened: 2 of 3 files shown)


mattt:
I don't want to get all "worse is better", but `List[Any]` is probably sufficient for the vast majority of use cases. The performance benefit for serializing directly to an `ndarray` is negligable, and the current approach wouldn't do much more to communicate the expected shape of the input or output than a `description` field. 

If we wanted to do this right, we could support arbitrary types defined by [pydantic-numpy](https://github.com/caniko/pydantic-numpy) (or manually defined with `Annotated`) like this:

```python
NpStrict1DArrayInt64 = Annotated[
    np.ndarray[tuple[int], np.dtype[np.int64]],
    NpArrayPydanticAnnotation.factory(data_type=np.int64, dimensions=1, strict_data_typing=True),
]

def predict(array: NpStrict1DArrayInt64):
    ...
```

That is blocked by #1687.

For now, I don't think there's much we can do with this PR, so I'm going to close it for now.

bfirsh:
`Any` types might help with this too!

My dream is arbitrary JSON encodable types in/out as the lowest level thing, then more specific types built on top of that. We've got that for outputs, but not inputs yet. :D

Issue #527, opened 2022-04-04
Array as inputs and outputs using Input method


simonMoisselin:
Hello,

Is it possible to add the type `array` as an input, in addition to images or files?

zeke:
Hi @simonMoisselin 👋🏼 

Arrays/lists as inputs and outputs are not currently supported in Cog, but this seems to me like a reasonable thing to support. Can you share a little about your intended use case?

For a list of types that are currently supported, see the docs on [Input and output types](https://github.com/replicate/cog/blob/684c735335d8a51676ab6b611d8535572ae8dd93/docs/python.md#input-and-output-types).

See also this open issue about supporting multiple file outputs: https://github.com/replicate/cog/issues/411

dashstander:
I think this is a great idea. I don't know necessarily if this would work the best (in particular it might be somewhat extraneous to using models from replicate.com), but would be awesome to support [Arrow](https://arrow.apache.org/docs/index.html) for serializing and de-serializing array data.

anotherjesse:
This is still the case - we do not yet support lists/arrays of input:

```
TypeError: Unsupported input type typing.List[float] for parameter `nums`. Supported input types are: str, int, float, bool, cog.File, cog.Path.
ⅹ Failed to get container status: exit status 1
```

jayaffine:
This would be a great idea to support different formats.
I wanted to host some derivatives of Segment Anything Model  where the results  are in `dict` format.
So the results will have numpy arrays, boolean arrays, etc.
It would be great if I can pass these dict files as results.

> Hi @simonMoisselin 👋🏼
> 
> Arrays/lists as inputs and outputs are not currently supported in Cog, but this seems to me like a reasonable thing to support. Can you share a little about your intended use case?
> 
> For a list of types that are currently supported, see the docs on [Input and output types](https://github.com/replicate/cog/blob/684c735335d8a51676ab6b611d8535572ae8dd93/docs/python.md#input-and-output-types).
> 
> See also this open issue about supporting multiple file outputs: #411

bfirsh:
Related / dupe of https://github.com/replicate/cog/issues/608

alexblattner:
hi, I'd like to be able to upload multiple pictures for a single argument. This would be very useful for that

camenduru:
Hi @zeke 👋 Dust3r ( https://replicate.com/camenduru/dust3r ) needs multiple images as input.

yosun:
[...]

PR #542, opened 2022-04-11
experiment: support List[str] as an input type

zeke:
Nothing to see here yet. This is an experiment Inspired by https://github.com/replicate/cog/issues/527

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -17,7 +17,7 @@ from .errors import ConfigDoesNotExist, PredictorNotSet
 from .types import Input, Path as CogPath, File as CogFile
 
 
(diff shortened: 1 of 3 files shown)


zeke:
Using `List[str]` as an input type annotation appears to produce a correct-looking OpenAPI schema:

```
{
  "Input": {
    "title": "Input",
    "required": [
      "fruits"
    ],
    "type": "object",
    "properties": {
      "fruits": {
        "title": "Fruits",
        "type": "array",
        "items": {
          "type": "string"
        },
        "description": "array of fruit strings",
        "x-order": 0
      }
    }
  }
}
```

but `cog build` fails with this error:

```
ⅹ Failed to get type signature: invalid character 'I' looking for beginning of value
```

Issue #539, opened 2022-04-11
document that `cog.Input()` is not required when defining a predictor


zeke:
I was under the impression that using [`cog.Input()`](https://github.com/replicate/cog/blob/main/docs/python.md#inputkwargs) was **required** when defining parameters to the `predict` method in pydantic Cog predictors. But then I noticed in our integration tests that a number of our test predictors don't use it. For example:

This predictor is valid:
```py
from cog import BasePredictor

class Predictor(BasePredictor):
    def predict(self, input: str) -> str:
        return "hello " + input
``` 

and produces this schema:

```json
{
  "components": {
    "schemas": {
      "Input": {
        "title": "Input",
        "required": ["input"],
        "type": "object",
        "properties": {
          "input": {
            "title": "Input",
            "type": "string",
            "x-order": 0
          }
        }
      },
      "Output": {
        "title": "Output",
        "type": "string"
      }
    }
  }
}
```

It seems like this should be documented, so users know that `Input` is optional.

Is there some reason why we would want to discourage users from defining simple inputs that don't use `cog.Input()`?

zeke:
Looks like the default value is handled here: https://github.com/replicate/cog/blob/684c735335d8a51676ab6b611d8535572ae8dd93/python/cog/predictor.py#L124-L131

bfirsh:
> Is there some reason why we would want to discourage users from defining simple inputs that don't use cog.Input()?

Nope, and this is intentional! Philosophically, we're trying to lean on vanilla Python as much as possible and only use these functions if additional information is required. `input: str`, `input: str = "default"` should all work. Same as Pydantic.

PR #544, opened 2022-04-11
docs: `cog.Input()` is optional

zeke:
Resolves https://github.com/replicate/cog/issues/539

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -103,6 +103,19 @@ The `Input()` function takes these keyword arguments:
 
 Each parameter of the `predict()` method must be annotated with a type like `str`, `int`, `float`, `bool`, etc. See [Input and output types](#input-and-output-types) for the full list of supported types.
 
+Using the `Input` function is recommended because it provides better documentation and validation constraints to the users of your model, but it is not strictly required. You can also specify default values for your parameters using plain Python, or omit default assignment entirely:
+
+```py
+class Predictor(BasePredictor):
+    def predict(self, 
+        prompt: str = "default prompt", # this is valid
+        iterations: int                 # also valid
+    ) -> str:
+        for x in (iterations or 1:)
+            return "hello, " + prompt
+
+```
+
 ## Output
 
 Cog predictors can return a simple data type like a string, number, float, or boolean. Use Python's `-> <type>` syntax to annotate the return type.
--- a/docs/python.md
+++ b/docs/python.md
@@ -111,9 +111,7 @@ class Predictor(BasePredictor):
         prompt: str = "default prompt", # this is valid
         iterations: int                 # also valid
     ) -> str:
-        for x in (iterations or 1:)
-            return "hello, " + prompt
-
+        # ...
 ```
 
 ## Output
--- a/docs/python.md
+++ b/docs/python.md
@@ -103,7 +103,7 @@ The `Input()` function takes these keyword arguments:
 
 Each parameter of the `predict()` method must be annotated with a type like `str`, `int`, `float`, `bool`, etc. See [Input and output types](#input-and-output-types) for the full list of supported types.
 
-Using the `Input` function is recommended because it provides better documentation and validation constraints to the users of your model, but it is not strictly required. You can also specify default values for your parameters using plain Python, or omit default assignment entirely:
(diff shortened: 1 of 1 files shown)


docs/python.md
```diff
@@ -103,6 +103,19 @@ The `Input()` function takes these keyword arguments:
...
+Using the `Input` function is recommended because it provides better documentation and validation constraints to the users of your model, but it is not strictly required. You can also specify default values for your parameters using plain Python, or omit default assignment entirely:
+
+```py
+class Predictor(BasePredictor):
+    def predict(self, 
+        prompt: str = "default prompt", # this is valid
+        iterations: int                 # also valid
+    ) -> str:
+        for x in (iterations or 1:)
+            return "hello, " + prompt
```
evilstreak, line 115: Should this be `yield`, for the spirit of the example? As is, the number of iterations won't matter.

docs/python.md
```diff
@@ -103,6 +103,19 @@ The `Input()` function takes these keyword arguments:
...
+
```
evilstreak, line 116: Spare line here

evilstreak (changes requested):
I like the docs improvement 👍

I think the example needs a few minor changes.

docs/python.md
```diff
@@ -103,6 +103,19 @@ The `Input()` function takes these keyword arguments:
...
 
+Using the `Input` function is recommended because it provides better documentation and validation constraints to the users of your model, but it is not strictly required. You can also specify default values for your parameters using plain Python, or omit default assignment entirely:
+
+```py
+class Predictor(BasePredictor):
+    def predict(self, 
+        prompt: str = "default prompt", # this is valid
+        iterations: int                 # also valid
+    ) -> str:
+        for x in (iterations or 1:)
```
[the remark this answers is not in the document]
bfirsh: 👍

`(iterations or 1:)` [sic -- should be `):`] is also not really Pythonic. Python prefers clear, explicit things rather than clever one liners. It also wouldn't work in the case where it is `0`, because that is falsy.

bfirsh (approved):
Looks great!

docs/python.md
```diff
@@ -103,6 +103,19 @@ The `Input()` function takes these keyword arguments:
 
 Each parameter of the `predict()` method must be annotated with a type like `str`, `int`, `float`, `bool`, etc. See [Input and output types](#input-and-output-types) for the full list of supported types.
 
+Using the `Input` function is recommended because it provides better documentation and validation constraints to the users of your model, but it is not strictly required. You can also specify default values for your parameters using plain Python, or omit default assignment entirely:
```
bfirsh, line 106: Not a big deal, but I'm not sure it is recommended? You get validation just with plain Python because of the type annotations. I really like the fact that it's Just Python Type Annotations, because it makes it understandable. Then, you can layer on the more complicated `cog.Input()` thing on top of that if you need extra annotation.

This really is a very minor point though so maybe over time we can iterate on this.

bfirsh: E.g. look at the headline examples on pydantic and fastapi:

https://pydantic-docs.helpmanual.io/
https://fastapi.tiangolo.com/

They don't use any of their clever default functions anywhere, and it makes it look more like simple, understandable, vanilla Python.

docs/python.md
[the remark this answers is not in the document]
zeke: Yeah I knew this was a bad example! Couldn't think of a great way to include a parameter that doesn't have a default value, but still do something sensible with it. Maybe the body of the function should just be `# ...` to avoid distraction. The import bit is really just illustrating the way params can be written.

docs/python.md
zeke: I think it's recommended primarily because we want the descriptions to propagate through to the OpenAPI schema, to API clients, to CLIs, to docs, for the benefit of end users.

docs/python.md
zeke: Removed the function body for clarity ad75f4d6d0e5751254c05b0ea1f192b768fa2fa9

docs/python.md
zeke: Changed the wording a bit here and removed the "recommended" bit.

bfirsh: Yeah, I suppose encouraging people to write descriptions is a good reason to encourage `Input`.

PR #547, opened 2022-04-12
Add logs to queueing API responses

evilstreak:
There's no easy way to do this with incremental changes, so one large change plus test updates all in one is how we're doing it. More details in the commit messages (particularly the first one: e3dc51f4ef79b00e04bdeb7ea7c4143384e23eee).

Implements the logging changes in #437.

The second commit (976931316a93fa70aad4fb95811653221b6444e8) raises a question: should empty values in our responses have keys? Would we prefer:

```json
{
  "status": "failed",
  "error": "oops"
}
```

or:

```json
{
  "status": "failed",
  "output": null,
  "logs": [],
  "error": "oops"
}
```

Given we want to include output or logs if they exist, I prefer always including the keys, which is the approach this change takes. It would be a pretty minor change to adopt the former example instead.


test-integration/test_integration/fixtures/failing-project-after-output/predict.py
```diff
@@ -0,0 +1,12 @@
+import time
```
zeke, line 1: Maybe it doesn't matter, but this project name `failing-project-after-output` differs from all the others which end with `-project`. This could be `failing-after-output-project` for consistency.

test-integration/test_integration/test_redis_queue.py
```diff
@@ -208,23 +210,78 @@ def test_queue_worker_yielding(docker_network, docker_image, redis_client):
...
+        interactive=True,
+        network=docker_network,
+        command=[
+            "python",
+            "-m",
+            "cog.server.redis_queue",
+            "redis",
+            "6379",
+            "predict-queue",
+            "",
```
zeke, line 245: What's this empty string for?

zeke (approved):
I'll leave the critical technical reviewing to @andreasjansson and @bfirsh, but this looks good to me. I love the code style, docstrings, comments, etc. Easy to follow. 🌟

test-integration/test_integration/fixtures/failing-project-after-output/predict.py
evilstreak: Good catch, thank you! Fixed in 6a4551b

test-integration/test_integration/test_redis_queue.py
evilstreak: It's an endpoint for Cog to upload files to. It's set to something other than an empty string in tests that deal with file uploads. (TBH, I just copy/pasted this whole block from the other `worker_error` test!)

andreasjansson (commented):
Unfortunately I think multiprocessing might break a lot of models, especially when using CUDA. If there's no way around using multiprocessing we should at least try to manually upgrade and run predictions on a few GPU models.

python/cog/server/runner.py
```diff
@@ -0,0 +1,135 @@
...
+        self._is_output_generator = None
+
+        # We haven't encountered an error yet
+        self._error = None
+
+        self.predictor_pipe_reader, predictor_pipe_writer = multiprocessing.Pipe(
+            duplex=False
+        )
+        self.error_pipe_reader, error_pipe_writer = multiprocessing.Pipe(duplex=False)
+        self.predictor_process = multiprocessing.Process(
```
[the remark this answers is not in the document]
evilstreak: I've reworked this in 7b6ce0a to run the predictor in a long running subprocess, where it does `setup()` and then multiple `predict()` calls.

python/cog/server/log_capture.py
```diff
@@ -0,0 +1,116 @@
+import sys
```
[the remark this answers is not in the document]
evilstreak: Removed it!

PR #549, opened 2022-04-13
Document the Redis queue API

evilstreak:
A first version of docs for the queue API. We'll probably want to revisit these after folks start using them for real.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -1,5 +1,97 @@
-# Redis queue API reference
+# Redis queue API
 
 Long-running deep learning models or batch processing is best architected as a queue. Cog has a built-in queue worker that can process predictions from a Redis queue, and return the output back to the queue.
 
-It's currently undocumented, but documentation is coming soon!
+The request queue is implemented with [Redis streams](https://redis.io/topics/streams-intro).
+
+## Start up the model
+
+The entrypoint to run a Cog model against a queue is `cog.server.redis_queue`. You need to provide the following positional arguments:
+
+- `redis_host`: the host your redis server is running on.
+- `redis_port`: the port your redis server is listening on.
+- `input_queue`: the queue the Cog model will listen to for prediction requests. This queue should already exist.
+- `upload_url`: the endpoint Cog will `PUT` output files to. <!-- link to file handling documentation -->
+- `consumer_id`: The name the Cog model will use to identify itself in the Redis group (also called "consumer name" by Redis).
+- `model_id`: a unique ID for the Cog model, used to label setup logs.
+- `log_queue`: the queue the Cog model should send setup logs to (prediction logs are sent as part of prediction responses).
+
+Note: logging is changing as part of 0.3.0, so the `model_id` and `log_queue` arguments are likely to change soon.
+
+You can optionally provide the following positional argument:
+
+- `predict_timeout`: the maximum time in seconds a prediction will be allowed to run for before it is terminated.
+
+For example:
+
+    docker run python -m cog.server.redis_queue \
+        redis 6379 my-predict-queue \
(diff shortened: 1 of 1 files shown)


docs/redis.md
```diff
@@ -1,5 +1,94 @@
...
+
+Note: logging is changing as part of 0.3.0, so the `model_id` and `log_queue` arguments are likely to change soon.
+
+You can optionally provide the following positional argument:
+
+- `predict_timeout`: the maximum time in seconds a prediction will be allowed to run for before it is terminated.
+
+For example:
+
+    docker run python -m cog.server.redis_queue \
```
zeke, line 27: Not sure if commentary on the API/usage itself is out of scope for this PR since we're just dealing with docs here, but I'm curious why these are all positional arguments rather than [flags](https://clig.dev/#arguments-and-flags). Took me a bit of back and forth to understand what each argument represents is in this example.

docs/redis.md
```diff
@@ -1,5 +1,94 @@
...
+    {
+        "status": "processing",
+        "output": [
+            "https://example.com/ab48b7ff-1589-4360-a54b-47f9d8d3f6b7/0.jpg",
+            "https://example.com/ab48b7ff-1589-4360-a54b-47f9d8d3f6b7/20.jpg",
+            "https://example.com/ab48b7ff-1589-4360-a54b-47f9d8d3f6b7/40.jpg",
+        ],
+        "logs": [
+            "Creating model and diffusion.",
+            "Done.",
```
zeke, line 87:
```suggestion
            "Done creating model and diffusion.",
```

zeke (commented):
This is looking great so far. Left a few comments inline.

docs/redis.md
zeke: Just to clarify that the prediction itself is not done.

docs/redis.md
bfirsh: This is indeed weird. I'm guessing just tech debt. @andreasjansson do know why this is like this?

docs/redis.md
```diff
@@ -1,5 +1,94 @@
...
+            "Iteration: 0, loss: -0.767578125",
+            "Iteration: 20, loss: -1.2333984375",
+            "Iteration: 40, loss: -1.380859375"
+        ]
+    }
+
+TODO: document `setup_time` and `run_time` queues?
```
bfirsh, line 94: I think we're going to get rid of this in favor of putting metrics in the response? May as well leave undocumented.

bfirsh (approved):
This looks great!

docs/redis.md
evilstreak: [The code itself](https://github.com/replicate/cog/blob/main/python/cog/server/redis_queue.py#L309) acknowledges that "This is intensely fragile. This should be kwargs or JSON or something like that." Definitely worth rewriting this at some point.

docs/redis.md
```diff
@@ -1,5 +1,95 @@
...
+
+- `id`: ...
+- `inputs`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
+- `response_queue`: ...
+
+You can enqueue the request using the `XADD` command:
+
+    redis:6379> XADD predict-queue * value {"id":"b7731050","inputs":{"tolerance":{"value":0.05}},"response_queue":"response-queue"}
+
+TODO: rename `inputs` to `input` to match the HTTP API (and drop the `value` wrapper)
```
zeke, line 48: WIP: https://github.com/replicate/replicate-web/pull/1192

bfirsh:
@evilstreak All looks good, and I have added a couple of extra notes!

Issue #541, opened 2022-04-11
Double check getting started guide still works


bfirsh:
Just a reminder for us to check this before we announce. See ~~#539~~ https://github.com/replicate/cog/issues/540

PR #553, opened 2022-04-14
Recursively encode tuple outputs

### LINKED ISSUE #499: Dear friend,please tell me why I can't run it from cog example.
I  am a newbie.
I run the code from cog examples.
I can  run "cog run python",but I can't run following command.
input:
sudo cog predict -i @input.jpg
resluts:
image

bfirsh:
Fixes the getting started guide, which returns numpy floats in tuples.

Based on FastAPI's json_encodable:
https://github.com/tiangolo/fastapi/blob/26f725d259c5dbe3654f221e608b14412c6b40da/fastapi/encoders.py#L111

Ref #541
Closes #499

### DIFF
--- a/python/tests/test_json.py
+++ b/python/tests/test_json.py
@@ -9,6 +9,11 @@ import numpy as np
 from pydantic import BaseModel
 
 
+def test_encode_json_recursively_encodes_tuples():
+    result = encode_json((np.float32(0.1), np.float32(0.2)), None)
+    assert type(result[0]) == float
+
+
 def test_encode_json_encodes_pydantic_models():
     class Model(BaseModel):
         text: str
--- a/python/cog/json.py
+++ b/python/cog/json.py
@@ -1,5 +1,6 @@
 from enum import Enum
 import io
+from types import GeneratorType
 from typing import Any
 
 from pydantic import BaseModel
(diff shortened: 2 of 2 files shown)


evilstreak (approved):
Nice! 👍

python/tests/test_json.py
```diff
@@ -9,6 +9,11 @@
 from pydantic import BaseModel
 
 
+def test_encode_json_recursively_encodes_tuples():
+    result = encode_json((np.float32(0.1), np.float32(0.2)), None)
```
zeke, line 13: What's the `None` for? Incidentally making sure this function doesn't blow up when given empty values?

bfirsh: It's a callback for uploading files which isn't used.

zeke: Got it. https://github.com/replicate/cog/blob/2d86d20947d66fbb8870ada667f32f033508eea7/python/cog/json.py#L17:L18

zeke (approved):
Looks good.

Does this need any accompanying documentation?

bfirsh:
> Does this need any accompanying documentation?

Thanks for the reminder, but I don't think so. Just broken.

PR #554, opened 2022-04-15
fix notebook docs

zeke:
`notebook.md` was in the wrong directory, and the notebook docs in the README were commented out.

🍐 @andreasjansson


bfirsh:
I actually intentionally commented that out. I kinda like how the readme is a really straight-forward "model -> docker image" story.

As soon as development environments get weaved in, it gets a bit confusing. If we're going to include this stuff, we should explain that it can be used in development.

Let's roll with it and iterate though!

zeke:
Ah I see. I thought we discussed notebook support being an essential thing to mention in the README, but yeah let's iterate.

bfirsh:
It's in the docs table of contents, but maybe that'll get skipped over... I wonder if we shoud pull the docs TOC up to the main TOC... 🤔

Issue #567, opened 2022-04-26
Suggest pinned versions for `system_packages`


iperdomo:
As of v0.3.0, there is a requirement that `python_packages` versions are _pinned_. `cog init` generates pinned versions for pip packages
https://github.com/replicate/cog/blob/1cde0ae054f119ea850bf8bce49b3ed64196f691/pkg/cli/init-templates/cog.yaml#L17-L20

The same _good practice_ applies to `system_packages`. Right now the `cog init` generates non-pinned versions for `system_packages`
https://github.com/replicate/cog/blob/1cde0ae054f119ea850bf8bce49b3ed64196f691/pkg/cli/init-templates/cog.yaml#L09-L11

The version available depends on the base Docker image.

```bash
# docker run --rm <image> bash -c "apt-get update -qq && apt-cache show <package-name> | grep Version"
$ docker run --rm python:3.8 bash -c "apt-get update -qq && apt-cache show libgl1-mesa-glx | grep Version"
Version: 20.3.5-1
$ docker run --rm python:3.8 bash -c "apt-get update -qq && apt-cache show libglib2.0-0 | grep Version"
Version: 2.66.8-1
```

Reference: https://github.com/hadolint/hadolint/wiki/DL3008

markphelps:
Thanks for the suggestion. As explored in PR #570, pinning apt packages is significantly more complicated than pip packages because versions are tightly coupled to the specific Debian/Ubuntu suite and architecture of the base image. Given that this issue has been inactive for over 4 years with no clear resolution path, I'm closing it as stale. If you have a concrete proposal for how to handle this robustly across different base images, feel free to reopen or start a new discussion.

PR #570, opened 2022-04-26
[#567] Use pinned versions for `apt-get` packages

iperdomo:
Based on Dockerfile best practices:

> Version pinning forces the build to retrieve a particular version
> regardless of what’s in the cache. This technique can also reduce
> failures due to unanticipated changes in required packages.
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

Signed-off-by: Iván Perdomo <ivan@perdomo.me>

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,8 +28,8 @@ Define the Docker environment your model runs in with `cog.yaml`:
 build:
   gpu: true
   system_packages:
-    - "libgl1-mesa-glx"
-    - "libglib2.0-0"
+    - "libgl1-mesa-glx=20.3.5-1"
+    - "libglib2.0-0=2.66.8-1"
   python_version: "3.8"
   python_packages:
     - "torch==1.8.1"
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -10,8 +10,8 @@ build:
   python_packages:
     - pytorch==1.4.0
   system_packages:
-    - "ffmpeg"
-    - "libavcodec-dev"
+    - "ffmpeg=7:4.3.3-0+deb11u1"
+    - "libavcodec-dev=7:4.3.3-0+deb11u1"
 predict: "predict.py:Predictor"
 ```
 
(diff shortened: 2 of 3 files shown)


iperdomo:
> One problem I've noticed with pinning apt dependencies is old versions sometimes get removed from the repositories. I wonder what Ubuntu's policy is about this, and whether this will remain reliable.

AFAIK, Alpine Linux is the distribution that removes _old_ packages. That's why they introduced `=~` when matching a package version in `apk`, e.g. `apk add --no-cache curl~=7`

I just tested a 2018 `Dockerfile` with a pinned package version and still builds successfully, e.g.
https://github.com/akvo/akvo-dockerfiles/blob/3410c095cd55db7830b21bf0fc427eb239e6109f/run-as-user/Dockerfile#L6=

I was based on `debian:strech` not `ubuntu` base image.

I've searched around but I couldn't find any documentation on their policy.

bfirsh (approved):
Super! I might be mistaken then. I have vague memories of this, but perhaps it was a third-party repository.

@andreasjansson Any thoughts about this?

iperdomo:
I reached to a friend that is closer to the Debian community and knows more about the package management part and he suggested that:

* By default old packages will not be reachable, e.g. Unbuntu 18.04 libglib2.0-0, only has some versions available - https://packages.ubuntu.com/bionic/libglib2.0-0, not all the history of versions http://changelogs.ubuntu.com/changelogs/pool/main/g/glib2.0/glib2.0_2.56.4-0ubuntu0.18.04.9/changelog
* By pinning the package to a specific version you implicitly discard security updates and minor fixes
* While it's true that you _loose_ reproducibility, the fact that you get security updates outweighs the potential problem of an update breaking your code
* If you use an LTS version, the packages updates are quite stable

I can update to the PR to use versions like `libglib2.0-0=2.*` (it's possible), or you can just ignore this changes and close the PR without merging.

bfirsh:
@iperdomo Really appreciate the contribution here, but I'm going to close this because this does seem this is more complex than just using pinned versions like this. I'm going to close this and we can continue the discussion in #567.

Thank you for helping us figure this out! 😄

mattt:
I'd like to reopen this PR for consideration.

Although there are some complexities involved with pinning system packages, the motivations are sound, and I think this is something we should do — and not just for user-specified system packages, but for anything installed in Cog's base image, too.

GitHub recently made it easier to report package dependencies in repos (https://github.com/replicate/cog/issues/1185), which helps resolve the central dilemma of pinning, which is that pinned dependencies can't be updated. If Cog users got Dependabot Alerts when vulnerabilities were reported, then I think that'd be really compelling.

PR #571, opened 2022-04-27
Update company name

bfirsh:


LICENSE
```diff
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021, Expanded Works, Inc.
+   Copyright 2022, Replicate, Inc.
```
evilstreak, line 190: Should this be a range rather than a single year? IANAL, but AIUI the year is the year of first publication, or you can use a range in the case of something that's continually updated (like software).

PR #573, opened 2022-04-27
Use multiprocessing spawn method to support CUDA

evilstreak:
## The problem

If you try to fork a new process after CUDA has been initialized, it raises an error. From [the PyTorch docs](https://pytorch.org/docs/stable/notes/multiprocessing.html#cuda-in-multiprocessing):

> The CUDA runtime does not support the fork start method; either the spawn or forkserver start method are required to use CUDA in subprocesses.

We're already doing setup within the subprocess, but it seems that some imports will start initialization of CUDA, and we're loading the predictor outside of the subprocess.

## The fix

This PR moves the loading into the subprocess and switches to using spawn to start the subprocess. Doing the former without the latter is possibly enough, but we'd need to spend more time testing that.

We should release this promptly as 0.3.1 as (some?) CUDA models won't with the queue API with Cog 0.3.0.

## Side effects

Some of the timings in the timeout tests have changed. I'm not entirely sure why the process of running a prediction is slower, but it is. If we can test it thoroughly with fork instead of spawn, we might be able to go back to the old (faster) timings.

### DIFF
--- a/python/cog/server/log_capture.py
+++ b/python/cog/server/log_capture.py
@@ -75,7 +75,13 @@ def capture_log(logs_dest):
             os.close(old_fds[out_name])
 
 
-class LogProcess(multiprocessing.Process):
+# `multiprocessing.get_context("fork")` returns the same API as
+# `multiprocessing`, but will use the fork method when creating any subprocess.
+# Although fork is the default, we need this here because the log process gets
+# created from the predictor subprocess, which is set to use the spawn method.
+# As currently written, this log process relies on shared state in a way that
+# doesn't work with the spawn method.
+class LogProcess(multiprocessing.get_context("fork").Process):
     def __init__(
         self,
         logs_dest,
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -7,6 +7,7 @@ import types
 
 from pydantic import BaseModel
 
+from ..predictor import load_predictor
 from .log_capture import capture_log
 
 
@@ -18,8 +19,7 @@ class PredictionRunner:
         SINGLE = 1
         GENERATOR = 2
 
-    def __init__(self, predictor):
-        self.predictor = predictor
+    def __init__(self):
         self.logs_pipe_reader, self.logs_pipe_writer = multiprocessing.Pipe(
             duplex=False
         )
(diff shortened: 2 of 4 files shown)


python/cog/server/log_capture.py
```diff
@@ -75,7 +75,7 @@ def capture_log(logs_dest):
             os.close(old_fds[out_name])
 
 
-class LogProcess(multiprocessing.Process):
+class LogProcess(multiprocessing.get_context("fork").Process):
```
andreasjansson, line 78: Could you add a line of documentation here explaining what `multiprocessing.get_context("fork").Process` does? It's a fairly unusual pattern and we might forget the rationale.

python/cog/server/runner.py
```diff
@@ -42,12 +42,13 @@ def setup(self):
         Sets up the predictor in a subprocess. To start a prediction after
         setup call `run()`.
         """
-        self.predictor_process = multiprocessing.Process(
+        self.predictor_process = multiprocessing.get_context("spawn").Process(
```
andreasjansson, line 45: Same here, would be nice to document `multiprocessing.get_context("spawn").Process? It's a fairly unusual pattern and we might forget the rationale.

Also, why is it `"spawn"` here and `"fork"` in the base class of `LogProcess`?

andreasjansson (approved):
This looks good, would just be nice with a couple of comments to explain spawn vs fork.

I'm not worried about the extra 400ms if it's just an absolute increase on booting the model. It shouldn't affect individual prediction times since the predictor process is kept alive for the duration of the parent process.

I assume you've tested this on @afiaka87's model and it works? If yes then :shipit:

python/cog/server/log_capture.py
evilstreak: Comment added!

python/cog/server/runner.py
evilstreak: Comment added about `get_context` and about the choice of methods used.

evilstreak:
> I'm not worried about the extra 400ms if it's just an absolute increase on booting the model. It shouldn't affect individual prediction times since the predictor process is kept alive for the duration of the parent process.

This is the weird thing. The timeout only applies to the `run()` method – we don't start counting until after the model has finished booting. I'm not sure what about the spawn method slows things down – I'm assuming it's due to differences in how the processes need to communicate and/or share resources, but that's a very untested assumption.

I'd like to spend some time finding (or putting together) a much smaller CUDA model for testing this more thoroughly, but I think this change is Good Enough for now.

> I assume you've tested this on @afiaka87's model and it works? If yes then :shipit:

Yep! @afiaka87 used this change to build a model that successfully runs via the queue API.

andreasjansson:
Yes definitely Good Enough, let's ship it!

PR #577, opened 2022-04-28
Upgrade github.com/opencontainers/image-spec

bfirsh:
Dependabot was struggling to do this by itself and throwing a minor
security warning (which was annoying).


zeke (approved):
👍🏼

bfirsh:
This is failing because NVIDIA has broken their apt package signing (sigh): https://forums.developer.nvidia.com/t/gpg-error-http-developer-download-nvidia-com-compute-cuda-repos-ubuntu1804-x86-64/212904

bfirsh:
Fixed here https://github.com/replicate/cog/pull/579

PR #580, opened 2022-05-02
Revert "Remove NVIDIA apt sources/key"

bfirsh:
This reverts commit 67c5b31a143d6601df58ddd705bd707680359d6b.

#579 removed the nvidia apt sources because they were broken upstream. When CI passes here, they have been fixed and we can merge this.

See also https://github.com/NVIDIA/nvidia-docker/issues/1631


bfirsh:
Seeing if the tests pass again...

bfirsh:
Tests pass now but judging by https://github.com/NVIDIA/nvidia-docker/issues/1631 it still seems to be an issue for some base images. Maybe the tests are running on one which happens to work for now.

Leaving this as a draft out of caution, and let's maybe merge this when that nvidia issue is resolved.

evilstreak:
@bfirsh should we get this in now? It looks like the linked issue has been pretty quiet for a good long while.

bfirsh:
Yeah I suppose so? I have rebased on main.

bfirsh:
I reckon let's get this in and we can always revert if there are errors.

PR #585, opened 2022-05-04
Drops --gpus flag, updates formatting

charlesfrye:
This PR removes the `--gpus all` flag in the `docker run` example in the Getting Started guide.
It also makes some changes to the formatting to make it easier to use.

First, the `cog.yaml` example doesn't include `gpu: true`
and second, that causes the command to fail on systems without GPUs

That plus the voluminous TensorFlow warnings convinced me I had something wrong in my Cog setup
-- that I needed `nvidia-docker` or hadn't set an env variable correctly, but it was just the flag.

You could put a specific note about the warnings from TF and direct folks to the `#gpus` section of the BYOmodel start guide.

I also updated the formatting, using some ideas from
[this style guide](https://arcticicestudio.github.io/styleguide-markdown/rules/code.html)
for code in Markdown.

The code snippets are now all copyable directly from GitHub --
outputs are separated from the commands or commented out
and `$`s are removed --
and they have terminal newlines so that the commands run immediately in some settings.

For testing, I ran in two environments:
- bare metal on a Linux machine with single GPU (Ubuntu18.04, NVIDIA107M)
- bare metal on Macbook Air

I then added `gpu: true` to the `cog.yaml` and confirmed that the commands still ran in the Linux environment.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -11,18 +11,21 @@ This guide will walk you through what you can do with Cog by using an example mo
 
 First, install Cog:
 
-```sh
+```bash
 sudo curl -o /usr/local/bin/cog -L https://github.com/replicate/cog/releases/latest/download/cog_`uname -s`_`uname -m`
 sudo chmod +x /usr/local/bin/cog
+
 ```
 
 ## Create a project
 
 Let's make a directory to work in:
 
-    mkdir cog-quickstart
-    cd cog-quickstart
+```bash
+mkdir cog-quickstart
+cd cog-quickstart
 
+```
 ## Run commands
 
 The simplest thing you can do with Cog is run a command inside a Docker environment.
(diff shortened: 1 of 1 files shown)


charlesfrye:
(PS I did this edit in the GitHub web UI because the goal was to change the rendering and UX on the site, and I got briefly blocked by the DCO check and had to force-push. Looks like there's [very recent experimental support for signing off in the UI](https://github.com/todogroup/gh-issues/issues/50#issuecomment-1110299580), and it might be worth looking into adding it for these kinds of quick edits/GH-focused content)

bfirsh:
Ah, brilliant thank you!

Great tip about the automatic DCO sign-off too. I've asked them to flip that flag...

zeke:
@all-contributors please add @charlesfrye for docs 🙏🏼

PR #594, opened 2022-05-09
Update remote image test to recent Cog model

evilstreak:
The old model was from before Cog 0.2.0.


bfirsh:
Looks like this is skipped and not run anyway.

PR #599, opened 2022-05-10
Include timestamps in Redis responses

### LINKED ISSUE #589: exposing timing metrics
From @evilstreak 

> Prediction time is not accurate. It depends on the user polling to update. If all predictions take 1 second but the user polls after 10 seconds, all predictions will have a prediction time of 10 seconds.

evilstreak:
Timestamps are useful for calculating how long a prediction has taken to run. There are other ways we could get this data (using HTTP streams instead of queueing, and doing accurate timing from the outside), but this is probably the simplest to get started with and arguably the simplest for all sorts of users in the future.

Resolves #589

### DIFF
--- a/test-integration/test_integration/test_redis_queue.py
+++ b/test-integration/test_integration/test_redis_queue.py
@@ -1,6 +1,8 @@
 import json
 from pathlib import Path
+import re
 import subprocess
+import unittest.mock as mock
 
 from .util import docker_run, random_string
 
@@ -52,6 +54,9 @@ def test_queue_worker_files(docker_image, docker_network, redis_client, upload_s
 
         response = json.loads(redis_client.blpop("response-queue", timeout=10)[1])
         assert response == {
+            "x-experimental-timestamps": {
+                "started_at": mock.ANY,
+            },
             "status": "processing",
             "output": None,
             "logs": [],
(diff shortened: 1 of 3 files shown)


zeke (approved):
Looks good to me!

test-integration/test_integration/test_redis_queue.py
```diff
@@ -51,6 +52,7 @@ def test_queue_worker_files(docker_image, docker_network, redis_client, upload_s
         )
         response = json.loads(redis_client.brpop("response-queue", timeout=10)[1])
         assert response == {
+            "timestamp": mock.ANY,
```
zeke, line 55: It looks like all the assertions are using `mock.ANY`. Do you think it's worth having at least one test that asserts the value is indeed an ISO formatted date string?

bfirsh:
I wonder if we want to record some more granular timestamps and consider that to be a permanent part of the prediction?

```
{
  "metrics": {
    "started_at": ...,
    "completed_at": ...,
  }
}
```

Or, whatever.

A couple of things I can think of that might not work with this approach:

1. Maybe a message that says `succeeded` with a timestamp doesn't necessarily correspond to the point the model is succeeded. It might be a message with a log line that was created after the model succeeded, or something like that.
2. Currently a Redis response maps 1-1 with Cog HTTP responses and Replicate API responses (I think?). A timestamp field makes a lot less sense for those things.

Happy to get this in if it's a bodge and undocumented, but if we intend for this to be something to be used properly, maybe needs a bit more thought.

andreasjansson (approved):

bfirsh:
If we want to ship this without thinking abotu the API we might want to prefix it with `x-experimental-` or whatever.

evilstreak:
I've reworked this to include timestamps in the format suggested by @bfirsh:

```
{
  "metrics": {
    "started_at": ...,
    "completed_at": ...
  },
  ...
}
```

Including those two times means we don't have to worry about delays or failures in message delivery or processing affecting recorded timestamps. That might be the case if, for example, we were unable to deliver the first "processing" message and only delivered the final "succeeded" message instead.

We could name this `x-experimental-metrics`, but I'm inclined to just go with `metrics` on this.

evilstreak:
Adding a test for the timestamp format, which is a good job, because it uncovered a bug!

bfirsh:
All sounds sensible. This means we'll have to change the API in Replicate, so I guess this is a decision between something marginally better and additional work to migrate Replicate, or something marginally worse and not having to migrate Replicate.

I wonder if `created_at` means different things in Replicate and Cog. In Replicate, it's the point where it's created in Replicate's system. In Cog it's the point it's created in Cog's system. But maybe you want to know the latter from Replicate? Or perhaps the only thing you care about is when `predict()` started. Hmm.

andreasjansson (approved):
Nice! This'll be great

PR #616, opened 2022-05-16
Upgrade to Go 1.18

bfirsh:
Should fix the issue we're seeing in #610.

I also ran `go mod tidy` because there seem to be changes to `go.mod` somewhere between 1.16 and 1.18.


evilstreak (approved):
👍

Issue #597, opened 2022-05-10
Bad error message from some unallowed input types


evilstreak:
Using `Any` or `List[SomeObject]` as an input type makes Cog throw an error when, ironically, it's trying to produce a human readable typename for the bad type:

```
Traceback (most recent call last):
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/cog/command/openapi_schema.py", line 24, in <module>
    app = create_app(predictor)
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/cog/server/http.py", line 34, in create_app
    InputType = get_input_type(predictor)
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/cog/predictor.py", line 126, in get_input_type
    f"Unsupported input type {human_readable_type_name(InputType)} for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}."
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/site-packages/cog/predictor.py", line 215, in human_readable_type_name
    return module + "." + t.__qualname__
  File "/root/.pyenv/versions/3.8.13/lib/python3.8/typing.py", line 760, in __getattr__
    raise AttributeError(attr)
AttributeError: __qualname__
```

TommyDew42:
Hi! It looks like a good issue I can help.

Would be happy if I can work on it!

bfirsh:
Thanks again for this @TommyDew42!

PR #618, opened 2022-05-17
Fix bad type names

TommyDew42:
A PR for [issue 597](https://github.com/replicate/cog/issues/597).

We return `str(t)` from `human_readable_type_name` if there's no `__qualname__` attribute.


TommyDew42:
Do we wanna write tests for the change?

bfirsh:
A test would be nice, but this is pretty minor so it's not essential. It'd probably be best to catch the exact exception here (`AttributeError`?) instead of a plain except.

TommyDew42:
Sounds good. If it is not an `AttributeError`, we'll just let python help us throw that error.

zeke (approved):
Looks good to me. Is this ready to go?

bfirsh:
Yes! Thanks @TommyDew42!

zeke:
@all-contributors please add @TommyDew42 for code

PR #620, opened 2022-05-19
Make default input first input

### LINKED ISSUE #381: What should `cog predict` do if you don't pass an input name?
The syntax of `cog predict` is this:

    cog predict -i image=@image.jpg

But you can also do this:

    cog predict -i @image.jpg

Which implicitly means an input name of `input`. This is a neat short hand but a bit weird for a few reasons:

- `input` is a Python built-in, so we should really be encouraging using that as a variable name.
- It is a magic name.
- For a sufficiently complex model, you probably don't want to call it `input`.

What could we do that is better here? Maybe if you don't pass a name, it defaults to the _first_ input defined, rather than a magic name? This is vaguely backwards compatible, which is neat.

bfirsh:
Having a magic `input` input was weird and collided with Python's
built-in `input()`.

Closes #381

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,10 +49,10 @@ class Predictor(BasePredictor):
 
     # The arguments and types the model takes as input
     def predict(self,
-          input: Path = Input(title="Grayscale input image")
+          image: Path = Input(title="Grayscale input image")
     ) -> Path:
         """Run a single prediction on the model"""
-        processed_input = preprocess(input)
+        processed_image = preprocess(image)
         output = self.model(processed_input)
         return postprocess(output)
 ```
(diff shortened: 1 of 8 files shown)


evilstreak (changes requested):
These changes look sensible 👍 Even without clashing with Python's magic `input`, seeing docs where the parameter is named something generic like `input` makes me wonder whether it's a special value or something I can change. I think the new parameter names in the examples make that clearer.

Added a few suggestions where the examples need updating to fully use the new variable name.

bfirsh:
Whoopsy, thank you.

evilstreak (approved):

Issue #633, opened 2022-06-03
Make HTTP concurrency tunable


bfirsh:
We make a naive attempt to limit number of workers:

https://github.com/replicate/cog/blob/4628b0e085417fdff46f80ba974d8dc281b31760/python/cog/server/http.py#L112-L113

But this is async! It'll run loads of async workers.

See https://github.com/tiangolo/fastapi/issues/4221 for how we might fix this. This shoud also be configurable, and default to 1 for GPU models.

Issue #521, opened 2022-04-01
Async HTTP API


bfirsh:
As mentioned in #475, we probably want an async HTTP API. There might be clever things we can do with bi-directional streams. [Needs expanding with things from #475.]

## User data

> After thinking about it a bit more, an async API in cog would be more helpful than a synchronous one in Replicate. I did a little benchmarking and it didn’t take very many concurrent requests to get cog to crash for using too much memory. It seems like the queuing system you’ve already built would be the best way to keep that from happening. Also, even if predictions are usually quick, sometimes they’ll be slow under load or while scaling up and it’s probably better to build for that case.

bfirsh:
Fixed in https://github.com/replicate/cog/pull/870

PR #634, opened 2022-06-03
Limit number of threads to 1

bfirsh:
The Cog HTTP server uses async I/O with many threads, but predictors are blocking. This will cause unexpected behavior, and may be causing Cog to run out of memory if too many requests come in. (#521)

We should support async predictors for concurrency (#633) but for now limit threads to 1 so this isn't broken.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -1,3 +1,5 @@
+from anyio import CapacityLimiter
+from anyio.lowlevel import RunVar
 import logging
 import os
 import types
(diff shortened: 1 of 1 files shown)


zeke (approved):
Looks good?

Is there a way to test this?

bfirsh:
Ah, yes, I forgot to mention. I just added `time.sleep()` to the hello world model:

```python
import time


    def setup(self):
        self.prefix = "hello"

    def predict(self, text: str = Input(description="Text to prefix with 'hello '")) -> str:
        time.sleep(5)
        return self.prefix + " " + text
```

Then ran this in several shells to confirm they ran one after the other instead of the same time:

```
curl http://localhost:5000/predictions -X POST -H "Content-Type: application/json" \                                    !10139
  -d '{"input": {
    "text": "..."
  }}'
```

PR #635, opened 2022-06-03
Make number of threads configurable

### LINKED ISSUE #633: Make HTTP concurrency tunable

bfirsh:
FastAPI uses threads for concurrency when the view function is synchronous, which is the case for `POST /predictions`. By default, the thread pool is 40. This won't work for GPU models which usually need to be run serially, and is too high for CPU models which are usually CPU bound. This may also be causing it to run out of memory (see #521).

This pull request sets the number of threads to be the number of CPUs, or just 1 if it's a GPU model. It is also configurable with the `--threads` option.

This is hard to test with an automated test, but open to ideas.

I tested locally with a simple model that sleeps:

```python
import time


    def setup(self):
        self.prefix = "hello"

    def predict(self, text: str = Input(description="Text to prefix with 'hello '")) -> str:
        time.sleep(5)
        return self.prefix + " " + text
```

Then ran this in several shells:

```
curl http://localhost:5000/predictions -X POST -H "Content-Type: application/json" \                                    !10139
  -d '{"input": {
    "text": "..."
  }}'
```

When `--threads=1` then they run one after another. With `--threads=5` or not passed at all they run at the same time.

Closes #633.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -4,6 +4,8 @@ Cog containers are Docker containers that serve an HTTP server for running predi
 
 This guide assumes you have a model packaged with Cog. If you don't, [follow our getting started guide](getting-started-own-model.md), or use [an example model](https://github.com/replicate/cog-examples).
 
+## Getting started
+
 First, build your model:
 
     cog build -t my-model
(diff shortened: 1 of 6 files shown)


docs/deploy.md
```diff
@@ -30,3 +32,17 @@ To run a prediction on the model, call the `/predictions` endpoint, passing inpu
...
+
+## Options
+
+Cog Docker images have `python -m cog.server.http` set as the default command, which gets overridden if you pass a command to `docker run`. When you use command-line options, you need to pass in the full command before the options.
+
+### `--threads`
+
+This controls how many threads are used by Cog, which determines how many requests Cog serves in parallel. If your model uses a CPU, this is the number of CPUs on your machine. If your model uses a GPU, this is 1, because typically a GPU can only be used by one process.
+
+You might need to control this if you want to control how much memory your model uses, or other similar constraints. To do this, you can use the `--threads` option.
```
zeke, line 44:
```suggestion
You might need to adjust this if you want to control how much memory your model uses, or other similar constraints. To do this, you can use the `--threads` option.
```

zeke (approved):
Looks good.

Thanks for the docs and the helpful context in the PR body. ⭐

PR #636, opened 2022-06-05
Check for new versions

### LINKED ISSUE #605: Automatically check for new Cog versions
And, explain how to upgrade or do it automatically. Cog is moving fast and we've been running into problems with people not upgrading to the latest version.

bfirsh:
Super simple thing to get this working. It works a bit like this:

- In the background, it hits `https://update.cog.run/v1/check?version=1.2.3` (easily cacheable on a CDN!)
- That endpoint responds with `{"message": "..."}` which gets stored in `~/.config/cog/update-check.json` along with a timestamp. 
- On subsequent runs, that message gets displayed before any Cog command. "A new version of Cog is available! Go to https://cog.run/docs/upgrade to upgrade."
- After an hour, it hits that endpoint again.

Any errors are swallowed and sent to debug logging so anything broken in here shouldn't break anything for the user.

### Alternative solutions & future

I considered encoding versions in the API somehow, but this seemed like the simplest, most flexible solution. You could imagine a future where `cog update` existed and this API had some smarts to inform how that worked.

### Backend

https://update.cog.run is currently running as a simple Cloudflare Worker, which runs on the edge to reduce latency, but this could be anything.

### Testing

Edit the Cloudflare Worker to return `{"message": "hello!"}` and Cog will say hello.

Closes #605

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -6,6 +6,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/replicate/cog/pkg/global"
+	"github.com/replicate/cog/pkg/update"
 	"github.com/replicate/cog/pkg/util/console"
 )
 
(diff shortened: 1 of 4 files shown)


zeke (approved):
It works!

```
$ cog --version
cog version v0.3.5+dev (built 2022-06-06T09:47:58-0700)
```

```
$ cog help
Hello

Containers for machine learning.
...
```

```
$ cog run echo this is a test
Hello

Building Docker image from environment in cog.yaml...
```

PR #653, opened 2022-06-13
Support lists of complex outputs in queue worker

### LINKED ISSUE #652: Support lists of complex `BaseModel` for outputs
The following model fails on Replicate.com with a cryptic `Can't pickle <class 'predict.Output'>: it's not the same object as predict.Output` error:

```python
class Output(BaseModel):
    foo: Optional[str]
    bar: Optional[Path]

    def predict(self) -> List[Output]:
        # [...]
        outputs = [Output("foo", some_path)]
        return outputs
```

This is [documented deep in Cog's Python layer](https://github.com/replicate/cog/blob/main/python/cog/server/runner.py#L238).

evilstreak:
An output type of `List[Output]` would break because of the problems that `make_pickleable` works around. This change makes that work, by making pickleable each item in the list.

This is still a bandaid.

Fixes #652

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -248,5 +248,7 @@ def make_pickleable(obj: Any) -> Any:
     """
     if isinstance(obj, BaseModel):
         return obj.dict(exclude_unset=True)
+    elif isinstance(obj, List):
+        return [make_pickleable(item) for item in obj]
     else:
         return obj


bfirsh:
Do you know why we have a separate function for the redis worker here? Could we just use the same `encode_json()` that HTTP uses?

https://github.com/replicate/cog/blob/db9bf0bb6b3a324b4abbd31f4098b1a125115c56/python/cog/json.py#L18-L46

bfirsh:
Oooh, right, this is the worker in the worker isn't, ho hum. I wonder if we could encode_json before sending it back to the parent process. 🤔

bfirsh:
Oh, yeah. I wonder if we should split out those two things.

Anyway didn't mean to derail this. This looks good. Not going to merge because I figure this needs shipping into a release etc.

bfirsh:
My subconscious just merged it. Oh dear.

_steps away from the computer_

PR #655, opened 2022-06-14
Annotate iterators so we can differentiate from lists

bfirsh:
Iterators represent progressive output, lists just represent several
outputs. On Replicate, we want to be able to differentiate between these
two so iterators have a slider and lists are displayed as a grid.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -5,12 +5,12 @@ import importlib.util
 import inspect
 import os.path
 from pathlib import Path
-from pydantic import create_model, BaseModel
+from pydantic import create_model, BaseModel, Field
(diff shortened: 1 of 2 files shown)


zeke (approved):
Magical. 🐇 

How did you figure this out? Just from the pydantic docs? Stack overflow? Reading the pydantic source?

bfirsh:
I wanted to apply `Field()` annotations to that list. I was considering doing something like this, which is what we do further down in this file:

```
class OutputType(BaseModel):
    __root__: List[get_args(OutputType)[0] = Field(...)
```

But then while searching through the pydantic docs for `__root__` I stumbled across `Annotated[]` which is a shortcut for this.

PR #656, opened 2022-06-15
document how to optionally set output properties

zeke:
People often try to set output values to `None`, only to discover that type checking fails. This PR updates the documentation for `Output` to clarify that unwanted values can be omitted.

cc @vccheng2001 @chenxwh

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -12,6 +12,7 @@ Tip: Run [`cog init`](getting-started-own-model#initialization) to generate an a
     - [Progressive output](#progressive-output)
 - [`Input(**kwargs)`](#inputkwargs)
 - [Output](#output)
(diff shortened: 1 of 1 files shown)


bfirsh:
I'm not sure they can be omitted, can they? The type needs to be set to `Optional[...]` in the model.

That's what @chenxwh did here:

https://github.com/borisdayma/dalle-mini/pull/253/files#diff-73c1982d8a085dc10fda2ac7b6f202ae3ff9530ee6a15991c5339051eb10a49aR24-R26

(Although in this case I think it's only `clip_score` which needs to be optional.)

vccheng2001:
awesome! thanks!

bfirsh:
:sparkles

PR #657, opened 2022-06-16
Send response when queue worker starts predict

evilstreak:
Currently the only way to know whether Cog has picked up a request is to watch for jobs to be claimed from the request queue. Sending an empty processing message is a much cleaner way to handle that, and will be picked up by whatever system folks are using to listen to non terminal responses.

PR also includes a commit to swap `brpop` to `blpop`, to fix a timing issue which was exposed by this change.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -250,6 +250,8 @@ class RedisQueueWorker:
                 "logs": logs,
             }
 
+            self.redis.rpush(response_queue, json.dumps(response))
+
             # just send logs until output starts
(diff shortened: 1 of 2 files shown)


andreasjansson (approved):
Looks good!

I'm curious about the blpop/brpop change. Did it previously work because there was only one item on the queue?

evilstreak:
> I'm curious about the blpop/brpop change. Did it previously work because there was only one item on the queue?

Yep, I think the tests were managing to make the request when there were either 0 or 1 items on the queue, so `brpop` and `blpop` were equivalent.

PR #659, opened 2022-06-20
Terminate worker if it can't receive messages

evilstreak:
When the worker can't receive messages because, for example, it cannot connect to Redis, the current behaviour is to output a stack trace to stderr and try again. It does this continuously, in a tight loop, which generates an awful lot of log output and disguises the fact that it's not working properly.

A more useful behaviour would be for the worker to crash when it can't connect, so the container shows as non-running. That's what this change does.

A few questions for reviewers:

- Is this feature even desirable? We need to not uncontrollably dump stack traces to stderr, and it'd be nice to have some sort of strong signal that something is wrong, but there are probably other ways to handle both of those needs. Is this a reasonable way to approach it?
- Should we have some sort of sleep/backoff in that failure loop? This will currently exit very quickly – do we want to be able to stay running through Redis blips of up to 15s? 60s? 300s?
- Should we try connecting to Redis *before* we run `setup()`, so we can crash early before spending potentially a few minutes working?
- What would a reasonable way to test this be? Try running it via Docker, wait a few seconds, and check the container has exited?

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -160,13 +160,29 @@ class RedisQueueWorker:
             sys.stderr.write(f"Setup time: {setup_time:.2f}\n")
 
         sys.stderr.write(f"Waiting for message on {self.input_queue}\n")
+        consecutive_failures = 0
         while not self.should_exit:
             try:
                 message_id, message_json = self.receive_message()
-                if message_json is None:
-                    # tight loop in order to respect self.should_exit
+                consecutive_failures = 0
+            except:
+                # Give up and exit after a few retries. The cost of running
+                # `setup()` is significant, so we don't want to exit due to an
+                # intermittent error. But, we do want to crash if we can't
+                # connect to Redis, so the failure becomes evident.
+                consecutive_failures += 1
+                if consecutive_failures > 10:
+                    sys.stderr.write("Failed to receive message from Redis; giving up")
+                    break
+                else:
+                    sys.stderr.write("Failed to receive message from Redis; retrying")
                     continue
 
+            if message_json is None:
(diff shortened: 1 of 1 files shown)


python/cog/server/redis_queue.py
```diff
@@ -160,13 +160,29 @@ def start(self) -> None:
...
+                # `setup()` is significant, so we don't want to exit due to an
+                # intermittent error. But, we do want to crash if we can't
+                # connect to Redis, so the failure becomes evident.
+                consecutive_failures += 1
+                if consecutive_failures > 10:
+                    sys.stderr.write("Failed to receive message from Redis; giving up")
+                    break
+                else:
+                    sys.stderr.write("Failed to receive message from Redis; retrying")
                     continue
```
andreasjansson, line 179: A short sleep would be nice in case of an intermittent blip. Backoffs are probably not necessary – in the case where the Cog model is deployed on Kubernetes, the cluster will CrashLoopBackoff.

andreasjansson (commented):
What you describe sounds like a good way to test this.

python/cog/server/redis_queue.py
evilstreak: I was thinking about the case where we've got a model that takes a couple of minutes to run `setup()`, and Redis goes down for 30s. We'd prefer to keep the model alive through that, rather than have to restart it and rerun `setup()`, I think. Weighed against: if there's a problem with config, which means it'll never be able to connect to Redis, we want it to crash sooner rather than later.

I wonder about doing a connection check before running `setup()`, so we can fail quickly, and then allowing rather longer before failing after `setup()` is complete.

PR #663, opened 2022-06-21
Fix Redis worker yielding list of complex outputs

bfirsh:
Starting by refactoring the Redis queue worker so it uses the same encoding as HTTP to hopefully remove a whole class of problems.

The last commit is the real change. Other commits are refactoring preludes.

Based on https://github.com/replicate/cog/pull/665

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -7,6 +7,8 @@ from typing import Any, Dict, List, Optional
 
 from pydantic import BaseModel
 
+
+from ..json import make_encodeable
 from ..predictor import load_config, load_predictor
 from .log_capture import capture_log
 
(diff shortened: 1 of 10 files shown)


python/cog/server/runner.py
```diff
@@ -232,13 +234,13 @@ def _run_prediction(
                     while True:
                         try:
                             self.predictor_pipe_writer.send(
-                                next(make_pickleable(output))
```
evilstreak, line 235: Would this have worked if it were `make_pickleable(next(output))` instead? Feeling a little 🤦 about that!

evilstreak (approved):
Looks great! Really appreciate the well broken down refactor.

python/cog/server/runner.py
bfirsh: Hah, maybe? I think it returns the same iterator either way doesn't it.

Issue #671, opened 2022-06-28
Redis issue causing CI to fail


bfirsh:
Logs are too long to display on GitHub actions but this is the error:

```
Failed to handle message: Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/cog/server/redis_queue.py", line 166, in start
    message_id, message_json = self.receive_message()
  File "/usr/local/lib/python3.8/site-packages/cog/server/redis_queue.py", line 128, in receive_message
    key, raw_message = raw_messages[0]
ValueError: too many values to unpack (expected 2)
```

I thought it might be that [the redis 7 docker image was pushed a few days ago](https://hub.docker.com/layers/redis/library/redis/7.0/images/sha256-31120dcdd310e9a65cbcadd504f4fe60a185bd634ab7c6a35e3e44a941904d97?context=explore) but pinning to `redis:6` doesn't seem to fix it.

andreasjansson:
Something has changed between `redis==4.3.3` and `redis==4.3.4`. If I downgrade to 4.3.3 things work as before. Tested locally with https://github.com/replicate/cog-redis-example and a custom build of Cog.

Issue #409, opened 2022-02-08
Vendor Pydantic and FastAPI


bfirsh:
This is bound to cause problems at some point colliding with user code. We have some prior art for vendoring Python packages in Keepsake that works pretty well.

Migrated from a todo in #378.

bfirsh:
And typing-extensions... https://canary.discord.com/channels/775512803439280149/979390547477274644/981627994097467482

anish-sahoo:
Closing as the dependency-collision concern was addressed by replacing the Python HTTP server with the Rust-based Coglet runtime and making Pydantic an optional dependency.

PR #672, opened 2022-06-28
Handle new XAUTOCLAIM response introduced in redis>=4.3.4

andreasjansson:
[A breaking change](https://github.com/redis/redis-py/pull/2252) in `redis==4.3.4` breaks Cog.

This PR handles the new XAUTOCLAIM response format in a backwards-compatible way.

Closes https://github.com/replicate/cog/issues/671

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -124,7 +124,8 @@ class RedisQueueWorker:
             1,
         )
         # format: [[b'1619393873567-0', [b'mykey', b'myval']]]
-        if raw_messages and raw_messages[0] is not None:
+        # since redis==4.3.4 an empty response from xautoclaim is indicated by [[b'0-0', []]]
+        if raw_messages and raw_messages[0] is not None and len(raw_messages[0]) == 2:
             key, raw_message = raw_messages[0]
             assert raw_message[0] == b"value"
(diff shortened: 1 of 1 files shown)


bfirsh:
Looks good. Perhaps we should also pin `redis` in `setup.py` to avoid this issue in the future? It seems unlikely that redis will be installed for any other reason.

The real fix: #409

bfirsh:
I guess this is a good reason to use #605 :)

andreasjansson:
> Perhaps we should also pin redis in setup.py to avoid this issue in the future?

I decided against this since redis-py is used by [a lot](https://github.com/redis/redis-py/network/dependents?dependent_type=PACKAGE&package_id=UGFja2FnZS01MjUzNzc4OA%3D%3D) of other packages.

But yeah, this is another reminder that we really should vendor dependencies.

bfirsh:
I wonder if they are going to be used in ML models though?

andreasjansson:
> I wonder if they are going to be used in ML models though?

Ray, Optuna, etc.

It's probably as likely that any of the other packages we depend on introduce breaking changes. I think we should just get this issue fixed and then move to vendor our dependencies.

bfirsh:
100% all looks good.

PR #673, opened 2022-06-28
Reset upgrade state when upgraded

bfirsh:
I made the update check display a message for #672 but the message
doesn't reset when people upgrade.

### DIFF
--- a/pkg/update/update.go
+++ b/pkg/update/update.go
@@ -26,15 +26,23 @@ func DisplayAndCheckForRelease() error {
 		return fmt.Errorf("update check disabled")
 	}
 
-	state, err := loadState()
+	s, err := loadState()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


pkg/update/update.go
```diff
@@ -30,6 +30,15 @@ func DisplayAndCheckForRelease() error {
 	if err != nil {
 		return err
 	}
+
+	if state.Version != global.Version {
+		console.Debugf("Resetting update message because Cog has been upgraded")
+		state.Message = ""
```
andreasjansson, line 36: Should we not also set `state.Version = global.Version` here?

bfirsh: It's a no-op because no message means it doesn't do anything but I suppose we may as well.

bfirsh: Ah, yes we should otherwise it'll rewrite the config on each command until the next version check.

PR #678, opened 2022-07-07
Better error handling

evilstreak:
Some improvements to error handling, including:

- an opentelemetry fix
- tracebacks in logs for errors
- an `IrrecoverablePredictorError` to signal serious problems with a predictor

Details in the commits.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -16,6 +16,7 @@ Tip: Run [`cog init`](getting-started-own-model#initialization) to generate an a
 - [Input and output types](#input-and-output-types)
 - [`File()`](#file)
 - [`Path()`](#path)
+- [Error handling](#error-handling)
 
 ## `BasePredictor`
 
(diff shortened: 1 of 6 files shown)


evilstreak:
It looks it's not the timing error I was trying to fix in 8f9ff5f – it seems the traceback isn't getting sent through as logs at all. @andreasjansson do the tests pass locally for you?

docs/python.md
```diff
@@ -217,3 +218,21 @@ class Predictor(BasePredictor):
         upscaled_image.save(output)
         return Path(output_path)
 ```
+
+## Error handling
+
+If your predictor encounters an error, Cog will catch it and handle it gracefully. You can generally ignore error handling and trust Cog to do the right thing.
+
+If your predictor is being run by the queue worker, it will returned a failed prediction along with the error message, and then prepare to pick up the next prediction request. If your predictor has encountered an error so serious that it cannot finish the current prediction, nor will it be able to finish any future prediction, it needs to terminate.
```
andreasjansson, line 226: s/returned/return/

andreasjansson (approved):
This looks good, I like the idea of being able to tell the predictor to restart from inside the predictor. I just had a couple minor nits.

The test `test_integration/test_build.py::test_build_gpu_model_on_cpu` hangs for me locally. I'm not sure if that's a real error or a red herring of some sort.

evilstreak:
This is mostly obsolete after #795. If we want to keep something like `IrrecoverablePredictorFailure` then we'll need to rework this.

PR #679, opened 2022-07-12
Fix timeout behaviour

evilstreak:
When the Redis queue worker is responsible for handling timeouts, the predictor process continues running after the queue worker sends the timeout message. If another job gets picked up before the queue worker is finished, it'll send back outputs from the previous job.

Moving it into the runner means the predictor itself is timed out and stops running. Expand the test case to demonstrate the issue and prevent regression.

Contains a couple of the preliminary fixes from #678, and a semi-related fix to timeout logic. Best viewed commit-by-commit with whitespace diffs turned off.

### DIFF
--- a/test-integration/test_integration/test_redis_queue.py
+++ b/test-integration/test_integration/test_redis_queue.py
@@ -338,11 +338,22 @@ def test_queue_worker_error(docker_network, docker_image, redis_client):
         assert response == {
             "x-experimental-timestamps": {
                 "started_at": mock.ANY,
+            },
+            "status": "processing",
+            "output": None,
+            "logs": mock.ANY,  # includes a stack trace
+        }
+        assert "Traceback (most recent call last):" in response["logs"]
+
+        response = json.loads(redis_client.blpop("response-queue", timeout=10)[1])
+        assert response == {
+            "x-experimental-timestamps": {
+                "started_at": mock.ANY,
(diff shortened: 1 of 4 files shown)


test-integration/test_integration/test_redis_queue.py
```diff
@@ -421,6 +432,16 @@ def test_queue_worker_error_after_output(docker_network, docker_image, redis_cli
...
         }
 
+        response = json.loads(redis_client.blpop("response-queue", timeout=10)[1])
+        assert response == {
+            "x-experimental-timestamps": {
+                "started_at": mock.ANY,
+            },
+            "status": "processing",
+            "output": ["hello bar"],
+            "logs": mock.ANY,  # includes a stack trace
```
zeke, line 442: The surrounding responses have assertions for the stack trace but this one doesn't. Is that intentional?

zeke (approved):
I'm not so familiar with this code but the change looks good. 👍🏼

bfirsh:
This is exactly the sort of thing it'd be nice to shim into old models. 🤔

test-integration/test_integration/test_redis_queue.py
evilstreak: It didn't seem worth doing that assertion twice, but I've moved it to the earliest logs it should appear in for both tests 👍

PR #684, opened 2022-07-22
Use a string key for responses instead of a list

evilstreak:
We're sending complete snapshots, so we don't need to keep intermediate states. The downside of putting the complete states onto a regular list is that we're adding monotonically increasing snapshots to it. In extreme cases, the size and number of those items can overwhelm the Redis server.

Using a string key should also be slightly more resilient than popping off a list, as we don't lose the latest state if something goes wrong.

## TODO

- [x] Add tests
- [x] Update documentation
- [x] Finish serving support for Replicate

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -92,13 +92,15 @@ If the model yields [progressive output](python.md#progressive-output) then a mi
         ]
     }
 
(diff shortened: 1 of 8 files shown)


andreasjansson (approved):
This looks great! The `response_iterator` makes the tests really nice and readable.

Issue #694, opened 2022-08-04
Latest tensorflow has no compatible CUDA versions


FrobtheBuilder:
I'm trying to wrap a model I made with Keras, but I can't get `cog build` to work because apparently the version of tensorflow I'm using: `tensorflow==2.9.1` (the latest version) has no compatible CUDA versions. I can't override it manually either by specifying a version in my cog.yaml file, same error either way. Why is this? I know for a fact my 11.7 installation works fine with this version of Tensorflow.
![image](https://user-images.githubusercontent.com/2354433/182748694-85f3c30c-932d-48f1-9c88-0e7865ea4559.png)

bfirsh:
Thanks for the report!

Cog 0.4.1 has been pushed with a couple of fixes for this. We now support Tensorflow 2.9.0 now (but not 2.9.1 because that doesn't seem to be picked up by our version compatibility scraper yet). https://github.com/replicate/cog/pull/697

We now also support unsupported tensorflow versions so the cog.yaml you posted above will now work. https://github.com/replicate/cog/pull/696 You'll need to use CUDA 11.6 though, because 11.7 isn't supported by the NVIDIA Docker base images yet.

FrobtheBuilder:
>You'll need to use CUDA 11.6 though, because 11.7 isn't supported by the NVIDIA Docker base images yet.

b-but...
![image](https://user-images.githubusercontent.com/2354433/182961547-4aa5f67b-1321-4d86-a54a-fd13aa89894a.png)

bfirsh:
Seems to be broken then. Would appreciate help if you were willing to do a PR. :) https://github.com/replicate/cog/issues/700

PR #696, opened 2022-08-04
Allow unsupported tensorflow versions

### LINKED ISSUE #181: Don't be tied hard to new Tensorflow/PyTorch versions

bfirsh:
Tensorflow counterpart of #229.
Closes #181

This code is quite spaghetti-like now, so there are a few random bodges in there to make the errors work right. I just bashed away at this until the tests passed and the project reported in #694 worked. Could do with refactor at some point, as mentioned in #181.

### DIFF
--- a/pkg/config/compatibility.go
+++ b/pkg/config/compatibility.go
@@ -138,7 +138,7 @@ func cudaFromTF(ver string) (cuda string, cuDNN string, err error) {
 			return compat.CUDA, compat.CuDNN, nil
 		}
 	}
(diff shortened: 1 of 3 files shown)


bfirsh:
Getting this in and shipped because it should be pretty minor.

Issue #535, opened 2022-04-04
Pick the latest ubuntu version for GPU images


bfirsh:
There are various different Ubuntu versions for the Nvidia images (`-ubuntu18.04`, -ubuntu20.04`, etc). Cog non-deterministically picks these, instead of the latest. It should pick the latest, or at least be consistent.

anish-sahoo:
Closing as CUDA base image selection is now deterministic, choosing the latest matching Ubuntu image.

PR #697, opened 2022-08-04
Deterministically select base images & update compatibility matrices

bfirsh:
I tried updating base images with `make generate` but tests broke because we don't deterministically pick Ubuntu image.

Cloes #535
Ref #694


bfirsh:
Getting this in and shipped because it should be pretty minor.

PR #705, opened 2022-08-23
Output multiple files

bfirsh:
Lots of models output multiple files -- either progressive output or batches of files.

This pull request makes `cog predict` output them as `output.0.png`, `output.1.png`, etc. The test is pretty illustrative.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -12,7 +12,8 @@ Tip: Run [`cog init`](getting-started-own-model#initialization) to generate an a
     - [Progressive output](#progressive-output)
 - [`Input(**kwargs)`](#inputkwargs)
 - [Output](#output)
-    - [Optional properties](#optional-output-properties)
+    - [Returning an object](#returning-an-object)
+    - [Returning a list](#returning-a-list)
 - [Input and output types](#input-and-output-types)
 - [`File()`](#file)
 - [`Path()`](#path)
@@ -79,6 +80,8 @@ class Predictor(BasePredictor):
             yield Path(output_path)
 ```
 
+
+
 ## `Input(**kwargs)`
 
 Use cog's `Input()` function to define each of the parameters in your `predict()` method:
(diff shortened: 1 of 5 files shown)


pkg/cli/predict.go
```diff
@@ -195,17 +207,39 @@ func predictIndividualInputs(predictor predict.Predictor, inputFlags []string, o
...
+
+	for i, output := range outputs {
+		outputString := output.(string)
+		dataurlObj, err := dataurl.DecodeString(outputString)
+		if err != nil {
+			return fmt.Errorf("Failed to decode dataurl: %w", err)
+		}
+		out := dataurlObj.Data
+		extension := mime.ExtensionByType(dataurlObj.ContentType())
+		outputPath := fmt.Sprintf("output.%d%s", i, extension)
```
zeke, line 234: Do we wanna put some leading zeros on this file format so the files will sort alphabetically beyond 9? `output.000.png`

zeke (approved):
I took the liberty of adding some docs.

Inline question about leading zeros on the file format, but otherwise this looks good to go.

docs/python.md
```diff
@@ -145,6 +150,27 @@ class Predictor(BasePredictor):
...
+        output = []
+        for i, prediction in enumerate(predictions):
+            out_path = Path(f"/tmp/out-{i}.txt")
+            with out_path.open("w") as f:
+                f.write(prediction)
+            output.append(out_path)
+        return output
+```
+
+Files are named in the format `output.<index>.<extension>`, e.g. `output.0.txt`, `output.1.txt`, and `output.2.txt` from the example above.
```
bfirsh, line 172: Not a blcoker to shipping, but this is a bit out of context -- nothing mentions `cog predict` anywhere.

bfirsh: Maybe this should live in the CLI documentation, but is currently just `cog predict --help`.

zeke: Ah yeah good point. Should we take this last line out and add a TODO on https://github.com/replicate/cog/issues/433?

PR #714, opened 2022-09-07
Resolve dependencies in `python_requirements`

bfirsh:
This is a partial implementation of #157 that is mostly backwards compatible.

Previously, we read `python_packages`, did our clever processing on it to resolve PyTorch/Tensorflow versions, then inserted it into the `Dockerfile` as a long `pip install` line. There was also an undocumented `python_requirements` option that naively did `pip install -r requirements.txt` without any dependency resolution.

In #157, we want to migrate to using `python_requirements` instead, and do our clever processing on `requirements.txt`. This pull request is a step towards that.

It does a few things:

- `python_requirements` is read into the system that does our clever dependency resolution. Lines that are not able to be processed by Cog (anything that's not `package==version`) are ignored.
- `python_packages` is fed into the same system for backwards compatibility, as before.
- The requirements are written to a temporary `requirements.txt` file instead of a long `pip install` command, which makes it more robust and means we aren't trying to do shell escaping.

We're still missing a few things like informational messages to the user and documentation. It's backwards compatible-ish, so should be fine to leave that. The rest of the work is tracked in #157.

Commits are significant and deliberate.

## Backwards incompatibility

The minor thing that is not backwards incompatible is that in previous versions of Cog, if you did something like `torch>1.12`, Cog would complain hard that it did not understand it.

In this PR, it will silently be passed straight through and Cog won't do any dependency resolution. #157 intends to add some more useful messages to tell the user what is going on here.

## Todo

- [x] Define Python requirements with `requirements.txt` and `python_requirements`
- [x] Make `python_packages` generate a `requirements.txt` for backwards compatibility
- [x] Parse `python_requirements` to determine versions

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -163,7 +163,8 @@ func (c *Config) ValidateAndComplete(projectDir string) error {
 	return nil
 }
 
-func (c *Config) PythonPackagesForArch(goos string, goarch string) (packages []string, indexURLs []string, err error) {
+// PythonRequirementsForArch returns a requirements.txt file with all the GPU packages resolved for given OS and architecture.
+func (c *Config) PythonRequirementsForArch(goos string, goarch string) (packages []string, indexURLs []string, err error) {
 	packages = []string{}
 	indexURLSet := map[string]bool{}
 	for _, pkg := range c.Build.pythonRequirementsContent {
@@ -183,6 +184,8 @@ func (c *Config) PythonPackagesForArch(goos string, goarch string) (packages []s
 	return packages, indexURLs, nil
 }
 
+// pythonPackageForArch takes a package==version line and
+// returns a package==version and index URL resolved to the correct GPU package for the given OS and architecture
 func (c *Config) pythonPackageForArch(pkg string, goos string, goarch string) (actualPackage string, indexURL string, err error) {
 	name, version, err := splitPythonPackage(pkg)
 	if err != nil {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -164,13 +164,13 @@ func (c *Config) ValidateAndComplete(projectDir string) error {
 }
 
 // PythonRequirementsForArch returns a requirements.txt file with all the GPU packages resolved for given OS and architecture.
-func (c *Config) PythonRequirementsForArch(goos string, goarch string) ([]string, error) {
+func (c *Config) PythonRequirementsForArch(goos string, goarch string) (string, error) {
 	packages := []string{}
 	indexURLSet := map[string]bool{}
 	for _, pkg := range c.Build.pythonRequirementsContent {
 		archPkg, indexURL, err := c.pythonPackageForArch(pkg, goos, goarch)
 		if err != nil {
-			return nil, err
+			return "", err
 		}
 		packages = append(packages, archPkg)
 		if indexURL != "" {
@@ -188,7 +188,7 @@ func (c *Config) PythonRequirementsForArch(goos string, goarch string) ([]string
 	// Then, everything else
 	lines = append(lines, packages...)
 
-	return lines, nil
+	return strings.Join(lines, "\n"), nil
 }
 
 // pythonPackageForArch takes a package==version line and
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -124,47 +126,68 @@ func (c *Config) ValidateAndCompleteConfig() error {
...
-		return err
+	if len(c.Build.PythonPackages) > 0 && c.Build.PythonRequirements != "" {
+		return fmt.Errorf("Only one of python_packages or python_requirements can be set in your cog.yaml, not both")
 	}
 
-	if c.Build.GPU {
-		if err := c.validateAndCompleteCUDA(); err != nil {
+	// Load python_requirements into memory to simplify reading it multiple times
+	if c.Build.PythonRequirements != "" {
+		b, err := os.ReadFile(path.Join(projectDir, c.Build.PythonRequirements))
```
nickstenning, line 135: It might be worth considering `bufio.Scanner` here to handle the possibility that someone's checked in a file with CRLF endings.

pkg/config/config.go
```diff
@@ -124,47 +126,68 @@ func (c *Config) ValidateAndCompleteConfig() error {
...
 		}
 	}
-	indexURLs = []string{}
+
+	// Create final requirements.txt output
+	// Put index URLs first
+	lines := []string{}
 	for indexURL := range indexURLSet {
-		indexURLs = append(indexURLs, indexURL)
+		lines = append(lines, "--find-links "+indexURL)
```
nickstenning, line 175: Do we want to take the opportunity to clean up the nomenclature confusion here (`--index-url` is a different thing to `--find-links`) or is that more trouble than it's worth?

pkg/config/load_test.go
```diff
@@ -34,20 +34,12 @@ func TestGetConfigShouldLoadFromCustomDir(t *testing.T) {
...
+	err = ioutil.WriteFile(path.Join(dir, "requirements.txt"), []byte("torch==1.0.0"), 0o644)
+	require.NoError(t, err)
 	conf, _, err := GetConfig(dir)
 	require.NoError(t, err)
-	want := &Config{
-		Predict: "predict.py:SomePredictor",
-		Build: &Build{
-			PythonVersion:      "3.8",
-			PythonRequirements: "requirements.txt",
-			SystemPackages: []string{
```
nickstenning, line 44: Do we care about no longer checking `SystemPackages` here?

nickstenning (commented):
With the (substantial) caveat that I understand next to nothing about the torch/cuda rewriting magic that we do here, this looks mostly sensible to me.

One thing I think we _should_ do as you've mentioned in the PR description is emit warnings for situations that aren't expected (unpinned requirements, `torch` lines that already have a CUDA version, etc.).

pkg/config/config.go
bfirsh: We need `--index-url` for Torch 1.12/1.13 I think, so let's maybe clean up in a follow-up PR for that.

pkg/dockerfile/generator.go
```diff
@@ -250,6 +233,19 @@ This is the offending line: %s`, run)
...
+// It returns the lines to add to Dockerfile to make it available and the filename it ends up as inside the container
+func (g *Generator) writeTemp(filename string, contents []byte) ([]string, string, error) {
+	path := filepath.Join(g.tmpDir, filename)
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return []string{}, "", fmt.Errorf("Failed to write %s: %w", filename, err)
+	}
+	if err := os.WriteFile(path, contents, 0o644); err != nil {
+		return []string{}, "", fmt.Errorf("Failed to write %s: %w", filename, err)
+	}
+	return []string{fmt.Sprintf("COPY %s /tmp/%s", filepath.Join(g.relativeTmpDir, filename), filename)}, "/tmp/" + filename, nil
```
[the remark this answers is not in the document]
bfirsh: Hmm, perhaps. This is kind of normal in Docker builds isn't it for temporary things that are needed in the build process?

pkg/config/load_test.go
bfirsh: I don't think so -- this test isn't testing this. I was intentionally making it less brittle.

pkg/config/config.go
bfirsh: Good idea! Fixed.

PR #721, opened 2022-09-20
Push package to PyPi

### LINKED ISSUE #146: Automatically publish Python package on PyPi on release
Currently there is a package that has been manually released, but it is not automated.

https://pypi.org/project/cog/

We should also set some expectation around what this package is. Just the Python package doesn't actually do very much.

bfirsh:
Push the exact thing we install inside the container to PyPi.

This is not well documented – it's not clear how to use the PyPi package standalone. In short, you can run `python -m cog.server.http`, so long as there's a `cog.yaml` with a `predictor` key in it. Not ideal, but we can fix this up later.

I have tested this little script locally, but not tested the actual push to PyPi obviously. We can fix it live if it doesn't work. I have set the `PYPI_TOKEN` secret.

Closes #146

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -121,3 +121,17 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build Python package
+        run: |
+          # clean package built for go client
+          rm -rf python/dist
+          # set version
+          sed -i -E "s/0.0.1/${GITHUB_REF_NAME:1}/" setup.py
+          # build package
+          python setup.py bdist_wheel
+      - name: Push Python package
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -121,3 +121,17 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build Python package
+        run: |
+          # clean package built for go client
+          rm -rf python/dist
+          # set version
+          sed -i -E "s/0.0.1/${GITHUB_REF_NAME:1}/" setup.py
```
zeke, line 129: What does the `:1` expansion do here?

zeke (approved):
Looks good to me.

.github/workflows/ci.yaml
bfirsh: Removes first character. The tag is `v1.2.3`.

bfirsh:
https://pypi.org/project/cog/

PR #724, opened 2022-09-20
Alternate method for fetching Torch versions

bfirsh:
https://pytorch.org/get-started/previous-versions/ is no longer kept up to date.

The best solution is probably to manually keep the Torch compatibility matrix up to date. Scraping is so messy and error prone, and releases are infrequent enough.

An alternate approach could be to scrape all the possible package versions: https://download.pytorch.org/whl/torch_stable.html

This pull request is the start of an attempt to do that. I stopped half way through because I realized that this page doesn't gives all the information we need except which torch/vision/audio versions are compatible with each other.

Back to manually updating...

### DIFF
--- a/tools/generate_compatibility_matrices/main.go
+++ b/tools/generate_compatibility_matrices/main.go
@@ -170,6 +170,25 @@ type pypiResponse struct {
 	} `json:"info"`
 }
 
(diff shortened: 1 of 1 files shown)


yorickvP:
I'd like to get them from conda

mattt:
Superceded by #812

PR #727, opened 2022-09-20
Send queue worker responses to webhooks

evilstreak:
~~Some test shenanigans to work out still!~~

~~Opening to get feedback on the interface, because it feels a bit Marmite.~~

This is ready to go! Details in the commit message.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -42,7 +42,11 @@ After starting, [the `setup()` method of the predictor](python.md#predictorsetup
 The message body should be a JSON object with the following fields:
 
 - `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
-- `response_queue`: the Redis queue Cog will send responses to
+- `webhook`: the URL Cog will send responses to.
+
+There's also one deprecated field:
+
+- `response_queue`: the Redis key to send responses to; ignored if `webhook` is set.
 
 You can enqueue the request using the `XADD` command:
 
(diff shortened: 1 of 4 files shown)


bfirsh:
We're intending to have a consistent API throughout the whole stack, so could we support the same webhook API as Replicate? https://github.com/replicate/replicate-web/issues/2115

I wonder if we can deprecate putting the response back on a queue for simplicity. Or if we do want to support it, maybe it's a separate special case, hidden behind an extensions key.

More generally, I'm thinking there's a subset of the API we want all things to support (Replicate, Cog HTTP, Cog Redis, internal webhooks, etc), but each of these things might have some special cases (Redis needs to do Redis things, Replicate needs to know a version, internal webhooks might need extra metadata, etc) so there should be a way to extend it.

I have written this up in here: https://github.com/replicate/replicate-web/issues/1474

evilstreak:
@bfirsh @andreasjansson This is tidied up based on our sync, and ready to review.

It's backwards compatible, so there shouldn't be an issue with shipping a new version with this change in.

There's probably a nicer way to handle the tests, by running a separate service similar to the `file-upload` service, dumping JSON webhooks received to disk, and then checking those in tests. However, what's here works, and I'd rather ship this than spend several more hours making that alternative work.

andreasjansson (approved):
Looks good!

🤔 Could we keep one or two tests for the legacy response queue format since we'll probably need to keep that around for a little while?

evilstreak:
> 🤔 Could we keep one or two tests for the legacy response queue format since we'll probably need to keep that around for a little while?

Yep, I added one in there already. I figure just the one is probably OK, because the difference in codepath between webhooks and redis keys is pretty minor.

PR #729, opened 2022-09-22
Update response schema

evilstreak:
Two changes:

- Add `started_at` and `completed_at` and `metrics` to the queue worker response, to match [the API provided by Replicate](https://replicate.com/docs/reference/http).
- Echo back, as part of the webhook response, any and all fields send as part of the request.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -198,9 +198,9 @@ class RedisQueueWorker:
                     except Exception as e:
                         response["status"] = Status.FAILED
                         response["error"] = str(e)
-                        response["x-experimental-timestamps"][
-                            "completed_at"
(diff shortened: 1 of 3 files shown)


andreasjansson (approved):
Looks great!

python/cog/server/redis_queue.py
```diff
@@ -242,9 +243,7 @@ def handle_message(
 
         self.runner.run(**input_obj.dict())
 
-        response["x-experimental-timestamps"] = {
-            "started_at": datetime.datetime.now().isoformat()
-        }
+        response["started_at"] = started_at.isoformat() + "Z"
```
andreasjansson, line 246: Maybe `.isoformat() + "Z"` should go in a function `format_datetime` or something, where can also document why we include the "Z" suffix.

PR #730, opened 2022-09-22
Disable flaky stacktrace logging test

evilstreak:
There's what looks like a timing issue with this test, where occasionally the error from the prediction runner will get processed before the log capture has managed to send back logs, and the response is sent back without the stacktrace being logged.

Having the stacktrace is nice, so longer term we should fix this (hence the TODO). For now, having a flaky test on main is something best avoided.


andreasjansson (approved):
👍 no flakey tests!

PR #735, opened 2022-09-28
Send file uploads as a basic PUT

### LINKED ISSUE #734: Cog Redis needs to be able to put files to API's file upload endpoint
It'll be a redirect to a signed GCP URL.

We already have a thing for the HTTP API to PUT files. Might be able to reuse same thing so Redis and HTTP are consistent.

This: 

https://github.com/replicate/cog/blob/12dfcaa4d47adb586b75a60842e22f0e5d3a41f5/python/cog/files.py#L12-L17

Except it needs to put the file as the request body rather than a multipart form. And needs to be able to follow a 307 redirect.

evilstreak:
Using a simple PUT request with the file contents as the body, rather than as multipart/form-data, means we can upload directly to GCS with a little redirecting.

Closes #734 

See also #496. We could take this opportunity to adjust the API, which I've not done here yet.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -15,7 +15,7 @@ The entrypoint to run a Cog model against a queue is `cog.server.redis_queue`. Y
 - `redis_host`: the host your redis server is running on.
 - `redis_port`: the port your redis server is listening on.
 - `input_queue`: the queue the Cog model will listen to for prediction requests. This queue should already exist.
-- `upload_url`: the endpoint Cog will `PUT` output files to. (Note: this will change in the near future. [See this pull request for more details.](https://github.com/replicate/cog/issues/496))
+- `upload_url`: the endpoint base Cog will `PUT` output files to. See below for more details.
 - `consumer_id`: The name the Cog model will use to identify itself in the Redis group (also called "consumer name" by Redis).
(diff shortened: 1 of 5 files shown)


evilstreak:
Has a minor conflict with #729, which will need resolving when we merge these two.

docs/redis.md
```diff
@@ -37,6 +37,12 @@ For example:
 
 After starting, [the `setup()` method of the predictor](python.md#predictorsetup) is called. When setup is finished the model will start polling the input queue for prediction request messages.
 
+### File uploads
+
+Cog will make a PUT request to the `upload_url` provided, with the filename appended. For example, if you provide an `upload_url` of `https://example.com/upload/` then Cog might send a PUT request to `https://example.com/upload/output.png`. After the PUT request is finished, it will include the final URL it sent to as the value in the output.
+
+It respects redirects, so you probably want to issue a 307 redirect to a unique URL that Cog can PUT the file to. This also gives an opportunity to sign the URL, if you need to provide permissions, for example to a GCS or S3 bucket. Cog will strip query strings from the final URL, to get rid of any signing information that might be included.
+
```
bfirsh, line 45: Should these docs live in HTTP docs too?

kris7ian: I think so that's where I was looking for them :)

Issue #713, opened 2022-09-02
CORS fix for use on locahost


GrantSparks:
I had to add the following code to allow me to execute XHR requests from a local html/javascript file.

cog/python/cog/server/http.py

...

from fastapi.middleware.cors import CORSMiddleware

...

    app.add_middleware(
        CORSMiddleware,
        allow_origins=["*"],
        allow_credentials=True,
        allow_methods=["*"],
        allow_headers=["*"],
    )

zeke:
Thanks for sharing, @GrantSparks 👍🏼

andrewnc:
Quick question here, how does this work exactly? I understand the fix, but I installed cog from the binary release and don't have a python version installed on my machine. 

Did you build cog directly from source? What did that process look like? I appreciate any insights here

anotherjesse:
I did an initial pass at integrating this: https://github.com/replicate/cog/pull/737

markphelps:
Closing this as stale because there has been no activity for over 3 years. There was an initial attempt to address this in PR #737. If adding CORS support for localhost development is still needed on the current version of Cog, feel free to reopen or submit a fresh PR.

Issue #687, opened 2022-07-28
feature request: able to serve static files


harshzalavadiya:
Thanks for making cog it makes deploying models very easy

wgen cog model gets generated it automatically creates openapi endpoints, can we add static rule that if there's `public` directory given that while building all files gets copied fastify can serve if public directory is present for example `/public/index.html` can be accessible at `localhost:5000/public/index.html`

or even better except `openapi.json` and `/docs` everything will be inside `public` directly will be accessible as url

this way from simple forms to complex demo applications can be designed and hosted instead of running another container just to host a static file

bfirsh:
Thanks for the suggestion! Related: #648 #649

anish-sahoo:
We have moved over from fastify to a rust-based implementation. If this is still a problem, please open a new ticket.

PR #737, opened 2022-10-01
allow docker to bypass CORS w/env var DISABLE_CORS

anotherjesse:
It isn't documented in the README as it won't work on existing catalog of images.  It is only for new images.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -7,6 +7,7 @@ import types
 from typing import Any, Optional
 
 from fastapi import Body, FastAPI, HTTPException
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel, ValidationError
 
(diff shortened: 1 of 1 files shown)


anotherjesse:
I haven't looked at an automated test yet

anotherjesse:
Thinking about this further it seems like having a command like:

```
cog serve r8.im/zeke/haiku-standard@sha256:78458ba6976fb7d159f9f3a336754d9e8467110c0fb7e8a2c86d93ccdb524f03
```

in `cog serve` you can address issues not directly related to the model:

- handle CORS #713
- static files #687 
- file handling #496 - and additionally where we can upload to GCS/S3/R2 or store in a local filesystem
- make api consistent between requests against a local docker deploy and replicate deployment

Why not do this in python `cog.server.http`

1. baking non-model specific code into models couple / complicate improvements to serving experience / existing cog'd model will always have older versions of "serve" (docker images)
2. move file access/uploading outside of docker images to focus solely on running the model, not integrations beyond model

I haven't looked at the go code beyond determining how the python code is injected during model build time. Perhaps this is a poor idea that I'll see once I try to look at prototyping this.

anotherjesse:
see POC in #738

zeke:
The name `DISABLE_CORS` is a bit ambiguous, as it could be taken to mean "do not allow cross-origin requests" or "allow any cross-origin request". Maybe something more explicit like `BYPASS_CORS` or `CORS_ALLOW_ALL`?

sepehr:
Blocked by this as well. Generally running Cog-powered models on production seem to be very restricted due to the fact that the http server is not modifiable. Allowing to modify the underlying FastAPI server would be greatly helpful.

soshiharami:
I am patching and using it myself. I hope this PR will be incorporated into production soon.

dotKokott:
Is this getting in any time soon? Or is there an alternative way to change CORS settings on the server?

PR #740, opened 2022-10-02
update docs to 1.18 version of Go (used in CI)

anotherjesse:
Contributing current specifies that you should use a different version of golang than the CI system uses


bfirsh:
Nice thanks!

PR #743, opened 2022-10-04
Send logs lines as newline-separated string instead of list

andreasjansson:
This PR makes Cog consistent with the [Replicate API](https://replicate.com/docs/reference/http#get-prediction) by sending logs as a string instead of a list of strings.

Log lines are separated by a single newline character.


evilstreak (approved):
👍

PR #745, opened 2022-10-05
Throttle webhook responses

andreasjansson:
Introduces a new environment variable `COG_THROTTLE_RESPONSE_INTERVAL` that lets the user control the rate at which non-terminal webhooks are sent.

### DIFF
--- a/python/cog/response.py
+++ b/python/cog/response.py
@@ -10,7 +10,7 @@ class Status(str, enum.Enum):
     FAILED = "failed"
 
     @staticmethod
-    def is_terminal(status):
+    def is_terminal(status: "Status") -> bool:
         return status in (Status.SUCCEEDED, Status.FAILED)
 
 
--- a/python/cog/response.py
+++ b/python/cog/response.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import enum
 from typing import Optional, Type, Any
 
@@ -10,7 +11,7 @@ class Status(str, enum.Enum):
     FAILED = "failed"
 
     @staticmethod
-    def is_terminal(status: "Status") -> bool:
+    def is_terminal(status: Status) -> bool:
         return status in (Status.SUCCEEDED, Status.FAILED)
 
 
--- a/python/cog/response.py
+++ b/python/cog/response.py
@@ -9,6 +9,10 @@ class Status(str, enum.Enum):
     SUCCEEDED = "succeeded"
     FAILED = "failed"
 
+    @staticmethod
+    def is_terminal(status):
+        return status in (Status.SUCCEEDED, Status.FAILED)
+
 
 def get_response_type(OutputType: Type[BaseModel]) -> Any:
     class Response(BaseModel):
--- a/python/cog/response.py
+++ b/python/cog/response.py
@@ -1,4 +1,3 @@
-from __future__ import annotations
 import enum
 from typing import Optional, Type, Any
 
@@ -11,7 +10,7 @@ class Status(str, enum.Enum):
     FAILED = "failed"
 
     @staticmethod
-    def is_terminal(status: Status) -> bool:
+    def is_terminal(status: "Status") -> bool:
         return status in (Status.SUCCEEDED, Status.FAILED)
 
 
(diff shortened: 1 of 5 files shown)


python/cog/server/response_throttler.py
```diff
@@ -0,0 +1,22 @@
...
+import os
+
+from ..response import Status
+
+
+class ResponseThrottler:
+    def __init__(self):
+        self.last_sent_response_time = 0
+        self.response_interval = float(
+            os.environ.get("COG_THROTTLE_RESPONSE_INTERVAL", 0)
```
evilstreak, line 11: I feel like this should be passed in as a parameter rather than pulled from the environment. I can imagine wanting to do this at predict time, to be more reactive to managing a cluster of models without having to reload the containers to get a new environment variables. In that situation passing it in makes it more obvious how/where that fit in.

python/cog/server/response_throttler.py
```diff
@@ -0,0 +1,22 @@
...
+        )
+
+    def should_send_response(self, response):
+        if Status.is_terminal(response["status"]):
+            return True
+
+        seconds_since_last_response = time.time() - self.last_sent_response_time
```
evilstreak, line 18: Nit: my instinct is to make `seconds_since_last_response` a method (possibly private).

python/tests/server/test_response_throttler.py
```diff
@@ -0,0 +1,41 @@
+import time
+
+from cog.server.response_throttler import ResponseThrottler
+from cog.response import Status
+
+def test_zero_iterval():
```
evilstreak, line 6:
```suggestion
def test_zero_interval():
```

python/tests/server/test_response_throttler.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_terminal_status():
+    throttler = ResponseThrottler()
+    throttler.response_interval = 10
+
+    assert throttler.should_send_response({"status": Status.PROCESSING})
+    throttler.update_last_sent_response_time()
+    assert not throttler.should_send_response({"status": Status.PROCESSING})
+    throttler.update_last_sent_response_time()
```
evilstreak, line 22: I don't think we need this line

python/tests/server/test_response_throttler.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_nonzero_internal():
+    throttler = ResponseThrottler()
+    throttler.response_interval = 0.2
+
+    assert throttler.should_send_response({"status": Status.PROCESSING})
+    throttler.update_last_sent_response_time()
+    assert not throttler.should_send_response({"status": Status.PROCESSING})
+    throttler.update_last_sent_response_time()
```
evilstreak, line 33: I don't think we need this line

python/tests/server/test_response_throttler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    time.sleep(0.3)
+
+    assert throttler.should_send_response({"status": Status.PROCESSING})
+    throttler.update_last_sent_response_time()
```
evilstreak, line 38: I think we can drop lines from here onwards, for a more focused test.

python/cog/response.py
```diff
@@ -9,6 +9,10 @@ class Status(str, enum.Enum):
     SUCCEEDED = "succeeded"
     FAILED = "failed"
 
+    @staticmethod
+    def is_terminal(status: "Status") -> bool:
```
evilstreak, line 13: I didn't realise you could use strings to do type hinting! 🧠

evilstreak (approved):
Looks great! I added a bunch of reckons on the code and tests, none of which need prevent us from shipping this. Do with them as you will 😅

python/tests/server/test_response_throttler.py
andreasjansson: I added those to make the test work like the class is used in practice. But no strong opinions.

python/tests/server/test_response_throttler.py
andreasjansson: Done!

PR #746, opened 2022-10-05
Handle cancelations in Redis queue worker

### LINKED ISSUE #723: Handle cancellation in Redis queue worker
Replicate currently kills the whole pod when the user cancels a prediction. Instead, we should be able to handle cancellations inside the running process.

One solution: Create a separate per-prediction cancellation key in Redis, and listen for keyspace notifications inside the Redis queue worker. When a cancellation occurs, cancel the running prediction and send a terminal ~~"cancelled"~~ "canceled" status to the output queue/webhook.
* As a first step we can kill the prediction subprocess which will re-run `setup()`.
* A refinement of that would be to listen for a signal inside the `predict()` function and only kill the current prediction (without re-running `setup()`).

evilstreak:
Allow a Redis key to be provided for each prediction. If that key is created at any point during the prediction, the worker will interrupt the predictor and return a terminal `canceled` response.

Closes #723

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -49,6 +49,7 @@ The message body should be a JSON object with the following fields:
 
 - `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
 - `webhook`: the URL Cog will send responses to.
+- `cancel_key` (optional): a Redis key to watch to signal that this prediction should be canceled. If the key is created the prediction will be canceled.
 
 There's also one deprecated field:
 
--- a/python/cog/response.py
+++ b/python/cog/response.py
@@ -8,6 +8,7 @@ class Status(str, enum.Enum):
     PROCESSING = "processing"
     SUCCEEDED = "succeeded"
     FAILED = "failed"
+    CANCELED = "canceled"
 
 
 def get_response_type(OutputType: Type[BaseModel]) -> Any:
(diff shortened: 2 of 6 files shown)


docs/redis.md
```diff
@@ -49,6 +49,7 @@ The message body should be a JSON object with the following fields:
 
 - `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
 - `webhook`: the URL Cog will send responses to.
+- `cancel_key` (optional): a Redis key to watch to signal that this prediction should be canceled. If the key is created the prediction will be canceled.
```
bfirsh, line 52: Hmm – not sure about this API because it's Redis specific but in a generic top-level key on the prediction object.

I've made a note here and we can iterate before we ship the next version of cog: https://github.com/replicate/replicate-web/issues/2146#issuecomment-1269706369

python/cog/server/redis_queue.py
```diff
@@ -274,6 +286,13 @@ def handle_message(
 
             while self.runner.is_processing():
                 # TODO: restructure this to avoid the tight CPU-eating loop
+                if should_cancel():
```
bfirsh, line 289: Does this add any latency by checking for this on each iteration? Not important, but might be something for future optimization.

python/cog/response.py
```diff
@@ -8,6 +8,7 @@ class Status(str, enum.Enum):
     PROCESSING = "processing"
     SUCCEEDED = "succeeded"
     FAILED = "failed"
+    CANCELED = "canceled"
```
bfirsh, line 11: It's been a todo for a while to use these objects in web. Now it's in Go too, perhaps we have a shared OpenAPI spec somewhere.

bfirsh (approved):
All seems sensible! Haven't tested or reviewed the canceling logic in great detail, but assuming it works looks good.

PR #747, opened 2022-10-06
Allow usage of password to connect to Redis

mathyouf:
I have been using an external machine to run Cog, which attempts to connect to a Redis Stream running on a separate machine. A password has been necessary to prevent random access to that machine by things other than this Cog container running externally on another machine.

Let me know if I am misinterpreting how to use this, and if you have recommendations on how to leverage machines other than the one the Redis Stream is hosted on to handle work orders in the queue, or other ways to secure the redis instance that doesn't involve password protecting it. 

Thank you!


zeke:
Hi @mathyouf! Thanks for opening this PR.

This change looks sensible to me, but I'll defer to @evilstreak for a proper review.

evilstreak:
Hey @mathyouf 👋 Thanks for this!

In principle, I'm totally up for this change.

One thing you're missing is adding the parameter to `_queue_worker_from_argv` at the bottom of the file. That's the bit that takes the command line arguments and provides them to the `RedisQueueWorker` initialiser.

I've got two reckons about this:

1. I think this would probably be nicer as a connection string, a la `redis://[[username]:[password]]@localhost:6379/0` instead of multiple separate components. The pypi client can take that URL directly, using [`from_url`](https://redis.readthedocs.io/en/latest/connections.html#redis.Redis.from_url).
2. It would be a lot more robust, and a lot easier to ship changes, if the worker used keyword arguments instead of positional arguments for invoking it. That's a bigger change to make, but once done it makes doing the Redis connection string change in a backwards compatible way much easier.

If you're up for picking up 2, that'd be wonderful. Otherwise, I'll aim to open a PR for that next week.

zeke:
Hey @mathyouf 👋🏼 

I just kicked off CI for this and it looks like it failed:

```
python/cog/server/redis_queue.py:544: error: Argument 2 to "_queue_worker_from_argv" has incompatible type "*List[str]"; expected "int"
```

See https://github.com/replicate/cog/actions/runs/3304036915/jobs/5494748346#step:9:24

evilstreak:
I've opened #775 to switch to keyword arguments. If you base your change on top of that, it should be straightforward to add a `--redis-url` argument. If `--redis-host` and `--redis-port` are provided instead, you can manually constructing the Redis connection URL from those two arguments.

evilstreak:
Closing this for now. Very happy with the change in principle though, if anyone wants to add this in the future.

PR #748, opened 2022-10-06
Retry webhook for terminal responses

### LINKED ISSUE #742: If webhook fails, model blows up
It should retry with back-off and eventually fail gracefully so model doesn't get stuck in reboot loop.

evilstreak:
For the terminal response, sending the webhook is pretty important. Retry up to a dozen times over the course of 5 minutes, with exponential backoff. This will get through temporary issues very quickly (5 retries within ~3 seconds) and through longer lived issues eventually, with progressively less pressure on the downstream service.

For non-terminal responses, we just try them once. Another response will be along soon enough, and as we're sending full snapshots the loss of an intermediate response isn't serious.

Fixes #742

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -16,6 +16,8 @@ import contextlib
 from pydantic import ValidationError
 import redis
 import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
 
 from ..predictor import BasePredictor, get_input_type, load_predictor, load_config
 from ..json import upload_files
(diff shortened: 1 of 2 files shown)


andreasjansson (approved):
Lovely!

bfirsh:
It retries in the background so it doesn't block other jobs, right?

evilstreak:
> It retries in the background so it doesn't block other jobs, right?

Sort of. The predictor is running in a separate process, so that'll continue without blocking. This blocks all the work of the queue worker though, which means it won't be processing any more outputs (uploading files, sending more webhooks) until this resolves. It also means it won't process a cancellation, as per #746, until this resolves.

Perhaps more problematically, because the queue worker is providing timestamps, rather than the runner in the separate process, it could affect the completed at timestamp.

I'd like to refactor the runner to provide full response objects, rather than just raw output and logs, which would address that. It's not a trivial piece of work though.

PR #756, opened 2022-10-13
Fix deadlock for large outputs

evilstreak:
`pipe.send()` is non-blocking when you send less than 64K of data. When you send more than that, the call blocks until something calls `recv()`. That causes a deadlock, where the predictor process is blocked on `send()`, and the parent process is waiting for it to send the `PROCESSING_DONE` signal.

This rejigs the non-generator output path to ignore the `PROCESSING_DONE` signal, because there's only going to be a single output.

I don't think there's a deadlock in the generator path, but I haven't tested it to be sure.

This whole lot of code is due a rewrite pretty soon anyway!

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -338,19 +338,6 @@ class RedisQueueWorker:
             send_response(response)
 
         else:
-            # just send logs until output ends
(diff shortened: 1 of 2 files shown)


nickstenning (approved):
LGTM

PR #760, opened 2022-10-14
Handle dead predictor process

andreasjansson:
Continuously check if the runner has a subprocess that's alive, and return an error if not.


evilstreak (approved):
Great! 🐛

PR #779, opened 2022-11-01
python-openssl system package not used

anotherjesse:
- container is built with pyenv
- pyenv doesn't need python-openssl installed to function
- appears that cog models do not rely on a side-effect of python-openssl install (libssl-dev is installed explicitly)
- if cog models need python-openssl, it can be installed by updating the model's cog.yaml

```
    python_packages:
      - "pyOpenSSL==22.1.0"
```

This change is because python-openssl was renamed to python3-openssl in more recent versions of ubuntu


nickstenning:
Nice, thank you!

zeke:
@allcontributors please add @anotherjesse for docs, code, tests

PR #780, opened 2022-11-01
Remove python_extra_index_urls and python_find_links

bfirsh:
Undocumented, and GitHub code search implies unused. If you want to do this, you can use `run`, and when we have #714 you can put it in `requirements.txt`.

Yanked out of #714

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,17 +17,15 @@ import (
 // TODO(andreas): suggest valid torchvision versions (e.g. if the user wants to use 0.8.0, suggest 0.8.1)
 
 type Build struct {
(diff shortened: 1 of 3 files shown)


zeke (commented):
Here's what I'm gathering from this:

- `pip` has these two options that can be used during install.
- we were allowing users to configure these options in `cog.yaml`
- we're yanking them out in favor of letting users define that stuff right in `requirements.txt`

Is that correct?

bfirsh:
Yep! Can define it in requirements.txt when #714 lands. In the meantime you can just do a `run` command, which has what people have typically need doing.

zeke (approved):
👍🏼

PR #781, opened 2022-11-01
Use ValidateAndCompleteConfig in config tests

bfirsh:
Just a more complete tests. Also surfaced a couple of little test bugs.

Cherry picked from #714


zeke (approved):
Looks good.

PR #782, opened 2022-11-01
handle case where runner dies during setup

anotherjesse:
in cases where runner OOMs or fails during setup, worker would loop forever waiting for "ready"


nickstenning:
I'm going to close this as we completely rewrote the runner/worker piece of cog in #795 (and should have fixed this issue there, too!).

Issue #783, opened 2022-11-01
Share request/response objects between HTTP and Redis


bfirsh:
It is intended that the HTTP and Redis request/response objects are consistent. Currently they independently generate their own request and response objects, but we could enforce this consistency in code by using the same request/response Pydantic objects between the two implementations.

Aryan8912:
please assign to me so I can work on this issue

PR #784, opened 2022-11-02
Use shared prediction object instead of arbitrary dicts

bfirsh:
Playing around with the idea of having a shared `Prediction` object as a potential implementation of #783.

Another line of thinking about having this object defined in code is you could imagine this being a base object for various predictions all the way across the stack (Cog Python clients, Replicate clients, etc).

Got the first redis test almost working, except running into an issue where our Pydantic objects are helpfully downloading the input files and rewriting them to be local files, which is not the thing we want to return as the response. Needs a bit of thought as to how this could work.

### DIFF
--- a/python/cog/prediction.py
+++ b/python/cog/prediction.py
@@ -1,0 +1,50 @@
+from datetime import datetime
+import enum
+from typing import Optional, Type, Any
+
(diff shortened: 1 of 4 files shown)


bfirsh:
Yup!

PR #791, opened 2022-11-05
Address timing issue with cancel signal

evilstreak:
We're planning to rewrite this subprocess communication Real Soon Now, but there's a bug with cancellation timing we need to fix.

If the runner gets sent a cancel signal (SIGUSR1) whilst it's outside of the _run_prediction try block, the exception isn't caught and it's then possible for the next prediction to end up with incorrect input/output.

Rejigging where the cancel signal is registered, and where it's caught, should make the runner behave as expected.


nickstenning (approved):
LGTM

PR #795, opened 2022-11-11
New Worker implementation (replaces Runner)

nickstenning:
This commit replaces the existing Runner with a completely rewritten Worker process. The main motivation is to fix a cancellation-related bug where prediction outputs from subsequent predictions occasionally get mixed up one with another (see also https://github.com/replicate/cog/commit/ae2cb4474af711f162dbc79d986c5566aaf6a943).

We decided to approach this as a complete rewrite largely because the existing Runner code was fragile, and this approach has allowed us to write a bunch of tests to verify the behaviour of Worker.
Here is a summary of the most important changes from the old code:

1. The predictor to run is passed into the Worker as a string (the value of `predict` from `cog.yaml`), so that the loading of predictor code can happen in the subprocess, allowing us to detect and handle a broader class of failures (e.g. SystemExit, SIGKILL) that might occur at setup time.

2. All communication with the worker subprocess occurs over a single bidirectional pipe object, and this is an implementation detail entirely hidden between the public API of Worker.

3. All outputs from the worker process -- including prediction outputs, log lines, and terminal status -- are represented by named objects (see `cog.server.eventtypes`) which are yielded from Worker's `predict(...)` method.

   This greatly simplifies the code for handling each prediction. For example, handling all prediction log output is now just two lines of code in `cog.server.redis_worker`.

4. Worker no longer needs to handle timeout code, this can be coordinated from the top level (`cog.server.redis_worker`) in much the same way as user-requested cancelation.

It's also worth noting a little bit about the testing strategy for Worker. We have a handful of unit tests covering specific behaviours, but there is also a test (written with [Hypothesis]) which performs randomly-generated sequences of operations on Worker in an attempt to break it or put it into an unknown state. This has helped us find and fix a number of synchronization/concurrency-related bugs.

[Hypothesis]: https://hypothesis.readthedocs.io/en/latest/

Lastly, I'll just note here that one of the longer-term goals of this work is to continue decoupling the code responsible for fetching work (currently, the code interacting with Redis) with the code responsible for orchestrating Worker. A very rough and ready sketch of this might look something like the following:

```python
class Arbiter:

    def __init__(self, work_queue, worker):
        self._queue = work_queue
        self._worker = worker
        self._terminating = False

        signal.signal(signal.SIGTERM, lambda _, _: self._terminating = True)

    def loop(self):
        self._setup()

        while not self._terminating:
            item = self._queue.next()
            self._predict(item)

        self._queue.shutdown()
        self._worker.shutdown()

        sys.exit(143)

    def _setup(self):
        for event in self.worker.setup():
            match event:
                case Log():
                    # send logs places
                case Done():
                    # set error status?

    def _predict(self, item):
        for event in self.worker.predict(item.payload):
            match event:
                case Heartbeat():
                    # check for cancelation, etc.
                case Log():
                    # send logs places
                case Output():
                    # accumulate outputs, start uploads?
                case Done():
                    # set error/cancellation status
```


nickstenning:
For posterity, although it _looks_ like we added a bunch of code, that's mainly tests. :)

### Code
```
 python/cog/server/eventtypes.py         |  45 +++++
 python/cog/server/exceptions.py         |  10 +
 python/cog/server/helpers.py            | 153 ++++++++++++++
 python/cog/server/http.py               |  10 +-
 python/cog/server/log_capture.py        | 124 ------------
 python/cog/server/redis_queue.py        | 289 ++++++++++----------------
 python/cog/server/response_throttler.py |   2 +-
 python/cog/server/runner.py             | 348 --------------------------------
 python/cog/server/webhook.py            |  71 +++++++
 python/cog/server/worker.py             | 238 ++++++++++++++++++++++
 10 files changed, 633 insertions(+), 657 deletions(-)
```

### Tests
```
 python/tests/server/fixtures/catch_in_predict.py   |  11 +
 python/tests/server/fixtures/complex_output.py     |  14 +
 python/tests/server/fixtures/count_up.py           |   7 +
 python/tests/server/fixtures/exc_in_predict.py     |   6 +
 python/tests/server/fixtures/exc_in_setup.py       |   6 +
 .../server/fixtures/exc_in_setup_and_predict.py    |   6 +
 python/tests/server/fixtures/exc_on_import.py      |   1 +
 python/tests/server/fixtures/exit_in_predict.py    |   8 +
 python/tests/server/fixtures/exit_in_setup.py      |   8 +
 python/tests/server/fixtures/exit_on_import.py     |   3 +
 python/tests/server/fixtures/hello_world.py        |   7 +
 python/tests/server/fixtures/killed_in_predict.py  |   9 +
 python/tests/server/fixtures/logging.py            |  34 ++
 python/tests/server/fixtures/missing_predict.py    |   5 +
 python/tests/server/fixtures/missing_predictor.py  |   0
 python/tests/server/fixtures/missing_setup.py      |   5 +
 python/tests/server/fixtures/simple.py             |   6 +
 python/tests/server/fixtures/sleep.py              |   9 +
 python/tests/server/fixtures/slow_predict.py       |  11 +
 python/tests/server/fixtures/steps.py              |  13 +
 python/tests/server/test_helpers.py                | 185 +++++++
 python/tests/server/test_http.py                   |   7 +-
 python/tests/server/test_http_input.py             |  13 +-
 python/tests/server/test_http_output.py            |   5 +-
 python/tests/server/test_response_throttler.py     |   2 +-
 python/tests/server/test_webhook.py                |  73 +++
 python/tests/server/test_worker.py                 | 559 +++++++++++++++++++++
 27 files changed, 1000 insertions(+), 13 deletions(-)
```

PR #799, opened 2022-11-14
Refactor redis_queue to extract response building

evilstreak:
Rejig the functions to end up with a `run_prediction()` function that takes the message and yields responses. My intent here is to wrap up the event stream from `Worker` with something that yields the responses we want to send back to the webhooks. These responses should also be the same thing we want to return to HTTP requests, allowing us to reuse this wrapping.

There shouldn't be any behaviour change in this PR.

It feels like it should probably live in a separate class (and be unit tested!), though I'm not sure what that'd be called.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -11,7 +11,7 @@ import types
 from argparse import ArgumentParser
 from mimetypes import guess_type
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple
 from urllib.parse import urlparse
 
 import redis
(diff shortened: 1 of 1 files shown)


python/cog/server/redis_queue.py
```diff
@@ -189,80 +189,49 @@ def start(self) -> None:
...
-            span.record_exception(e)
-            span.set_status(TraceStatus(status_code=StatusCode.ERROR))
-            return
+            yield response
 
-        cleanup_functions.append(input_obj.cleanup)
+            try:
+                input_obj.cleanup()
+            except Exception as e:
+                sys.stderr.write(f"Cleanup function caught error: {e}")
```
nickstenning, line 227: This should ideally be `print(..., file=sys.stderr)` ideally to get its autoflush behaviour.

python/cog/server/redis_queue.py
```diff
@@ -189,80 +189,49 @@ def start(self) -> None:
...
+        response["logs"] = ""
+
         started_at = datetime.datetime.now()
 
         try:
             input_obj = self.InputType(**response["input"])
-        except ValidationError as e:
-            tb = traceback.format_exc()
-            sys.stderr.write(tb)
+        except Exception as e:
```
nickstenning, line 219: I guess the question of what to do if we get a non-`ValidationError` exception here in terms of flagging it as an infrastructure failure is one we can keep for our future selves...

python/cog/server/redis_queue.py
```diff
@@ -309,14 +278,15 @@ def handle_message(
...
-                        send_response(response)
+                        yield response
                     else:
                         assert (
                             response["output"] is None
                         ), "Predictor unexpectedly returned multiple outputs"
                         response["output"] = output
 
                 elif isinstance(event, Done):
+                    # TODO assert we only get one Done
```
nickstenning, line 289: I mean to be honest we can just do it here! `assert not done_event` should be fine here and we probably should crash the worker if it happens.

python/cog/server/redis_queue.py
```diff
@@ -347,7 +317,12 @@ def handle_message(
             response["status"] = Status.FAILED
             response["error"] = str(e)
         finally:
-            send_response(response)
+            yield response
+
+            try:
+                input_obj.cleanup()
```
nickstenning, line 323: This shouldn't still be here, should it?

python/cog/server/redis_queue.py
```diff
@@ -189,80 +189,49 @@ def start(self) -> None:
...
             response["error"] = str(e)
-            send_response(response)
-            span.record_exception(e)
-            span.set_status(TraceStatus(status_code=StatusCode.ERROR))
-            return
+            yield response
 
-        cleanup_functions.append(input_obj.cleanup)
+            try:
+                input_obj.cleanup()
```
nickstenning, line 225: I'm not sure I know what the implications are of calling this before we pass `input_obj` into `.predict(...)`. Is that guaranteed to be safe?

nickstenning (commented):
This is a nice clarity improvement. Left a handful of comments!

python/cog/server/redis_queue.py
evilstreak: Longer term, I think any validation errors should be considered infrastructure failures, because we should be validating inputs against the OpenAPI schema before it ever reaches a Cog worker.

python/cog/server/redis_queue.py
evilstreak: This is only called if input validation fails, and we're about to bail out.

python/cog/server/redis_queue.py
evilstreak: The call to `input_obj.cleanup()` further up is only run if the input validation fails – this one handles the case where validation passed and we actually ran the prediction.

python/cog/server/redis_queue.py
evilstreak: It definitely shows it could be clearer! Tbh, the main reason I didn't do go further on inputs is because we talked about moving that inside the predictor anyway, at which point this code is going away.

nickstenning (approved):
LGTM!

PR #800, opened 2022-11-14
Add option to exit redis worker after too many failures

evilstreak:
We're occasionally seeing models get into a zombie state in Replicate, where the model will keep accepting jobs but it fails every single one. When it's failing them very quickly, and survives as the final worker after the model is downscaled, we end up serving only failures for the affected model.

This change allows a maximum consecutive number of failures to be set. If the worker fails more than that many predictions in a row, it'll exit, forcing a new (hopefully healthy) worker to be started to take its place.

---

Opening this as a draft because it's based on top of #799, which will need reviewing first.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -59,6 +59,7 @@ class RedisQueueWorker:
         predict_timeout: Optional[int] = None,
         redis_db: int = 0,
         report_setup_run_url: Optional[str] = None,
+        max_failure_count: Optional[int] = None,
     ):
         self.worker = Worker(predictor_ref)
         self.redis_host = redis_host
@@ -71,6 +72,7 @@ class RedisQueueWorker:
         self.predict_timeout = predict_timeout
         self.redis_db = redis_db
         self.report_setup_run_url = report_setup_run_url
+        self.max_failure_count = max_failure_count
         if self.predict_timeout is not None:
             # 30s grace period allows final responses to be sent and job to be acked
             self.autoclaim_messages_after = self.predict_timeout + 30
@@ -181,6 +183,8 @@ class RedisQueueWorker:
             )
             sys.stderr.write(f"Setup time: {setup_time:.2f}\n")
 
+        failure_count = 0
+
         sys.stderr.write(f"Waiting for message on {self.input_queue}\n")
         while not self.should_exit:
             try:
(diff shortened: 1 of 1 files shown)


python/cog/server/redis_queue.py
```diff
@@ -192,6 +196,17 @@ def start(self) -> None:
...
                         send_response(response)
 
+                    if self.max_failure_count is not None:
+                        # Keep track of runs of failures to catch the situation
+                        # where the worker has gotten into a bad state where it can
+                        # only fail predictions, but isn't exiting.
+                        if response["status"] == Status.FAILED:
+                            failure_count += 1
+                            if failure_count > self.max_failure_count:
+                                self.should_exit = True
```
nickstenning, line 206: I know we don't have proper telemetry in here yet, but we should probably at least log something so that we can see why the pod stopped and/or search the logs to find out how often this is happening.

nickstenning (approved):
We should emit some logs to let ourselves know that we hit this code, but otherwise LGTM.

python/cog/server/redis_queue.py
```diff
@@ -57,6 +57,7 @@ def __init__(
         log_queue: Optional[str] = None,
         predict_timeout: Optional[int] = None,
         redis_db: int = 0,
+        max_failure_count: int = None,
```
nickstenning, line 60:
```suggestion
        max_failure_count: Optional[int] = None,
```

python/cog/server/redis_queue.py
```diff
@@ -71,6 +72,7 @@ def __init__(
         self.predict_timeout = predict_timeout
         self.redis_db = redis_db
         self.report_setup_run_url = report_setup_run_url
+        self.max_failure_count = max_failure_count
```
nickstenning, line 75: Did we want to check that it's `> 0` too?

evilstreak: With the implementation here, 0 is a valid number. Negatives aren't, though they'll behave the same as 0.

`max_failure_count` is the maximum number of concurrent failures we'll accept before it exiss. So, 0 would mean exit after any failure. Perhaps it would be clearer renamed to `exit_after_failure_count`, and then making 1 mean exit after any failure (and making 0 invalid).

nickstenning: Ah sorry, yes! Makes sense.

nickstenning (approved):

evilstreak:
I'm not 100% on the name of this config parameter (see comments), but this queue worker is slightly experimental and probably due for obsolescence after some planned refactoring, so shipping this as is for now.

PR #803, opened 2022-11-15
Send timings and logs for setup to a URL

evilstreak:
Expose setup logs in a similar way to prediction outputs.

### DIFF
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -36,7 +36,7 @@ from ..predictor import (
 from ..response import Status
 from .eventtypes import Done, Heartbeat, Log, PredictionOutput, PredictionOutputType
 from .probes import ProbeHelper
-from .webhook import webhook_caller
+from .webhook import requests_session, webhook_caller
 from .worker import Worker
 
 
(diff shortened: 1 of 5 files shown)


python/cog/server/redis_queue.py
```diff
@@ -133,14 +135,34 @@ def receive_message(self) -> Tuple[Optional[str], Optional[str]]:
     def start(self) -> None:
         with self.tracer.start_as_current_span(name="redis_queue.setup") as span:
             signal.signal(signal.SIGTERM, self.signal_exit)
-            start_time = time.time()
+            started_at = datetime.datetime.now()
 
-            # TODO(bfirsh): setup should time out too, but we don't display these logs to the user, so don't timeout to avoid confusion
+            setup_logs = ""
+            # TODO might want to be more defensive consuming this stream
```
nickstenning, line 141: Haha yes, we will! https://github.com/replicate/cog/blob/a2df984a4b8eed9d405bf16e8d30c019a962b169/python/cog/server/worker.py#L126

nickstenning: Which is to say that any setup failure will raise, so we need to wrap the whole thing in a try/except if we want to reliably send logs on failure.

nickstenning (approved):
LGTM

PR #805, opened 2022-11-16
Write a file when setup is complete

nickstenning:
This can be checked by an exec readiness probe and used to control deployment rollout for model pods.


evilstreak (approved):
Great! This is probably worth documenting in docs/redis.md.

zeke:
@allcontributors please add @nickstenning for code, docs, design, infra, tests

PR #807, opened 2022-11-17
Accept list of events to send webhooks for

### LINKED ISSUE #786: Make it possible for queue worker to only send webhooks for new output
Currently the queue worker sends a webhook whenever there's new output **or** new logs. For some use cases, getting those logs immediately is great. For others, it's just noise. It'd be great to have a way to pick between those behaviours by passing a parameter with the queued request.

evilstreak:
The main reason for this is to allow us to not send a webhook every time there's a new log line (which can happen a lot!) when all the consumer cares about is new outputs.

Closes #786.

---

I'm interested in feedback on the API this provides as much as the implementation. `webhook_events` feels like a natural accompaniment to `webhook`, and is something we would want to provide through other interfaces like the HTTP API.

It can also naturally encompass the `webhook_completed` property currently exposed through the Replicate API, by passing `{"webhook": "...", "webhook_events": ["done"]}`.

### DIFF
--- a/docs/redis.md
+++ b/docs/redis.md
@@ -49,6 +49,7 @@ The message body should be a JSON object with the following fields:
 
 - `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
 - `webhook`: the URL Cog will send responses to.
+- `webhook_events_filter` (optional): a list of events to send webhooks for. May contain `start`, `output`, `logs` and `completed`. Defaults to all events. Will always send `completed` events, even if omitted from the list.
 - `cancel_key` (optional): a Redis key to watch to signal that this prediction should be canceled. If the key is created the prediction will be canceled.
 
 There's also one deprecated field:
--- a/python/cog/server/redis_queue.py
+++ b/python/cog/server/redis_queue.py
@@ -33,7 +33,7 @@ from ..predictor import (
     load_config,
     load_predictor_from_ref,
 )
-from ..response import Status
+from ..response import Event, Status
 from .eventtypes import Done, Heartbeat, Log, PredictionOutput, PredictionOutputType
 from .probes import ProbeHelper
 from .webhook import requests_session, webhook_caller
@@ -226,8 +226,17 @@ class RedisQueueWorker:
 
                     should_cancel = self.cancelation_checker(message.get("cancel_key"))
 
-                    for response in self.run_prediction(message, should_cancel):
-                        send_response(response)
+                    if "webhook_events_filter" in message:
+                        # this will raise an exception if the list isn't valid
+                        events_filter = Event.validate(message["webhook_events_filter"])
+                    else:
+                        events_filter = Event.default_events()
+
+                    for response_event, response in self.run_prediction(
+                        message, should_cancel
+                    ):
+                        if response_event in events_filter:
+                            send_response(response)
 
                     if self.max_failure_count is not None:
                         # Keep track of runs of failures to catch the situation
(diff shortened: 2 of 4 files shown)


bfirsh:
Nice. I wonder if this is a "filter"? I wonder if there is any prior art for this.

docs/redis.md
```diff
@@ -49,6 +49,7 @@ The message body should be a JSON object with the following fields:
 
 - `input`: a JSON object with the same keys as the [arguments to the `predict()` function](python.md). Any `File` or `Path` inputs are passed as URLs.
 - `webhook`: the URL Cog will send responses to.
+- `webhook_events` (optional): a list of events to send webhooks for. Must contain `done`, can also contain `start`, `output` and `logs`. Defaults to all events.
```
nickstenning, line 52: I agree with Ben that

- `webhook_events_filter` might be a more self-documenting name for this, and
- `done` should be `completed` as that's the word we use externally.

I'm also not sure how I feel about requiring that the completed event (whatever we call it) should be in this list, as that seems to me to make it easier to provide invalid input for no obvious benefit. I would suggest that we both:

- always send a `completed` event
- allow `completed` in this list

python/cog/server/redis_queue.py
```diff
@@ -245,14 +245,22 @@ def run_prediction(
...
+
         started_at = datetime.datetime.now()
 
         try:
             input_obj = self.InputType(**response["input"])
         except Exception as e:
             response["status"] = Status.FAILED
             response["error"] = str(e)
-            yield response
+            if WebhookEvent.DONE in webhook_events:
```
nickstenning, line 262: I don't think this is a blocking observation for this change, but it feels odd to me that we have inverted control of webhook sending through the conversion of `run_prediction` into a generator method, and now are putting aspects of that back into this method.

It feels like like there might be a nicer refactoring here where all of this gets wrapped up with `WebhookSender` and a `PredictionResponse` class something like this:

```
class WebhookSender:
    def __init__(self, events_filter=None):
        self.events_filter = ...

    def send(self, event, payload):
        ...

class PredictionResponse:
    def __init__(self, data, webhook_sender):
        self.data = data
        self.webhook_sender = webhook_sender

    def start(self):
        self.data['started_at'] = ...
        self.webhook_sender.send("start", self.data)

    def append_logs(self, logs):
        self.data['logs'] += logs
        self.webhook_sender.send("logs", self.data)

    ...
```
What do you reckon?

test-integration/test_integration/test_redis_queue.py
```diff
@@ -2084,6 +2084,80 @@ def capture_response(request):
         )
 
 
+def test_queue_worker_webhook_events(
```
nickstenning, line 2087: Are you sure we want to add an integration test for this? If it's not possible to unit test it I get it, but it seems like maybe a refactoring like the one I suggested above might allow us to gain confidence that it works without having to add another (slow) integration test.

nickstenning (commented):
Generally looks good. Basically two observations from me: one about naming, and the other about how we're implementing the filtering and how to test that it works.

python/cog/server/redis_queue.py
evilstreak: Yeah, that's nicer. I like your proposed refactor, though it's a bigger change than I was planning to make at the moment – I think a simpler step towards it, to keep the inversion of control we have, is to yield an `(event, response)` tuple instead of just yielding the response. That'll allow the caller to decide what to do with the information.

Refactoring into a nicer `PredictionResponse` class (and replacing most of the integration tests) can wait, I think.

test-integration/test_integration/test_redis_queue.py
evilstreak: Going to leave the integration test in for now, awaiting the large refactor described above.

nickstenning (approved):
Nice. 🚀

PR #810, opened 2022-11-25
Expand the BasePredictor API to configure the FastAPI app #713

fozziethebeat:
This adds a new _optional_ method to the `BasePredictor` class: `configure_app`.  Any predictor can implement `configure_app` to change how the Fast API server handles requests.  The documented example shows how to solve issue #713 by allowing for cross origin requests.

With this users can customize how cross origin requests are handled without limitation and without having to rely on complex environment variables.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -24,12 +24,23 @@ You define how Cog runs predictions on your model by defining a class that inher
 
 ```python
 from cog import BasePredictor, Path, Input
+from fastapi.middleware.cors import CORSMiddleware
 import torch
 
 class Predictor(BasePredictor):
     def setup(self):
         """Load the model into memory to make running multiple predictions efficient"""
         self.model = torch.load("weights.pth")
+
+    def configure_api(self, app):
+        """Configure FastAPI CORS."""
+        app.add_middleware(
+            CORSMiddleware,
+            allow_origins=["*"],
+            allow_credentials=True,
+            allow_methods=["*"],
+            allow_headers=["*"],
+        )
 
     def predict(self,
             image: Path = Input(description="Image to enlarge"),
@@ -42,7 +53,7 @@ class Predictor(BasePredictor):
         return output
 ```
 
-Your Predictor class should define two methods: `setup()` and `predict()`.
+Your Predictor class should define `predict()` and optionally `setup()` and `configure_api()`.
 
 ### `Predictor.setup()`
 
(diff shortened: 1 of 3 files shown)


bfirsh:
Nice, thanks!

There's an interesting design question in here. The `Predictor` is not supposed to be tied to any particular runtime.

The intended way to configure the app would be to extend `cog.server.http` somehow rather than modify the predictor. I wonder if there's an elegant way to do that.

fozziethebeat:
Right, this feels odd modifying Predictor to control the server.

A very different approach would be to expand the cog.yaml configuration to take server configs.  This would be similar to how serverless handles this (example below).

That alternate approach wouldn't require changes to BasePredictor, just some hopefully simple changes to the config and cog.server.http

```
  httpApi:
    cors:
      allowedOrigins:
        - https://www.something.com
      allowCredentials: true
      allowedHeaders:
        - authorization
        - auth-provider
        - content-type
        - X-Amz-Date
        - X-Api-Key
        - X-Amz-Security-Token
        - X-Amz-User-Agent
```

fozziethebeat:
@bfirsh, Looking at the current yaml design, I'd suggest expanding the config to have a stanza like the following (mostly copied from serverless)

```
predict: '.....'

server:
  cors:
      allowedOrigins:
        - https://www.something.com
      allowCredentials: true
      allowMethods:
         - GET
      allowedHeaders:
        - authorization
        - auth-provider
        - content-type
        - X-Amz-Date
        - X-Api-Key
        - X-Amz-Security-Token
        - X-Amz-User-Agent
``` 

Each sub-stanza for `cors` just corresponds to a field needed by `CORSMiddleware`.  If the sub-stanza isn't specified it would default to the empty list. 

An example config for a fully open cors setting would then be something like

```    
server:
  cors:
      allowedOrigins:
        - *
      allowCredentials: true
      allowMethods:
         - *
      allowedHeaders:
        - *
```

and a simpler configuration of

```
server:
  cors: true
```

Would use a bunch of defaults (probably equivalent to what serverless does to have consistency with something stable).

The code changes would then be limited to a new method in cog.server.http to pull out the config fields and setup the defaults then add the middleware if desired.

sepehr:
While I agree that configuration of the underlying http server should not be part of the predictor object, I'd vote for a imperative mechanism that liberates the http server for proper production use.

sepehr:
To have a proper seam here, I'd say a configurable CMD would work best:

```yaml
build:

predict:

image:
  cmd: python -m 'myapp.server.http'
```

Where `myapp.server.http` is either an extension of `cog.server.http`, or a complete overwrite if necessary.

PR #812, opened 2022-11-26
Fix PyTorch 1.11, 1.12 and 1.13

### LINKED ISSUE #789: PyTorch 1.12 and 1.13 might need extra arguments for CUDA 11.6
The default PyTorch 1.12 binary doesn't work with CUDA 11.6, it seems.

Screenshot 2022-11-04 at 15 24 32

This is why there's this bodge:

https://github.com/replicate/cog-stable-diffusion/blob/bbedf972a628f7e3befcf8acb36efdb95a9bebc6/cog.yaml#L5-L7

Looking at the PyTorch website, looks like selecting 11.6 you need that argument:

Screenshot 2022-11-04 at 16 27 46

But for 11.7 you don't need the argument:

Screenshot 2022-11-04 at 16 27 15

We should represent this in the compatibility tables.

### LINKED ISSUE #814: Could not find a version that satisfies the requirement `torch==1.11.0+cu113`
Looking to set up [pyannote-audio](https://github.com/pyannote/pyannote-audio#installation) with a specific version of `torch`.

```
build:
  gpu: true
  python_version: "3.8"
  python_packages:
    - "pyannote.audio==2.1.1"
    - "torch==1.11.0"
    - "torchaudio==0.11.0"
    - "torchvision==0.12.0"
  system_packages:
    - "libsndfile1"
```

```
 > [stage-0 8/9] RUN --mount=type=cache,target=/root/.cache/pip pip install   pyannote.audio==2.1.1 torch==1.11.0+cu113 torchaudio==0.11.0 torchvision==0.12.0+cu113:
#14 0.922 Collecting pyannote.audio==2.1.1
#14 0.926   Using cached pyannote.audio-2.1.1-py2.py3-none-any.whl (390 kB)
#14 1.011 ERROR: Could not find a version that satisfies the requirement torch==1.11.0+cu113 (from versions: 1.4.0, 1.5.0, 1.5.1, 1.6.0, 1.7.0, 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.9.1, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0)
#14 1.016 WARNING: You are using pip version 22.0.4; however, version 22.3.1 is available.
#14 1.016 You should consider upgrading via the '/root/.pyenv/versions/3.8.15/bin/python3.8 -m pip install --upgrade pip' command.
```

I believe the warning at the end may be relevant, it looks it may have been recently fixed pypa/pipenv#5022

bfirsh:
The correct install commands for 1.12 and 1.13 weren't being generated, partially because the correct `--extra-index-url` option wasn't being passed, and partially because the way we were fetching the installation instructions for the latest PyTorch wasn't working any longer.

Details are in commits, and this'll be much easier to review commit by commit. Following the output in the compatibility matrix helps to understand what each change does.

I have tested this by pushing cog-stable-diffusion with PyTorch 1.12 automatically resolving and the model runs on Replicate. PyTorch 1.13 defaults to CUDA 11.7 which isn't supported on Replicate yet.

Closes #789 
Closes #814


pkg/config/compatibility.go
```diff
@@ -260,18 +261,18 @@ func tfGPUPackage(ver string, cuda string) (name string, cpuVersion string, err
...
 	for _, compat := range TorchCompatibilityMatrix {
 		if compat.TorchVersion() == ver && compat.CUDA == nil {
-			return "torch", torchStripCPUSuffixForM1(compat.Torch, goos, goarch), compat.FindLinks, nil
+			return "torch", torchStripCPUSuffixForM1(compat.Torch, goos, goarch), compat.FindLinks, compat.ExtraIndexURL, nil
 		}
 	}
 
 	// Fall back to just installing default version. For older pytorch versions, they don't have any CPU versions.
-	return "torch", ver, "", nil
+	return "torch", ver, "", "", nil
```
nickstenning, line 272: Not gonna block but this is definitely getting to the point where this should be a struct :)

nickstenning (approved):
In all honesty I don't understand half of what is going on here beyond "we're scraping the pytorch website to work out which packages are compatible with which other packages" but it all looks reasonably sensible to me.

Next time we add another string to the `(string, string, string, string, error)` return type we should do the work of making that a struct with named fields.

bfirsh:
Some context: https://github.com/replicate/cog/issues/820

Really we should get rid of this thing, but for the sake of getting it working I got it working.

Issue #822, opened 2022-12-02
Create Homebrew formula


mattt:
[Homebrew](https://brew.sh) is a popular way to install developer tools on macOS. It would be great to be able to install `cog` in a single command, like this:

```console
$ brew install cog
```

zeke:
> 1. Can either of you please cut a release when you have a chance?

Fine by me! We've historically kept the main branch releasable, but we've recently had a few situations where changes accumulated on main that required a bit of team coordination before releasing. @evilstreak @bfirsh @nickstenning are we out of the woods now that 0.5.0 has shipped?

zeke:
Hey @mattt. Sorry the delay. I cut a new release! You should be good to go.

https://github.com/replicate/cog/releases/tag/v0.5.1

PR #823, opened 2022-12-02
Refactor `Makefile`

mattt:
This PR implements some changes I needed to get a Homebrew formula working (see https://github.com/replicate/cog/issues/822), specifically:

- Defining constants for executables like `PYTHON`. When Homebrew installs Python 3, my understanding is that it links to `python3`, keeping the default `python` (2.7) executable accessible in the default `PATH`. Because Python 3 is a prerequisite for building from source, we need a way to override this when running `make`.
- Adding an `install` target. In this PR, it's configured to be overridden by conventional `PREFIX` / `BINDIR` constants. It also supports [staged installs](https://www.gnu.org/prep/standards/html_node/DESTDIR.html) using a `DESTDIR` constant.

This PR includes some other changes to keep the rest of the `Makefile` more consistent:

- Adding an `uninstall` target, to keep symmetry with `install`.
- Defining and using constants for `GO` and other executables.
- Adding a new default `all` target that builds `cog` as a prerequisite.
- Adding `go clean` to the `clean` target.

Finally, I noticed a TODO comment from @bfirsh for the `test-integration` target:

> `TODO(bfirsh): use local copy of cog so we don't have to install globally`

With the new `cog` target, we can run integration tests against local executables without installing them by prepending `PWD` to `PATH`. (afbd925bdb2458291493dc6b939d37b9a7d018e5)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,8 +40,8 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-${{ secrets.CACHE_VERSION }}-
-      - name: "Build"
-        run: make build
+      - name: Build
+        run: make cog
       - name: Test
         run: make test-go
 
@@ -118,8 +118,8 @@ jobs:
         with:
           go-version-file: go.mod
           cache: true
-      - name: Build dependencies
-        run: make build-dependencies
+      - name: Build
+        run: make cog
       - uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-release
+cog
 .ipynb_checkpoints/
 Untitled*.ipynb
 __pycache__
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,10 @@ MAIN := cmd/cog/cog.go
 BUILD_TIME := $(shell date +%Y-%m-%dT%H:%M:%S%z)
 LDFLAGS := -ldflags "-X github.com/replicate/cog/pkg/global.Version=$(COG_VERSION) -X github.com/replicate/cog/pkg/global.BuildTime=$(BUILD_TIME) -w"
 
+default: all
 
-default: build
+.PHONY: all
+all: cog
 
 pkg/dockerfile/embed/cog.whl: python/* python/cog/* python/cog/server/* python/cog/command/*
 	@echo "Building Python library"
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,11 @@
 SHELL := /bin/bash
 
+PREFIX = /usr/local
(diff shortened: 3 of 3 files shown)


bfirsh:
😍 thank you!

One thing, which I'm curious if you have opinions on: we were considering just using `script/` files, because we don't really use makefile as a makefile. (I think we build the Python wheel as an actual target, but it's somewhat tenuous).

Any opinions on makefiles full of phonies vs just having some scripts?

zeke:
Nice! I pressed the "Approve and Run" button to kick off CI here.

I know Makefiles are a tried and true old-school hacker convention, but I'm generally more in favor of [scripts to rule them all](https://github.com/github/scripts-to-rule-them-all) because they're just easier to comprehend than Makefiles, with less arcane syntax and none of that phony business. But yeah.. is it worth making the change? I dunno.

Looks like CI is 🍏, but when I ran this locally I hit a snag on `make install`:

```
$ make install
install -m 0755 cog /usr/local/bin/cog
install: /usr/local/bin/cog: Permission denied
make: *** [install] Error 71
```

Maybe just a permissions thing that's unique to my setup? `make install` does work for me on the main branch:

```
$ make install            
go install -ldflags "-X github.com/replicate/cog/pkg/global.Version=v0.5.0+dev -X github.com/replicate/cog/pkg/global.BuildTime=2022-12-04T22:31:18-0800 -w" cmd/cog/cog.go

$ cog --version
cog version v0.5.0+dev (built 2022-12-04T22:31:18-0800)

$ go version
go version go1.19.1 darwin/arm64
```

nickstenning (approved):
Perhaps let's add a `make install-dev` which does the `go install` this was previously doing? That would address @zeke's comment, I think.

Otherwise, LGTM.

zeke:
> But before you do that, could you try running `make install INSTALL=cp` to check that the `install` command isn't the culprit?

```
~/git/replicate/cog pr/823
$ make install INSTALL=cp
CGO_ENABLED=0 go build -ldflags "-X github.com/replicate/cog/pkg/global.Version=v0.5.0+dev -X github.com/replicate/cog/pkg/global.BuildTime=2022-12-05T12:02:57-0800 -w" -o cog cmd/cog/cog.go
cp cog /usr/local/bin/cog
cp: /usr/local/bin/cog: Permission denied
make: *** [install] Error 1
```

> Running `sudo make install` should do the trick.

```
$ sudo make install
Password:
fatal: unsafe repository ('/Users/z/git/replicate/cog' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /Users/z/git/replicate/cog
install -m 0755 cog /usr/local/bin/cog
```

zeke (approved):
That worked!

I'm all for shipping this. If it has any rough edges, or if we decide we want to "add a `make install-dev` which does the `go install` this was previously doing" as @nickstenning suggested, we can iterate.

zeke:
@allcontributors please add @mattt for code, docs, infra

PR #824, opened 2022-12-05
Converge Cog and Replicate HTTP APIs; use Worker in Cog HTTP server.

nickstenning:
This switches the old, synchronous predict API over to use the new `Worker` code for running predictions in a subprocess so we're sharing a common approach to running predictions between the redis worker code and the Cog HTTP API.

At the same time, it defines prediction request and response types, and uses these in the Cog HTTP API (annotating/extending them with the model input and output types as appropriate). This is the first step in making
the Cog HTTP API more similar to Replicate's own API.

The change is intended to be backwards compatible, but as a nod the the possibility that we'll break existing users of the HTTP API despite every effort, we suggest that this gets released at v0.6.0.

There is one slightly surprising wrinkle in this change: with `Worker` on the far side of a process boundary, prediction inputs to must survive a pickle/unpickle. Until now, a validated `File` input contained a file handle, and so could never meet this requirement.

This PR addresses the problem for `File`, by mapping URLs provided for file inputs to `URLFile`, a "file-like" object that lazily executes the HTTP request when it is first used. This works, but does have some downsides:

- `File` and `Path` now behave differently: `File` inputs are downloaded in the predictor process, `Path` inputs are downloaded at validation-time. At some point in the future we should likely normalise this.

- We now don't detect problems with the URL (e.g. it 404s) until we actually call `predict()`.

🍐 @evilstreak


nickstenning:
Once #827 is merged and we can rebase this on top of that, there's only one more test that's broken/one more tricky thing to fix before I think we should try and get this reviewed and merged.

python/tests/test_types.py
```diff
@@ -0,0 +1,65 @@
...
+def test_urlfile_can_be_pickled_even_once_loaded():
+    responses.get(
+        "https://example.com/some/url",
+        json={"message": "hello world"},
+        status=200,
+    )
+
+    u = URLFile("https://example.com/some/url")
+
+    result = pickle.loads(pickle.dumps(u))
```
evilstreak, line 63: This looks like the same test as the one above. Does this need a call to `u.read()` or something before the pickling?

nickstenning: Whoops! You're absolutely right. Thank you :)

zeke:
> This is the first step in making the Cog HTTP API more similar to Replicate's own API.

✨❤️ ✨

PR #826, opened 2022-12-06
Make cog HTTP server port configurable

nickstenning:
Does what it says on the tin.


zeke (approved):
Looks good. 👍🏼 

This reminded me that we don't have a doc about all the env vars, so I opened this: https://github.com/replicate/cog/issues/829

bfirsh:
ronseal_frame_b_0

PR #827, opened 2022-12-06
Don't pin dependencies that are unpinned in setup.py

nickstenning:
Pinning dependencies in requirements-dev.txt more strictly than we do in cog's setup.py gives us a false sense of security about whether recent updates to those dependencies have broken anything.

Development installations should use

    pip install -e python -r requirements-dev.txt


bfirsh:
Seems reasonable!

PR #828, opened 2022-12-06
Update install instructions in `README`

### LINKED ISSUE #818: Install instructions don't work on fish shell
`fish` doesn't support backticks, so this doesn't work:

```
sudo curl -o /usr/local/bin/cog -L https://github.com/replicate/cog/releases/latest/download/cog_`uname -s`_`uname -m`
```

One way to work around this would be to echo the contents of the script and pipe it into bash.

Prior art: https://github.com/dependabot/cli/tree/afb2e1ba322344c90106357d058c15193404fd1a#installation

cc @mattt

mattt:
Fixes #818 

The current instructions use [backquoted command substitution](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_03) (<code>``</code>) to select the appropriate release artifact for the current machine. This PR updates this command to use dollar sign syntax (`$()`) instead, which is more widely supported.

Following up on the changes in #823, this PR also updates the "Install" section of the README to provide instructions for building from source. This PR also consolidates the "Upgrade" section, since you can re-run the same install command to get the latest version. Finally, I went through and annotated terminal commands with the <code>```console</code>, which adds some syntax highlighting on GitHub.

Feel free to cherry-pick only the `$()` change (6855f2310bae76c6286137269deb5cb631e0436e) if you disagree with the more subjective copy editing changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,8 @@ Hit us up if you're interested in using it or want to collaborate with us. [We'r
 
 ## Install
 
+<a id="upgrade"></a>
+
 You can download and install the latest release of Cog
 by running the following commands in a terminal:
 
--- a/README.md
+++ b/README.md
@@ -115,22 +115,18 @@ Hit us up if you're interested in using it or want to collaborate with us. [We'r
 
 ## Install
 
-First, [install Docker if you haven't already](https://docs.docker.com/get-docker/). Then, run this in a terminal:
+First, [install Docker if you haven't already](https://docs.docker.com/get-docker/). 
+
+Next, 
+download and install a precompiled executable
+by running the following terminal commands:
 
 ```console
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,21 +115,19 @@ Hit us up if you're interested in using it or want to collaborate with us. [We'r
...
+by running the following commands in a terminal:
 
-```
-sudo curl -o /usr/local/bin/cog -L https://github.com/replicate/cog/releases/latest/download/cog_`uname -s`_`uname -m`
+```console
+sudo curl -o /usr/local/bin/cog -L "https://github.com/replicate/cog/releases/latest/download/cog_$(uname -s)_$(uname -m)"
 sudo chmod +x /usr/local/bin/cog
 ```
 
-## Upgrade
```
zeke, line 125: I'm sure if there are many links to `#upgrade` out in the wild, but maybe we could have an invisible anchor here just in case?

PR #843, opened 2022-12-12
Configure Cog python package version from git

### LINKED ISSUE #816: Set python package version explicitly and expose in package
The cog python package sets version metadata but this has never been updated:

```python
In [1]: from importlib.metadata import version

In [2]: version('cog')
Out[2]: '0.0.1'
```

In addition, there's no `__version__` property on the package. This isn't essential but it would be nice to have this too:

```python
In [3]: import cog

In [4]: cog.__version__
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [4], line 1
----> 1 cog.__version__

AttributeError: module 'cog' has no attribute '__version__'
```

It would be really nice to do this in a way that:

- returns the same version from both of the above
- returns the tagged version in tagged builds (e.g. `0.3.4`)
- appends git metadata when not on a tagged build (e.g. `0.3.4-dev+630e696`)

nickstenning:
This commit allows the Cog python package to derive its own version from git metadata and bake that into the installed package.

To do this, we first update the Cog python package over to `pyproject.toml` rather than `setup.py` (as is recommended for modern packages[1]) and then use setuptools_scm[2] to define the version dynamically.

This should mean that versions work correctly:

- on tagged releases
- in development
- when installed directly from git
- when installed from a git archive zip/tarball

[1]: https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/
[2]: https://github.com/pypa/setuptools_scm

Fixes #816.


nickstenning:
Ways to test this works as expected:

1. Install from git: `pip install 'git+https://github.com/replicate/cog@python-version#egg=cog&subdirectory=python'`
2. Checkout and editable install: `pip install -e python/`
3. Checkout and non-editable install: `pip install python/`
4. Install from wheel: `pip install build && python -m build python/ --wheel && pip install python/dist/<name>.whl`
5. Tagged build: `git tag v0.6.0 && pip install -e python/ && git tag -d v0.6.0`

nickstenning:
Tagging @zeke and @bfirsh FYI!

bfirsh:
Excellent! Good riddance setup.py!

PR #844, opened 2022-12-12
Install tini as image entrypoint

### LINKED ISSUE #840: Add tini (or similar appropriate PID 1 supervisor) to cog images
It would probably be sensible for the images that Cog builds to be suitable for production environments without further tweaking. At the moment the default command (`["python", "-m", "cog.server.http"]`) doesn't set up appropriate signal handlers to be running as PID 1. It cannot be killed from outside without sending it a SIGKILL.

The most robust way for us to do this would probably be to use something like https://github.com/krallin/tini or https://github.com/Yelp/dumb-init as the container entrypoint. Both of these are ~20KiB statically compiled binaries.

nickstenning:
Configure krallin/tini as the Docker image entrypoint. Ensures that signals are handled appropriately and any child processes are correctly reaped inside Cog containers.

Fixes #840.

Signed-off-by: Nick Stenning <nick@whiteink.com>

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -89,6 +89,7 @@ func (g *Generator) GenerateBase() (string, error) {
 		"# syntax = docker/dockerfile:1.2",
 		"FROM " + baseImage,
 		g.preamble(),
+		g.installTini(),
 		installPython,
 		installCog,
 		aptInstalls,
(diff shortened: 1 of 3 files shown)


nickstenning:
If you want to demonstrate to yourself that this fixes the signal handling issues you can do something like this:

```console
$ cog build -t signalstest
...
$ docker run --name=signalstest-1 --rm signalstest sh -c 'sleep 30'  # don't attach a TTY with -t as this changes the shell's default signal handling!
...
```

and then in another shell, try

```console
$ docker kill --signal=TERM signalstest-1
```

On `main`, this won't have the intended effect. The container will ignore the signal and keep running. On this branch, `tini` will handle the signal, forward it to the shell, and the shell will exit.

evilstreak (approved):
:+1:

nickstenning:
Nvidia just broke the world by adding entrypoints to every CUDA image, old and new. Merging this as it will coincidentally fix cog-built images.

PR #847, opened 2022-12-13
Use faster GitHub runners

bfirsh:


nickstenning:
TBH I'm not really sure this is working very well at the moment. We would at least need to bump up the maximum number of runners at https://github.com/organizations/replicate/settings/actions/github-hosted-runners/3, as this job runs 12 different jobs in parallel.

I'm also not really sure how much speedup we're going to get on the Go and Python test suites by running on bigger boxes. Maybe a sensible approach here would be to run the Go/Python matrices on `ubuntu-latest-4-cores` (or even the default GitHub-hosted runners) and run the actually slow part -- the integration suite -- on the big box with pytest parallelization?

bfirsh:
~11min -> ~8min

zeke (approved):
Ship it?

PR #851, opened 2022-12-14
Update instructions for installing `cog` for development

### LINKED ISSUE #848: Update CONTRIBUTING (and possibly Makefile) to reflect the new dev installation process
https://github.com/replicate/cog/pull/823 made some changes to the build process (to pave the way for `brew install cog` becoming a thing). The [dev env docs](https://github.com/replicate/cog/blob/0bb3ace3d22458f3929368737ad20fa595a4c946/CONTRIBUTING.md#development-environment) slipped through the cracks, and they're now out of date. Let's update them to include the proper invocation for building and installing cog locally.

A few ideas were floated about how to keep the local dev install working:

From @nickstenning:

> Perhaps let's add a `make install-dev` which does the `go install` this was previously doing?

From @mattt :

> we can get the old behavior with `make install PREFIX=$GOPATH`. What do you think about that convention?

I attempted both of those (with various prependings of `sudo`) but wasn't able to get either of them working. I figured I should stop and open an issue for guidance before digging any further.

mattt:
Resolves #848

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -101,7 +101,7 @@ Install the Python dependencies:
 
 Once you have Go installed, run:
 
-    make install
+    make install PREFIX=$(go env GOPATH)
 
 This installs the `cog` binary to `$GOPATH/bin/cog`.
 


zeke:
Sorry I missed this. Looks good! I took it for a spin locally and it worked like a charm.

PR #854, opened 2022-12-16
Run Python tests in parallel

bfirsh:
I think this is the bottleneck in CI now.


bfirsh:
64 cores seems to reduce Python tests time by a couple of mins.

zeke (approved):
🐍 🚀

PR #857, opened 2022-12-22
Allow redis connection URL to connect to redis.

bbedward:
Related to #747 

This is implementing `--redis-url` CLI option to connect to redis as described in the comments of #747 

It's useful because allowing a URL as an alternative to just host/port allows for more options when using redis
- Using SSL to connect
- Using username/password to connect
- Using a different database

Existing `--redis-host` and `--redis-port` adds a deprecation warning but still works as before.


bbedward:
I kept 1 of the integration tests using `--redis-host` and `--redis-port` and some of the integration tests still use positional arguments. So it should cover all use cases including the new one.

evilstreak (approved):
Great, thank you @bbedward!

evilstreak:
@all-contributors please add @bbedward for code

PR #868, opened 2023-01-09
Dependency fixes

evilstreak:
Some miscellaneous issues to get my dev environment (mostly) working again.

### DIFF
--- a/python/tests/server/test_http.py
+++ b/python/tests/server/test_http.py
@@ -3,6 +3,7 @@ import io
 
 from PIL import Image
 import pytest
+import unittest.mock as mock
 
 from .conftest import make_client, match, uses_predictor
 
@@ -30,7 +31,7 @@ def test_openapi_specification(client):
                 "responses": {
                     "200": {
                         "description": "Successful Response",
-                        "content": {"application/json": {"schema": {}}},
+                        "content": {"application/json": {"schema": mock.ANY}},
                     }
                 },
             }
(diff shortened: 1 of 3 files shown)


evilstreak:
Dang, the last version of numpy to support Python 3.7 is 1.21.6. Going to need to rethink that change.

nickstenning:
> Dang, the last version of numpy to support Python 3.7 is 1.21.6. Going to need to rethink that change.

We can probably get away with this as `requirements-dev.txt` is only used in our dev and CI environments:

```
numpy==1.21.6; python_version == '3.7'
numpy==1.24.1; python_version > '3.7'
```

(Requires `pip>=6`)

evilstreak:
> ```
> numpy==1.21.6; python_version == '3.7'
> numpy==1.24.1; python_version > '3.7'
> ```

TIL, thanks!

requirements-dev.txt
```diff
@@ -2,7 +2,8 @@ build==0.9.0
 hypothesis==6.56.4
 httpx==0.23.1
 mypy==0.990
-numpy==1.21.5
+numpy==1.21.6; python_version == '3.7'
```
zeke, line 5: Today I also learned.

https://peps.python.org/pep-0508/#environment-markers

zeke (approved):

nickstenning (approved):

python/tests/server/test_http.py
```diff
@@ -3,6 +3,7 @@
 
 from PIL import Image
 import pytest
+import unittest.mock as mock
```
nickstenning, line 6: This is a slightly unusual way of saying `from unittest import mock` but not a big deal in test code :)

PR #870, opened 2023-01-10
Add support for async prediction in cog HTTP API

evilstreak:
This large PR adds support for running predictions "asynchronously" (i.e. in the background) using the Cog HTTP API. It adds the following features:

1. Async prediction create: clients can request an asynchronous response from the `POST /predictions` endpoint.
2. Webhook support: clients can specify an HTTP endpoint which will receive periodic requests carrying a payload indicating prediction status.
3. Prediction cancelation: clients can cancel an in-flight prediction via a `POST /predictions/<prediction_id>/cancel` endpoint.

### Async prediction create

Callers can specify an HTTP header of `Prefer: respond-async` when calling the `POST /predictions` endpoint. If provided, the request will return immediately after starting the prediction with an HTTP `202 Accepted` status and a prediction object in status `processing`.

```http
POST /predictions HTTP/1.1
Content-Type: application/json; charset=utf-8
Prefer: respond-async

{
    "input": {"prompt": "A picture of an onion with sunglasses"}
}
```

The only supported mechanism for receiving updates on the status of predictions started asynchronously is via webhooks. There is as yet no support for polling for prediction status.

**Note 1:** that while this allows clients to create predictions "asynchronously," cog can only run one prediction at a time, and it is currently the caller's responsibility to make sure that earlier predictions are complete before new ones are created.

**Note 2:** predictions created asynchronously use a different mechanism for file upload than those created using the synchronous API. You must specify an `--upload-url` when running the cog server process. All uploads will be `PUT` using the provided `--upload-url` as a prefix, in much the same way that `output_file_prefix` worked. There is currently no single upload mechanism which works the same way for both synchronous and asynchronous prediction creation. This will be addressed in a future version of cog.

### Webhook support

Clients can (and should, if a prediction is created asynchronously) provide a `webhook` parameter at the top level of the prediction request, e.g.

```http
POST /predictions HTTP/1.1
Content-Type: application/json; charset=utf-8
Prefer: respond-async

{
    "input": {"prompt": "A picture of an onion with sunglasses"},
    "webhook": "https://example.com/webhook/prediction"
}
```

Cog will make requests to the URL supplied with the state of the prediction object in the request body. Requests are made when specific events occur during the prediction, namely:

- `start`: immediately on prediction start
- `output`: each time a prediction generates an output (note that predictions can generate multiple outputs)
- `logs`: each time log output is generated by a prediction
- `completed`: when the prediction reaches a terminal state (succeeded/canceled/failed)

Requests for event types `output` and `logs` will be sent at most once every 500ms. This interval is currently not configurable. Requests for event types `start` and `completed` will always be sent.

By default, cog will send requests for all event types. Clients can change which events trigger webhook requests by specifying `webhook_events_filter` in the prediction request. For example, if you only wanted requests to be sent at the start and end of the prediction, you would provide:

```http
POST /predictions HTTP/1.1
Content-Type: application/json; charset=utf-8
Prefer: respond-async

{
    "input": {"prompt": "A picture of an onion with sunglasses"},
    "webhook": "https://example.com/webhook/prediction",
    "webhook_events_filter": ["start", "completed"]
}
```

### Prediction cancelation

While an asynchronous prediction is running, clients can cancel it by making a request to `POST /predictions/<prediction_id>/cancel`. The prediction `id` must have been supplied when creating the prediction. Predictions created without a supplied `id` field will not be cancelable.

For example, if a prediction is created with

```http
POST /predictions HTTP/1.1
Content-Type: application/json; charset=utf-8
Prefer: respond-async

{
    "id": "abcd1234",
    "input": {"prompt": "A picture of an onion with sunglasses"},
}
```

it can be canceled with

```http
POST /predictions/abcd1234/cancel HTTP/1.1
```

Use of the cancelation API to cancel predictions started "synchronously" (i.e. without the `Prefer: respond-async` header) is currently not supported. This may change in a future release of cog.

## Director

This PR also includes the code for "director," which is what we at Replicate use to run predictions using the Cog HTTP API in our production environment. Readers should note that `cog.director` is not considered part of the public API of the Cog package, and we warn anyone who uses this code that it is provided as-is, without warranty of any kind, and will soon be removed from the Cog repository. Take dependencies on this code entirely at your own risk.

🍐 @nickstenning, @erbridge


python/cog/schema.py
```diff
@@ -66,6 +67,8 @@ class PredictionResponse(PredictionBaseModel):
     error: t.Optional[str]
     status: t.Optional[Status]
 
+    webhook: t.Optional[pydantic.AnyHttpUrl]
```
nickstenning, line 70: I'm wondering if this actually needs to be in the response object? Or did it get added because we're trying to set a field that doesn't exist and that was bombing out?

bfirsh:
:tada:

This is a gloriously clear description of the change – thank you.

I wonder what things we need to change still to make the Replicate and Cog APIs consistent. E.g. I think Replicate returns 201 but Cog returns 202. Maybe it'd be easier to change Cog to return 201 rather than change Replicate to return 202. 🤔

PR #874, opened 2023-01-19
Fix example code in README

williamluer:
Signed-off-by: williamluer <85975676+williamluer@users.noreply.github.com>


zeke:
@all-contributors please add @williamluer for docs

PR #876, opened 2023-01-19
cog/async-api: log prediction id

sirupsen:
Very simple change.

It would be slightly more work to also re-write Cog to include this when it logs to stdout; but it's certainly something I could do in a follow-up.

@nickstenning @evilstreak


nickstenning:
Took the liberty of adding your DCO acknowledgement in the form of the `Signed-off-by` line in the commit. I'll assume that's fine with you unless I hear otherwise :)

PR #877, opened 2023-01-19
Report setup runs from director

evilstreak:
This matches the behaviour of the existing RedisQueueWorker. It also causes the director to shutdown if setup fails.


nickstenning:
👍

nickstenning:
As soon as https://github.com/replicate/cluster/pull/96 merges this is done.

PR #878, opened 2023-01-19
Tighten up error handling in PredictionRunner

evilstreak:
The general principles here:

- if we're at all able to, trigger a final "failed" webhook with the error and all the logs and output so far
- if the error is irrecoverable, crash the server -- the director has to be able to handle a crashed sibling container anyway, so don't try any heroics to stay alive

---

This is untested – consider this a sketch for feedback rather than a polished-and-ready-to-go piece of code!

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -6,12 +6,16 @@ from multiprocessing import Event
 from multiprocessing.pool import ThreadPool, AsyncResult
 from typing import Any, Callable, Dict, Optional, Tuple
 
+import structlog
+
 from .. import schema
 from ..files import put_file_to_signed_endpoint
 from ..json import upload_files
 from .eventtypes import Done, Heartbeat, Log, PredictionOutput, PredictionOutputType
 from .webhook import webhook_caller, webhook_caller_filtered
 from .worker import Worker
+
+log = structlog.get_logger("cog.server.runner")
 
 
 class PredictionRunner:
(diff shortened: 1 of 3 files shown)


python/cog/server/runner.py
```diff
@@ -221,52 +226,74 @@ def predict(
...
+            "Got unexpected error during prediction response handling", error=error
+        )
 
-    return event_handler.response
+        # Attempt to fail the prediction and trigger a webhook, but if that
+        # also fails just move on
+        try:
+            # TODO: include a stack trace
+            event_handler.failed(
+                error="Got unexpected error during prediction response handling: {error}"
```
nickstenning, line 289: You're missing an `f` at the start of this format string!

nickstenning (approved):
👍

python/cog/server/runner.py
```diff
@@ -206,13 +211,13 @@ def _upload_files(self, output: Any) -> Any:
...
-            # re-raise and crash the container
-            # FIXME: ensure director gracefully handles this container crashing
-            raise error
+            # If something goes wrong uploading a file, it's irrecoverable.
+            # The re-raised exception will be caught and cause the prediction
+            # to be failed, with a useful error message.
+            raise FileUploadError("Got error trying to upload output files") from error
 
 
 def predict(
```
nickstenning, line 223: I think my preference here might be to wrap the existing code up as `_predict` and then have `predict` be an error handling wrapper around `_predict` -- that might make it easier to verify that all errors are caught and handled appropriately, rather than indenting the entire function body.

evilstreak: Ah yeah, that's much nicer, thanks

PR #882, opened 2023-01-20
observability: add utilization span + prediction span

sirupsen:
@nickstenning @evilstreak 

This adds a `cog.prediction` and `cog.director.idle` spans from Cog

**Should we set the parent span on `cog.prediction`?** We could probably pass down the parent span id in the `Prediction` message and then set the parent span. I didn't want to go through this effort of changing yet another repo if you guys didn't think it was worth it! Let me know.

**Why do you need both `cog.director.utilization` and `cog.prediction`?** I want to be able to get a utilization number between 0 and 1, and we can't use `CONCURRENCY(cog.prediction) / availableReplicas` or something to compute it. We can get away without it, but I would love to have this 0..1 utilization number.

**How many events will this generate per month?** This will generate [around 100M events ](https://share.cleanshot.com/CqHzC538) per month, or [6% of our quota](https://share.cleanshot.com/Y7KpzkLZ). We could reduce by sampling less often.

### DIFF
--- a/python/cog/director/director.py
+++ b/python/cog/director/director.py
@@ -1,3 +1,4 @@
+import datetime
 import json
 import queue
 import signal
@@ -6,6 +7,7 @@ from typing import Any, Callable, Dict, List, Optional
 
 import requests
 import structlog
+from opentelemetry import trace
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry  # type: ignore
 
@@ -15,6 +17,7 @@ from ..server.probes import ProbeHelper
 from ..server.webhook import requests_session, webhook_caller
 from .eventtypes import HealthcheckStatus, Webhook
 from .healthchecker import Healthchecker
+from .monitor import Monitor
 from .prediction_tracker import PredictionTracker
 from .redis import EmptyRedisStream, RedisConsumerRotator
 
@@ -49,6 +52,7 @@ class Director:
         self,
         events: queue.Queue,
         healthchecker: Healthchecker,
+        monitor: Monitor,
         redis_consumer_rotator: RedisConsumerRotator,
         predict_timeout: int,
         max_failure_count: int,
@@ -56,6 +60,7 @@ class Director:
     ):
         self.events = events
         self.healthchecker = healthchecker
+        self.monitor = monitor
         self.redis_consumer_rotator = redis_consumer_rotator
         self.predict_timeout = predict_timeout
         self.max_failure_count = max_failure_count
@@ -64,6 +69,7 @@ class Director:
         self._failure_count = 0
         self._should_exit = False
         self._shutdown_hooks: List[Callable] = []
+        self._tracer = trace.get_tracer("cog-director")
 
         self.cog_client = _make_local_http_client()
         self.cog_http_base = "http://localhost:5000"
@@ -156,6 +162,7 @@ class Director:
             structlog.contextvars.bind_contextvars(
                 queue=redis_consumer.redis_input_queue,
             )
+            self.monitor.set_current_prediction(None)
 
             self._confirm_model_health()
 
@@ -173,7 +180,10 @@ class Director:
 
             try:
                 log.info("received message")
-                self._handle_message(message_id, message_json)
+                with self._tracer.start_as_current_span(
+                    name="cog.prediction"
+                ) as span:
+                    self._handle_message(message_id, message_json, span)
             except Exception:
                 self._record_failure()
                 log.error("caught exception while running prediction", exc_info=True)
@@ -183,7 +193,9 @@ class Director:
                 redis_consumer.ack(message_id)
                 log.info("acked message")
 
-    def _handle_message(self, message_id: str, message_json: str) -> None:
+    def _handle_message(
+        self, message_id: str, message_json: str, span: trace.Span
+    ) -> None:
         message = json.loads(message_json)
         prediction_id = message["id"]
 
@@ -193,9 +205,6 @@ class Director:
         )
 
         log.info("running prediction")
-        should_cancel = self.redis_consumer_rotator.get_current().checker(
-            message.get("cancel_key")
-        )
 
         # Tracker is tied to a single prediction, and deliberately only exists
         # within this method in an attempt to eliminate the possibility that we
@@ -203,6 +212,12 @@ class Director:
         tracker = PredictionTracker(
             response=schema.PredictionResponse(**message),
             webhook_caller=webhook_caller(message["webhook"]),
+        )
+        self.monitor.set_current_prediction(tracker._response)
+        self._set_span_attributes_from_tracker(span, tracker)
+
+        should_cancel = self.redis_consumer_rotator.get_current().checker(
+            message.get("cancel_key")
         )
 
         # Override webhook to call us
@@ -216,15 +231,15 @@ class Director:
                 headers={"Prefer": "respond-async"},
                 timeout=PREDICTION_CREATE_TIMEOUT,
             )
-        except requests.exceptions.RequestException:
+        except requests.exceptions.RequestException as e:
             tracker.fail("Unknown error handling prediction.")
             log.error("prediction failed: could not create prediction", exc_info=True)
-            self._record_failure()
+            self._record_failure(span, e)
             return
 
         try:
             resp.raise_for_status()
-        except requests.exceptions.RequestException:
+        except requests.exceptions.RequestException as e:
             # Special case validation errors
             if resp.status_code == 422:
                 tracker.fail(f"Prediction input failed validation: {resp.text}")
@@ -240,7 +255,7 @@ class Director:
                     status_code=resp.status_code,
                     response=resp.text,
                 )
-            self._record_failure()
+            self._record_failure(span, e)
             return
 
         tracker.start()
@@ -262,6 +277,7 @@ class Director:
                         tracker.fail("Model stopped responding during prediction.")
                         self._abort(
                             "prediction failed: model container failed healthchecks",
+                            span=span,
                             health=event.health.name,
                         )
                 else:
@@ -269,7 +285,7 @@ class Director:
 
             if should_cancel():
                 log.info("prediction cancelation requested")
-                self._cancel_prediction(prediction_id)
+                self._cancel_prediction(prediction_id, span)
                 break
 
             if self.predict_timeout and tracker.runtime > self.predict_timeout:
@@ -280,7 +296,7 @@ class Director:
                 # Mark the prediction as timed out so we handle the
                 # cancelation webhook appropriately.
                 tracker.timed_out()
-                self._cancel_prediction(prediction_id)
+                self._cancel_prediction(prediction_id, span, timed_out=True)
                 break
 
         # Wait up to another CANCEL_WAIT seconds for cancelation if necessary
@@ -298,20 +314,42 @@ class Director:
         # and we should abort.
         if not tracker.is_complete():
             tracker.force_cancel()
-            self._abort("prediction failed to complete after cancelation")
+            self._abort("prediction failed to complete after cancelation", span)
 
         # Keep track of runs of failures to catch the situation where the
         # worker has gotten into a bad state where it can only fail
         # predictions, but isn't exiting.
         if tracker.status == schema.Status.FAILED:
             log.warn("prediction failed")
-            self._record_failure()
+            self._record_failure(span)
         elif tracker.status == schema.Status.CANCELED:
             log.info("prediction canceled")
             self._record_success()
         else:
             log.info("prediction succeeded")
             self._record_success()
+
+        self._set_span_attributes_from_tracker(span, tracker)
+
+    # OpenTelemetry is very picky about not accepting None types
+    def _set_span_attributes_from_tracker(self, span, tracker):
+        span.set_attribute("prediction.uuid", tracker._response.id)
+        if tracker._response.version:
+            span.set_attribute("model.version", tracker._response.version)
+        if tracker._response.started_at:
+            span.set_attribute(
+                "prediction.started_at", tracker._response.started_at.timestamp()
+            )
+        if tracker._response.created_at:
+            span.set_attribute(
+                "prediction.created_at", tracker._response.created_at.timestamp()
+            )
+        if tracker._response.status:
+            span.set_attribute("prediction.status", tracker.status)
+        if tracker._response.completed_at:
+            span.set_attribute(
+                "prediction.completed_at", tracker._response.completed_at.timestamp()
+            )
 
     def _confirm_model_health(self) -> None:
         self.healthchecker.request_status()
@@ -346,7 +384,13 @@ class Director:
             wait_seconds=HEALTHCHECK_WAIT,
         )
 
-    def _cancel_prediction(self, prediction_id: Any) -> None:
+    def _cancel_prediction(
+        self, prediction_id: Any, span: trace.Span, timed_out=False
+    ) -> None:
+        span.set_attribute("prediction.status", "canceled")
+        if timed_out:
+            span.set_attribute("prediction.timed_out", True)
+
         resp = self.cog_client.post(
             self.cog_http_base + "/predictions/" + prediction_id + "/cancel",
             timeout=1,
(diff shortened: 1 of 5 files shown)


evilstreak:
> **Should we set the parent span on `cog.prediction`?** We could probably pass down the parent span id in the `Prediction` message and then set the parent span. I didn't want to go through this effort of changing yet another repo if you guys didn't think it was worth it! Let me know.

There's some prior art for this in `RedisQueueWorker`: https://github.com/replicate/cog/blob/08f4c8f909c2a67ae2c8e531fd7b03514ed65252/python/cog/server/redis_queue.py#L193-L203. We might want to consider a linked span rather than a parent span – viewing very long async traces in Honeycomb isn't great, so setting the parent would it harder to reason about the bits of the trace leading up enqueueing the prediction.

Also worth noting: we're not passing this value through for Replicate yet, but it wouldn't be a big change to do so.

sirupsen:
@nickstenning @evilstreak updated for Nick's changes

python/cog/director/director.py
```diff
@@ -258,8 +265,13 @@ def _handle_message(self, message_id, message_json) -> None:
         # predictions, but isn't exiting.
         if tracker.status == schema.Status.FAILED:
             self._record_failure()
+            span.set_attribute("prediction.status", 'success')
```
evilstreak, line 268: You've got these statuses inverted I think. Given we can also have `canceled` as a status, I think it'd be cleaner to just use `span.set_attribute("prediction.status", tracker.status)` outside this conditional.

python/cog/director/director.py
```diff
@@ -170,6 +174,9 @@ def _handle_message(self, message_id, message_json) -> None:
         )
         should_cancel = self.redis_consumer.checker(message.get("cancel_key"))
         prediction_id = message["id"]
+        span.set_attribute("prediction.id", prediction_id)
+        span.set_attribute("prediction.created_at", message.get("created_at", None))
+        span.set_attribute("prediction.started_at", message.get("started_at", None))
```
evilstreak, line 179: `started_at` won't be set in the message, you'll need to get it from the tracker. For ease and consistency, it might be better to add all these attributes after the tracker is created, by just pulling them from the tracker.

python/cog/director/director.py
```diff
@@ -258,8 +265,13 @@ def _handle_message(self, message_id, message_json) -> None:
         # predictions, but isn't exiting.
         if tracker.status == schema.Status.FAILED:
             self._record_failure()
+            span.set_attribute("prediction.status", 'success')
         else:
             self._record_success()
+            span.set_attribute("prediction.status", 'failure')
+
+        time_rfc3339 = datetime.datetime.now(datetime.timezone.utc).isoformat()
+        span.set_attribute("prediction.completed_at", time_rfc3339)
```
evilstreak, line 274: We're pretty scattered in our approach to timing events but I think we might be settling on millisecond precision timestamps. @nickstenning might have other thoughts!

python/cog/director/director.py
```diff
@@ -298,6 +310,12 @@ def _abort(self, message=None):
         self._should_exit = True
         raise Abort(message)
 
+    def _utilization_thread_loop(self) -> None:
+        while not self._should_exit:
+            with self._tracer.start_as_current_span(name="cog.director.idle") as span:
+                span.set_attribute("idle", self.idle)
```
evilstreak, line 316: Would this be better as a `threading.Event` rather than a boolean, as we're using it in two different threads?

I know this matches the way we're emitting node utilisation metrics, but I wonder if there's a better way to handle it. This isn't going to give us the data in Honeycomb for a full picture in one graph, because these events are separate to the node utilisation events.

I'll have a noodle about how else we might construct useful data in Honeycomb for this before tomorrow morning.

evilstreak (commented):
I think this approach would work, but I wonder if it'd be nicer to use an approach closer to that of the `Healthchecker` – it's its own thing, initialised from `__main__` with the events stream, and then it just sends events to that stream when it has something to say. I'm not sure exactly what that'd look like for the utilisation reporter, as it needs to know something about the state of the prediction, rather than just reporting something to that events thread.

I'd like to think about whether attaching attributes to the prediction event would get us the data in a format that's useful for us: one extra attribute for `time_spent_idle_before_this_prediction` would certainly give us everything we needed to calculate utilisation manually, but I'm not sure it'd be any use for Honeycomb graphs.

One comment about formatting: we tend to use Black for Python formatting, but for some reason it looks like we haven't set that up on this repo. Either way, there's a few lines you're churning here that are formatted by Black, so they'll churn straight back the next time they're edited. It'd be great if you could run Black over your changes, or at least drop the churned lines.

python/cog/director/director.py
sirupsen: It's ok that this is "racy" because there's only one setter, but I changed it to an event now just so it doesn't confuse anyone.

sirupsen:
@evilstreak I believe all concerns have been addressed, let me know what you think

> I think this approach would work, but I wonder if it'd be nicer to use an approach closer to that of the Healthchecker 

Done

> I know this matches the way we're emitting node utilisation metrics, but I wonder if there's a better way to handle it. This isn't going to give us the data in Honeycomb for a full picture in one graph, because these events are separate to the node utilisation events.

Yeah, I've been thinking about this, and I cannot think of any other way to get a nice number between 0 and 1, which is what we want. Ideally we can use `CONCURRENCY` or some kind of query, but Honeycomb has nothing for it. There may be a clever way, but I have definitely thought about it and couldn't come up with anything.

> One comment about formatting: we tend to use Black for Python formatting, but for some reason it looks like we haven't set that up on this repo.

I do run `black(1)` locally. The reason it was churning is that I have mine set to always run with a line-length of 120. I've changed that to whatever the default is.

python/cog/director/director.py
```diff
@@ -162,7 +167,16 @@ def _loop(self):
             structlog.contextvars.bind_contextvars(message_id=message_id)
 
             try:
-                self._handle_message(message_id, message_json)
+                span_attributes = {
+                    "idling_ms": (
```
sirupsen, line 171: I added this, but I don't think it's going to help. Even emitting a span for the duration where we're idle also doesn't do much. It's just an unfortunate limitation of Honeycomb to my knowledge, and will also not give us the most accurate utilization we could get. That said, it is a sampling every 15s, which is not awful.

sirupsen:
Can we merge this into the branch today, unless there are any other changes you'd like me to make? Otherwise I'll keep chasing your changes :)

python/cog/director/director.py
```diff
@@ -56,10 +57,12 @@ def __init__(
         self.predict_timeout = predict_timeout
         self.max_failure_count = max_failure_count
         self.report_setup_run_url = report_setup_run_url
+        self.idle = threading.Event()  # is_set() means a prediction is running!
```
evilstreak, line 60: Is this comment right? From my reading of the code, it looks like a prediction is running when the event is cleared, and it's idle when set.

python/cog/director/director.py
```diff
@@ -165,7 +170,16 @@ def _loop(self):
 
             try:
                 log.info("received message")
-                self._handle_message(message_id, message_json)
+                span_attributes = {
+                    "idling_ms": (
+                        datetime.datetime.utcnow() - started_idling_at
+                    ).total_seconds()
```
evilstreak, line 176: To get millisecond precision I think we'd need  `(datetime.datetime.utcnow() - started_idling_at) / timedelta(microseconds=1)`

python/cog/director/monitor.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def join(self) -> None:
+        if self._thread is not None:
+            self._thread.join()
+
+    def _run(self) -> None:
+        last_span_at = time.perf_counter() - UTILIZATION_SPAN_INTERVAL
+        while not self._should_exit.is_set():
+            if time.perf_counter() >= last_span_at + UTILIZATION_SPAN_INTERVAL:
+                idle = self._director.idle.is_set()
```
evilstreak, line 33: This is the only thing we're using `director` for right? Would it be cleaner to pass the `idle` threading event into both `Director` and `Monitor`, rather than passing the entire object?

python/cog/director/monitor.py
```diff
@@ -0,0 +1,42 @@
...
+                with self._tracer.start_as_current_span(
+                    name="cog.director.idle"
```
evilstreak, line 35: Would `cog.director.utilization` be a more descriptive name for this?

python/cog/director/monitor.py
```diff
@@ -0,0 +1,42 @@
...
+                ) as span:
+                    span.set_attribute("idle", idle)
```
evilstreak, line 37: It'd be great to some basic dimensions in here, like the model name or version ID or pod name, so that we can group by things in Honeycomb to get more understanding than just a top level number.

python/cog/director/director.py
```diff
@@ -301,6 +319,13 @@ def _handle_message(self, message_id, message_json) -> None:
             log.info("prediction succeeded")
             self._record_success()
 
+        span.set_attribute("prediction.status", tracker.status)
+        if tracker._response.completed_at:
+            completed_at_rfc3999_ms = tracker._response.completed_at.strftime(
```
evilstreak, line 324: Why is this datestamp handled differently the the `started_at` and `created_at` datestamps?

evilstreak (approved):
A few questions and comments, but I think you could address those in a follow up if you wanted to get this in to start collecting metrics.

python/cog/director/director.py
sirupsen: ah yes I had it as `self.active` first but decided to invert

python/cog/director/director.py
sirupsen: It's right, it's because `total_seconds(0)` returns a float with ms precision. tbh copilot gave me this, I tested it, and it was fine 👍🏻

python/cog/director/monitor.py
sirupsen: I thought it was cleaner to pass the whole object because monitor might do other things in the future. I don't have a strong opinion

python/cog/director/monitor.py
sirupsen: Sure, but that changes things a bit... Probably better though. Was probably a bit too biased by what we do so far

python/cog/director/monitor.py
sirupsen: yeah good call

python/cog/director/director.py
sirupsen: Yeah no good reason... I need to see if OpenTelemetry in Python by default emits millisecond precision and then I'll fix it all up

sirupsen:
@evilstreak updated to:

1. Use events instead of a `threading.Event`, this could make monitor take various other events from various other places. I've written the code so it'd be trivial to hook that up.
2. Add more metadata like model versions etc.
3. Renamed to `cog.director.utilization`
4. Rebase with latest changes

sirupsen:
@evilstreak @nickstenning now monitors and emits the correct utilization inside each model, and has a test.

What started as a 10 line hack to get some numbers has morphed quite a bit, but for the better 🙃

python/cog/director/monitor.py
```diff
@@ -0,0 +1,93 @@
...
+                    name="cog.director.utilization"
+                ) as span:
+                    if current_prediction_started_at:  # event didn't finish in window
+                        active_sec += (
+                            time.perf_counter() - current_prediction_started_at
+                        )
+
+                    utilization_for_window = min(
+                        active_sec / self.utilization_interval, 1.0
+                    )
```
sirupsen, line 72: The line length configuration for black really makes this way harder to read than it should be... I think we should move to 120 for the repo

python/cog/director/director.py
```diff
@@ -241,6 +263,7 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
                     response=resp.text,
                 )
             self._record_failure()
+            span.set_attribute("prediction.status", "failure")
```
nickstenning, line 266:
```suggestion
            span.set_attribute("prediction.status", "failed")
```

python/cog/director/director.py
```diff
@@ -270,6 +293,7 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
             if should_cancel():
                 log.info("prediction cancelation requested")
                 self._cancel_prediction(prediction_id)
+                span.set_attribute("prediction.status", "cancelled")
```
nickstenning, line 296:
```suggestion
                span.set_attribute("prediction.status", "canceled")
```

python/cog/director/director.py
```diff
@@ -281,6 +305,7 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
                 # cancelation webhook appropriately.
                 tracker.timed_out()
                 self._cancel_prediction(prediction_id)
+                span.set_attribute("prediction.status", "timeout")
```
nickstenning, line 308:
```suggestion
                span.set_attribute("prediction.status", "canceled")
                span.set_attribute("prediction.timed_out", True)
```

python/cog/director/director.py
```diff
@@ -313,6 +338,28 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
             log.info("prediction succeeded")
             self._record_success()
 
+        self._set_span_attributes_from_tracker(span, tracker)
+
+    # OpenTelemetry is very picky about not accepting None types
+    def _set_span_attributes_from_tracker(self, span, tracker):
+        span.set_attribute("prediction.id", tracker._response.id)
```
nickstenning, line 345:
```suggestion
        span.set_attribute("prediction.uuid", tracker._response.id)
```

python/cog/director/director.py
```diff
@@ -313,6 +338,28 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
...
+        if tracker._response.version:
+            span.set_attribute("prediction.model.version", tracker._response.version)
```
nickstenning, line 347: I think we _probably_ want this to be
```suggestion
            span.set_attribute("model.version", tracker._response.version)
```

python/cog/director/director.py
```diff
@@ -204,6 +220,12 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
...
         )
+        self.monitor.set_current_prediction(tracker._response)
+        self._set_span_attributes_from_tracker(span, tracker)
+
+        should_cancel = self.redis_consumer_rotator.get_current().checker(
+            message.get("cancel_key")
+        )
 
         # Override webhook to call us
         message["webhook"] = "http://localhost:4900/webhook"
```
nickstenning, line 231: In the `try: ... except: ...` blocks below here where we early-return it's _especially_ important to capture the failure (and ideally stuff like exception tracebacks if relevant, or HTTP status codes, etc.). I guess we want to make sure somehow that we set the status `succeeded/canceled/failed` properties and whatever other relevant info we can on every possible exit path from this method.

sirupsen:
* Point at refinery

python/cog/director/director.py
nickstenning: Do we even still need this?

python/cog/schema.py
```diff
@@ -58,6 +58,7 @@ class PredictionResponse(PredictionBaseModel):
     output: t.Any
 
     id: t.Optional[str]
+    version: t.Optional[str]
```
nickstenning, line 61: I *think* this is ok, as it shouldn't change anything about what we're sending back in webhooks (because of https://github.com/replicate/cog/blob/b4a940efd8a7790f96d5fcda7b50be68751ca875/python/cog/json.py#L28) but let's just check that.

sirupsen: If the input prediction comes with the version (which it does in Replicate's case), then it will be sent in the webhook too. I'm not sure whether that was the case before or not, or if API just overrides it or not. But it is [in the schema](https://replicate.com/docs/reference/http#get-prediction), so I'm not sure I understand what the problem here might be, or how that code is relevant?

sirupsen: It seems like `version` is the only field that wasn't typed out here. Might it be that we just aren't sending it in webhooks right now..? Wouldn't that be a bug that this is fixing?

nickstenning (approved):
LGTM

PR #884, opened 2023-01-22
Fix a bunch of error handling edge cases in director

nickstenning:
First of all, I apologise for opening a pull request on the weekend. I forgot that I'm out on Monday, and I had a moment of inspiration (or at least it felt like it to me!) today and figured I would try and get ahead of things.

If I haven't completely messed this up, this PR should:

1. Address most of the remaining error handling we need to do on the director side. It handles:
    - sending failure webhooks for input validation errors
    - sending failure webhooks when the model container ignores cancelations
    - ensuring that handled failure modes result in the redis message being ACKed

2. Make it much clearer how to integrate detection of failure of the model container while a prediction is running. This isn't done yet, but the model container healthchecking moves into its own logical thread, and there's a `TODO` in `director.py` which notes where we'd do this.

3. Make it much harder to accidentally mix up state between predictions (e.g. as you saw at the weekend, @evilstreak).

4. Untrust timestamps set by the model container.

5. Make it a lot easier to test the various components involved in director.

The critical change that makes all of this a lot easier is moving all control of the lifecycle of the process into Director itself. All communication from the auxiliary threads is mediated through a shared queue.

For example, the webserver code is now extremely uncomplicated. It receives webhook requests and publishes them onto a queue.

The healthchecker thread periodically makes an HTTP request to the model container to establish its health, and publishes changes of health status onto the queue.

The shutdown of these auxiliary threads is coordinated by allowing the registration of "shutdown hooks" with the director. In this way, the director doesn't need to know anything about the auxiliary threads, it just reads from a shared queue which they can publish to.

This will likely be easier to review commit by commit.

### DIFF
--- a/python/cog/director/__main__.py
+++ b/python/cog/director/__main__.py
@@ -12,6 +12,7 @@ import uvicorn
 from ..logging import setup_logging
 from .redis import RedisConsumer
 from .director import Director
+from .http import Server, create_app
 
 log = structlog.get_logger("cog.director")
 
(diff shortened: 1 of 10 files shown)


evilstreak (approved):
This makes sense to me, and seems like a much cleaner approach 👍

PR #886, opened 2023-01-24
Abort prediction on a failed healthcheck event

evilstreak:
Increase the number of retries so that the healthcheck endpoint has to fail for roughly ~6s before emitting a `HealthcheckStatus` event with a status of `Health.UNKNOWN`. That roughly mirrors the ~5s interval we poll at when the model container is healthy.

If we get a non-healthy status event during a prediction, abort the prediction and shut everything down.

### DIFF
--- a/python/cog/director/director.py
+++ b/python/cog/director/director.py
@@ -218,8 +218,10 @@ class Director:
                 if isinstance(event, Webhook):
                     tracker.update_from_webhook_payload(event.payload)
                 elif isinstance(event, HealthcheckStatus):
-                    # TODO: handle these appropriately
                     log.info("received healthcheck status update", data=event)
+                    if event.health != Health.HEALTHY:
+                        tracker.fail("Model stopped responding during prediction.")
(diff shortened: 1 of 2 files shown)


nickstenning:
Based on my reading of `urllib3` source, this would actually fail permanently after `0 + 0.1 + 0.2 + 0.4 + 0.8 + 1.6 = 3.1` seconds. If we also change `backoff_factor` to `0.2` that would then fail permanently after `6.2` seconds. Shall we do that?

evilstreak:
> Based on my reading of `urllib3` source, this would actually fail permanently after `0 + 0.1 + 0.2 + 0.4 + 0.8 + 1.6 = 3.1` seconds. If we also change `backoff_factor` to `0.2` that would then fail permanently after `6.2` seconds. Shall we do that?

Oh good find, thanks. Yeah, let's change it to `0.2`. If we get two bad responses back to back then trying again after `0.1` seconds doesn't seem all that useful.

nickstenning (approved):
💥

PR #895, opened 2023-01-25
Optionally add authorization header to all outgoing HTTP calls

andreasjansson:
This will add the header `Authorization: Bearer <webhook-auth-token>` to all outgoing calls from the director, if `WEBHOOK_AUTH_TOKEN` is defined.


python/cog/director/director.py
```diff
@@ -312,14 +312,25 @@ def _shutdown_model(self):
...
-    adapter = HTTPAdapter(
-        max_retries=Retry(
-            total=3,
-            backoff_factor=0.1,
-            status_forcelist=[429, 500, 502, 503, 504],
-            allowed_methods=["POST"],
-        ),
-    )
+    auth_token = os.environ.get("WEBHOOK_AUTH_TOKEN")
+    adapter = WebhookHTTPAdapter(auth_token)
```
nickstenning, line 316: Ah, slight misunderstanding here. This is the HTTP client used to talk to the model container. It looks to me like we use a different client for webhook sending (grep for `webhook_sender`) and for reporting setup runs. I suspect there's a bit of reorganization needed to get both of those using this token.

nickstenning (changes requested):

python/cog/director/director.py
andreasjansson: Ah right, thank you!

PR #896, opened 2023-01-25
Improve server shutdown behaviour

nickstenning:
This commit attempts to improve how server shutdown occurs, in a few different ways:

1. Exceptions thrown while running a prediction are caught, logged, and used to trigger a clean shutdown of the server. Unfortunately we cannot currently recover from an exception thrown while a prediction is running, because we have no way to resume and drain the prediction event stream.

2. Server shutdown tries to proceed gracefully, but will escalate to sending SIGKILL to the process if needs be.

3. SIGINT at the terminal is now always respected if possible.

One minor note of caution here: we are running Worker.terminate() to kill the worker, which is not a graceful shutdown. In-flight predictions will be killed immediately, and no final status webhooks will be sent. It is beholden on the director to ensure that webhooks are sent for predictions running when a shutdown request is sent.

Note also that an exception during even synchronous prediction processing will shut down the webserver. This is likely not expected behaviour, and we will likely want to revisit it in future.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -5,7 +5,7 @@ import signal
 import textwrap
 import threading
 from datetime import datetime, timezone
-from typing import Any, Optional, Union
+from typing import Any, Callable, Optional, Union
 
 import pydantic
 import structlog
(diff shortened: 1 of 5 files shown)


nickstenning:
Merging this as the current code is really broken and this is unlikely to make things worse. Let's review this together tomorrow either way @evilstreak.

evilstreak (approved):
With the caveat that I'm well past my best at this point in the day, this looks good to me!

PR #898, opened 2023-01-26
Explicitly healthcheck the model container before each prediction

nickstenning:
We wait up to 10 seconds for an explicit confirmation of the model container's health before fetching the next prediction from Redis. This should hopefully prevent us from dequeueing predictions that are bound to fail (e.g. when the model container has crashed but we have not yet detected that with the regular background checks).

### DIFF
--- a/python/cog/director/director.py
+++ b/python/cog/director/director.py
@@ -38,6 +38,11 @@ POLL_INTERVAL = 0.1
 # uploads, we should likely try and reduce this to a smaller number, e.g. 5s.
 CANCEL_WAIT = 30
 
+# How long to wait for an explicit healthcheck request before aborting. Note
+# that to avoid failing prematurely this should be at least as long as the it
+# takes for the complete chain of retries configured for the Healthchecker.
+HEALTHCHECK_WAIT = 10
+
 
 class Abort(Exception):
     pass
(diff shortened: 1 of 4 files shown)


python/cog/director/director.py
```diff
@@ -309,6 +315,32 @@ def _handle_message(self, message_id, message_json) -> None:
...
+        mark = time.perf_counter()
+
+        while time.perf_counter() - mark < HEALTHCHECK_WAIT:
+            try:
+                event = self.events.get(timeout=POLL_INTERVAL)
+            except queue.Empty:
+                continue
+
+            if not isinstance(event, HealthcheckStatus):
+                log.warn(
```
evilstreak, line 329: The only other event we can get here is a webhook, which would indicate something is *very* wrong. I think this is probably fine as is, but maybe something to keep an eye on and make more defensive (ie abort) if we see it being an issue in prod.

evilstreak (approved):

python/cog/director/director.py
nickstenning: Yup, I went back and forth on this one. I guess let's see what happens in prod. If we are regularly getting webhook events here we probably need to dig into why in addition to possibly crashing here instead.

PR #900, opened 2023-01-26
Make integration tests work with director

evilstreak:


test-integration/test_integration/test_redis_queue.py
```diff
@@ -126,11 +183,11 @@ def capture_final_response(request):
         assert waiting.result
 
         assert re.match(
-            r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{6}Z",
+            r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{6}\+00:00",
```
nickstenning, line 186: Only a nit really but maybe we should make this `(Z|\+00:00)` as I'm 90% sure both are accepted by downstream systems.

evilstreak:
> Do we care about `redis_queue` any more? Do we want to make the integration tests run on both? I'm totally fine with the answer being "not worth it," FWIW.

I don't think we do – I've broken down the commits and left a note to that effect in the commit message.

PR #902, opened 2023-01-27
Tidy up some issues raised by linters and type checkers

erbridge:


python/cog/director/__main__.py
```diff
@@ -47,7 +47,7 @@ def _die(signum: Any, frame: Any) -> None:
 
 args = parser.parse_args()
 
-events = queue.Queue(maxsize=128)
+events: queue.Queue = queue.Queue(maxsize=128)
```
nickstenning, line 50: lol, so glad that the python typing ecosystem has such powerful type inference :P

python/cog/server/runner.py
```diff
@@ -72,7 +69,7 @@ def setup(self) -> Tuple[schema.Status, str]:
     # no longer have to support Python 3.8
     def predict(
         self, prediction: schema.PredictionRequest, upload: bool = True
-    ) -> AsyncResult:
```
erbridge, line 75: Note this is a change, not just additive.

python/cog/director/director.py
```diff
@@ -371,15 +362,15 @@ def _record_failure(self):
                 "saw too many failures in a row", failure_count=self._failure_count
             )
 
-    def _record_success(self):
+    def _record_success(self) -> None:
         self._failure_count = 0
 
-    def _abort(self, message=None, **kwds):
+    def _abort(self, message: str = None, **kwds: Any) -> None:
```
nickstenning, line 368: I'm not quite sure why `message` was optional before but I don't think it needs to be or should be.
```suggestion
    def _abort(self, message: str, **kwds: Any) -> None:
```

python/cog/director/__main__.py
erbridge: Seems to have problems with `queue.Queue` but basically nothing else 🤷

nickstenning (approved):
LGTM.

PR #908, opened 2023-01-30
Optionally add authorization header to all webhook HTTP calls

andreasjansson:


nickstenning (approved):
LGTM. It makes me a touch nervous that this code is used by both director and the model container, but I guess we're only going to set the env var in the director container so that should be fine for now.

PR #909, opened 2023-01-31
Swallow connection errors for non-terminal webhooks

evilstreak:
We're happy to ignore any status response failures with this webhook, so we should be happy to ignore any connection errors too.

Without this we're actually less tolerant of network failures than the retry logic for terminal webhooks would suggest, because a single connection error on a non-terminal webhook is enough to interrupt a prediction and crash the model.

### DIFF
--- a/python/cog/server/webhook.py
+++ b/python/cog/server/webhook.py
@@ -2,11 +2,14 @@ import os
 from typing import Any, Callable, Set
 
 import requests
+import structlog
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry  # type: ignore
 
 from ..schema import Status, WebhookEvent
 from .response_throttler import ResponseThrottler
+
+log = structlog.get_logger(__name__)
 
 
 def _get_version() -> str:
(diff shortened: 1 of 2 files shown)


python/cog/server/webhook.py
```diff
@@ -56,8 +59,11 @@ def caller(response: Any) -> None:
                 # For terminal updates, retry persistently
                 retry_session.post(webhook, json=response)
             else:
-                # For other requests, don't retry
-                default_session.post(webhook, json=response)
+                # For other requests, don't retry, and ignore any errors
+                try:
+                    default_session.post(webhook, json=response)
+                except Exception:
```
erbridge, line 65: Do we really want to catch all exceptions, not explicitly the connection errors?

evilstreak: I think yes – sending this webhook is not critical to the whole thing. The main thing we're guarding against here are temporary issues, where really any error can be ignored as long as it's not our final webhook.

There's an argument that in the future we might want to be more selective here (eg, fail the prediction early if we get a 413 or something, on the basis that it won't recover during the prediction run and we'd be better off just giving up now). But I think that's something to explicitly fail on, on a case by case basis.

nickstenning: Yup, I think this should probably be catching `requests.exceptions.RequestException`.

python/cog/server/webhook.py
```diff
@@ -2,12 +2,15 @@
...
 
 import requests
+import structlog
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry  # type: ignore
 
 from ..schema import Status, WebhookEvent
 from .response_throttler import ResponseThrottler
 
+log = structlog.get_logger("cog.server.webhook")
```
nickstenning, line 12: Not a big deal, but in files other than the entrypoint you can get away with `...get_logger(__name__)`.

PR #915, opened 2023-02-01
Build director image after release

evilstreak:
We get the version of the published package by looking at the filename of the wheel during the python package build step. We could try loading that wheel and checking `cog.__version__`, but that requires installing a bunch of dependencies which is more of a faff than just looking at the file. This may end up being error prone, in which case we can revisit it!


nickstenning (approved):
LGTM

PR #917, opened 2023-02-02
Incorporate READY/BUSY status into model healthcheck

nickstenning:
This commit tightens up the contract between director and the model container's `/health-check` endpoint.

Specifically, the `status` field of the healthcheck endpoint now reports one of four health statuses: STARTING, READY, BUSY, SETUP_FAILED.

The expected state transitions between these four states are as follows:

                             ┌────────┐
                             ▼        │
            STARTING ────► READY     BUSY
                │            │        ▲
                │            └────────┘
                ▼
          SETUP_FAILED

Once setup is complete, we expect the model pod to cycle between READY and BUSY. If any other status is seen, or if BUSY is seen when we expect the container to be READY, we abort the pod.

### DIFF
--- a/python/cog/director/director.py
+++ b/python/cog/director/director.py
@@ -10,9 +10,10 @@ from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry  # type: ignore
 
 from .. import schema
+from ..server.http import Health
 from ..server.probes import ProbeHelper
 from ..server.webhook import requests_session, webhook_caller
-from .eventtypes import Health, HealthcheckStatus, Webhook
+from .eventtypes import HealthcheckStatus, Webhook
 from .healthchecker import Healthchecker
 from .prediction_tracker import PredictionTracker
 from .redis import EmptyRedisStream, RedisConsumerRotator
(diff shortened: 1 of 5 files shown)


python/cog/logging.py
```diff
@@ -85,5 +85,5 @@ def setup_logging(*, log_level: int = logging.NOTSET) -> None:
         logging.getLogger(l).propagate = True
 
     # Reconfigure log levels for some overly chatty libraries
-    logging.getLogger("uvicorn.access").setLevel(logging.WARNING)
+    # logging.getLogger("uvicorn.access").setLevel(logging.WARNING)
```
evilstreak, line 88: Is this intentional?

python/cog/director/director.py
```diff
@@ -262,7 +258,7 @@ def _handle_message(self, message_id: str, message_json: str) -> None:
                     tracker.update_from_webhook_payload(event.payload)
                 elif isinstance(event, HealthcheckStatus):
                     log.info("received healthcheck status update", data=event)
-                    if event.health != Health.HEALTHY:
+                    if event.health not in {Health.BUSY, Health.READY}:
```
evilstreak, line 261: Do we want to handle the model container saying it's `READY` without having sent a terminal webhook? My instinct is we worry about then if and when we observe it being a problem.

evilstreak (approved):
👍

One thing that might need addressing, otherwise ship it!

python/cog/director/director.py
nickstenning: I think there is a race condition here: we might get the READY healthcheck update before the webhook event, so I think we have to allow this here.

evilstreak: Yeah, there's definitely a race condition here. It'd have to be some more complicated logic, like if it flips to ready we should receive the terminal webhook within 5s otherwise we'd abort it, similar to the cancelation bit. Which feels like a lot of work for something that may not even be a problem in practice!

PR #918, opened 2023-02-02
Ignore XREADGROUP error on missing queue

evilstreak:
We've given director the ability to cycle through multiple Redis queues to make it possible to migrate to another Redis instance. For that to work it needs to not blow up when the queue is missing!

### DIFF
--- a/python/cog/director/redis.py
+++ b/python/cog/director/redis.py
@@ -48,13 +48,21 @@ class RedisConsumer:
         # None of this is ideal, and in future we likely want to improve
         # director's ability to recover and retry predictions that failed
         # exceptionally.
-        raw_messages = self.redis.xreadgroup(
-            groupname=self.redis_input_queue,
-            consumername=self.redis_consumer_id,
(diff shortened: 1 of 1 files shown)


python/cog/director/redis.py
```diff
@@ -48,13 +48,21 @@ def get(self) -> Tuple[str, str]:
...
+                streams={self.redis_input_queue: ">"},
+                count=1,
+                block=1000,
+            )
+        except redis.exceptions.ResponseError as e:
+            # treat a missing queue the same as an empty queue
+            if str(e).startswith("NOGROUP No such key"):
+                raise EmptyRedisStream()
+            else:
+                raise e
```
nickstenning, line 64:
```suggestion
                raise
```

It's easier to interpret the call stack if you just do a bare reraise here, otherwise you'll get the "While handling this exception, another exception was raised" frames.

evilstreak: Ah neat, I couldn't remember if Python supported the bare raise

PR #919, opened 2023-02-02
Allow customization of director and Cog ports

nickstenning:
This makes it a lot easier to test integration between director and Cog locally if port 5000 is in use (looking at you, Apple).

### DIFF
--- a/python/cog/director/__main__.py
+++ b/python/cog/director/__main__.py
@@ -30,6 +30,7 @@ signal.signal(signal.SIGTERM, _die)
 
 parser = ArgumentParser()
 
+parser.add_argument("--cog-url", type=str, default="http://localhost:5000")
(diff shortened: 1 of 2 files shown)


python/cog/director/__main__.py
```diff
@@ -49,12 +50,13 @@ def _die(signum: Any, frame: Any) -> None:
 
 events: queue.Queue = queue.Queue(maxsize=128)
 
-config = uvicorn.Config(create_app(events=events), port=4900, log_config=None)
+port = int(os.getenv("PORT", 4900))
```
erbridge, line 53: Is it weird that this is an environment variable while everything else is an argument? Should we make this an arg, too?

erbridge (approved):
One comment, but otherwise looks good to me.

mattt:

PR #923, opened 2023-02-03
Stop leaking Redis secrets on connection

erbridge:
We provide the host as context instead.


nickstenning (approved):

zeke:
@all-contributors please add @erbridge for bugs and code.

PR #924, opened 2023-02-03
record complete ts earlier

anotherjesse:
we should record completed_at timestamp earlier in the flow

### DIFF
--- a/python/cog/director/prediction_tracker.py
+++ b/python/cog/director/prediction_tracker.py
@@ -72,9 +72,11 @@ class PredictionTracker:
         return self._response.status
 
     def _update(self, mapping: Dict[str, Any]) -> None:
+        # we capture timestamp prior to updating response as copying response can take time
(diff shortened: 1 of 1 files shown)


anotherjesse:
fixed

erbridge (approved):
Seems sensible to me.

python/cog/director/prediction_tracker.py
```diff
@@ -72,9 +72,11 @@ def status(self) -> Optional[schema.Status]:
         return self._response.status
 
     def _update(self, mapping: Dict[str, Any]) -> None:
+        # we capture timestamp prior to updating response as copying response can take time
```
mattt, line 75:
```suggestion
        # capture timestamp prior to updating response, since copying response can take time
```

mattt (approved):
Thanks for contributing this, @anotherjesse! Merging this now.

mattt:
The code for Replicate's director was removed from Cog's codebase with #989.

PR #935, opened 2023-02-10
director: change default trace exporter to http

sirupsen:
Given that e.g. Refinery disables GRPC by default, I anticipate Protobuf over HTTP is the more common protocol—so we're going to default to that. I can't find an easy way for the Python SDK to properly respect `OTEL_EXPORTER_OTLP_PROTOCOL` that wouldn't be super hacky, so we'll default to this for now until the use-case arises for someone to need GRPC.

@nickstenning

### DIFF
--- a/python/cog/director/__main__.py
+++ b/python/cog/director/__main__.py
@@ -9,7 +9,7 @@ from typing import Any
 import structlog
 import uvicorn
 from opentelemetry import trace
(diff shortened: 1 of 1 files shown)


nickstenning (approved):
👍

sirupsen:
@nickstenning should I change it in the model runner too? That's where I copied the code with GRPC from (which is then copied from the official documentation). I didn't because it's technically a breaking change, and I'm unsure if anyone uses it

PR #936, opened 2023-02-10
director: set span attributes from env

sirupsen:
This sets span attributes from the environment if they're set. I'm not using `None` because OpenTelemetry emits warnings for span attributes set to `None`.

I could now make `set_current_prediction` in `Monitor` be replaced with `working` and `idle` methods, but it could be useful in the future for other monitoring of e.g. emitting utilization by user (highly sampled). Doesn't really do any damage IMO.

@nickstenning @evilstreak


nickstenning:
Retroactive LGTM! :)

sirupsen:
@nickstenning ugh sorry I think I confused the email on the other PR with an approval on this one 😳

nickstenning:
It's fine! As discussed, it's totally fine to merge your own stuff as long as you own it.

PR #943, opened 2023-02-15
director: only emit utilization after model is ready

sirupsen:
This changes it so that utilization is only emitted after the model is ready.

But shouldn't a model be penalized in its utilization for booting slowly? Yes, but not in this metric. That would typically be something you'd want to monitor as part of the utilization of the node itself that this is running on: a node is utilized if a model is _ready_ and booted on the node. Then your final utilization is `node_utilization * pod_utilization`.

**Edit:** I realized it's a bit more complicated, we need to make one of these two definitions come true, either (1):

```
Node Idle: When a Director pod is not up
Cog Idle: When a Model pod is not running a prediction, emitted from when Director is up 
```

OR (2)

```
Node Idle: When a Model pod is up
Cog Idle: When a Model pod is not running a prediction, emitted from when Model is up
```

Neither is true right now. This would enable us to do (2), or we change our own `Node Idle` to honour (1). Either way, our current situation is not correct.

@evilstreak @nickstenning

### DIFF
--- a/python/cog/director/monitor.py
+++ b/python/cog/director/monitor.py
@@ -54,12 +54,17 @@ class Monitor:
         active_sec = 0.0
         current_prediction: Optional[schema.PredictionResponse] = None
         current_prediction_started_at: Optional[float] = None
+        model_ready = False
 
         while not self._should_exit.is_set():
             try:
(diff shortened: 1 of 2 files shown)


mattt:

PR #944, opened 2023-02-16
Run setup on a background thread

nickstenning:
Currently, `Runner.setup()` blocks. When we call it in the FastAPI app `startup` hook, we block the initialisation of the HTTP server itself.

That in turn means that any attempts to connect to the HTTP interface while setup is running are refused, so we cannot successfully shut down a model container while setup is running. This leaves model pods in a broken state for up to 30m (the K8S termination grace period) until they are killed.

This changes `Runner.setup()` to a non-blocking method that returns an AsyncResult. We can then poll this object when the healthcheck endpoint is called to find out whether setup is complete.

Note that it is not a requirement to call the healthcheck endpoint. While setup is running, prediction endpoints will return a 409 Conflict indicating that the predictor is busy, and will start accepting requests once setup has completed. (Although note that currently, they will still accept requests even if setup has failed.)

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -58,6 +58,7 @@ def create_app(
 
     app.state.health = Health.STARTING
     app.state.setup_result = None
+    app.state.setup_result_payload = None
 
     runner = PredictionRunner(
         predictor_ref=predictor_ref,
@@ -80,21 +81,7 @@ def create_app(
         # https://github.com/tiangolo/fastapi/issues/4221
         RunVar("_default_thread_limiter").set(CapacityLimiter(threads))  # type: ignore
 
-        setup_start = datetime.now(timezone.utc)
-        setup_status, setup_logs = runner.setup()
-        setup_complete = datetime.now(timezone.utc)
-
-        app.state.setup_result = {
-            "logs": setup_logs,
-            "status": setup_status,
-            "started_at": setup_start,
-            "completed_at": setup_complete,
-        }
-
-        if setup_status == schema.Status.SUCCEEDED:
-            app.state.health = Health.READY
-        else:
-            app.state.health = Health.SETUP_FAILED
+        app.state.setup_result = runner.setup()
 
         probes = ProbeHelper()
         probes.ready()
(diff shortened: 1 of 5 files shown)


python/cog/server/http.py
```diff
@@ -241,6 +229,25 @@ def start_shutdown() -> Any:
...
         return JSONResponse({}, status_code=200)
 
+    def _check_setup_result() -> Any:
+        if app.state.setup_result is None:
+            return
+
+        if not app.state.setup_result.ready():
+            return
+
+        result = app.state.setup_result.get()
```
sirupsen, line 239: `AsyncResult` will allow you to consume the result repeatedly, right? It seems like it's just a wrapper object so I'm sure it's fine

python/cog/server/runner.py
```diff
@@ -276,6 +281,37 @@ def _upload_files(self, output: Any) -> Any:
             raise FileUploadError("Got error trying to upload output files") from error
 
 
+def setup(*, worker: Worker):
+    logs = []
+    status = None
+    started_at = datetime.now(tz=timezone.utc)
+
+    try:
+        for event in worker.setup():
```
sirupsen, line 290: This still has to finish before it shuts down, right? Or do the signals/events propegate in such a way that we interrupt the model booting, before we shut it down? Not obvious to me

sirupsen (approved):

python/cog/server/http.py
nickstenning: Yup, this is fine to call repeatedly.

python/cog/server/runner.py
nickstenning: This is running in a thread managed by `PredictionRunner`. When director requests a shutdown, `PredictionRunner.shutdown()` is called, which terminates the worker process and the thread running this code: https://github.com/replicate/cog/blob/d574a3496947eea7ae09929c2544c80bbf23e7d7/python/cog/server/runner.py#LL138C7-L138C7

So no, it should shut straight down, and testing that locally seems to validate that.

PR #945, opened 2023-02-16
Configure cog base URI via args to director

erbridge:
While we run these on the same network, it can be useful to not have to. For example, when using Docker Compose to spin up a local development environment, the model and director might use different services. Being able to configure the endpoints will make setting that up a lot simpler.


evilstreak (approved):

Lal119833 (approved):

mattt:

PR #947, opened 2023-02-20
runner: gracefully fail if failing to get url path

sirupsen:
@evilstreak

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -316,7 +316,14 @@ def _predict(
 
     for k, v in input_dict.items():
         if isinstance(v, types.URLPath):
-            input_dict[k] = v.convert()
+            try:
+                input_dict[k] = v.convert()
+            except requests.exceptions.RequestException as e:
+                tb = traceback.format_exc()
+                event_handler.append_logs(tb)
+                event_handler.failed(error=str(e))
+                log.warn("failed to download url path from input", exc_info=True)
+                return event_handler.response
 
     for event in worker.predict(input_dict, poll=0.1):
         if should_cancel.is_set():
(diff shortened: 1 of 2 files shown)


python/cog/server/runner.py
```diff
@@ -316,7 +316,16 @@ def _predict(
 
     for k, v in input_dict.items():
         if isinstance(v, types.URLPath):
-            input_dict[k] = v.convert()
+            try:
+                input_dict[k] = v.convert()
+            except Exception as e:
+                # Failed to get remote file.
+                # We do not re-raise since it is not an inference error.
```
evilstreak, line 323: We don't reraise for inference errors either, only for irrecoverable errors, so I think we can drop this comment.

python/cog/server/runner.py
```diff
@@ -316,7 +316,16 @@ def _predict(
 
     for k, v in input_dict.items():
         if isinstance(v, types.URLPath):
-            input_dict[k] = v.convert()
+            try:
+                input_dict[k] = v.convert()
+            except Exception as e:
```
evilstreak, line 321: I think we probably want a more tightly scoped exception here, to match what we do with webhooks and in director.

```suggestion
            except requests.exceptions.RequestException as e:
```

evilstreak (approved):
Great!

python/cog/server/runner.py
sirupsen: Ah... Fair, got it!

python/cog/server/runner.py
sirupsen: I just tried a couple of different things to try to break it, and Requests seems to do a good job of re-wrapping exceptions—so doesn't appear to be a problem, at least in some light testing. I still suspect we might hit exceptions not wrapped here, but it might be the better trade-off for now

zeke:
@all-contributors please add @sirupsen for code

PR #949, opened 2023-02-22
Repeatedly claim Redis messages during processing

evilstreak:
If director catastrophically fails (oom killed, preempted, etc) then the only signal we have that something has gone wrong is that the prediction never finishes and the message in the queue is never acked. To make use of that signal we have to wait longer than the prediction could possibly take.

This changes the behaviour of director so that it uses `XCLAIM` to continuously update the idle time of the message it's processing. If the idle time of any message is longer than the `claim_interval` used (with a resonable margin for error) then we can assume the director responsible for it has somehow died, and we can either retry it or fail it from whatever system monitors the queue.

This is done in a thread so that issues with, eg, webhook delivery don't stop the idle time from being updated at the interval.

### DIFF
--- a/python/cog/director/redis.py
+++ b/python/cog/director/redis.py
@@ -2,6 +2,8 @@ from typing import Callable, List, Optional, Tuple
 
 import redis
 import structlog
+import threading
+import time
 
 log = structlog.get_logger(__name__)
 
@@ -70,6 +72,11 @@ class RedisConsumer:
         self.message_id = key.decode()
         message = raw_message[b"value"].decode()
 
+        self._claimer = RedisMessageClaimer(
+            claim_message=self.claim_message(self.message_id)
+        )
+        self._claimer.start()
+
         return message
 
     def ack(self) -> None:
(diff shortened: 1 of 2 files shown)


nickstenning:
Hmmmm, I'm not sure how I feel about doing this in another thread. It seems to me that opens up possibilities of (e.g.) the main thread getting stuck in a loop somewhere while we happily claim the message every few seconds as if nothing is wrong.

I had imagined that we would bake this into the main `while not tracker.is_complete():` loop somehow. Perhaps you'd be up for pairing with me on this tomorrow to see if that's possible?

Alternatively, if we're keen to get *something* in to mitigate possible issues for ThumbnailAI, I can probably stomach it for now :)

evilstreak:
> Hmmmm, I'm not sure how I feel about doing this in another thread. It seems to me that opens up possibilities of (e.g.) the main thread getting stuck in a loop somewhere while we happily claim the message every few seconds as if nothing is wrong.
> 
> I had imagined that we would bake this into the main `while not tracker.is_complete():` loop somehow. Perhaps you'd be up for pairing with me on this tomorrow to see if that's possible?

Happy to pair on it! I first tried to put it into the main loop, but the way we're currently sending webhooks risks false positives on dropped predictions. If we can work out how to get around that, then I agree that having it in the main loop is much safer.

I figure the thread is OK, at least for now, because the main things we're guarding against are the node being preempted or the director being oom killed (or similar), both of which will destroy this thread at the same time.

python/cog/director/redis.py
```diff
@@ -103,3 +126,32 @@ def rotate(self) -> None:
...
+        last_claimed_at = time.perf_counter()
+
+        while not self._should_exit.is_set():
+            now = time.perf_counter()
+            if now >= last_claimed_at + self._claim_interval:
+                self._claim_message()
+                last_claimed_at = now
+
+            # Only sleep for a short time to respect _should_exit
+            time.sleep(0.01)
```
erbridge, line 157: This is so we don't block for 2 seconds when acking, right? Do we need to sleep at all if this is running on a thread?

nickstenning (approved):
Tested locally and this seems to work well.

PR #961, opened 2023-03-08
Tag builds of director image with latest

evilstreak:
This will be helpful in development, where manually updating to new pinned versions is more of a faff than we want.


nickstenning (approved):
LGTM, but I think we could maybe tweak the `buildx` task to tag with git tags as well (if present) and use that instead? It would be great to get rid of CloudBuild entirely.

PR #965, opened 2023-03-23
fix invalid empty lists in cog.yaml

### LINKED ISSUE #368: Lists in cog.yaml should be allowed to be empty

pysqz:
resolves #368. pls review it, @bfirsh

### DIFF
--- a/pkg/cli/init-templates/cog.yaml
+++ b/pkg/cli/init-templates/cog.yaml
@@ -6,21 +6,26 @@ build:
   gpu: false
 
   # a list of ubuntu apt packages to install
-  # system_packages:
+  system_packages:
     # - "libgl1-mesa-glx"
     # - "libglib2.0-0"
+    # - "cowsay"
+
+  # a list of setup commands to run in the environment before your python packages are installed
+  pre_install:
+    # - "cowsay moo"
 
   # python version in the form '3.8' or '3.8.12'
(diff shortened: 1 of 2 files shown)


pysqz:
> resolves #368. pls review it, @bfirsh

bfirsh:
Nice, thanks! Would you mind writing a test?

ccxx0:
> > Nice, thanks! Would you mind writing a test?
> 
> sure, test `cog init` is enough?

@bfirsh @zeke

mattt:
Hi @pysqz. Thanks so much for your contribution. I just merged https://github.com/replicate/cog/pull/1073, which specifies [an array for `type`](https://json-schema.org/understanding-json-schema/reference/type.html#type) instead of using `anyOf`.

PR #968, opened 2023-03-24
allow arbitrary input types for model inputs

jgershen:
This is especially useful for transformers, where you often want to generate multiple text completions with the same settings. With this change, you can now define cog models which take in lists of strings as parameters.


jgershen:
Example of use at https://github.com/replicate/cog-flan-t5/pull/1

jgershen:
this probably doesn't fix all the test errors; I need to fix some package versions in my local dev env

bfirsh:
Whomp looks like more failures. :(

zeke:
> you can now define cog models which take in lists of strings as parameters.

Nice. Would you mind also updating the docs around this?

mattt:
Closing in favor of https://github.com/replicate/cog/pull/1311 and/or https://github.com/replicate/cog/pull/1313. See https://github.com/replicate/cog/pull/1292#issuecomment-1730714860.

Issue #969, opened 2023-03-24
Streaming output for language models


bfirsh:
Large language models typically stream tokens as they are running.

We've called this "progressive" output in the past in Cog, but "streaming" is the more common term, so perhaps we switch to that.

## Prior art
- [OpenAI's API](https://platform.openai.com/docs/api-reference/completions/create) returns server-side events as JSON messages in the format `{"data": "..."}` with each token.

## Proposed design

Models return an iterator of strings that are concatenated together. There is a special type called `Stream[str]` that works the same way as `Iterator[str]`, except it implies that the outputs should be concatenated together to get the final result.

For example:

```python
from cog import BasePredictor, Stream

class Predictor(BasePredictor):
    def predict() -> Stream[str]:
        yield "The "
        yield "quick "
        yield "brown "
        yield "fox "
```

The output type looks something like:

```json
{
    "type": "array",
    "items": {
        "type": "string",
    },
    "x-cog-array-type": "stream",
}
```

## Other designs considered

- We just use `Iterator[str]`. The problem with this is that the current semantics of this is in-progress snapshots of the output (e.g. diffusion models diffusing), not parts of the output that need concatenating together. You could also imagine situations where you return distinct strings as the output that you don't want concatenated together. So, we need a bit of annotation to hint that the output needs concatenating for things like Replicate's web interface.
- We return an iterator of output in the the format `["The", "The quick", "The quick brown"]`. The advantage of this is you don't have to concatenate output together to get the completed output (it's just `output[-1]`) but this returns a lot of data in each response and makes it harder to process each token as its coming in (you have to keep track of your position in the string).
- We modify Cog to output a string that can change over time. E.g. `{"output": "The"}`, `{"output": "The quick"}` etc. This means that the default case where you just want to get the complete output is really simple (no concatenation required) but it'll require some rearchitecting of Cog and makes it harder to process each token as its coming in (you have to keep track of your position in the string).
- We have a similar server-side events API to OpenAI. This is a big change.

## See also
- https://github.com/replicate/cog-llama/pull/4

daanelson:
@bfirsh I like this. I think the one outstanding thread here is handling the case where a user generates more than 1 completion for a given prompt. OpenAI returns a stream of jsons, each of which are basically a `(completion_id, token, metadata)` tuple, where `completion_id` is between 0 and the number of requests generated. Quick example: 

```(base_ml) dans-mbp:analytics dan$ curl https://api.openai.com/v1/completions \
[...]

PR #975, opened 2023-03-29
Add ConcatenateIterator type

bfirsh:
Implements https://github.com/replicate/cog/issues/969 

It's the same thing, just with a different name. `Stream` is a new noun, whereas `ConcatenateIterator` is a verbose description of what it does: it is an iterator that gets concatenated.

This is an experiment at passing hints about how outputs should be displayed to Replicate. You can imagine other hints about display that we might want to pass.

### DIFF
--- a/python/cog/types.py
+++ b/python/cog/types.py
@@ -6,7 +6,7 @@ import pathlib
 import requests
 import shutil
 import tempfile
-from typing import Any, Callable, Dict, Iterator, List, Union
+from typing import Any, Callable, Dict, Iterator, List, TypeVar, Union
 from urllib.parse import urlparse
 
 from pydantic import Field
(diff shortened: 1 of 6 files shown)


python/cog/types.py
```diff
@@ -6,7 +6,7 @@
 import requests
 import shutil
 import tempfile
-from typing import Any, Callable, Dict, Iterator, List, Union
+from typing import Any, Callable, Dict, Iterator, List, Union, Iterator
```
daanelson, line 9: nit - importing Iterator twice

daanelson:
thinking out loud here - why not call this a `Stream`? `ConcatenateIterator` would be a replicate-specific bit of jargon, `Stream` is widely used. 

There is the counterargument that we should preserve `Stream` for output sequences which aren't necessarily supposed to be concatenated - i.e. when streaming batch/multiple completions. IMO that's just a bog standard `Iterator`, which we already support; not particularly different from yielding multiple image generations. 

Ok, I'm coming around to it. The name feels a bit long still, but I can't think of anything else offhand.

bfirsh:
Agreed the name is a bit heavy-handed, but it's hard to disambiguate why `Stream` and `Iterator` are different, unless you explain it really carefully in some documentation.

Maybe we both agree that the name is crap, but names are also easily changeable so maybe we run with this and let a better name percolate? 😄

At some point when we figure it out, we just do `ConcatenateIterator = MuchBetterName` for backwards compatibility.

PR #976, opened 2023-03-29
Training 🚂

bfirsh:
This adds a `train` option to `cog.yaml` and a `cog train` command to run it.

There are a few bits of refactoring to make this possible, contained in some significant commits. Notably, predictors can now just be functions if you have no `setup()`.

## How it works

`cog.yaml`:

```yaml
build:
  python_version: "3.10"
train: "train.py:train"
```

`train.py`:

```python
from cog import BasePredictor, File
import io

def train(param: str) -> File:
    return io.StringIO("hello " + param)
```

Then you can run it like this:

```
$ cog train -i param=train
...

$ cat weights
hello train
```

## API

The HTTP server can be booted up in two different modes, specified with the `--x-mode` argument: `predict` and `train`.

In `predict` mode, it behaves like it did before.

In `train` mode, if you call `POST /predictions` it actually runs the thing specified by `train` in `cog.yaml`.

## Notes

- The way the API works is obviously... suboptimal. Ideally we'd have something like a separate `/trainings` endpoint and both predictions and trainings can be run when the container is booted up.
  - Unfortunately this is fiddly, so I opted to not do it this way. For example, `Predictor.setup()` gets called on boot, but we don't want to do that if we're training. Also, the worker would require a lot of rewiring to support both things.
  - So – I opted for a bodge where the HTTP server can be booted up in either prediction or training mode, and in training mode `POST /predictions` does a training (bleurgh).
  - This should probably be tidied up at some point but this will get us going.
- Weights need to be loaded in `setup()` somehow. This is not done yet, and I have a half-working version locally.
- We probably want to support directories as `Path`, because this is often the format that weights are in. Perhaps we can add support in Cog to turn a `Path` that points to a directory into a tarball, which is a common pattern (e.g. zip files in DreamBooth).
- We probably want to pass the weights as an input to the trainer to use as the base weights.
- `cog train` passes around files with data URLs, which ain't going to scale. We probably want to use `file://` and store it on the shared Docker volume.
- There aren't any schemas! The model will have to boot up to get validation errors, we can't generate GUIs, etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ pkg/dockerfile/embed/cog.whl: python/* python/cog/* python/cog/server/* python/c
 	mkdir -p pkg/dockerfile/embed
 	cp python/dist/*.whl $@
 
+.PHONY: cog
 cog: pkg/dockerfile/embed/cog.whl
 	$(eval COG_VERSION ?= $(shell git describe --tags --match 'v*' --abbrev=0)+dev)
 	CGO_ENABLED=0 $(GO) build -o $@ \
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -40,11 +40,12 @@ https://github.com/replicate/cog`,
 	rootCmd.AddCommand(
 		newBuildCommand(),
 		newDebugCommand(),
+		newInitCommand(),
+		newLoginCommand(),
 		newPredictCommand(),
 		newPushCommand(),
 		newRunCommand(),
-		newLoginCommand(),
-		newInitCommand(),
+		newTrainCommand(),
 	)
 
 	return &rootCmd, nil
--- a/pkg/cli/train.go
+++ b/pkg/cli/train.go
@@ -1,0 +1,99 @@
+package cli
+
+import (
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/replicate/cog/pkg/config"
+	"github.com/replicate/cog/pkg/docker"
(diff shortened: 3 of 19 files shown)


bfirsh:
I have pushed a commit that attempts to read weights from the `COG_WEIGHTS` environment variable and passes it to `setup()`, but you can't seem to read environment variables from within the worker (both from the tests and running the fixture manually). Maybe somebody who knows more about how the worker works can figure this out.

anotherjesse:
@nickstenning  / @andreasjansson I think we should re-write the URLs before downloading (either in COG or replicate API)

It is 5x faster to download from GCS directly than to stream through the replicate.delivery CDN

```
15:     if strings.HasPrefix(url, "https://replicate.delivery/pbxt/") {
16:             url = strings.Replace(url, "https://replicate.delivery/pbxt/", "https://storage.googleapis.com/replicate-files/", 1)
```

nickstenning:
> It is 5x faster to download from GCS directly than to stream through the replicate.delivery CDN

Interesting! If you get a chance could you test that with the new CDN URLs? `https://pbxt.replicate.delivery/` rather than `https://replicate.delivery/pbxt/`.

andreasjansson:
I'm going to merge and deploy this. There are some known things we want to change (e.g. booting models in "training mode"), but it holds together enough that we can test with actual models in production. I have tested a few existing deployments by manually pointing to this branch, and it works.

PR #978, opened 2023-03-31
Return better error code when setup is running 

ruravi:
If the server is still running setup and we issue a new predict call, there is an exception thrown because the workers aren't ready yet.
```
 cog.server.exceptions.InvalidStateException: Invalid operation: state is WorkerState.NEW (must be WorkerState.READY)
```

This commit catches that exception and returns a 503 - retryable server error to ask the client to retry prediction when the server is ready.

---

Note: The runner doesn't seem to throw a RunnerBusyError if it's doing setup. Instead it throws an InvalidStateException.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -30,6 +30,7 @@ from ..predictor import (
     load_config,
     load_predictor_from_ref,
 )
+from .exceptions import InvalidStateException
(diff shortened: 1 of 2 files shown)


ruravi:
Closing this because my commit didn't have a sign-off

Issue #966, opened 2023-03-23
Server reports "Already running a prediction" during setup


evilstreak:
It should say it's doing setup and isn't ready, instead!

mattt:
It looks like this was introduced by #944 with these lines:

https://github.com/replicate/cog/blob/dfff002e275cdb6af8af4cc290e976813a4a8735/python/cog/server/runner.py#L57-L59

@nickstenning Would it make sense to raise a different error here? 

Also, it looks like `is_busy` has side effects here, which I wasn't expecting...

https://github.com/replicate/cog/blob/dfff002e275cdb6af8af4cc290e976813a4a8735/python/cog/server/runner.py#L141-L142

Pankaj-Samyak:
Does anyone know how long it takes to Cog for starting up the first time when running it in a Docker container?

PR #979, opened 2023-03-31
Return better error code when setup is running #978

ruravi:
If the server is still running setup and we issue a new predict call, there is an exception thrown because the workers aren't ready yet.
```
 cog.server.exceptions.InvalidStateException: Invalid operation: state is WorkerState.NEW (must be WorkerState.READY)
```

This commit catches that exception and returns a 503 - retryable server error to ask the client to retry prediction when the server is ready.

---
Note: The runner doesn't seem to throw a RunnerBusyError if it's doing setup. Instead it throws an InvalidStateException.


mattt:
Hi, @ruravi. Thanks for contributing this PR. Apologies for not responding sooner.

I just pushed a merge commit to get this up to date with the latest `origin/main`.

I'm still trying to understand the timing of when the server raises `InvalidStateException` vs. `RunnerBusyError`, which is also discussed in https://github.com/replicate/cog/issues/966... Can you help share the steps you took to run the model and issue a new predict call?

PR #989, opened 2023-04-10
Remove cog.director and cog.server.redis_queue

### LINKED ISSUE #985: Remove cog.director and cog.server.redis_queue
We've extracted director into its own repository from where it's deployed independently of Cog.

We should:

- remove `cog.director` from this repository
- remove `cog.server.redis_queue` as it will no longer be supported
- update Cog's dependencies to be minimal -- several dependencies that were brought in for director are not needed by Cog itself

nickstenning:
Director is Replicate-only code and has moved to its own codebase. `cog.server.redis_queue` is unused (replaced by director) and was marked as experimental so it should be reasonable to remove it.

This also allows us to remove several dependencies from cog's Python package, namely:

- `opentelemetry-*` (telemetry is handled by director)
- `protobuf` (transitive dependency via opentelemetry)
- `redis`

Closes #985.


erbridge (approved):
🎉

PR #990, opened 2023-04-11
fix(python): put file to endpoint `file.seek(0)`

jianghushinian:
I deploy models with cog use: `docker run -d -p 5000:5000 my-model python -m cog.server.http --upload-url http://host:port/file`.

I found that when I used an asynchronous method to generate images and then uploaded to 'upload-url', the result was empt.

when I modified it using `fh.seek(0)`, I succeeded.

### DIFF
--- a/python/cog/files.py
+++ b/python/cog/files.py
@@ -40,6 +40,8 @@ def guess_filename(obj: io.IOBase) -> str:
 def put_file_to_signed_endpoint(
     fh: io.IOBase, endpoint: str, client: requests.Session
 ) -> str:
+    fh.seek(0)
+
     filename = guess_filename(fh)
     content_type, _ = mimetypes.guess_type(filename)
 


mattt:
Hi @jianghushinian. Thanks for submitting this PR. It looks like the method for uploading files already seeks to the beginning of the file [on this line in `upload_file`](https://github.com/replicate/cog/blob/5e3fe5b09da709768293d81cee085293ce448a3c/python/cog/files.py#L11). Do you have a sense of why the change in this PR would be necessary?

jianghushinian:
> Hi @jianghushinian. Thanks for submitting this PR. It looks like the method for uploading files already seeks to the beginning of the file [on this line in `upload_file`](https://github.com/replicate/cog/blob/5e3fe5b09da709768293d81cee085293ce448a3c/python/cog/files.py#L11). Do you have a sense of why the change in this PR would be necessary?

Hello, based on my test, the asynchronous file upload code does not execute `upload_file`, but instead executes `put_file_to_signed_endpoint`. The `upload_file` function is only executed when the file is uploaded to a specific directory.

Furthermore, I conducted another test where I called `fh.seek(0)` before asynchronously uploading the file in the `predict` method, and it also took effect.

Adding `fh.seek(0)` in `put_file_to_signed_endpoint` can be considered as an additional precautionary measure.

mattt (approved):

mattt:
> Hello, based on my test, the asynchronous file upload code does not execute `upload_file`, but instead executes `put_file_to_signed_endpoint`. The `upload_file` function is only executed when the file is uploaded to a specific directory.

Ah, I see that now. I didn't appreciate how these were parallel code paths. In that way, adding the `file.seek(0)` harmonizes the implementation of `put_file_to_signed_endpoint` with `upload_file`. Thanks again for your contribution and taking the time to explain that to me.

PR #991, opened 2023-04-11
Document environment variables.

### LINKED ISSUE #829: Document all the environment variables
There are a handful of environment variables that can be used to configure Cog's behavior. We should document them all in one place.

From a quick search, I found these:

- `CGO_ENABLED`
- `COG_LOG_LEVEL`
- `COG_NO_UPDATE_CHECK`
- `OTEL_SERVICE_NAME`
- `PORT `

joannejchen:
Introduces another documentation file to list the environment variables used by Cog, how they are used, and possible values for them.

Closes #829.


zeke (commented):
Thanks so much for opening this PR. This is a great start.

I think it would be useful to have each variable be a heading, so that we can deep link to it. Something like this, with a heading and a paragraph for each:

```
### `PORT`

This defines what port is exposed from the container for the HTTP server to be hosted on. Defaults to 5000.
```

Then we could link to it, like `docs/environment.md#port`

docs/environment.md
```diff
@@ -0,0 +1,108 @@
+# Environment Variables
+
+This guide provides a list of environment variables that can be set that impact the execution of Cog. This assumes that you have already followed the [main steps to set up Cog](https://github.com/replicate/cog/blob/main/docs/getting-started.md) and have a general familiarity with how to run Cog.
+
+
+## General
+
+This section lists the relevant environment variables for general execution of Cog.
+
+### CGO_ENABLED
```
zeke, line 10:
```suggestion
### `CGO_ENABLED`
```

Let's put these all in backticks to make it clear that they're "code".

zeke (commented):
This is looking great so far!

docs/environment.md
joannejchen: Updated!

zeke:
@all-contributors please add @joannejchen for docs 💛

PR #995, opened 2023-04-14
Feat: Deprecate Tini usage since already in Docker

mihaigalos:
This removes the reference to Tini, since it is already in Docker.
See: https://github.com/krallin/tini#using-tini


nickstenning:
Thanks for the PR @mihaigalos but we need the container images that `cog` builds to run correctly under `containerd` and not just Docker, so I don't think we'd merge this.

PR #996, opened 2023-04-14
Add /trainings API and integrate with CLI

andremoeller:
## Background

#976 added a `cog train` command to the CLI pointed at the `/predictions` API, allowing the user to specify a value for `train` in `cog.yaml` specifying a path to a callable.

## Overview

This PR adds a `/trainings` API to Cog containers and points the CLI to it. It additionally defines a base class `Trainer` that has an optional `cancel` method.

```python
from cog import BaseTrainer

class Trainer(BaseTrainer):
    @abstractmethod
    def train(self, **kwargs: Any) -> Any:
        """
        Train a model.
        """

    def cancel(self) -> None:
        """
        An optional method to handle training cancellation to save model weights.
        """
```

The `cancel` method is run upon a cancellation request being received. This is an especially important use case for training, when one may want to save model weights or other state when, for example, a job is  running for considerably longer than expected, or when the user inspects the training logs and suspects the model is being over-trained.

## Remarks

* This PR doesn't solve the problem Ben notes in #976 about being able to run both predictions and trainings in a single container, since this would require loading and unloading the model a Predictor would load during set-up, making interleaving predictions and trainings complicated.
* There's still work to do in the CLI to disentangle the `train` command from the `predict` package.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -14,21 +14,23 @@ from typing import Any, Callable, Dict, List, Optional, Type, Union
 from typing_extensions import get_origin, get_args, Annotated
 import yaml
 
-from .errors import ConfigDoesNotExist, PredictorNotSet
+from .errors import ConfigDoesNotExist, PredictorNotSet, TrainerNotSet
 from .types import (
     Input,
     Path as CogPath,
     File as CogFile,
-    URLFile,
     URLPath,
-    get_filename,
 )
 
 
 ALLOWED_INPUT_TYPES = [str, int, float, bool, CogFile, CogPath]
 
 
-class BasePredictor(ABC):
+class BaseRunnable(ABC):
+    pass
+
+
+class BasePredictor(BaseRunnable):
     def setup(self, weights: Optional[Union[CogFile, CogPath]] = None) -> None:
         """
         An optional method to prepare the model so multiple predictions run efficiently.
@@ -41,12 +43,33 @@ class BasePredictor(ABC):
         """
 
 
-def run_setup(predictor: BasePredictor) -> None:
-    weights_type = get_weights_type(predictor.setup)
+class BaseTrainer(BaseRunnable):
+    def setup(self, weights: Optional[Union[CogFile, CogPath]] = None) -> None:
+        """
+        An optional method to initialize the trainer.
+        """
+
+    @abstractmethod
+    def train(self, **kwargs: Any) -> Any:
+        """
+        Train a model.
+        """
+
+    def cancel(self) -> None:
+        """
+        An optional method to handle training cancellation to save training artifacts.
+        """
+
+
+Runnable = Union[BasePredictor, BaseTrainer]
+
+
+def run_setup(runnable: Runnable) -> None:
+    weights_type = get_weights_type(runnable.setup)
 
     # No weights need to be passed, so just run setup() without any arguments.
     if weights_type is None:
-        predictor.setup()
+        runnable.setup()
         return
 
     weights: Union[io.IOBase, Path, None]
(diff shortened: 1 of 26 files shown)


python/cog/predictor.py
```diff
@@ -184,15 +199,18 @@ def cleanup(self) -> None:
...
 
-def get_predict(predictor):
-    if hasattr(predictor, "predict"):
-        return predictor.predict
-    return predictor
+def get_run_method(runnable: Runnable):
+    if hasattr(runnable, "predict"):
+        return runnable.predict
+    if hasattr(runnable, "train"):
+        return runnable.train
```
nickstenning, line 206: We might want be more explicit here about which one we expect on which type:

```suggestion
    if isinstance(runnable, BasePredictor) and hasattr(runnable, "predict"):
        return runnable.predict
    if isinstance(runnable, BaseTrainer) and hasattr(runnable, "train"):
        return runnable.train
```

python/cog/server/http.py
```diff
@@ -111,6 +127,86 @@ def healthcheck() -> Any:
...
+        # [compat] If no body is supplied, assume that this model can be run
+        # with empty input. This will throw a ValidationError if that's not
+        # possible.
+        if request is None:
+            request = TrainingRequest(input={})
+        # [compat] If body is supplied but input is None, set it to an empty
+        # dictionary so that later code can be simpler.
+        if request.input is None:
+            request.input = {}
+
```
nickstenning, line 161: We might be able to get away with dropping these for trainings.

python/cog/server/http.py
```diff
@@ -111,6 +127,86 @@ def healthcheck() -> Any:
...
+            )
+        if runner.is_busy():
+            return JSONResponse({"detail": "Already training."}, status_code=409)
+
+        respond_async = prefer == "respond-async"
+
+        return _train(request=request, respond_async=respond_async)
+
+    def _train(*, request: TrainingRequest, respond_async: bool) -> Response:
+        # [compat] If no body is supplied, assume that this model can be run
```
nickstenning, line 152: We might be able to get away with dropping these `[compat]` clauses for trainings.

python/cog/server/http.py
```diff
@@ -111,6 +127,86 @@ def healthcheck() -> Any:
...
+        try:
+            response = TrainingResponse(**async_result.get().dict())
+        except ValidationError as e:
+            _log_invalid_output(e)
+            raise HTTPException(status_code=500)
+
+        response_object = response.dict()
+        response_object["output"] = upload_files(
+            response_object["output"],
+            upload_file=lambda fh: upload_file(fh, request.output_file_prefix),  # type: ignore
```
nickstenning, line 184: Similarly, we can probably get rid of this part given that we don't support `output_file_prefix` on `TrainingRequest`s if I read the earlier code correctly. That would then mean that we can pass in `upload=True` to `runner.run(...)` a few lines earlier.

python/cog/predictor.py
andremoeller: Done, except for backwards compatibility, I kept

```
if hasattr(runnable, "predict"):
    return runnable.predict
```

Since it wasn't enforced that the user extend `BasePredictor`.

python/cog/server/http.py
andremoeller: Removed since there's no backwards compatibility to be maintained for this API.

python/cog/server/http.py
andremoeller: Removed, and set `upload=True`.

python/cog/schema.py
```diff
@@ -81,3 +78,21 @@ def with_types(cls, input_type: t.Type, output_type: t.Type) -> t.Any:
...
+
+
+class TrainingResponse(JobResponse):
+    pass
+
+
+class PredictionRequest(JobRequest):
+    # TODO: deprecate this
+    output_file_prefix: t.Optional[str]
+    pass
```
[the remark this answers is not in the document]
andremoeller: Done.

andremoeller:
For the second revision I removed `setup` from the `BaseTrainer` class to keep the API surface small; setup logic can be put in `train` for now instead.

PR #999, opened 2023-04-17
Vendor Python packages

### LINKED ISSUE #409: Vendor Pydantic and FastAPI

andreasjansson:
Cog model code is run in the same environment as the Cog server code. Sometimes this means that users will have conflicting dependencies, resulting in incredibly hard-to-debug issues that only appear once the model has been pushed to Replicate.

This PR vendors all Python packages using the [vendoring](https://pypi.org/project/vendoring/) library.

PR is broken up into two commits: First one adds all the files in the `_vendor` directory, second one is the actually changed code.

Closes #409


zeke:
@all-contributors please add @andreasjansson for maintenance 👷🏼

Makefile
```diff
@@ -100,3 +100,8 @@ lint: lint-go lint-python
 .PHONY: mod-tidy
 mod-tidy:
 	$(GO) mod tidy
+
+.PHONY: vendor-python
+vendor-python:
+	pip install vendoring
```
nickstenning, line 106: 1. Maybe this could be a script in `script/` as we're slowly trying to deprecate the `Makefile`, I think.

2. If you do want it to stay in here then please replace `pip` with `$(PYTHON) -m pip` and `vendoring` with `$(PYTHON) -m vendoring` if possible, otherwise the Makefile breaks when a custom `PYTHON` is supplied.

python/cog/_vendor/vendor.txt
```diff
@@ -0,0 +1,24 @@
+attrs==22.2.0
+fastapi==0.95.1
+    starlette==0.26.1
+    anyio==3.6.2
+    sniffio==1.3.0
+    httpx==0.24.0
+    httpcore==0.17.0
+    h11==0.14.0
+    curio==1.6
+hypothesis==6.72.0
```
nickstenning, line 10: This is a dev dependency, so I would have thought we don't want to vendor it, right?

python/cog/_vendor/vendor.txt
```diff
@@ -0,0 +1,24 @@
...
+requests==2.28.2
+    certifi==2022.12.7
+    chardet==5.1.0
+    idna==3.4
+    urllib3==1.26.15
+structlog==23.1.0
+typing_extensions==4.5.0
+uvicorn[standard]==0.21.1
+    click==8.1.3
+responses==0.23.1
```
nickstenning, line 23: Ditto here. I'm pretty sure this is a dev/test dependency only.

test-integration/Makefile
```diff
@@ -1,4 +1,4 @@
 .PHONY: test
 test:
 	pip install -r requirements.txt
-	pytest -n auto -vv
```
nickstenning, line 4: Did you intend to remove this?

nickstenning:
A couple of general thoughts:

Vendoring is a pretty big step to take with some reasonably large ongoing maintenance costs. Even the `vendoring` package itself [contains some fairly strongly-worded disincentives](https://pypi.org/project/vendoring/) against using it, even going so far as to say:

> As a general rule of thumb, if the project is going to be a PyPI package, it should not use this tool.

Are there ways we could achieve the same ends without going this route?

If we can avoid the need to load the user's code into the same process as most of `cog`, then we would likely be able to achieve similar (or better) isolation between `cog` and the user's code simply by installing and running `cog` in its own virtualenv.

To my knowledge, there are currently only two reasons we load the user's code:

1. To inspect the input schema of the Predictor.
2. When we actually run a prediction.

Both of these things need only relatively small parts of of `cog`. (1) Needs the `cog.schema` package, and (2) needs `cog.server.worker._ChildWorker` and its dependencies.

Perhaps that hints at another solution here, namely to split the `cog` package into two parts:

1. `cog`, which contains only that code which _must_ be loaded into the same Python runtime as the user's code. To my knowledge, the only external deps here are `attrs` and `pydantic`, or could easily be made so.
2. `cog_internal` (or similar) which contains the HTTP server code and its dependencies. This would _not_ need to be released as a `PyPI` package and could instead be embedded into the `cog` CLI itself, from where it could be injected into the built images and installed into its own virtualenv.

andreasjansson:
Thanks @nickstenning, you're making some really good points here.

I think we have a legitimate use case for vendoring here, in that we are running Cog alongside user code with arbitrary dependencies and that is sometimes causing models to fail boot.

You're right that vendoring creates a maintenance burden on keeping patches up to date when updating Cog's dependencies. In my experience from writing those patches yesterday, that was fairly trivial work, but there could of course be more complicated patches to write in the future.

I like the idea of limiting the amount of code that _has_ to interact with users' code, thereby limiting the number of dependencies we have to vendor.

Given that this currently works (if I can get the integration test to not hang on CI) and fixes user-facing issues, how about we ship this change now and then split Cog to reduce the number of packages we have to vendor?

nickstenning:
I'm ok with that approach in principle, though my comments on this PR still stand, and I'm also unsure on whether vendoring actually fixes https://github.com/replicate/cluster/issues/324 -- as I wasn't sure how to reproduce this.

mattt:
One alternative to vendoring our code would be to use something like [PEX](https://pex.readthedocs.io/) to package Cog's HTTP server into a [self-contained executable](https://pex.readthedocs.io/en/v2.1.137/recipes.html).

bfirsh:
In case some vague data is useful: we analyzed a few methods of doing this, including PEX, for Keepsake. We settled on the technique that Andreas is using in this PR as the best way of doing it, but I can't really remember why. This was ~3 years ago, so the decision might be out of date.

andreasjansson:
Superseded by https://github.com/replicate/cog/pull/1615

PR #1008, opened 2023-04-25
Record git commit on images

bfirsh:
If a `.git` directory exists when doing `cog build` or `cog push`, this pull request sets the label `run.cog.git_commit` from `HEAD`. This lets us link an image back to the code it was built from.

An alternate name could be `run.cog.git_head` to make it clear precisely what it is.

In the future, you could imagine some indication of whether the commit is dirty or not. I wonder whether there is a standard format for indicating this stuff beyond just appending `-dirty`?

### DIFF
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -4,6 +4,9 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"os"
+	"os/exec"
+	"path"
 
 	"github.com/replicate/cog/pkg/config"
 	"github.com/replicate/cog/pkg/docker"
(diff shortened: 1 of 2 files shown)


anotherjesse:
This looks super useful.  

I prefer `git_commit` over `git_head` (this is a soft preference)

anotherjesse:
lgtm! - if it fails to get a commit, it fails gracefully (meaning it will warn and keep going)

evilstreak:
> In the future, you could imagine some indication of whether the commit is dirty or not. I wonder whether there is a standard format for indicating this stuff beyond just appending `-dirty`?

The default for git is `-dirty`:

```
% git describe --tags --dirty
v0.7.0-beta16-2-g1fbdde7-dirty
```

mattt:
+1 on the idea and general approach.

The OCI Image Format Specification defines a [set of pre-defined annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys), including the Git revision and version.

The convention I'm most familiar with is to set `org.opencontainers.image.revision` to `$(git show --pretty=format:%H --no-patch)` and `org.opencontainers.image.version` to `$(git describe --tags --dirty)`.

mattt:
@bfirsh I just pushed two commits to the feature branch.

- 2c073217dd3456083e7ddda761153873a65b26dc adopts the OCI annotation for source revision (commit)
- 35d33c7d3359316b6546b93bf9b3d7586015ba4c adds a new OCI annotation for source version (tag) 

Feel free to back these out if you're not a fan 😄

mattt (approved):

bfirsh:
Oh, brilliant. That's exactly what I was looking for. (As somebody who worked on OCI specs it's a bit embarrassing I didn't think of that. 🙃)

PR #1010, opened 2023-04-25
Add support for build secrets

### LINKED ISSUE #1009: Build secrets
This syntax:

```
build:
  run:
    - command: pip install ...
      mounts:
       - type: secret
         id: someid
         target: path/in/container
```

Should turn into:

```
RUN --mount=type=secret,id=someid,target=path/in/container pip install ...
```

then to build you would use:

```
cog build --secret id=someid,src=path/in/host
```

todo:

- [ ] end-to-end documentation on this feature
- [ ] enable feature to work without swarm
- [ ] ensure `cog build` re-builds when contents of secret file changes

mattt:
Resolves #1009

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -19,16 +19,25 @@ import (
 // TODO(andreas): custom cpu/gpu installs
 // TODO(andreas): suggest valid torchvision versions (e.g. if the user wants to use 0.8.0, suggest 0.8.1)
 
+type RunItem struct {
+	Command string `json:"command,omitempty" yaml:"command"`
+	Mounts  []struct {
+		Type   string `json:"type,omitempty" yaml:"type"`
+		ID     string `json:"id,omitempty" yaml:"id"`
+		Target string `json:"target,omitempty" yaml:"target"`
+	} `json:"mounts,omitempty" yaml:"mounts"`
+}
+
 type Build struct {
-	GPU                bool     `json:"gpu,omitempty" yaml:"gpu"`
-	PythonVersion      string   `json:"python_version,omitempty" yaml:"python_version"`
-	PythonRequirements string   `json:"python_requirements,omitempty" yaml:"python_requirements"`
-	PythonPackages     []string `json:"python_packages,omitempty" yaml:"python_packages"` // Deprecated, but included for backwards compatibility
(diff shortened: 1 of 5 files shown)


pkg/config/data/config_schema_v1.0.json
```diff
@@ -85,6 +85,34 @@
...
+                          "type": "string"
+                        },
+                        "target": {
+                          "type": "string"
+                        }
+                      },
+                      "required": ["type", "id", "target"]
+                    }
+                  }
+                }
```
mattt, line 115: Setting `"command"` as a required key causes `gojsonschema` to emit a validation error, despite the key being present (at least I'm pretty sure it is). I ran this through an online validator and couldn't reproduce this problem.

pkg/config/config.go
```diff
@@ -19,16 +19,25 @@ import (
...
-	Run                []string `json:"run,omitempty" yaml:"run"`
-	SystemPackages     []string `json:"system_packages,omitempty" yaml:"system_packages"`
-	PreInstall         []string `json:"pre_install,omitempty" yaml:"pre_install"` // Deprecated, but included for backwards compatibility
-	CUDA               string   `json:"cuda,omitempty" yaml:"cuda"`
-	CuDNN              string   `json:"cudnn,omitempty" yaml:"cudnn"`
+	GPU                bool      `json:"gpu,omitempty" yaml:"gpu"`
+	PythonVersion      string    `json:"python_version,omitempty" yaml:"python_version"`
+	PythonRequirements string    `json:"python_requirements,omitempty" yaml:"python_requirements"`
+	PythonPackages     []string  `json:"python_packages,omitempty" yaml:"python_packages"` // Deprecated, but included for backwards compatibility
+	Run                []RunItem `json:"run,omitempty" yaml:"run"`
```
evilstreak, line 36: Does this require an update to the [run docs](https://github.com/replicate/cog/blob/main/docs/yaml.md#run) given the change in syntax? Do we want to continue to support the old syntax as well?

mattt: Good callout. I just added docs with 080e0fae4760f52b17649d0678abc1a305d4d4aa.

This new change is additive, so existing uses of `run` commands will continuing working. I don't have any plans to deprecate the old syntax.

PR #1013, opened 2023-04-26
Add `--secret` flag to `cog build` and `cog push` commands

### LINKED ISSUE #1009: Build secrets

mattt:
Resolves #1009

### DIFF
--- a/docs/private-package-registry.md
+++ b/docs/private-package-registry.md
@@ -34,3 +34,6 @@ $ cog build cog build --secret id=pip,source=/path/to/pip.conf
 ```
 
 Using a secret mount allows the private registry credentials to be securely passed to the `pip install` setup command, without baking them into the Docker image.
+
+> **Warning**
+> If you run `cog build` or `cog push` and then change the contents of a secret source file, the cached version of the file will be used on subsequent builds, ignoring any changes you made. To update the contents of the target secret file, either change the `id` value in `cog.yaml` and the `--secret` option, or pass the `--no-cache` option to bypass the cache entirely.
--- a/docs/private-package-registry.md
+++ b/docs/private-package-registry.md
@@ -4,7 +4,7 @@ This guide describes how to build a Docker image with Cog that fetches Python pa
 
 ## `pip.conf`
 
-Create a `pip.conf` file with an `index-url` set to the registry's URL with embedded credentials.
+In a directory outside your Cog project, create a `pip.conf` file with an `index-url` set to the registry's URL with embedded credentials.
 
 ```conf
 [global]
--- a/docs/private-package-registry.md
+++ b/docs/private-package-registry.md
@@ -11,6 +11,9 @@ In a directory outside your Cog project, create a `pip.conf` file with an `index
 index-url = https://username:password@my-private-registry.com
 ```
 
+> **Warning**
+> Be careful not to commit secrets in Git or include them in Docker images. If your Cog project contains any sensitive files, make sure they're listed in `.gitignore` and `.dockerignore`.
+
 ## `cog.yaml`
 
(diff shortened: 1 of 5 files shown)


anotherjesse:
@mattt the code lgtm

I think we should add to the docs.yaml how to use it:

```
cog build --secret id=foo,source=/path/to/foo
```

I want to add a note about how caching secrets works since it is different from files, but I'm not sure the best recommendation for telling users how to do so 

`docker rmi cog-{directory-name}` or `cog rmi r8.im/username/model_name` only deletes the top layer, but the underlying cached layers still remain, and the build still has the old secret.

From googling/chatgpting it seems like best practice is to use ARG before your RUN

```
ARG SECRET_VERSION
RUN --mount-type=secret ....
```

The other way is `--no-cache`, which as Ben mentioned we would need to add support for.

I'm only hesitant to ship without documentation/process for breaking cache as I ran into the issue within minutes of using it.

mattt:
@anotherjesse Agreed on the point about docs. I'll add that now.

I've opened PRs to this PR branch implementing both approaches you described:

- https://github.com/replicate/cog/pull/1014
- https://github.com/replicate/cog/pull/1015

docs/private-package-registry.md
```diff
@@ -0,0 +1,39 @@
+# Private package registry
+
+This guide describes how to build a Docker image with Cog that fetches Python packages from a private registry during setup.
+
+## `pip.conf`
+
+Create a `pip.conf` file with an `index-url` set to the registry's URL with embedded credentials.
```
anotherjesse, line 7: Great overview!

Question: Should we advise to either have a .dockerignore or (preferably?) keep this file outside of the cog project directory - as all content gets copied in your `cog build` step?

(Many times when I review customers' images they have lots of extra stuff that the they don't realize is copied in automatically.)

docs/private-package-registry.md
```diff
@@ -0,0 +1,39 @@
...
+          id: pip
+          target: /etc/pip.conf
+```
+
+## Build
+
+When building or pushing your model with Cog, pass the `--secret` option with an `id` matching the one specified in `cog.yaml`, along with a path to your local `pip.conf` file.
+
+```console
+$ cog build cog build --secret id=pip,source=/path/to/pip.conf
```
anotherjesse, line 33: extra `cog build` here

docs/private-package-registry.md
mattt: Great idea. I just made a few additions that call this out explicitly.

anotherjesse:
@mattt this works great.  I tested by using github private packages

https://docs.readthedocs.io/en/stable/guides/private-python-packages.html#from-a-git-repository

thanks for pushing it over the line.

bfirsh (approved):
Fantastic. Haven't done a detailed code review but sounds good in principle. This maps to the same `--secret` argument in `docker build`, doesn't it?

mattt:
> Fantastic. Haven't done a detailed code review but sounds good in principle. This maps to the same `--secret` argument in `docker build`, doesn't it?

Correct. Here's where we're passing them as args to that command: https://github.com/replicate/cog/pull/1013/files#diff-7579bff2107ab04f17023ce17e83ca76171c266529245b9a263fcf8eba2a5c65R23-R25.

Issue #1006, opened 2023-04-21
[enhancement] COG.YAML support trusted host and extra index for NVIDIA


djj0s3:
Trying to leverage the new TensorRT Diffusers support feature, it requires using NVidia's hosted package manager. My cog.yaml looks like this gross mess:
```
build:
  gpu: true
  cuda: "11.7"
  python_version: "3.9.16"
  system_packages:
    - "python3-pip"
  python_packages:
    - "diffusers==0.15.1"
    - "torch==2.0.0"
    - "ftfy==6.1.1"
    - "scipy==1.9.3"
    - "transformers==4.25.1"
    - "accelerate==0.15.0"
    - "huggingface-hub==0.13.2"
    - "nvidia-pyindex==1.0.9"
    - "tensorrt==8.6.0"
  run: 
    - pip3 install polygraphy --extra-index-url=https://pypi.ngc.nvidia.com --trusted-host pypi.ngc.nvidia.com
    - pip3 install onnx --extra-index-url=https://pypi.ngc.nvidia.com --trusted-host pypi.ngc.nvidia.com
    - pip3 install onnx_graphsurgeon --extra-index-url=https://pypi.ngc.nvidia.com --trusted-host pypi.ngc.nvidia.com
```
Would be great to specify in python_packages to use --extra index and --trusted-host in addition to pypi.

mattt:
Hi @djj0s3. I agree, that's a mess. Fortunately, Cog lets you define package dependencies in a `requirements.txt` file like this:

```yaml
build:
  gpu: true
  cuda: "11.7"
  python_version: "3.9.16"
  system_packages:
    - "python3-pip"
  python_requirements: requirements.txt
```

Then, in a sibling `requirements.txt` file:

```pip
diffusers==0.15.1
torch==2.0.0
ftfy==6.1.1
scipy==1.9.3
transformers==4.25.1
accelerate==0.15.0
huggingface-hub==0.13.2
nvidia-pyindex==1.0.9
tensorrt==8.6.0

--extra-index-url=https://pypi.ngc.nvidia.com --trusted-host pypi.ngc.nvidia.com
polygraphy
onnx
onnx_graphsurgeon
```

Unfortunately, this doesn't appear to be documented anywhere. So I opened https://github.com/replicate/cog/pull/1016 to fix that.

PR #1016, opened 2023-04-27
Document `python_requirements` field in `cog.yaml`

mattt:
Related to #1006

Most Cog examples use `python_packages`, but the `python_requirements` option is convenient when you already have a `requirements.txt` file or need to configure options like `--extra-index-url` or `--trusted-host`. 

This PR documents the `python_requirements` field in `yaml.md`.

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -58,6 +58,17 @@ build:
     - tensorflow==2.5.0
 ```
 
+### `python_requirements`
+
+A pip requirements file specifying the Python packages to install. For example:
+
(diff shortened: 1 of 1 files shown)


anotherjesse:
lgtm!

bfirsh:
This is intentionally not documented, because it doesn't behave quite as you expect. I forget the exact context.

The intention is to clean up `python_requirements` and deprecate `python_packages` in favor of it.

Some background is in https://github.com/replicate/cog/issues/157

Issue #308, opened 2021-10-22
Avoid pushing large files on `cog push`


andreasjansson:
Since code and weights are stored in the same Docker layer, making small changes can result in really long push times when model weights are in the hundreds of megabytes or gigabytes.

One potential solution could be a `build.model_files` option in cog.yaml, that saves those files in a separate layer before the code is added.

andreasjansson:
Build times are also affected by large weights files. Any change in the code or dependencies requires all files in the working directory (including giant ones) to be re-added to the image.

Perhaps we should automatically add any files >1GB to the image before installing system dependencies and Python packages? That should make both builds and pushes faster since the layer containing the large weights file will be cached both locally and remotely.

zeke:
Semi-related: https://github.com/replicate/cog/issues/209

afiaka87:
> 

I think specifying the model files in the configuration works well and could maybe even allow for some other functionality such as a "fork" of a project that only changes its `build.model_files` but uses the same Predictor.

I also agree that a simple fix in the meantime would be to just detect the bigger files and add those automatically. Could also do lookup via suffix (e.g. `.pt`,`.plk`, `.pth`) or check the date modified attribute on the file.

JosephCatrambone:
I like the idea of having a pre/post build stage for model copying (or just detecting ckpt/etc, like above).  Specifying `build.model_files` also seems fairly ergonomic and friendly for new people.

In terms of implementation, Docker itself doesn't deal very well with symlinks, otherwise we could just make and break symlinks when we needed to build the images.  (I tried to do `ln -s ~/.cache/huggingface ./models`, but it seems to be ignored.)  

Thoughts on configuration vs convention?  Picking a directory name like 'models' and using some of the built in Dockerfile functionality, like `COPY [^models]/*` seems like a low-effort solution, but also might break in the cases where people have their own implementations with a 'models' directory.

bfirsh:
A potential quick fix discussed with @anotherjesse IRL last week: when generating the Dockerfile, find all the files larger than, say, 100MB and add those in a layer before `COPY . /src`.

That'll mean the last layer is small and will be the only thing that changes if you change source code.

asadm:
Forgive for bumping this but is there a plan for this?

I assume this will also reduce some burden on replicate infra too.

kirillrogovoy:
Similar case here, and so a similar problem.

[...]

PR #1041, opened 2023-05-23
feat: prevent small code changes from copying entire model weights

hongchaodeng:
fix https://github.com/replicate/cog/issues/308

## Description
In the current implementation of cog, both the code and model weights are stored in the same Docker layer during the docker build process. Consequently, making minor changes to the code can lead to significant build and push times, especially when the model weights are >10GB in size.

## Proposed Solution

To address the issue of long build and push times when making small code changes due to large model weights, we propose separating the model weights into a separate image layer. This will allow us to reuse the "model weights image layer" across builds, reducing build times and network usage.

To identify large model weight files, we will use the following heuristics:

- Files with a size of larger than 10MB
- Files not located in the directories (e.g. ".cog", ".git", "pycache")
- Files with extensions not typically used for model weights (e.g.  ".py", ".ipynb", ".jpg", ".jpeg", ".png", ".webp", ".svg", ".gif", ".mp4", ".log")

Once we have identified the large model weight files, the image build process will be modified as follows:

- Build a temporary “model_weights_image” to include large model weights files
- Create or update the ".dockerignore" file to exclude the paths of the large model weight files
- Build the final runner image with the following modifications:
  - Copy the model weights to "/src" from the "model_weights_image" before installing dependencies
  - Copy the entire codebase, excluding the model weights as they are now in ".dockerignore"

The following diagram illustrates the workflow:
cog

### Setup

To show the performance improvement, we do a benchmark to compare the current `cog` cli and the `cog2` cli built with this change. We use the [cog-stable-diffusion repo](https://github.com/replicate/cog-stable-diffusion) as the test case.

To simulate user behavior, we will first do a `cog build`, then we make small changes, and then do `cog build` again.

The benchmark script goes as:

```bash
docker system prune
cog build
echo $(date) > small_change.txt
time cog build

docker system prune
cog2 build
echo $(date) > small_change.txt
time cog2 build
```

The model weights is 11GB:

```bash
$ du -sh diffusers-cache/
 11G	diffusers-cache/
```

The benchmark is run on Macbook-16 M1 with 16GB memory.

### Result

`cog build` (current release) after small change:

Snipaste_2023-05-23_13-54-06

`cog2 build`(with this change) rebuilds fast after small change:

Snipaste_2023-05-23_18-57-46

Key observation: 

- Overall, cog takes **101s** and cog2 takes **50s** to rebuild the image.
- The cog (current release) takes  **27s** to in `Copy . /src` only for a small change.
  - Note that this is relatively fast on my macbook. I have experienced >100s on Azure VM.
- The cog2 takes **0s** in `COPY . /src` because it doesn’t need to copy the model weights.

Double checked that the models in the runner image built by `cog2`:

```bash
$ docker run --rm -it cog-cog-stable-diffusion bash
root@7744fc63a8dc:/src# du -sh diffusers-cache/models--CompVis--stable-diffusion-safety-checker/blobs
1.2G	diffusers-cache/models--CompVis--stable-diffusion-safety-checker/blobs
```

Double checked docker image layers via `docker history <image>`:
Snipaste_2023-05-23_00-25-10

The layers show that:
- The model weights layer is under `pip install` and `COPY . /src`. This means small code changes or dependency update won't need to copy large model weights again.
- The `COPY . /src` layer is small which indicates that code changes should be fast to rebuild.

## FAQ

Q: Does `cog push` still work? It looks like it generates two images now.
A: Yes it works. The model_weights_image is temporary. After we copy the model weights into the runner image, then it is the only thing that we need to push.

Q: How much overhead does building model_weights_image incur?
A: Building model_weights_image only add small overhead (1s) given the buildkit caching.


hongchaodeng:
@mattt Rebased. Please rerun the test

mattt:
@hongchaodeng Thank you! Approved and running now.

hongchaodeng:
@mattt Please help re-approve it.

Tested it locally and passed:
Snipaste_2023-06-09_12-41-50

mattt:
@hongchaodeng Sure thing! I'm going through this PR now. Hoping to have some review comments later today.

pkg/modelfinder/modelfinder.go
```diff
@@ -0,0 +1,169 @@
...
+	FindModels() ([]string, []string, error)
+}
+
+type FileWalker func(root string, walkFn filepath.WalkFunc) error
+
+type modelFinder struct {
+	fw FileWalker
+}
+
+func New() ModelFinder {
```
mattt, line 32: What do you think about rewriting this as a top-level function instead of defining a new type? Any reason `FindModels` couldn't take a `FileWalker` argument directly?

pkg/dockerfile/generator.go
```diff
@@ -11,11 +11,35 @@ import (
...
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.log
+.git
+.mypy_cache
+.pytest_cache
+.hypothesis
+`
```
mattt, line 41: It'd be helpful to annotate what each of these are:

```suggestion
const DockerignoreHeader = `
# generated by replicate/cog

# Python cache files
__pycache__
*.pyc
*.pyo
*.pyd
.Python

# Virtual environments
env

# Log files
pip-log.txt

# Files generated by pip
pip-delete-this-directory.txt

# Test environments for Python
.tox

# Code coverage tools
.coverage
.coverage.*
.cache
nosetests.xml
coverage.xml
*.cover

# Log files
*.log

# Git directory
.git

# Cache for type check tool
.mypy_cache

# Pytest cache directory
.pytest_cache

# Hypothesis (property-based testing tool) directory
.hypothesis
`
```

mattt: Another version that groups entries in a slice (requires additional changes to generate `.dockerignore` file):

```suggestion
var DockerignoreEntries = []string{
	"__pycache__, *.pyc, *.pyo, *.pyd, .Python",  // Python cache files and debug info
	"env",  // Virtual environments
	"pip-log.txt, pip-delete-this-directory.txt",  // Pip log files and generated files
	".tox",  // Test environments for Python
	".coverage, .coverage.*, .cache, nosetests.xml, coverage.xml, *.cover",  // Code coverage tools
	"*.log",  // Log files
	".git",  // Git directory
	".mypy_cache",  // Cache for type check tool
	".pytest_cache",  // Pytest cache directory
	".hypothesis",  // Hypothesis (property-based testing tool) directory
}
```

pkg/modelfinder/modelfinder.go
```diff
@@ -0,0 +1,169 @@
...
+	return &modelFinder{
+		fw: filepath.Walk,
+	}
+}
+
+func (mf *modelFinder) FindModels() ([]string, []string, error) {
```
mattt, line 38: This method is called `FindModels`, but the scratch image we write to is called `modelweights`. Is there a single term that accurately describes what we're looking for?

mattt (commented):
Would it make sense to allow users to override this behavior? If so, then as a starting point maybe we could add a CLI flag to opt-out entirely? (And then as a follow-up, we could add `cog.yaml` configuration options to include / exclude specific files)

mattt:
@hongchaodeng Thanks again for your excellent work on this! I went ahead and resolved the feedback I had on the PR. The only bit I wasn't able to finish today was putting this functionality behind an out-out CLI flag. 

In the interest of getting this into everyone's hands to try out, I'm merging this in as-is, and will cut a new beta release. We can then work to add a CLI flag in a follow-up.

PR #1054, opened 2023-05-28
doc: add best practices to building images

hongchaodeng:
This relies on https://github.com/replicate/cog/pull/1041. We need to merge that first.
I separate the doc change out because every git push takes a long cycle on CI testing. I would like to decouple them since this change doesn't need any CI.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -217,8 +217,7 @@ cog push
 The Docker image is now accessible to anyone or any system that has access to this Docker registry.
 
 > **Note**
-> When you build an image, Cog copies large files to a separate layer before copying source code. This reduces the time needed to rebuild after making changes to code. For best results, put any weights, checkpoints, or other model files in a separate subdirectory.
-> If you put both weights and code in the same directory, Cog will NOT copy the weights to a distinct layer, and you will have to wait for the weights to be copied every time you make a change to the code.
+> Model repos often contain large data files, like weights and checkpoints. If you put these files in their own subdirectory, Cog can copy these files into a separate Docker layer, which reduces the time needed to rebuild after making changes to code.
 >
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -205,6 +216,32 @@ cog push
 
 The Docker image is now accessible to anyone or any system that has access to this Docker registry.
 
+> **Note**
+> When you build an image, Cog copies large files to a separate layer before copying source code. This reduces the time needed to rebuild after making changes to code. For best results, put any weights, checkpoints, or other model files in a separate subdirectory.
```
zeke, line 220: I love this documentation, but I'm missing something: What is the effect of putting weights and other files in a subdirectory? I think that should be clarified here.

hongchaodeng:
Thanks @mattt for your edits. That looks great!

------

@zeke That's an excellent question! Thank you for your review and for bringing this up. I really appreciate it. I have now addressed your concern by adding a description for it.

To clarify, if the other file is a type of code, the weights directory will not be separated into a distinct layer. However, if it is not code, the directory will be included as is.

docs/getting-started.md
mattt: Ah, I think I understand your feedback. If it's the case that Cog copies large files before code, then there shouldn't be a problem with them co-mingling with code. The reason why this matters is an implementation detail of how Docker `COPY` commands work — that is.

Rather than wade into that, I rewrote the callout to be explicit about the contract. "If you do this, then Cog can do this optimization".

mattt:
@zeke Thanks for your feedback. Merging this in now to lock in our gains. Happy to revisit this in a follow-up PR.

zeke:
Nice!

PR #1056, opened 2023-06-03
WIP: generating input and output types without loading model.

daanelson:
Need to do quite a bit more testing and still implement for output, but the initial idea seems to work.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+import ast
 from collections.abc import Iterator
 import enum
 import importlib.util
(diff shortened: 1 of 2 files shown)


nickstenning:
I think there might be an easier and more robust way to do this. We can use the schema annotation (I think we add one?) on the docker image rather than loading the predictor file.

mattt:
This is an interesting and clever approach, but I had the same reaction as @nickstenning. We can probably get better performance with less effort and risk by determining types once at build time and caching the results.

daanelson:
@nickstenning that's a good point, parsing and cacheing the schema at build time is safer than just parsing strings w/`ast`. Will close this out for now and work on something that looks like that instead.

PR #1059, opened 2023-06-05
Generate requirements files from `pyproject.toml` using pip-tools

mattt:
Depends on #1058

This PR updates `pyproject.toml` to declare dev dependencies, and then generates `requirements.txt` and `requirements-dev.txt` files using [pip-compile](https://github.com/jazzband/pip-tools).


nickstenning:
I'm not sure I quite understand this. When users install cog via `pip install cog`, these files won't be used. To me it seems like pinning in our own GHA runners makes it _more_ likely that we release a version of Cog that's broken, because if a point release in one of our deps breaks us, we won't see that until it's too late.

mattt:
`pyproject.toml` declares the package dependencies for Cog the library. Like other libraries, these are defined with minimal constraints to maximize compatibility with other dependencies used by an application.

The `requirements.txt` and `requirements-dev.txt` files are for Cog in the sense of being an application, which in our case means development and testing. These files are used by dev tooling to provide type hinting and by CI to produce reliable test behavior over time.

The problem you're describing where a user installs `pip install cog` and resolving to an untested combination of dependency versions is real, and is something I'm eager for us to address.

One solution is to try to keep on top of version updates, to ensure that the latest versions of all dependencies are known to work in combination with one another. Dependabot has helped to that end, but that's not a complete solution.

Another solution that we could explore is to expand our CI to test more combinations of things. For example, we could have a workflow that runs on a daily schedule to smoke test `pip install cog` without any pinned dependencies.

What I'd really like is a way to separate out the runtime (application) dependencies of the HTTP server, like FastAPI and uvicorn, from the type (library) dependencies, like Pydantic. You previously discussed this in #999, and I recently opened https://github.com/replicate/cog/issues/1142. 

I also like your idea of taking a first step to separating out the HTTP stuff into an extra requirement (e.g. `pip install cog[serve]`).

Let's talk more about how to solve this problem, and start working towards a solution.

nickstenning:
I don't disagree with anything you've written, but to put in public what we've already discussed in private: until we have a solution for the difference between production and our test suite, I don't think it's wise for us to pin.

We had a bug in production introduced by `pydantic>=0.99` (our `pyproject.toml` specified `<1`) which wasn't picked up by the tests _solely_ because we were pinning these deps.

PR #1061, opened 2023-06-05
Fix various mypy type checking errors

mattt:
Based on #1060 

This PR resolves the mypy type checking errors in `python/cog`. 

The only remaining issues have to do with the `weights` argument passed to setup and the dynamically defined `PredictionRequest` / `PredictionResponse` types:

```console
$ python -m mypy python/cog
python/cog/predictor.py:95: error: Argument "weights" to "setup" of "BasePredictor" has incompatible type "Union[IOBase, pathlib.Path, None]"; expected "Union[File, cog.types.Path, None]"  [arg-type]
python/cog/server/http.py:141: error: Variable "PredictionRequest" is not valid as a type  [valid-type]
python/cog/server/http.py:141: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/cog/server/http.py:147: error: PredictionRequest? has no attribute "id"  [attr-defined]
python/cog/server/http.py:161: error: PredictionRequest? has no attribute "id"  [attr-defined]
python/cog/server/http.py:169: error: Variable "PredictionRequest" is not valid as a type  [valid-type]
python/cog/server/http.py:169: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
python/cog/server/http.py:178: error: PredictionRequest? has no attribute "input"  [attr-defined]
python/cog/server/http.py:179: error: PredictionRequest? has no attribute "input"  [attr-defined]
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ dynamic = ["version"]
 write_to = "python/cog/_version.py"
 
 [tool.mypy]
+plugins = "pydantic.mypy"
 disallow_untyped_defs = true
 # TODO: remove this and bring the codebase inline with the current mypy default
 no_implicit_optional = false
--- a/python/cog/logging.py
+++ b/python/cog/logging.py
@@ -60,9 +60,9 @@ def setup_logging(*, log_level: int = logging.NOTSET) -> None:
     )
 
     if development_logs:
-        log_renderer = structlog.dev.ConsoleRenderer(event_key="message")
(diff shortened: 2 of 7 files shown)


mattt:
Punting on fixing the remaining mypy issues. Hoping that an alternate solution to https://github.com/replicate/cog/pull/1056 will help resolve this as well.

PR #1069, opened 2023-06-07
Automatically re-run command when GPU isn't available

### LINKED ISSUE #590: Make `cog run` and `cog predict` work on GPU images, even if you don't have a GPU
This works for `cog build` but not `cog run` and `cog predict`. There's a little thing that checks for the error returned, and re-runs it without `--gpus`.

Requested by @charlesfrye.

mattt:
Fixes #590

I was able to reproduce the described behavior with the following example project:

```yaml
# cog.yaml
build:
  python_version: "3.8"
  gpu: true
  python_packages:
    - torch
    - numpy
```

```python
# predict.py
from cog import BasePredictor
import torch

    def predict(self, size: int) -> int:
        n = torch.randn([size])

        if torch.cuda.is_available():
            n = n.cuda()

        return n.sum().item()
```

With the latest version of Cog, `cog predict` fails on my local machine with the following error:

```console
$ cog predict -i size=100
Starting Docker image cog-gpu-project-base and running setup()...
docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].
ⅹ Failed to start container: exit status 125
```

This PR updates the `cog run` and `cog predict` subcommands to automatically re-run themselves in the event of this error, bringing them in line with `cog build` which already has this behavior.

### DIFF
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
@@ -124,7 +125,7 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 	if err := predictor.Start(&stderr); err != nil {
 		console.Debug(stderr.String())
 
-		if gpus != "" && strings.Contains(err.Error(), docker.ErrMissingDeviceDriver.Error()) {
+		if gpus != "" && errors.Is(err, docker.ErrMissingDeviceDriver) {
 			console.Debug("Missing device driver, re-trying without GPU")
 
 			_ = predictor.Stop()
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -126,7 +126,7 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 		console.Debug(stderr.String())
 
 		if gpus != "" && errors.Is(err, docker.ErrMissingDeviceDriver) {
-			console.Debug("Missing device driver, re-trying without GPU")
+			console.Info("Missing device driver, re-trying without GPU")
 
 			_ = predictor.Stop()
 			predictor = predict.NewPredictor(docker.RunOptions{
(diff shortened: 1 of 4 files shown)


zeke:
I took this for a spin locally on my M1 Mac but ran into this error:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].
ⅹ Failed to start container: exit status 125
```

zeke:
Hey it worked on the non-GPU codespace!

```
Starting Docker image cog-cog-run-without-gpu-base and running setup()...
Error response from daemon: page not found
Running prediction...
-7
```

Not sure what that `page not found` is about though.

mattt:
> I took this for a spin locally on my M1 Mac but ran into this (unrelated) error:

This looks like the original error. Something I ran into during development was that I wasn't actually calling the new version of Cog. In my case, `cog` resolved to the Homebrew installation. Try doing a `which cog` to make sure it's the right one, and do `[sudo] make install` as necessary.

zeke:
I'm pretty sure I was calling `~/go/bin/cog` explicitly, but maybe I forgot that time. I can try it again.

pkg/cli/predict.go
```diff
@@ -120,8 +120,25 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	if err := predictor.Start(os.Stderr); err != nil {
-		return err
+	var stderr bytes.Buffer
+	if err := predictor.Start(&stderr); err != nil {
+		console.Debug(stderr.String())
+
+		if gpus != "" && strings.Contains(err.Error(), docker.ErrMissingDeviceDriver.Error()) {
```
hongchaodeng, line 127: Use `errors.Is/As` for more accurate checking?

hongchaodeng (commented):
Nice PR. Add some minor suggestion and hope that would be helpful :)

zeke (approved):
Let's get this out the door!

pkg/cli/predict.go
mattt: Great idea!

ramesh144561:
hh

nasimkhan1:
A

PR #709, opened 2022-08-25
Report the reason for the Internal Server Error, if an Exception is thrown by predict()

cmdr2:
Hi,

This is a minor change for reporting the reason for an Internal Server Error, if an Exception is thrown by predict(). This is really helpful for REST API users to understand what failed, especially when using models created by others.

make test-python passed successfully.

Also tested this works correctly with a local docker container running the Stable Diffusion model from replicate, by triggering a scenario where it's supposed to throw an Exception.

I tested by modifying the docker container with this PR, created a docker image and container from that, and tested that it no longer threw an opaque "Internal Server Error" message, but instead described the error correctly.

Thanks

Signed-off-by: cmdr2 <secondary.cmdr2@gmail.com>

PR #1071, opened 2023-06-07
Provide details in Internal Server Error responses for predict endpoint

mattt:
Reimplementation of https://github.com/replicate/cog/pull/709

Things have moved around a bit in `http.py` since the original PR was opened. Notably, the predict function is now delegated to a helper that's shared by a `POST` and idempotent `PUT` endpoint. That `raise HTTPException` call now has exception chaining that includes details in the traceback.

This PR reapplies https://github.com/replicate/cog/pull/709/commits/893b28bf7b5c4a57f3a86727fd725421c351d22e to include the original error message in the raised exception as well as chaining to that error.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -195,7 +195,7 @@ def create_app(
             response = PredictionResponse(**async_result.get().dict())
         except ValidationError as e:
             _log_invalid_output(e)
-            raise HTTPException(status_code=500) from e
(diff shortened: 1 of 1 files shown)


mattt:
Hmm, GitHub isn't picking up the `Co-authored-by` as expected. Trying again by rebasing and cherry-picking.

PR #1095, opened 2023-06-18
feat: opt-out flag to disable weights optimization

hongchaodeng:
This is a follow-up for the [comment](https://github.com/replicate/cog/pull/1041#issuecomment-1591969088) from previous weights optimization PR.

With this change, the CLI flag will be provided to users with flexibility to opt out the weights speedup:
Snipaste_2023-06-18_16-05-06

Here's what the dockerfile will look like if the CLI flag is used (which is what previous version looks like):
Snipaste_2023-06-18_16-05-33


hongchaodeng:
Please help review it!
cc @mattt

mattt (approved):
Awesome. Thanks, @hongchaodeng! Merging this now.

Issue #1096, opened 2023-06-18
Build fails when using `cog version 0.8.0-beta4`


andreasjansson:
<img width="1414" alt="image" src="https://github.com/replicate/cog/assets/713993/b67717dd-99be-40f6-a6d2-acd698ad9b85">

For some reason it's trying to pull the intermediate weights image from Docker Hub?

cc @hongchaodeng @mattt

hongchaodeng:
Let me look into it.
Last time I tried it worked. There are new problems uncovered here. I will try to reproduce it first using https://github.com/andreasjansson/cog-musicgen-looper

hongchaodeng:
Fix submitted: https://github.com/replicate/cog/pull/1097
I tried it locally and it worked. Also included tests.
cc @andreasjansson

PR #1097, opened 2023-06-18
fix:  take into account user specified image name

hongchaodeng:
fix https://github.com/replicate/cog/issues/1096


pkg/cli/debug.go
```diff
@@ -25,17 +27,18 @@ func newDebugCommand() *cobra.Command {
 	}
 
 	cmd.AddCommand(debug)
+	cmd.Flags().StringVarP(&imageName, "image-name", "", "", "The image name to use for the generated Dockerfile")
 
 	return cmd
 }
 
 func cmdDockerfile(cmd *cobra.Command, args []string) error {
-	config, projectDir, err := config.GetConfig(projectDirFlag)
```
mattt, line 33: Huh, I guess we were shadowing this before. I wonder if this was being caught by linters and not being surfaced in an actionable way...

mattt (approved):
Great work, @hongchaodeng!

PR #1112, opened 2023-06-21
Update `cog init` template to use Python 3.11

mattt:
This PR updates the `cog.yaml` file template created when running `cog init` to use [Python 3.11](https://www.python.org/downloads/release/python-3110/). It's currently set to use Python 3.8.


mattt:
CI is set up to run Go and integration tests using Python 3.8. Does it make sense to continue to use the oldest supported version of Python for these? Or should we switch to the newest? Thoughts, @hongchaodeng @andreasjansson?

PR #1125, opened 2023-06-22
Separate integration test dependencies from dev dependencies

### LINKED ISSUE #1102: Extract integration test dependencies into separate requirements file
Cog uses NumPy for integration testing only. However, because it's listed as a dev dependency, we get Dependabot PRs like https://github.com/replicate/cog/pull/1100.

We should separate those out into a separate requirements file and configure Dependabot to ignore these. Or better yet, install those requirements on-demand from each integration test based on the dependencies declared in their respective `cog.yaml` file.

mattt:
Resolves #1102


mattt:
Closing in favor of https://github.com/replicate/cog/pull/1126

Issue #209, opened 2021-08-18
Document `.dockerignore`


bfirsh:
At the moment we don't mention that you can use .dockerignore files to exclude files from the Cog-built Docker image. We should document that feature.

(edited by @bfirsh and @andreasjansson)

bfirsh:
@charlesfrye mentioned this the other day.

afiaka87:
Would it be unwise to have `cog init` create a .dockerignore file for the user? It could contain popular editor folders, maybe some commented out examples like
```sh
# Avoid pushing the user's IDE settings.
.vscode/
.idea/

# Prevent accidental key push.
# id_rsa*
# credentials.txt
# .env
# source_credentials.sh

# Prevent .venv push to slim image, as cog uses a different interpreter anyways. 
# .venv/

# Prevent any test outputs created from being in the image.
# outputs/
# predictions/
```

bfirsh:
Great idea.

mattt:
Reopening to track writing something down in addition to adding it to `cog init` template.

PR #1136, opened 2023-06-26
Add `.dockerignore` file to `cog init` template

mattt:
Resolves https://github.com/replicate/cog/issues/209

### DIFF
--- a/pkg/cli/init-templates/.dockerignore
+++ b/pkg/cli/init-templates/.dockerignore
@@ -1,0 +1,5 @@
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+.DS_Store
+.git
+.github
+.gitignore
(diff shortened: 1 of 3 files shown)


pkg/cli/init-templates/.dockerignore
```diff
@@ -0,0 +1,5 @@
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+.DS_Store
+.git
+.github
+.gitignore
```
zeke, line 5: Are there some other good defaults we could add here?

cc @replicate/models

zeke (approved):
Looks good to me. 👍🏼 

It would be nice to get some other good defaults in the ignore file, if there are some no-brainer file types that fit the bill.

pkg/cli/init-templates/.dockerignore
mattt: @zeke I just updated this based on [other `.dockerignore` files in our org](https://github.com/search?q=org%3Areplicate+path%3A.dockerignore&type=code).

zeke:
Nice!

PR #1155, opened 2023-06-28
parse type signature without running container 

technillogue:
This will allow generating the OpenAPI schema at build time, by parsing the AST. 

It's tedious to wait for docker run and the torch imports inside the container every time you change a comma in predict.py. Having the option of parsing it statically makes the feedback loop shorter.


nickstenning:
I think there might be a slight misunderstanding here. I might be missing something, but I don't think generating the OpenAPI schema at build time by running the model is a problem?

The problem is loading the model at _runtime_ in order to derive its type signature so that we can validate inputs, as happens [here](https://github.com/replicate/cog/blob/5186e251e7d5be897d470166e608608587abbfab/python/cog/server/http.py#L64).

We could avoid this by ensuring that the generated OpenAPI schema is embedded into the image in a way that the Python code running inside the image can retrieve. The `Input` schema from said schema can then be used directly for input validation.

Apologies if the way I originally explained this was confusing. 🙃

anotherjesse:
I personally love this idea.  As @technillogue shared, waiting for cog to do a bunch of work is VERY tedious.

having something like this would be needed to support modifying existing cogs without having to rebuild everything

( that said, I could see saying "let's explore this in 'cog-lite' rather than core  ... )

mattt:
Echoing @nickstenning, I think there's some misunderstanding about what problem we're trying to solve here.

For the purpose of improving cold boot times, the important thing is how fast a schema gets loaded at runtime. Since an OpenAPI spec is generated at build time and set to an image label, our thinking was that we could save that file to a well-known location, load that at run time, and construct the model schema from that.

As I've mentioned before, I have concerns about the inherent limitations of ASTs; they can produce the right answer in some cases, but can fail if the `predict` function signature isn't purely declarative (it references a variable, contains a conditional, etc.). We could use the slower path as a fallback, but then we introduce a second source of truth.

In our conversations, @technillogue, you've described the poor UX of slow rebuilds due to regenerating the OpenAPI spec. Similar to @nickstenning, I'm still not convinced this is a problem we need to solve right now. But I do think there's a potential compromise that can reuse this work: 

Instead of using an AST to generate the spec, we could use it determine whether the `predict` method signature has changed between builds, and avoid regenerating the OpenAPI schema. @technillogue @nickstenning What do y'all think about an approach like that?

pkg/cli/build.go
```diff
@@ -76,6 +78,10 @@ func addSeparateWeightsFlag(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&buildSeparateWeights, "separate-weights", false, "Separate model weights from code in image layers")
 }
 
+func addSchemaFlag(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&buildSchemaFile, "openapi-schema", "", "Read OpenAPI schema from a file")
```
mattt, line 82:
```suggestion
	cmd.Flags().StringVar(&buildSchemaFile, "openapi-schema", "", "Load OpenAPI schema from a file")
```

python/tests/server/fixtures/openapi_output_type.py
```diff
@@ -4,7 +4,7 @@
 
 # An output object called `Output` needs to be special cased because pydantic tries to dedupe it with the internal `Output`
 class Output(BaseModel):
-    foo_number: int = "42"
+    foo_number: int = "42" # this can't be right... it's a string?
```
mattt, line 7:
```suggestion
    foo_number: int = "42"
```

mattt (approved):

python/tests/server/fixtures/openapi_output_type.py
technillogue: what was going on here?

python/tests/server/fixtures/openapi_output_type.py
```diff
@@ -4,7 +4,7 @@
 
 # An output object called `Output` needs to be special cased because pydantic tries to dedupe it with the internal `Output`
 class Output(BaseModel):
-    foo_number: int = "42" # this can't be right... it's a string?
```
mattt: FYI: In very JavaScript ["wat"](https://www.destroyallsoftware.com/talks/wat) fashion, Pydantic automatically casts "42" into the int 42

zeke:
@all-contributors please add @technillogue for code

PR #1156, opened 2023-06-28
don't install curl when building from python:3

technillogue:
`python:3` already includes curl. This lets users skip an unnecessary apt update, and makes the tini layer faster to pull. 

for this cog.yaml:
```yaml
build:
  gpu: false
  python_version: "3.10"
```

before: 4.82MB
after: 23.1KB

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -262,7 +262,9 @@ func (g *Generator) installTini() string {
 	// N.B. If you remove/change this, consider removing/changing the `has_init`
 	// image label applied in image/build.go.
 
+    // python:3 includes curl, so it does not need to be installed
 	install := `RUN set -eux; \`
+	// nvidia/cuda does not include curl, so it must be installed
 	if g.Config.Build.GPU {
 		install = `RUN --mount=type=cache,target=/var/cache/apt set -eux; \
 apt-get update -qq; \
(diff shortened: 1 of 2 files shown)


pkg/dockerfile/generator.go
```diff
@@ -261,12 +261,17 @@ func (g *Generator) installTini() string {
 	//
 	// N.B. If you remove/change this, consider removing/changing the `has_init`
 	// image label applied in image/build.go.
-	lines := []string{
-		`RUN --mount=type=cache,target=/var/cache/apt set -eux; \
+
+	install := `RUN set -eux; \`
+	if g.Config.Build.GPU {
```
mattt, line 266: It'd be nice to provide some context for why a GPU image needs to run this code. A code comment would help. Or what do you think about updating the `installTini` function to take an `installCurl` Boolean parameter, to make the intention clearer?

mattt:
Hi, @technillogue. Thanks for the PR! I left a comment about providing some more context around this change. I think you may also need to update a couple tests to remove the expected `RUN apt-get install` lines from the generated Dockerfiles. The one other thing is that you'll need to [sign your commit](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to pass the DCO check.

Once we have all that, I'd be happy to approve a CI run and get this merged 😃

pkg/dockerfile/generator.go
technillogue: is this comment enough?

mattt: That works, yeah. Thanks!

mattt:
Closing in favor of #1161

PR #1159, opened 2023-06-30
Install tini with apt

mattt:
Alternative to #1156 

The [tini README](https://github.com/krallin/tini#debian) provides installation instructions for installing with `apt`. I believe this should work for both the Python and CUDA base images.


mattt:
Closing since the point of #1156 was to avoid apt update in the first place.

PR #1160, opened 2023-06-30
Use `docker run --init` instead of installing tini

mattt:
The [tini README] includes the following note:

> **Note**
> If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just [pass the --init flag to docker run](https://docs.docker.com/engine/reference/commandline/run/).

This PR removes tini (see #844) in favor of using the `--init` option for Docker.

### DIFF
--- a/pkg/docker/run.go
+++ b/pkg/docker/run.go
@@ -52,6 +52,7 @@ func generateDockerArgs(options internalRunOptions) []string {
 	// Use verbose options for clarity
 	dockerArgs := []string{
 		"run",
+		"--init",
 		"--rm",
 		"--shm-size", "8G", // https://github.com/pytorch/pytorch/issues/2244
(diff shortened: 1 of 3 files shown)


mattt:
@nickstenning In #840, you wrote that "It would probably be sensible for the images that Cog builds to be suitable for production environments without further tweaking." If Docker ships with built in support for tini through the `--init` flag, do you think that's sufficient? Or does that fall under "further tweaking".

mattt:
Looking around, I see software like Deno moving away from `--init` to providing their own init handler. #1161 takes the approach I saw there, where tini is install using curl in a separate stage.

nickstenning:
Yup, #1161 or something like it is the way to do. We can't rely on `docker --init` for the simple reason that we don't use Docker in production.

PR #1161, opened 2023-06-30
Download `tini` in separate stage

mattt:
Alternative to #1156

This PR downloads `tini` using `curl` in a separate build stage, and then copies it over. This avoids running `apt update` or relying on the existence of `curl`.


mattt:
@technillogue What do you think about this approach?

technillogue:
nice, exactly what I was thinking of https://github.com/replicate/cog/compare/main...syl/no-base-curl

technillogue:
https://hub.docker.com/r/appropriate/curl/tags is slightly smaller

Issue #1162, opened 2023-07-01
All builds on Cuda 11.2.0 fail because Cog references a non-existing Docker image


andreemic:
Hey! Since yesterday I'm unable to build any image using Cog and therefore deploy any models to Replicate.

As far as I understand, an external Nvidia Docker image which is referenced [here](https://github.com/replicate/cog/blob/67f607c6c7b7d0b0916b7bc1834b3199b18ff307/pkg/config/cuda_base_image_tags.json#L61C37-L61C37) got [removed from Docker Hub](https://hub.docker.com/layers/nvidia/cuda/11.2.0-cudnn8-devel-ubuntu20.04/images/sha256-780f4c3f099464364ab2449fad3f5b68d46c5d11da2c6f48ec51e4b59aa5cde7?context=explore) and causes builds on certain CUDA versions to fail.

If so, this prevents all Lambda Labs users, and probably more people, from using Cog and deploying to Replicate.

I am not entirely confident in my explanation but maybe we can figure out together.

## The Error:
![image](https://github.com/replicate/cog/assets/27695722/2cd4830f-cdf0-40a3-bda3-28b2a39a3460)

andreemic:
@mattt

jabedbd:
facing the same issue today. can not push any model to the replicate due to this issue.

zba:
i add this version of cuda to build and it seems fixed it, at least temporary. 
```
build:
  cuda: "11.2.2"
```

zsxkib:
I was dealing with this too. It was weird because everything was working fine and then it randomly started happening. Upon further inspection, I think Nvidia have edited/removed this image.

It's strange because it even fails on the Cog Hello World example:
```sh
$ mkdir hello-world-cog
$ cd hello-world-cog/
```

```sh
$ cog init

Setting up the current directory for use with Cog...

✅ Created /home/sakib/hello-world-cog/cog.yaml
✅ Created /home/sakib/hello-world-cog/predict.py

Done! For next steps, check out the docs at https://cog.run/docs/getting-started
```

```sh
$ sed -i 's/gpu: false/gpu: true/g' cog.yaml
```

```sh
$ cat cog.yaml
# Configuration for Cog ⚙️
# Reference: https://github.com/replicate/cog/blob/main/docs/yaml.md

build:
  # set to true if your model requires a GPU
  gpu: true

  # a list of ubuntu apt packages to install
  # system_packages:
    # - "libgl1-mesa-glx"
[...]

PR #1163, opened 2023-07-03
Update CUDA base images

mattt:
Related to #1162

Nvidia recently unpublished some tagged images from Docker Hub, which is breaking cog build for some users.

This PR regenerates the list of available CUDA images to ensure that missing images like `11.2.1-cudnn8-devel-ubuntu20.04` aren't requested.


Makefile
```diff
@@ -28,6 +28,10 @@ pkg/dockerfile/embed/cog.whl: python/* python/cog/* python/cog/server/* python/c
 	mkdir -p pkg/dockerfile/embed
 	cp dist/*.whl $@
 
+.PHONY: compatibility-matrices
+compatibility-matrices:
+	$(GO) run tools/generate_compatibility_matrices/main.go
+
```
bfirsh, line 34: `make generate` does this I think?

bfirsh (approved):
Ah, nvidia, this has broken a bunch of times before.

I suppose this means CUDA 8-10 is not supported in Cog any longer? We should probably note that in the release notes.

zeke:
> Nvidia recently unpublished some tagged images from Docker Hub

Do we know why?

Makefile
mattt: Good callout. Rebased out of #1170

mattt:
Superseded by #1170

PR #1169, opened 2023-07-03
use newer CUDA versions in tests

zeke:
Addendum to https://github.com/replicate/cog/pull/1163 to try to fix the broken tests.


zeke:
Changed the base from `mattt/update-cuda-base-images` to `main` to get CI to run.

mattt:
Cherry-picked 38f3fc4 into #1170. Thanks!

PR #1170, opened 2023-07-03
Refactor compatibility generation

mattt:
Offshoot of #1163 

After spending some time investigating #1162, I believe that some of the failing tests are correctly signaling new holes in our compatibility matrices. So while we could update the data files and get tests passing again (https://github.com/replicate/cog/pull/1169), I think we'll need to filter out those missing CUDA images.

This PR attempts to do this. It takes on a bigger refactoring than was necessary, based on a mistaken assumption about how I'd need to update the TF and Torch generator scripts to take CUDA image tags as arguments. It looks like we can filter in `init` instead, but we may want to stick with the refactoring anyway.


bfirsh:
Fantastic. This is a big mess.

This reminded me of something rattling around in my brain: https://github.com/replicate/cog/issues/1171

Issue #1168, opened 2023-07-03
Add `--separate-weights` flags to `cog debug` and other subcommands


mattt:
Following up on #1151, we should add this flag for any other subcommands that have to do with building images, including `cog debug`.

(Thanks to @philandstuff for reporting)

hongchaodeng:
Let me handle it :)

philandstuff:
Thanks both ❤️

PR #1173, opened 2023-07-04
fix: debug cmd should have --separate-weights

hongchaodeng:
fix https://github.com/replicate/cog/issues/1168

Actually the flag has been implemented. The code doesn't look right to extra command.
cc @mattt


philandstuff (approved):
haha, in the current release it turns out `cog debug debug` is a (useless) command

mattt (approved):
Awesome. Thanks, @hongchaodeng!

PR #1181, opened 2023-07-06
Fix input filename handling issues

nickstenning:
We have a bunch of issues with input filename handling. The most important is that we set no limits on filename length, which results in errors when the filename exceeds Linux's 255 byte limit.

This commit does the following:

- enforces a shorter 200-byte filename limit through truncation
- correctly decodes URL-encoded filenames
- protects against illegal characters in filenames (nul byte, slash)

In addition, it replaces a couple of ad-hoc data URI parsers with use of `urllib.request.urlopen`, which has supported data URIs since Python 3.4.

### DIFF
--- a/python/cog/types.py
+++ b/python/cog/types.py
@@ -5,11 +5,18 @@ import os
 import pathlib
 import shutil
 import tempfile
+import urllib
 from typing import Any, Dict, Iterator, List, Optional, TypeVar, Union
-from urllib.parse import urlparse
 
 import requests
 from pydantic import Field
+
+FILENAME_ILLEGAL_CHARS = set("\u0000/")
+
+# Linux allows files up to 255 bytes long. We enforce a slightly shorter
+# filename so that there's room for prefixes added by
+# tempfile.NamedTemporaryFile, etc.
+FILENAME_MAX_LENGTH = 200
 
 
 def Input(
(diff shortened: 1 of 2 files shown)


python/cog/types.py
```diff
@@ -198,15 +203,31 @@ def __repr__(self) -> str:
...
+    # If the filename is too long, we truncate it (appending '~' to denote the
+    # truncation) while preserving the file extension.
+    # - truncate it
+    # - append a tilde
+    # - preserve the file extension
+    if _len_bytes(basename) > FILENAME_MAX_LENGTH:
+        basename = _truncate_filename_bytes(basename, length=FILENAME_MAX_LENGTH)
+
+    for c in FILENAME_ILLEGAL_CHARS:
+        basename = basename.replace(c, "_")
```
anotherjesse, line 228: no strong preference on this, but nicer to replace multiple illegal chars in a row with a single `_`?

debugging with multiple _ might be complicated to count - but will probably never actually happen - so ignore me?

    pattern = "[" + re.escape(''.join(FILENAME_ILLEGAL_CHARS)) + "]+"
    basename = re.sub(pattern, "_", basename)

python/cog/types.py
```diff
@@ -47,12 +54,10 @@ def validate(cls, value: Any) -> io.IOBase:
...
 
-        parsed_url = urlparse(value)
+        parsed_url = urllib.parse.urlparse(value)
         if parsed_url.scheme == "data":
-            # https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
-            # TODO: decode properly. this maybe? https://github.com/fcurella/python-datauri/
-            header, encoded = parsed_url.path.split(",", 1)
-            return io.BytesIO(base64.b64decode(encoded))
+            res = urllib.request.urlopen(value)
+            return io.BytesIO(res.read())
```
anotherjesse, line 60: interesting change here - is this for efficiency/correctness?

nickstenning: Correctness. The old code would work with base64-encoded data URIs but not arbitrary data URIs.

python/cog/types.py
nickstenning: We could do that, but I'm not sure there's a meaningful user experience improvement here (how many people have file names with multiple NULs or `/`s in a row?) and it would make the code a little harder to grok.

anotherjesse: agreed

python/cog/types.py
anotherjesse: I just read your initial PR comment - thx!

anotherjesse:
this LGTM

anotherjesse (approved):

mattt (approved):
Excellent work, @nickstenning. Thanks so much for your initiative on this.

PR #1182, opened 2023-07-07
Fixed BaseImage Issue for Nvidia Cuda not being found

TeenPhilosopher:
Fixed the format of the baseImage urls. Changed cudnn8-devel to base.

### DIFF
--- a/pkg/config/cuda_base_images.json
+++ b/pkg/config/cuda_base_images.json
@@ -1,237 +1,237 @@
 [
   {
-    "Tag": "12.1.1-cudnn8-devel-ubuntu22.04",
+    "Tag": "12.1.1-base-ubuntu22.04",
     "CUDA": "12.1.1",
     "CuDNN": "8",
     "IsDevel": true,
     "Ubuntu": "22.04"
   },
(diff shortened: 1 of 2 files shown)


mattt:
Hi, @TeenPhilosopher. Thanks for your interest in contributing to Cog. The issue with base images not being found (#1162) has to do several versions being untagged from Docker Hub, which I believe is the case for all variants, including `base`. And despite their "BaseImage" naming, we actually do want the `devel` images, which include header files and development tools that models rely on.

TeenPhilosopher:
Okay but the matter of fact is that it’s not working as is. I tried it with this version of the source code and it built correctly and the model ran… I want to push this model to replicate (and indeed I did but it wouldn’t accept it as valid because cog had been modified by me) but I can’t as it stands. Any thoughts?

mattt:
@TeenPhilosopher What specifically wasn't working for you? Could you share the commands you tried and their output?

There was a recent issue with CUDA base images (#1162) that's now fixed by v0.8.0. Have you tried updating to the latest version of Cog?

PR #1203, opened 2023-07-12
don't import numpy at the top of json.py

technillogue:
It takes a little while to import numpy and it's not necessary for the server process to do up front. 

Signed-off-by: technillogue <technillogue@gmail.com>

### DIFF
--- a/python/cog/json.py
+++ b/python/cog/json.py
@@ -8,13 +8,6 @@ from pydantic import BaseModel
 
 from .types import Path
 
-try:
-    import numpy as np  # type: ignore
-
-    has_numpy = True
-except ImportError:
-    has_numpy = False
-
 
 def make_encodeable(obj: Any) -> Any:
     """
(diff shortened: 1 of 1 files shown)


python/cog/json.py
```diff
@@ -24,6 +17,13 @@ def make_encodeable(obj: Any) -> Any:
 
     Somewhat based on FastAPI's jsonable_encoder().
     """
+    try:
+        import numpy as np  # type: ignore
```
mattt, line 21: How do Python imports work? Would subsequent calls to `make_encodable` have the same overhead? Or should we move `has_numpy` to top-level scope and conditionalize the import?

mattt: Also, would it make sense to move this to the bottom of the conditional, like replaying `if has_numpy` with an `else`?

mattt (approved):
Perfect. Great work!

Issue #1189, opened 2023-07-10
[Cog v0.8.0 error after upgrading from v0.7.2] Error on Cog build: exec: /sbin/ldconfig.real: not found


Glavin001:
Impact: I'm unable to build any image using Cog and therefore deploy any models to Replicate.

---

On both Lambdalabs and TensorDock:

```bash
sudo cog build
```

`cog.yaml`:

```yaml
# Configuration for Cog ⚙️
# Reference: https://github.com/replicate/cog/blob/main/docs/yaml.md

build:
  # set to true if your model requires a GPU
  gpu: true

  cuda: "11.8"

  # python version in the form '3.8' or '3.8.12'
  python_version: "3.10"

  # a list of packages in the format <package-name>==<version>
  python_packages:
    - "torch==2.0.0"
    - "transformers==4.30.1"
    - "sentencepiece==0.1.97"
    - "accelerate==0.20.3"
    # https://github.com/oobabooga/text-generation-webui/blob/main/docs/LLaMA-model.md#option-2-convert-the-weights-yourself
    - "protobuf==3.20.1"
    - "auto-gptq==0.2.2"
  
# predict.py defines how predictions are run on your model
predict: "predict.py:Predictor"
```

I receive the following error logs:

```
 => ERROR [stage-1  3/11] RUN --mount=type=cache,target=/var/cache/apt apt-get update -qq && apt-get install -qqy --no-install-recom  16.4s
------                                                                                                                                      
 > [stage-1  3/11] RUN --mount=type=cache,target=/var/cache/apt apt-get update -qq && apt-get install -qqy --no-install-recommends      make        build-essential         libssl-dev      zlib1g-dev      libbz2-dev      libreadline-dev         libsqlite3-dev  wget    curl    llvm        libncurses5-dev         libncursesw5-dev        xz-utils        tk-dev  libffi-dev      liblzma-dev     git     ca-certificates    && rm -rf /var/lib/apt/lists/*:                                                                                                              
#0 13.37 debconf: delaying package configuration, since apt-utils is not installed    

....

#0 20.67 Setting up tk-dev:amd64 (8.6.11+1build2) ...
#0 20.67 Processing triggers for libc-bin (2.35-0ubuntu3.1) ...
#0 20.67 /usr/sbin/ldconfig: 16: exec: /sbin/ldconfig.real: not found
#0 20.67 /usr/sbin/ldconfig: 16: exec: /sbin/ldconfig.real: not found
[...]

PR #1208, opened 2023-07-14
fix: tiny installation should include TINI_ARCH

hongchaodeng:
This will fix https://github.com/replicate/cog/issues/1189

TLDR: missing the TINI_ARCH causes the apt-get commands fail.

See the details in [this discussion ](https://github.com/replicate/cog/issues/1189#issuecomment-1635244643).


ichernev:
Today I managed to test this patch (got the same issue as described in #1189) and I can report it does fix it!

hongchaodeng:
This has been fixed.

PR #1214, opened 2023-07-18
Add `--use-cuda-base-image` flag

mattt:
Alternative to #1205


test-integration/test_integration/test_predict.py
```diff
@@ -58,7 +58,9 @@ def test_predict_writes_files_to_files(tmpdir_factory):
         capture_output=True,
     )
     assert result.stdout == b""
-    with open(out_dir / "output.bmp", "rb") as f:
+    # something about this PR makes the default output be "output" instead of "output.bmp"
+    with open(out_dir / "output", "rb") as f:
+    # with open(out_dir / "output.bmp", "rb") as f:
```
hongchaodeng, line 63: remove commented code?

hongchaodeng (approved):
lgtm after nit!

test-integration/test_integration/test_predict.py
technillogue: do you have any idea what might be causing this error?

technillogue: it was getting reported as x-ms-bmp instead of bmp in python:3.8-slim for some reason

mattt:
Nice work, @technillogue. LGTM!

Issue #1218, opened 2023-07-20
Feature Request: Ignore weights when cog build the second time


hongchaodeng:
## Problem

When we `cog build`, it takes a long time to transfer weights into docker image due to the sheer size of weights files.

## Solution

We can skip transferring the weights into docker context by:
- using a separate weights image (with `--separate-weights` flag)
- skip the step to rebuild the weights image
- add weights to .dockerignore to avoid re-transferring

PR #1223, opened 2023-07-23
feat: skip building the weights image to speed up

hongchaodeng:
Fix https://github.com/replicate/cog/issues/1218

This is a problem that some user complains about. A simple solution would be that:

1. When user builds the image the first time, they do:

```
cog build --separate-weights
```

~2. When user builds the image the second time or more, they can avoid the **transferring context** wait time by doing: `cog build --separate-weights --skip-build-weights`~ <ins>@mattt: This is now done automatically</ins>


mattt:
@hongchaodeng Thanks for working on this. Is there any way that we can get this behavior automatically? For example, could the weights layer provide a checksum or manifest that we could diff against to tell if we need to rebuild?

hongchaodeng:
@mattt Thanks for the suggestion! I think that's a great idea.

I think we can make a hash for (all the file system hierarchy + file hashes). We can store the previous hash under `.cog/` and then compare it to avoid rebuilding the layer.

> Note that checking on the checksum of weights layer might not be optimal because that would involve transferring context somehow.

However, I just check that tar-ing and md5ing the 11G weights of stable diffusion takes **~60s** on my M1 Macbook.

I think using a flag manually would still be optimal. But the automation can still be provided. WDYT?

mattt:
> think we can make a hash for (all the file system hierarchy + file hashes). We can store the previous hash under `.cog/` and then compare it to avoid rebuilding the layer.

How about computing a [CRC32 hash](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) of the files matched by the file walker?

hongchaodeng:
> How about computing a [CRC32 hash](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) of the files matched by the file walker?

Of course, that should work too.
I just tried and for 11G weights it took ~30s on my M1 Mac. This might not be a big deal. So maybe we can start making this an automated option first?

mattt:
@hongchaodeng Yes! Please push any work in progress you have for doing this an automatic option.

I'm confident we can get checksum calculations fast enough to not be noticeable. For example, [this blog post](https://dougallj.wordpress.com/2022/05/22/faster-crc32-on-the-apple-m1/) claims to have CRC32 processing ~70GB/s on an M1.

hongchaodeng:
Of course! Just pushed!

The blog post you shared looks very promising. I will dig deeper to see how to optimize it as much as possible.

hongchaodeng:
@mattt The latest implementation is really fast. It is like instant to calculate all file hashes (<1s).

mattt (approved):
@hongchaodeng Great work on this! I just pushed a couple changes to align the naming and refactor the implementation. If everything looks good to you, I think this is ready to merge 😃

PR #1229, opened 2023-07-25
Allow `cog build` and `cog push` to specify path to custom Dockerfile

technillogue:
For building tiny images and modifying existing images with less pain

This is also intended as an escape hatch for when cog's dockerfile generation is inadequate. You can use `cog debug` as a starting point and edit that.

If this is merged, we can revert stuff like #1289. This provides for edge cases where you need a dependency to compile packages that you don't want at runtime, and various other problems. 

This is also mentioned in https://github.com/replicate/cog/blob/main/pkg/config/config.go#L20


zeke:
Hi there. Thanks for sharing your work. 👍🏼 

If you think this is something that could be generally useful for others, would you be willing to write some docs explaining the rationale behind these changes and what they make possible?

technillogue:
The main usecase for this is developing smaller images, such as https://github.com/replicate/cog-stable-diffusion/blob/syl/pget/Dockerfile

It also provides for doing this:

```
FROM r8.im/oranzino/tar-multi@sha256:21d2d80ab9b2129c959540e5ea04a027046f63928c20b36a4f01e32ffc0ed3ea
COPY ./fixed_predict.py /src/predict.py
```

There are issues like https://github.com/replicate/customer/issues/86, where you want to try small tweaks such as sprinkling in some `.to("cuda")` calls but would rather not redo the entire build process. Using FROM <existing image> makes that easier. 

To use this, all you need to do is add a Dockerfile in the working directory and not used --separate-weights.

technillogue:
@mattt,  so we have a few options about configuring it 

1. leave it as-is: if there's a file named Dockerfile, use that, otherwise do nothing
2. a section in cog.yaml under build
3. yet another flag

For (3) we could do `cmd.Flags().Lookup("dockerfile").NoOptDefVal = "Dockerfile"` so that you could either pass `--dockerfile <path>` or just `--dockerfile`. 


Other reference material includes flyctl's `--dockerfile string                Path to a Dockerfile. Defaults to the Dockerfile in the working directory`, and their behavior of using dockerfiles if present but otherwise trying to guess a buildpak for a given app.

mattt:
@technillogue I think `--dockerfile <path>` is the best option. It also lets you point at multiple Dockerfiles, allowing for things like dev/prod variants, easier debugging between two versions, and monorepos with multiple models.

pkg/cli/build.go
```diff
@@ -85,3 +87,8 @@ func addSchemaFlag(cmd *cobra.Command) {
 func addUseCudaBaseImageFlag(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&buildUseCudaBaseImage, "use-cuda-base-image", "auto", "Use Nvidia CUDA base image, 'true' (default) or 'false' (use python base image). False results in a smaller image but may cause problems for non-torch projects")
 }
+
+func addDockerfileFlag(cmd *cobra.Command) {
+	cmd.Flags().StringVar(&buildDockerfilePath, "dockerfile", nil, "Path to a Dockerfile. If the flag is passed but no value is provided, defaults to the Dockerfile in the working directory.")
```
mattt, line 92:
```suggestion
	cmd.Flags().StringVar(&buildDockerfilePath, "dockerfile", "", "Path to a Dockerfile. If the flag is passed but no value is provided, defaults to the Dockerfile in the working directory.")
```

pkg/image/build.go
```diff
@@ -70,9 +70,18 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 			return fmt.Errorf("Failed to build runner Docker image: %w", err)
 		}
 	} else {
-		dockerfileContents, err := generator.GenerateDockerfileWithoutSeparateWeights()
-		if err != nil {
-			return fmt.Errorf("Failed to generate Dockerfile: %w", err)
+		// if Dockerfile exists, use that instead
+		var dockerfileContents string
+		maybeDockerfile := path.Join(dir, "Dockerfile")
+		contents, err := os.ReadFile(maybeDockerfile)
```
mattt, line 76: I think this needs to be updated to use `buildDockerfilePath`, right?

pkg/docker/build.go
```diff
@@ -23,6 +23,11 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 		args = append(args, "--platform", "linux/amd64", "--load")
 	}
 
+	buildArg := os.Getenv("BUILD_ARG")
+	if buildArg != "" {
+		args = append(args, fmt.Sprintf("--build-arg=%s", buildArg))
+	}
+
```
mattt, line 30: This looks unrelated. Can we split this out into a separate PR?

pkg/image/build.go
```diff
@@ -131,6 +141,11 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 	if err != nil {
 		return fmt.Errorf("Failed to convert config to JSON: %w", err)
 	}
+	// useful for tricking kubernetes/version_deployment.go into not reinstalling pip
+	fakeCogVersion := os.Getenv("FAKE_COG_VERSION")
+	if fakeCogVersion != "" {
+		global.Version = fakeCogVersion
+	}
```
mattt, line 148: This also looks unrelated. Do we need this here?

mattt:
```suggestion
```

pkg/docker/build.go
mattt:
```suggestion
```

mattt:
@technillogue Any idea why integration tests are failing with these changes?

PR #1239, opened 2023-07-31
Fix weight manifest change check

ichernev:
The weights are changed if there is no cache (no previous value) or the value differs.

Introduced in #1223

On a tangential note -- if cog crashes around the `Equal` function, the `.dockerignore` is in an inconsistent state, which leads to weights being ignored (because it thinks the `.dockerignore` cog created is the real one, next time you run it) ... ideally there should be a finally block in the right place to restore the original `.dockerignore` if any, even if the code crashes.

### DIFF
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -50,8 +50,8 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 		if err != nil {
 			return fmt.Errorf("Failed to generate weights manifest: %w", err)
 		}
-		cachedManifest, err := weights.LoadManifest(weightsManifestPath)
-		changed := err != nil && weightsManifest.Equal(cachedManifest)
+		cachedManifest, _ := weights.LoadManifest(weightsManifestPath)
+		changed := cachedManifest == nil || !weightsManifest.Equal(cachedManifest)
 		if changed {
 			if err := buildWeightsImage(dir, weightsDockerfile, imageName+"-weights", secrets, noCache, progressOutput); err != nil {
 				return fmt.Errorf("Failed to build model weights Docker image: %w", err)


ichernev:
cc @hongchaodeng

mattt:
Thanks for sharing this, @ichernev. I recognize that as a change I introduced to @hongchaodeng's PR. Taking a look now.

pkg/image/build.go
```diff
@@ -51,7 +51,7 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 			return fmt.Errorf("Failed to generate weights manifest: %w", err)
 		}
 		cachedManifest, err := weights.LoadManifest(weightsManifestPath)
-		changed := err != nil && weightsManifest.Equal(cachedManifest)
+		changed := err != nil || !weightsManifest.Equal(cachedManifest)
```
hongchaodeng, line 54: Let's just return err if not nil. That's easier to tell 😄 
Then make:

```go
changed := !weightsManifest.Equal(cachedManifest)
```

hongchaodeng (commented):
@ichernev Thanks for sharing this! Add a small comment.
LGTM after nit

pkg/image/build.go
ichernev: Well, if there are no cached weights (i.e the first time this is executed), there will be `err`. Which I think is OK. I would have checked `cachedManifest == nil || !weights.Equal(cached)`. What do you think?

ichernev: Alternatively we can put a check for `nil` inside `Equal` method, after all why not compare against `nil`.

mattt: +1 for `cachedManifest == nil || !weights.Equal(cached)` as the check. That closely tracks my understanding for how the feature behaves.

hongchaodeng: Use os.IsNotExist() to tell if there is no cached weights (i.e. cachedManifest == nil)

mattt (approved):
Great work, @ichernev! Thanks again for your help with this. I really appreciate it 😃

PR #1245, opened 2023-08-04
Build Docker images with SBOM and provenance attestation

mattt:
From [Docker's docs](https://docs.docker.com/build/attestations/):

> Build attestations describe how an image was built, and what it contains. The attestations are created at build-time by BuildKit, and become attached to the final image as metadata.

> Attestations are stored as manifest objects in the image index, similar in style to OCI artifacts.

This PR adds injects the `--sbom=true` and `--provenance=true` flags into the `docker build` command.

### DIFF
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -56,8 +56,16 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 
 	args = append(args,
 		"--file", "-",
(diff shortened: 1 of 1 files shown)


mattt:
This seems like a concern best handled by a CD system. For example, using something like [actions/attest-build-provenance](https://github.com/actions/attest-build-provenance). Closing for now.

PR #1251, opened 2023-08-07
.github: add issue template file with bug report template

AChelikani:
Add a first issue template of type bug report - this create issues with a label of `type/bug` and gives users the below experience when opening issues.

Screen Shot 2023-08-07 at 4 16 06 PM

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,22 @@
+---
+name: Bug report
+about: Something isn't working
+title: ''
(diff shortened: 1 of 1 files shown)


mattt:
Hi @AChelikani. Thanks for opening this PR. Sorry for not responding sooner.

So far, I haven't found issues for this project to be especially noisy. Or at least, not enough that I think an issue or PR template (#1684) would improve the helpfulness of community contributions.

Now, that would be a nice problem to have, getting so much attention that we need more process to keep up. But for now, I'm happy to keep things as they are.

PR #1267, opened 2023-08-21
Add training schemas and operations to OpenAPI schema, when available

### LINKED ISSUE #1262: Include training interfaces in OpenAPI schema
Cog has experimental support for fine-tuning. See https://github.com/replicate/cog/pull/976

This involves adding a `train()` function, much like `predict()`:

```py
from cog import BasePredictor, File
import io

def train(param: str) -> File:
    return io.StringIO("hello " + param)
```

We should update Cog to parse that training function signature into a JSON schema describing its inputs and outputs, the way we do for predictors.

mattt:
Resolves #996 
Resolves #1262


python/cog/server/http.py
```diff
@@ -79,6 +81,44 @@ def create_app(
         input_type=InputType, output_type=OutputType
     )
 
+    if "train" in config:
```
mattt, line 84: So the contract is "If a `cog.yaml` file has a `train` key, then the resulting OpenAPI schema has `TrainingInput` and `TrainingOutput` components defined; otherwise, those components aren't defined."

bfirsh:
Nice. Some little API design noodlings:

- It's still an experimental API. I wonder if it should marked as such in the schema somehow?
- Having `Input` and `TrainingInput` seems an unfortunate asymmetry. Maybe it's fine if prediction feels like the primary action. Maybe it'll feel better when (if?) we rename "prediction" to "run".

Just thoughts.

test-integration/test_integration/fixtures/train-project/train.py
```diff
@@ -0,0 +1,15 @@
...
+
+class TrainingOutput(BaseModel):
+    weights: Path
+
+def train(
+    n: int,
+) -> TrainingOutput:
+    with open("weights.bin", "w") as fh:
+        for _ in range(n):
+            fh.write("a")
```
zeke, line 11: As all the way down 🐢

zeke (approved):
Lotta surgical Python method inspection stuff in here. Thank you for your bravery.

Not confident in my ability to review this code meaningfully, but the tests look like they prove that it works... 👍🏼

Pipazoul:
Thanks for this pull request !

I've build a cog version based on `matt/train-schema` and rebuilt the the [cog-sdxl](https://github.com/replicate/cog-sdxl) repo
the trainings route when called seems to fallback to the predict 
when calling `apiurl/trainings`
```json
{
    "id": null,
    "logs": "Traceback (most recent call last):
File \"/root/.pyenv/versions/3.9.17/lib/python3.9/site-packages/cog/server/worker.py\", line 217, in _predict
result = predict(**payload)
File \"/root/.pyenv/versions/3.9.17/lib/python3.9/site-packages/torch/utils/_contextlib.py\", line 115, in decorate_context
return func(*args, **kwargs)
TypeError: predict() got an unexpected keyword argument 'input_images'
",
    "error": "predict() got an unexpected keyword argument 'input_images'",
    "input": {
        "mask": null,
        "seed": null,
        "image": null,
        "width": 1024,
        "height": 1024,
        "prompt": "An astronaut riding a rainbow unicorn",
        "refine": "no_refiner",
        "scheduler": "K_EULER",
        "lora_scale": 0.6,
        "num_outputs": 1,
        "refine_steps": null,
        "guidance_scale": 7.5,
        "apply_watermark": true,
        "high_noise_frac": 0.8,
        "negative_prompt": "",
        "prompt_strength": 0.8,
        "num_inference_steps": 50
    },
    "output": null,
    "status": "failed",
    "metrics": null,
    "version": null,
    "webhook": null,
    "created_at": null,
    "started_at": "2023-08-26T09:08:25.990165+00:00",
    "completed_at": "2023-08-26T09:08:26.322538+00:00",
    "output_file_prefix": null,
    "webhook_events_filter": [
        "start",
        "output",
        "logs",
        "completed"
    ]
}

````

**Comment by mattt** [comment]
@Pipazoul Thanks for giving this a try. In this PR, the training routes are defined solely for the purpose of defining `TrainingInput` and `TrainingOutput` types. To do trainings, you would need to run the Cog model in training mode.

File: python/cog/predictor.py [hunk 3]
```diff
@@ -326,6 +332,80 @@ class Output(BaseModel):
...
+        n: int
+    ) -> Any:
+        ...
+"""
+        )
+    else:
+        OutputType = signature.return_annotation
+
+    return create_model(
+        "Trainingoutput",
```
zeke, line 401: This capitalization is wrong. Should. be `TrainingOutput`, but when I tried changing it, tests failed. 🤔

zeke (changes requested):
Need to figure out the `Trainingoutput` vs `TrainingOutput` casing issue before shipping this.

python/cog/predictor.py
```diff
@@ -326,6 +332,80 @@ class Output(BaseModel):
...
+    # class TrainingOutput(BaseModel):
+    #     weights: CogFile
+
+    # return TrainingOutput
+
+    train = get_train(predictor)
+    signature = inspect.signature(train)
+
+    # raise NotImplementedError(f"{signature.return_annotation}")
+
```
yorickvP, line 381: should this also check that this has a `weights` key? Currently omitting that causes image creation to fail in `api`

aron:
@allcontributors please add @username for doc,code,ideas

aron:
@allcontributors please add @aron for doc,code,ideas

python/cog/server/http.py
aron: I find it a bit weird that the prediction API uses the train one when run in predict mode, so you end up with two APIs that have the same interface.

Is this essentially for backwards compatibility?

aron:
Okay… I've managed to fix the failing tests. It was a bit of a journey to get there. But essentially we were having conflicts between the `Output` type created for the prediction output and the `TrainingOutput` type created for the training.

This was compounded by a couple of subtle bugs.

1. When we assigned the class in `get_output_type` using:
   ```
   class Output(BaseModel):
     __root__: OutputType
   ```
   This would still set the `TrainingOutput` type as the root, and as such the schema would
   look like this:
   ```
   {'title': 'Output', '$ref': '#/schema/components/TrainingOutput' ... }
   ```
   And so we'd still have a `TrainingOutput` component despite also having an `Output` component.
2. Then in `get_training_output_type` we used `create_model(__root__=OutputType)` but `create_model()` doesn't support `__root__` and so just created an empty schema:
   ```
   {'title': 'TrainingOutput' }
   ```
   Using the subclass pattern we use for `Output` works here. We also had the name of the object set to `Trainingoutput` which meant that for the moment we didn't have any conflicts. But the test was picking up the one generated for the prediction Output which is why it was passing.
3. When we renamed the training output model to `TrainingOutput` we suddenly had two TrainingOutput components and so the schema decided to namespace them by module.
   ```
   {'title': 'Output', '$ref': '#/schema/components/Predictor_predict_TrainingOutput' ... }
   ```
   and
   ```
   {'title': 'TrainingOutput', '$ref': '#/schema/components/train_TrainingOutput' ... }
   ```
   so the tests suddenly started failing.

So the solution was to avoid the use of `__root__` when the name of the output type is `Output` or `OutputType` and use inheritance instead:

```diff
- class Output(BaseModel)
-   __root__: OutputType
+ class Output(OutputType)
+   pass
```

So I think now all this is working, and 🤞 the tests are passing.

.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.20
```
zeke, line 1: Is this just an `asdf`-related file, or does it somehow impact the CI/CD process?

I'm wondering if it's necessary to have it in the repo, or if it's specific to your individal dev env. If we want to keep it here, it might be worth adding a comment to the file to explain what its for and when/where it applies.

zeke (approved):
Thanks @aron for digging into that hairy pydantic issue. I think this is good to go!  😅

PR #1271, opened 2023-08-23
Attempt to parse predict input as JSON

erbridge:
We want to unescape things like `\n` that are escaped to `\\n` when serialized, but still fall back to the raw value if that fails for any reason (like the input not deserializing being to a string).


erbridge:
After a conversation, I'm not actually convinced this is the right thing...

mattt:
@erbridge Thanks for jumping on this.

As we discussed, I think a better solution would be to follow the pattern of `echo` and other utilities, and add a `-e` flag (and its inverse, `-E`) to enable escape sequences. Attempting to do this automatically, such as by first parsing as JSON, could produce unexpected results (such as passing code like `print("Hello,\nthere")` as a text input).

PR #1277, opened 2023-08-28
document training API

zeke:
This PR adds some reference documentation for the experimental training API.

### DIFF
--- a/docs/training.md
+++ b/docs/training.md
@@ -1,0 +1,103 @@
+# Training interface reference
+
+> [!NOTE]  
+> The training API is still experimental, and is subject to change.
+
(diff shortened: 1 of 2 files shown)


docs/training.md
```diff
@@ -0,0 +1,103 @@
...
+
+Use Cog's `Input()` function to define each of the parameters in your `train()` function:
+
+```py
+from cog import Input, Path
+
+def train(
+    train_data: Path = Input(description="HTTPS URL of a file containg training data"),
+    learning_rate: float = Input(description="learning rate, for learning!", default=1e-4, ge=0),
+    seed: int = Input(description="random seed to use for training", default=42)
```
evilstreak, line 50:
```suggestion
    seed: int = Input(description="random seed to use for training", default=None)
```

I get that this is illustrative rather than real, but I've seen users get confused about seeds before, and we've certainly seen models with bugs related to this. If we want to use seed here, I think it'd be better to default it to `None`, as that's almost certainly the desired behaviour.

PR #1288, opened 2023-09-05
Add ENV NVIDIA_DRIVER_CAPABILITIES=all to permit Video SDK transcoding, 3D rendering, and X11 OpenGL Display

nelsonjchen:
https://github.com/NVIDIA/nvidia-container-runtime/issues/112

I'm interested in capabilities beyond the default Nvidia base images of just compute and utility. I'm particularly interested in using the video encoding and decoding functionality and OpenGL in an X11 application.

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -106,3 +106,13 @@ This section lists the relevant environment variables for Cog's built-in queue w
 This determines whether to enable or disable the usage of OpenTelemetry. OpenTelemetry (OTEL) is an open-source technology used to capture and measure metrics, traces, and logs. It is used for Cog's queue worker.
 
 This can either be set / unset in order to determine whether or not to enable OpenTelemtry. If it is set, then Cog will handle the necessary setup for OpenTelemtry. Otherwise, OpenTelemetry calls will be treated as no-ops. If OpenTelemetry is enabled, the OTLP exporter may also need to be [configured via environment variables](https://opentelemetry-python.readthedocs.io/en/latest/sdk/environment_variables.html).
+
+
+## Docker Image
+
+### `NVIDIA_DRIVER_CAPABILITIES`
+This [controls which Nvidia driver libraries/binaries will be mounted inside the container](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/docker-specialized.html#driver-capabilities). The generated Docker image will set this to  `all` which will mount all Nvidia driver libraries/binaries inside the container beyond the default `utility` and `compute` capabilities.
+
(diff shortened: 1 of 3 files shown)


nelsonjchen:
Hmm works locally on my WSL2 machine but not on replicate.com itself. Will need to continue investigating.

nelsonjchen:
https://replicate.com/p/jokv5fbbljv3mu64k4eyyyqjli

Nope, got it working after all. Pretty pleased :) . I would like to bring this up for re-consideration.

nelsonjchen:
Could there be some consideration of this? It is not possible to set this at runtime and get the same results as the Nvidia parts of the runtime will mount more files into the container if it sees these environment variables set. That is not something possible when the container is already running.

zeke:
Hi @nelsonjchen sorry your PR went unnoticed. This seems like a sensible change to me, but do you have a sense of whether this change might break anything?

Wold you be willing to update this PR with docs about the purpose of this env var in https://github.com/replicate/cog/blob/main/docs/environment.md ?

nelsonjchen:
> Hi @nelsonjchen sorry your PR went unnoticed. This seems like a sensible change to me, but do you have a sense of whether this change might break anything?

My sense is that it does not, but I am also not a typical/normal/daily GPU user. However, I do defer to this comment by an Nvidia employee in the thread I linked on Nvidia's repo about this which was made about 3 years ago with a general approval for this kind of change:

https://github.com/NVIDIA/nvidia-container-runtime/issues/112#issuecomment-712312284

> Wold you be willing to update this PR with docs about the purpose of this env var in https://github.com/replicate/cog/blob/main/docs/environment.md ?

I don't know. It's not really configurable which is unprecedented. The environment variables there are usually customizable. I suppose I could add a section of "interesting" environmental variables that can't be configured but might be of interest.

nelsonjchen:
@zeke Documented. Not sure if that's what you're looking for.

zeke (approved):
This change looks good to me, and the documentation is clear and helpful. 👍🏼 

That said, I would like to get a second opinion from folks who have more expertise on the matter and can speak to possible issues that could arise from enabling this.

cc @technillogue @mattt @joehoover @anotherjesse @daanelson

zeke:
Hey @nelsonjchen sorry there hasn't been much attention on this.

I think it'd be much less risky to ship this as an option than a new default. 👍🏼

zeke:
Hey @nelsonjchen sorry there hasn't been much attention on this.

I think it'd be much less risky to ship this as an option than a new default. 👍🏼

nelsonjchen:
OK, I'll see what I can do to propose it as an option.

technillogue:
I think I'm fine with this even without it being an option. Can you rebase on main please?

technillogue (approved):

nelsonjchen:
Rebased on main.

PR #1289, opened 2023-09-05
COG_EXPERIMENTAL_BUILD_STAGE_DEPS for dependencies required in the pip install stage

technillogue:
this is like #1233. however, there's no way to configure anything to be installed in the pip install build stage. llama requires sentencepiece, which requires cmake to build from source. we don't want to put cmake in system_packages (and it wouldn't help anyway). so there's this envvar as a stopgap until we decide to have a build_system_packages setting or something

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -352,9 +352,18 @@ func (g *Generator) pipInstallStage() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// Not slim, so that we can compile wheels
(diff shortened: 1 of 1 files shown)


pkg/dockerfile/generator.go
```diff
@@ -372,10 +372,14 @@ func (g *Generator) pipInstallStage() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	buildStageDeps := os.Getenv("BUILD_STAGE_DEPS")
```
technillogue, line 375: COG_EXPERIMENTAL_BUILD_STAGE_DEPS

mattt (approved):

Issue #608, opened 2022-05-12
Support `Any` as input


bfirsh:
I want it to be possible to start with the simplest, most flexible thing, then layer in things on top.

You should be able to start with this:

```python
def predict(name):
    return f"hello {name}"
```

Then, you can augment it. If you want to get some fancy output, add an output type. If you want a better form field and some validation, add an input type. If you need some setup processing, add a `Predictor` and a `setup()` function.

## See also

- #470
- #1312

bfirsh:
Implementation: https://github.com/replicate/cog/pull/968

bfirsh:
There are some tests in here representing the behavior: https://github.com/replicate/cog/pull/1313

bfirsh:
From #538:

> My dream is arbitrary JSON encodable types in/out as the lowest level thing, then more specific types built on top of that. We've got that for outputs, but not inputs yet. :D

markphelps:
Thank you for this feature request and the linked implementation explorations. Since this issue has been inactive for over 716 days and the related PRs appear to have stalled, we are closing it as stale. If work on those PRs resumes or there is renewed interest, please feel free to reopen or open a new issue referencing this one.

PR #1292, opened 2023-09-08
Allow `Any` inputs

mattt:
Rebase of https://github.com/replicate/cog/pull/968


bfirsh:
Wahey! Passing tests!

It's a shame that test was deleted in the other PR though. It should be testing the new behavior that it works.

This PR is a bit of a mess now so I might try and make a clean version...

bfirsh:
So I tried the test that was deleted and turns out arbitrary input objects don't work because runner doesn't typecast objects, so this doesn't work all.

I have written some tests to represent some of the behavior I really want from #608 in #1313, but I reckon this PR and #968 is the wrong approach so I reckon we can close them.

mattt:
@bfirsh Thanks so much for digging into that. I'll go ahead and close this PR and https://github.com/replicate/cog/pull/968 now.

PR #1304, opened 2023-09-14
Override `sys.argv` while importing predictor module

nickstenning:
In production, any sys.argv should not be exposed to user code. Partly because this might leak information about the production environment, but primarily because user code often has its own argument parsing code which gets confused when it sees our arguments to cog/server/http.py.

This uses mock.patch to ensure that sys.argv contains just sys.argv[0] when user code is executing.

Even though we (Cog) have already done all argument parsing long before this executes, we also put the original value back once we're done running user code.

### DIFF
--- a/python/tests/test_predictor.py
+++ b/python/tests/test_predictor.py
@@ -1,7 +1,10 @@
+import os
+import sys
 from typing import Optional
+from unittest.mock import patch
 
 from cog import File, Path
-from cog.predictor import get_weights_type
+from cog.predictor import get_weights_type, load_predictor_from_ref
 
 
 def test_get_weights_type() -> None:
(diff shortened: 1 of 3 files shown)


anotherjesse:
@nickstenning extra points for the tests

python/tests/test_predictor.py
```diff
@@ -24,3 +27,18 @@ def f(weights: Optional[File]) -> None:
...
 
     assert get_weights_type(f) == File
+
+
+def test_load_predictor_from_ref_overrides_argv():
+    with patch("sys.argv", ["foo.py", "exec", "--giraffes=2", "--eat-cookies"]):
+        predictor = load_predictor_from_ref(_fixture_path("argv_override"))
+
+        # check the predictor module saw no args
+        assert predictor.predict() == ["foo.py"]
```
anotherjesse, line 37: awesome test

anotherjesse (approved):
lgtm

mattt (approved):

PR #1311, opened 2023-09-21
Allow `Union` and `List` input types

mattt:
Alternative to ~#968~ https://github.com/replicate/cog/pull/1292

This PR extends Cog to support `List` and `Union` type inputs. 

For example, given the following predictor:

```python
from cog import BasePredictor, Path

from typing import List, Union

    def predict(self, args: Union[int, List[int]]) -> int:
        if isinstance(args, int):
            return args
        else:
            return sum(args)

```

The following assertions are satisfied:

```python
resp = client.post("/predictions", json={"input": {"args": 123}})
assert resp.status_code == 200
assert resp.json()["output"] == 123

resp = client.post("/predictions", json={"input": {"args": [1, 2, 3]}})
assert resp.status_code == 200
assert resp.json()["output"] == 6

resp = client.post("/predictions", json={"input": {"args": "abc"}})
assert resp.status_code == 422

resp = client.post("/predictions", json={"input": {"args": ["a", "b", "c"]}})
assert resp.status_code == 422
```

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -217,7 +217,7 @@ def get_predict(predictor: Any) -> Callable:
 def validate_input_type(type: Type, name: str) -> None:
     if type is inspect.Signature.empty:
             raise TypeError(
-                f"No input type provided for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}."
+                f"No input type provided for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}, or a Union or List of those types."
             )
     elif type not in ALLOWED_INPUT_TYPES:
         if hasattr(type, "__origin__") and (type.__origin__ is Union or type.__origin__ is list):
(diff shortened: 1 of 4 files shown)


bfirsh (approved):
Marvellous. Could do with some docs, but we can fix up later.

I wonder how this interacts with Replicate?

mattt:
@bfirsh `Union[str, List[str]]` is expressed in OpenAPI as `{"anyOf": [{"type": "string"}, {"type": "array", "items": "string"}}]`. Currently, Replicate renders inputs of that schema as a single text field. We can eventually support multiple inputs, but this seems sufficient for now, since users can send multiple inputs via the API.

peter65374:
only List[int] is supported so far? Does input type support List[Path] or List[str]? I am expecting to input multiple images to process as below
```
def predict(
        self,
        target_image: Path = Input(
            description="Faces in target image would be changed.",
        ),
        model_faces: list[Path] = Input(
            description="Model Faces would be swapped into target image.", 
            default=None,
        ),
        inference_mode: str = Input(
            default="swap",
            choices=["swap", "detect"],
            description="Face swap mode or detection mode. Default is swap.",
        ),
    ) -> Union[List[tuple], List[Path]]:
        # Check if inference_mode is detect or swap
        if inference_mode == "detect":
            detect_mode = True
        else: 
            detect_mode = False
```

PR #1313, opened 2023-09-22
Allow `Any` input with tests

bfirsh:
A version of https://github.com/replicate/cog/pull/1292 that retains the tests. A partial implementation of https://github.com/replicate/cog/issues/608.

Turns out it doesn't work at all. Objects are passed as plain JSON instead of being typecast because runner doesn't typecast inputs. To be revisited at some point.


mattt:
Closing in favor of #1311.

PR #1315, opened 2023-09-22
Hide `--dockerfile` flag

mattt:
Follow-up to #1229

The `--dockerfile` flag is intended primarily for debugging, and shouldn't be advertised in command usage output. This PR marks the flag as hidden.

```console
$ cog build --help
Build an image from cog.yaml

Usage:
  cog build [flags]

Flags:
  -h, --help                         help for build
      --no-cache                     Do not use cache when building the image
      --openapi-schema string        Load OpenAPI schema from a file
      --progress string              Set type of build progress output, 'auto' (default), 'tty' or 'plain' (default "auto")
      --secret stringArray           Secrets to pass to the build environment in the form 'id=foo,src=/path/to/file'
      --separate-weights             Separate model weights from code in image layers
  -t, --tag string                   A name for the built image in the form 'repository:tag'
      --use-cuda-base-image string   Use Nvidia CUDA base image, 'true' (default) or 'false' (use python base image). False results in a smaller image but may cause problems for non-torch projects (default "auto")

Global Flags:
      --debug     Show debugging output
      --version   Show version of Cog
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/moby/term v0.5.0
 	github.com/spf13/cobra v1.7.0
+	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.4
 	github.com/vincent-petithory/dataurl v1.0.0
 	github.com/xeipuuv/gojsonschema v1.2.0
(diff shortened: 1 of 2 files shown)


bfirsh (approved):

technillogue:
I would be in favor of letting people in general use it, maybe with a disclaimer saying the entrypoint needs to work and conform

mattt:
There's symmetry between `cog build --dockerfile=` and `cog debug`, which is similarly hidden. Whatever we do for one we should do for both.

technillogue:
It's also a shame cog debug is hidden

PR #1317, opened 2023-09-25
Unhide dockerfile commands

technillogue:
As per https://github.com/replicate/cog/pull/1315#issuecomment-1733358324, cog debug is very useful as a starting point for --dockerfile. We should make both visible.


technillogue:
I guess we're not doing this right now

PR #1327, opened 2023-10-03
Update deprecated list type usage

iamargentum:


iamargentum:
Just came across this while referring the docs, so thought of making the change.

iamargentum:
thanks @yorickvP !

PR #1345, opened 2023-10-23
use lifespan instead of deprecated startup/shutdown events

technillogue:
This ensures shutdown can't get called unless setup finished. This isn't strictly required, and if we don't change any other dependencies (unlikely) we can skip it to avoid upgrading fastapi. `predict` and `shutdown` will become async, but that doesn't mean lifespan is required.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -1,4 +1,5 @@
 import argparse
+import contextlib
 import logging
 import os
 import signal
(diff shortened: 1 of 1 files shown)


nickstenning (commented):
_Please_ include some motivation/context around the _why_ of pull requests like this. I have no idea whether this is needed to unblock something or whether it's just tidying. :)

If it is just tidying we may want to consider whether this is worthwhile given that it will force us update the FastAPI version pin to `>0.94.0` when `lifespan` was introduced.

PR #1350, opened 2023-10-25
support async predict functions 

technillogue:
this is the first step towards supporting continuous batching and concurrent predictions. eventually, you will be configure it so your predict function will be called concurrently. this will work with engines that implement batching internally like vLLM and TRT, and allow users to implement things like microbatching on their own. it may also be possible to have truly concurrent predictions without batching at all by using cuda streams.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -173,12 +173,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
         self._stream_redirector.start()
 
         self._setup()
-        try:
-            loop = asyncio.get_running_loop()
-        except RuntimeError:
-            loop = asyncio.new_event_loop()
-        loop.run_until_complete(self._loop())
-
+        asyncio.run(self._loop())
         self._stream_redirector.shutdown()
 
     def _setup(self) -> None:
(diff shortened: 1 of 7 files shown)


python/cog/server/worker.py
```diff
@@ -172,7 +173,11 @@ def run(self) -> None:
         self._stream_redirector.start()
 
         self._setup()
-        self._loop()
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
+            loop = asyncio.new_event_loop()
+        loop.run_until_complete(self._loop())
```
nickstenning, line 180: What's the reasoning behind this `try/except` vs calling `asyncio.run(self._loop())`?

nickstenning (approved):
One question but otherwise LGTM.

Would suggest adding a bit of context around the "why" to the PR and possibly the commit messages before we merge it.

python/cog/server/worker.py
technillogue: I wasn't sure if there would be an event loop already running but I think you're right and this is the top of the subprocess so unless multiprocesisng does something I don't understand that should be fine

PR #1352, opened 2023-10-28
async runner

technillogue:
stacked PR on top of #1350 as well as #1355, this similarly only takes steps to eventually allow concurrent predictions. this PR starts with changing multiprocessing.pool.AsyncResult to asyncio.Task, and propagates the necessary async changes

- have runner return asyncio.Task instead of AsyncFuture
- making tests async

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -203,7 +203,8 @@ def create_app(
             return JSONResponse(jsonable_encoder(initial_response), status_code=202)
 
         try:
-            response = PredictionResponse(**(await async_result).dict())
+            prediction = await async_result
+            response = PredictionResponse(**prediction.dict())
         except ValidationError as e:
             _log_invalid_output(e)
             raise HTTPException(status_code=500, detail=str(e)) from e
@@ -246,7 +247,8 @@ def create_app(
         if not app.state.setup_result.done():
             return
 
-        result = await app.state.setup_result
+        # this can raise CancelledError
+        result = app.state.setup_result.result()
 
         if result["status"] == schema.Status.SUCCEEDED:
             app.state.health = Health.READY
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -107,7 +107,7 @@ def create_app(
 
     @app.get("/health-check")
     async def healthcheck() -> Any:
-        _check_setup_result()
+        await _check_setup_result()
         if app.state.health == Health.READY:
             health = Health.BUSY if runner.is_busy() else Health.READY
         else:
@@ -137,7 +137,7 @@ def create_app(
         # TODO: spec-compliant parsing of Prefer header.
         respond_async = prefer == "respond-async"
 
-        return _predict(request=request, respond_async=respond_async)
+        return await _predict(request=request, respond_async=respond_async)
 
     @limited
     @app.put(
@@ -172,9 +172,9 @@ def create_app(
         # TODO: spec-compliant parsing of Prefer header.
         respond_async = prefer == "respond-async"
 
-        return _predict(request=request, respond_async=respond_async)
+        return await _predict(request=request, respond_async=respond_async)
 
-    def _predict(
+    async def _predict(
         *, request: PredictionRequest, respond_async: bool = False
     ) -> Response:
         # [compat] If no body is supplied, assume that this model can be run
@@ -203,7 +203,8 @@ def create_app(
             return JSONResponse(jsonable_encoder(initial_response), status_code=202)
 
         try:
-            response = PredictionResponse(**async_result.get().dict())
+            res = await async_result
+            response = PredictionResponse(res.dict())
         except ValidationError as e:
             _log_invalid_output(e)
             raise HTTPException(status_code=500, detail=str(e)) from e
@@ -239,14 +240,14 @@ def create_app(
             shutdown_event.set()
         return JSONResponse({}, status_code=200)
 
-    def _check_setup_result() -> Any:
+    async def _check_setup_result() -> Any:
         if app.state.setup_result is None:
             return
 
-        if not app.state.setup_result.ready():
+        if not app.state.setup_result.done():
             return
 
-        result = app.state.setup_result.get()
+        result = await app.state.setup_result
 
         if result["status"] == schema.Status.SUCCEEDED:
             app.state.health = Health.READY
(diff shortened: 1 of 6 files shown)


nickstenning:
I started rebasing this but without context it quickly turned into a bigger job than I hoped. Would you mind rebasing this on `main` so it's reviewable in its own right?

technillogue:
(did you mean something else about rebasing?)

nickstenning:
> (did you mean something else about rebasing?)

Yes! Your description here says:

> stacked PR on top of https://github.com/replicate/cog/pull/1350 as well as https://github.com/replicate/cog/pull/1355,

Currently this PR is for a set of commits that we'd not want to merge to main without a rebase. Please rebase this branch on `main` so that it's a reviewable unit. Merging `main` achieves the same overall diff but it doesn't allow this PR to be reviewed commit by commit.

technillogue:
had to brush up on how to rebase but once I remembered --onto I had it work without any conflicts. this is still stacked on #1355

python/tests/server/test_http.py
```diff
@@ -450,6 +450,7 @@ def test_prediction_idempotent_endpoint_conflict(client, match):
         json={"input": {"sleep": 1}},
         headers={"Prefer": "respond-async"},
     )
+    time.sleep(0.001)
```
nickstenning, line 453: Hmmmm. What's going on here?

python/tests/server/test_http.py
```diff
@@ -517,12 +518,12 @@ def test_prediction_cancel(client):
     )
     assert resp.status_code == 202
 
-    resp = client.post("/predictions/456/cancel")
-    assert resp.status_code == 404
-
     resp = client.post("/predictions/123/cancel")
     assert resp.status_code == 200
 
+    resp = client.post("/predictions/456/cancel")
```
nickstenning, line 524: Why has this moved?

python/cog/server/runner.py
```diff
@@ -144,8 +130,7 @@ def is_busy(self) -> bool:
 
     def shutdown(self) -> None:
         self._worker.terminate()
-        self._threadpool.terminate()
-        self._threadpool.join()
+        # TODO: cancel setup or predict task
```
nickstenning, line 133: I think we need to address this TODO before we can merge this.

python/cog/server/http.py
```diff
@@ -203,7 +203,7 @@ def _predict(
             return JSONResponse(jsonable_encoder(initial_response), status_code=202)
 
         try:
-            response = PredictionResponse(**async_result.get().dict())
+            response = PredictionResponse(**(await async_result).dict())
```
nickstenning, line 206:
```suggestion
            prediction = await async_result
            response = PredictionResponse(**prediction.dict())
```

python/cog/server/http.py
```diff
@@ -239,14 +239,14 @@ async def start_shutdown() -> Any:
...
+    async def _check_setup_result() -> Any:
         if app.state.setup_result is None:
             return
 
-        if not app.state.setup_result.ready():
+        if not app.state.setup_result.done():
             return
 
-        result = app.state.setup_result.get()
+        result = await app.state.setup_result
```
nickstenning, line 249: I think it might be clearer to do

```suggestion
        result = app.state.setup_result.result()
```

here, because as written, this kind of implies that we might actually be waiting for it, which shouldn't happen.

nickstenning: Also do we need to handle cancellation of this task here?

python/cog/server/runner.py
```diff
@@ -283,12 +268,13 @@ def _upload_files(self, output: Any) -> Any:
...
 
 
-def setup(*, worker: Worker) -> Dict[str, Any]:
+async def setup(*, worker: Worker) -> Dict[str, Any]:
     logs = []
     status = None
     started_at = datetime.now(tz=timezone.utc)
 
     try:
+        # will be async
```
nickstenning, line 277: Do we need an async yield (`await asyncio.sleep(0)`) in this loop as you've done in the predict loop?

python/tests/server/test_runner.py
```diff
@@ -65,53 +68,57 @@ def test_prediction_runner(runner):
...
-def test_prediction_runner_called_while_busy(runner):
+@pytest.mark.asyncio
+async def test_prediction_runner_called_while_busy(runner):
     request = PredictionRequest(input={"sleep": 0.1})
     _, async_result = runner.predict(request)
 
     assert runner.is_busy()
     with pytest.raises(RunnerBusyError):
-        runner.predict(request)
+        await runner.predict(request)[1]
```
nickstenning, line 78:
```suggestion
        _, task = runner.predict(request)
        await task
```

python/tests/server/test_runner.py
```diff
@@ -65,53 +68,57 @@ def test_prediction_runner(runner):
...
+@pytest.mark.asyncio
+async def test_prediction_runner_called_while_busy_idempotent(runner):
     request = PredictionRequest(id="abcd1234", input={"sleep": 0.1})
 
     runner.predict(request)
     runner.predict(request)
     _, async_result = runner.predict(request)
 
-    response = async_result.get(timeout=1)
+    response = await async_result
```
nickstenning, line 93: Let's not do away with the timeouts. From experience, the timeouts help make running the tests not an utterly miserable experience if you're working on the code and break it somehow.
```suggestion
    response = await asyncio.wait_for(async_result, timeout=1)
```

python/tests/server/test_http.py
technillogue: when I added it the test would fail because the second request would come in too quickly, but I think maybe the limiter change from async routes fixes it

python/tests/server/test_runner.py
technillogue: that's a lot nicer, thank you

python/cog/server/runner.py
technillogue: I added it in the predict loop because I was getting errors and didn't get any in this loop, but we may as well for completeness

python/cog/server/http.py
technillogue: hm, yes, both result and await can raise CancelledError (or errors from setup). that code path would happen if we get a healthcheck while shutting down after being told to shut down... is that SETUP_FAILED? there's not a shutting-down state

technillogue: it's possible just propagating CancelledError and setup errors through to the healthcheck isn't the worst behavior, it should return 500 which seems reasonable

nickstenning (approved):
LGTM.

technillogue:
is this good to merge then?

PR #1353, opened 2023-10-30
Add support for `.glb`, `.mtl` and `.obj` files

mattt:
This PR adds MIME type definitions some 3D assets.

| Content type | File extension |
|-|-|
| `model/gltf-binary` | `.glb` |
| `model/mtl` |         `.mtl` |
| `model/obj` |          `.obj`|

### DIFF
--- a/pkg/util/mime/mime.go
+++ b/pkg/util/mime/mime.go
@@ -68,6 +68,10 @@ var typeToExtension = map[string]string{
 	"image/vnd.microsoft.icon": ".ico",
 	"image/webp":               ".webp",
 
+	"model/gltf-binary": ".glb",
+	"model/mtl":         ".mtl",
+	"model/obj":         ".obj",
+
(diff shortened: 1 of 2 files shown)


yorickvP:
There are still some problems with glb files.

1. inside the container, python's mimetypes doesn't know about it, and returns `None` (https://github.com/replicate/cog/blob/main/python/cog/files.py#L28). We should add the mimetypes list inside the container and/or default to application/octet-stream.
2. Outside the container, dataurl.DecodeString on `data:model/gltf-binary;base64,...` URLs fails with "Failed to decode dataurl: invalid media type"

yorickvP:
What we could do is call mime.AddExtensionByType [here](https://github.com/replicate/cog/blob/main/pkg/util/mime/mime.go#L97) and then call python mimetypes.add_type for each of the extensions somewhere in the cog python.

PR #1355, opened 2023-10-30
make HTTP routes async

technillogue:
this doesn't change any logic, but as far as I understand sync routes use threads instead of the starlette event loop: https://fastapi.tiangolo.com/async/#path-operation-functions

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -13,8 +13,6 @@ from typing import Any, Callable, Dict, Optional, Union
 
 import structlog
 import uvicorn
-from anyio import CapacityLimiter
-from anyio.lowlevel import RunVar
 from fastapi import Body, FastAPI, Header, HTTPException, Path, Response
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import RequestValidationError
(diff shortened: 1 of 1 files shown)


nickstenning (commented):
I believe this may not be as harmless as it looks.

There's a `--threads` argument to `cog.server.http` and I believe that it functions as a concurrency control for the synchronous prediction create endpoint.

I believe at least some people rely on this behaviour so they can use the network socket's accept queue as a mechanism to queue up work. If this breaks that behaviour (i.e. those requests are immediately accepted and we serve a 409 instead) that may break those folks' expectations.

Is this change a requirement or just a nice-to-have for the batching/concurrent predictions work?

technillogue:
Do we have any specific examples of someone actually using that, or is hypothetical?

I _think_ it will be required somewhere down the line, I put it up front here because it can be reviewed as an independent PR

nickstenning:
Yes, in that if I remember correctly, we broke it previously while working on Cog and had to put it back.

technillogue:
we could stick a timeout or something where the endpoint waits for the worker for a similar time as what the tcp timeout would have been? do you remember what exactly the use case was and why they didn't want to use retries instead?

python/cog/server/http.py
```diff
@@ -79,9 +81,20 @@ def create_app(
...
+            async with http_semaphore:
+                return await f(*args, **kwargs)
+
+        return wrapped
+
     @app.on_event("startup")
     def startup() -> None:
         # https://github.com/tiangolo/fastapi/issues/4221
+        # potentially no longer necessary?
         RunVar("_default_thread_limiter").set(CapacityLimiter(threads))  # type: ignore
```
nickstenning, line 98: I think it's definitely no longer necessary if we've verified that the semaphore does the same job.

nickstenning (approved):
I think it would be nice if we could have a test for this behaviour, but I'm not going to block this on that. I've tested it locally and it works well!

PR #1366, opened 2023-11-07
create event loop before predictor setup

technillogue:
The current implementation (prior to #1350)  throws an error:
```
Traceback (most recent call last):
File "/usr/local/lib/python3.11/site-packages/cog/server/worker.py", line 226, in _predict
for r in result:
...
File "/src/src/download.py", line 118, in sync_download_file
return self.loop.run_until_complete(self.download_file(url))
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/asyncio/base_events.py", line 629, in run_until_complete
self._check_running()
File "/usr/local/lib/python3.11/asyncio/base_events.py", line 590, in _check_running
raise RuntimeError(
RuntimeError: Cannot run the event loop while another loop is running
```

The problem is that llama's [setup()](https://github.com/replicate/cog-llama-template/blob/main/predict.py#L35) creates a Downloader, which [can't find an event loop and creates one](https://github.com/replicate/cog-llama-template/blob/main/src/download.py#L21-L24), then tries to use that loop in predict, and by that time `_ChildWorker.run` has used `asyncio.run`

This solves that specific problem, and since we didn't see any other errors besides llama maybe few other models use an event loop. However, we may want/need to consider moving the check where we inspect if `predict()` is async earlier, and have more separate code paths so that we don't need to be async if the user isn't using the async feature. On the other hand, doing that could get messy and involve a lot of duplicated code, so if we're sure our users use `asyncio.get_running_loop` and not `asyncio.create_event_loop`, it would be best to not do that.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -80,7 +80,7 @@ async def run_setup_async(predictor: BasePredictor) -> None:
         return await maybe_coro
 
 
-def get_weights_argument(predictor: BasePredictor) -> Union[io.IOBase, CogPath, None]:
+def get_weights_argument(predictor: BasePredictor) -> Union[CogFile, CogPath, None]:
     # by the time we get here we assume predictor has a setup method
     weights_type = get_weights_type(predictor.setup)
     if weights_type is None:
@@ -112,7 +112,7 @@ def get_weights_argument(predictor: BasePredictor) -> Union[io.IOBase, CogPath, 
     return None
 
 
-def get_weights_type(setup_function: Callable) -> Optional[Any]:
+def get_weights_type(setup_function: Callable[[Any], Optional[Awaitable[None]]]) -> Optional[Any]:
     signature = inspect.signature(setup_function)
     if "weights" not in signature.parameters:
         return None
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -10,6 +10,7 @@ from collections.abc import Iterator, AsyncIterator
 from pathlib import Path
 from typing import (
     Any,
+    Awaitable,
     Callable,
     Dict,
     List,
@@ -48,7 +49,9 @@ ALLOWED_INPUT_TYPES: List[Type[Any]] = [str, int, float, bool, CogFile, CogPath]
 
 
 class BasePredictor(ABC):
-    def setup(self, weights: Optional[Union[CogFile, CogPath]] = None) -> None:
+    def setup(
+        self, weights: Optional[Union[CogFile, CogPath]] = None
+    ) -> Union[Awaitable[None], None]:
         """
         An optional method to prepare the model so multiple predictions run efficiently.
         """
@@ -63,15 +66,23 @@ class BasePredictor(ABC):
 
 
 def run_setup(predictor: BasePredictor) -> None:
+    weights = get_weights_argument(predictor)
+    if weights:
+        return predictor.setup()
+    return predictor.setup(weights=weights)
+
+
+async def run_setup_async(predictor: BasePredictor) -> None:
+    weights = get_weights_argument(predictor)
+    maybe_coro = predictor.setup(weights=weights) if weights else predictor.setup()
+    if maybe_coro:
+        return await maybe_coro
+
+
+def get_weights_argument(predictor: BasePredictor) -> Union[io.IOBase, CogPath, None]:
     weights_type = get_weights_type(predictor.setup)
-
-    # No weights need to be passed, so just run setup() without any arguments.
     if weights_type is None:
-        predictor.setup()
-        return
-
-    weights: Union[io.IOBase, Path, None]
-
+        return None
     weights_url = os.environ.get("COG_WEIGHTS")
     weights_path = "weights"
 
@@ -81,30 +92,25 @@ def run_setup(predictor: BasePredictor) -> None:
     # TODO: CogFile/CogPath should have subclasses for each of the subtypes
     if weights_url:
         if weights_type == CogFile:
-            weights = cast(CogFile, CogFile.validate(weights_url))
-        elif weights_type == CogPath:
+            return cast(CogFile, CogFile.validate(weights_url))
+        if weights_type == CogPath:
             # TODO: So this can be a url. evil!
-            weights = cast(CogPath, CogPath.validate(weights_url))
-        else:
-            raise ValueError(
-                f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
-            )
-    elif os.path.exists(weights_path):
+            return cast(CogPath, CogPath.validate(weights_url))
+        raise ValueError(
+            f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
+        )
+    if os.path.exists(weights_path):
         if weights_type == CogFile:
-            weights = cast(CogFile, open(weights_path, "rb"))
-        elif weights_type == CogPath:
-            weights = CogPath(weights_path)
-        else:
-            raise ValueError(
-                f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
-            )
-    else:
-        weights = None
-
-    predictor.setup(weights=weights)
+            return cast(CogFile, open(weights_path, "rb"))
+        if weights_type == CogPath:
+            return CogPath(weights_path)
+        raise ValueError(
+            f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
+        )
+    return None
 
 
-def get_weights_type(setup_function: Callable[[Any], None]) -> Optional[Any]:
+def get_weights_type(predictor: BasePredictor) -> Optional[Any]:
     signature = inspect.signature(setup_function)
     if "weights" not in signature.parameters:
         return None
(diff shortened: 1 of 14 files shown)


nickstenning:
Is this the fix for cog v0.9.0b9 being broken on `main` for some models? If so would you have a look at what's broken the tests here so we can get this merged? It's not great for `main` to be be unreleasable.

python/cog/predictor.py
```diff
@@ -46,7 +47,9 @@
 
 
 class BasePredictor(ABC):
-    def setup(self, weights: Optional[Union[CogFile, CogPath]] = None) -> None:
+    def setup(
+        self, weights: Optional[Union[CogFile, CogPath]] = None
+    ) -> Union[Awaitable[None], None]:
```
nickstenning, line 52: `Union[T, None]` is usually spelled `Optional[T]`, right? Any reason we wouldn't do that here?

python/cog/server/worker.py
```diff
@@ -171,18 +178,24 @@ def run(self) -> None:
...
+        self._loop()
         self._stream_redirector.shutdown()
 
+    # I would rather have _setup_async and have a single run_until_complete for setup + loop
+    # but we need to load predictor (and handle errors) to find out the code path
     def _setup(self) -> None:
         done = Done()
         try:
             self._predictor = load_predictor_from_ref(self._predictor_ref)
+            if is_async_predictor(self._predictor):
```
nickstenning, line 191: The fact that `is_async_predictor` checks both setup and predict makes this a little hard to read IMO.

I think it would be clearer if the setup path did:

```python
self._predictor = load_predictor_from_ref(self._predictor_ref)
# Could be a function or a class
if hasattr(self._predictor, "setup"):
    if is_async(self._predictor.setup):
        get_loop().run_until_complete(run_setup_async(self._predictor))
    else:
        run_setup(self._predictor)
```

and then similarly, later on

```python
def _loop(self) -> None:
        if is_async(get_predict(self._predictor)):
            return get_loop().run_until_complete(self._loop_async())
        return self._loop_sync()
```

python/cog/server/worker.py
```diff
@@ -198,50 +211,73 @@ def _setup(self) -> None:
...
-                    for r in result:
-                        self._events.send(PredictionOutput(payload=make_encodeable(r)))
                 elif inspect.isawaitable(result):
                     result = await result
                     self._events.send(PredictionOutputType(multi=False))
                     self._events.send(PredictionOutput(payload=make_encodeable(result)))
+
+    def _predict_sync(self, payload: Dict[str, Any]) -> None:
+        with self._handle_predict_error():
+            result = get_predict(self._predictor)(**payload)
```
nickstenning, line 272: I'm not sure this is more readable on one line than it was on two.

python/cog/server/worker.py
```diff
@@ -52,7 +59,7 @@ def __init__(self, predictor_ref: str, tee_output: bool = True) -> None:
         self._child = _ChildWorker(predictor_ref, child_events, tee_output)
         self._terminating = False
 
-    def setup(self) -> Iterable[_PublicEventType]:
+    def setup(self) -> Iterator[_PublicEventType]:
```
nickstenning, line 62: Oops! This was probably my mistake.

nickstenning (commented):
Just a couple of mostly superficial requests to make things easier to follow. Mostly it looks good.

python/cog/server/worker.py
technillogue: I think Iterable isn't wrong I just needed Iterator for the generator and didn't feel like importing both

python/cog/server/worker.py
technillogue: I think the rationale here is that is predict is async, the user has opted into the async world entirely, and we want to create an event loop before running setup, so get_running_loop in the users' setup code will find the same event loop we're going to use to run _loop_async. that may not be the right approach though

nickstenning: Ah, ok, that makes sense. That wasn't clear to me, but the comment you added is very helpful.

python/cog/server/worker.py
```diff
@@ -171,18 +178,30 @@ def run(self) -> None:
...
+            # that to find the same loop that will be be used for _loop_async
+            #
+            # otherwise, the user might new_event_loop to use with a ClientSession
+            # and then try to use the same ClientSession with a different event loop
+            if is_async_predictor(self._predictor):
+                self.loop = get_loop()
             # Could be a function or a class
             if hasattr(self._predictor, "setup"):
-                run_setup(self._predictor)
+                if inspect.iscoroutinefunction(self._predictor.setup):
```
nickstenning, line 201: I still think we should probably use `is_async` here?

technillogue: setup shouldn't be allowed to be a generator

python/cog/server/worker.py
technillogue: actually now that I think about it just using new_event_loop doesn't mean that get_running_loop will work, I might be completely wrong...

the crux of the issue is `def setup()` creating a Downloader, which does get_running_loop/new_event_loop, creates a ClientSession implicitly attached to that loop, uses sync_download_files_to_disk, and then (in the future), `async def predict()` might use await `downloader.download_file()` to download a finetune, and ClientSession will error because it's not the same event loop as when it was created. my specific downloader code tries to guard against this but a different user might not do that

maybe if predict is async we should just always wrap setup() in a coroutine so that any get_running_loop will do the right thing?

python/cog/server/worker.py
```diff
@@ -171,18 +178,30 @@ def run(self) -> None:
             [ws_stdout, ws_stderr], self._stream_write_hook
         )
         self._stream_redirector.start()
-
         self._setup()
-        asyncio.run(self._loop())
+        self._loop()
```
nickstenning, line 182: Another option that might (?) make things less icky, if I've finally understood the problem here.

```python
def run(self) -> None:
    ...
 
    self._stream_redirector.start()
    asyncio.run(self._run())
    self._stream_redirector.shutdown()

async def _run(self) -> Awaitable[None]:
    await self._setup()
    await self._loop()

async def _setup(self) -> Awaitable[None]:
    ...
```

technillogue: we don't know if we're async or not until we load_predictor_from_ref, which might throw an error, that should be handled in the same block as errors from calling setup

technillogue: we could duplicate the setup error handling with a similar pattern as _handle_predict_error and then this could work, I was just a little confused what was happening with Done and _cancelable

python/cog/server/worker.py
nickstenning: I think what you're doing here should work.

```python
In [24]: async def who_is_running():
    ...:     print(id(asyncio.get_running_loop()))
    ...:

In [25]: await who_is_running()
4338068384

In [26]: loop = asyncio.new_event_loop()

In [27]: id(loop)
Out[27]: 4342403856

In [28]: loop.run_until_complete(who_is_running())
4342403856
```

python/cog/server/worker.py
nickstenning: Maybe I'm missing something, but why does the predictor need to be async for these methods to be async?

technillogue: if the predictor isn't async, they haven't opted into the async world, and if we make these methods async we would get the same problems we were having before with there being a running event loop that the user isn't expecting, breaking unsuspecting code

technillogue: we could have a kind of silly check like `"async" in open("predict.py").read()`

python/cog/server/worker.py
technillogue: it's more like this

```
def get_loop():
    try:
        return asyncio.get_running_loop()
    except RuntimeError:
        return asyncio.new_event_loop()


async def who_is_running(msg):
    print(msg, id(asyncio.get_running_loop()))


def setup():
    user_loop = get_loop()
    user_loop.run_until_complete(who_is_running("user setup loop"))


async def predict():
    await who_is_running("predict loop")


loop = get_loop()
setup()
loop.run_until_complete(predict())
```

user setup loop 139844438211888
predict loop 139844449272592

if that's IPython with autoawait you can get confusing results

technillogue:
after reflecting on it, I think the current code has the correct behavior, but not for the reasons I initially thought. here are some some scenarios:

1) you use the same event loop from sync setup and from sync predict
2) you have sync setup that uses asyncio.run, and an async predict. your setup and predict will use different event loops, but if setup is only concerned with downloading and predict doesn't use the network, that's fine, and we'd rather not make people rewrite everything here. in this case there is no difference if we new_event_loop before the setup code or not.
3) you have both async setup and async predict. you can't use asyncio.run/run_until_complete, but you will be in the same running loop
4) if you really want, you can have async setup and sync predict, which is silly but won't break anything

python/cog/server/worker.py
yorickvP: We don't have [an existing contract](https://www.hyrumslaw.com/) on `async def setup` functions, so we're free to break people who are spawning their own event loops there. They'll discover and fix that during testing.

We do have an established behavior of allowing event loops in non-async `def setup` or predict functions, disallowing that now will probably break user code.
Maybe that means they should run in another thread so they're free to run their own loop?
(note that we should only call 1 at a time)

technillogue: I think https://github.com/replicate/cog/pull/1366#issuecomment-1811775725 addresses the motivation and the current code in this PR doesn't break people who spawn event loops in setup; running it in a thread shouldn't be needed

technillogue: if neither setup or predict is async, no event loop is created; if predict is async but setup isn't, then an event loop is created but not running, so asyncio.run will work, get_running_loop won't, and people will fall through to new_event_loop. this will be a different event loop than the one that will run an async predict, but that's tolerable

nickstenning (approved):
LGTM.

PR #1372, opened 2023-11-14
Get faster initial output from latency-sensitive models

nickstenning:
This is a bit of a hack, but it should help us get output sooner from models that are latency-sensitive and return their first output very shortly after prediction start.

If COG_THROTTLE_RESPONSE_INTERVAL is set to a value below 0.1 (i.e. <100ms), don't sent the initial "start" webhook. This should mean that the first output webhook will be sent as soon as it is received. Without this, there is a risk that it will be throttled by the ResponseThrottler.

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -16,7 +16,7 @@ from ..files import put_file_to_signed_endpoint
 from ..json import upload_files
 from .eventtypes import Done, Heartbeat, Log, PredictionOutput, PredictionOutputType
 from .probes import ProbeHelper
-from .webhook import webhook_caller_filtered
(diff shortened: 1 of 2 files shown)


technillogue (approved):
My only concern is if the rest of the system will accurately record status_processing_at and stuff; otherwise this seems like a great idea

nickstenning:
> My only concern is if the rest of the system will accurately record status_processing_at and stuff; otherwise this seems like a great idea

Good question! But that's done [in director](https://github.com/replicate/director/blob/12d1627015118943e34db173ec42bdfdab0a0289/director/director.py#L349) and is independent of any of the webhooks that come back from Cog.

PR #1379, opened 2023-11-15
allow None argument

technillogue:
Sometimes you want to distinguish None from 0 or "", and this isn't currently allowed


yorickvP:
Tests fail

mattt (changes requested):
I got tests passing, but I'm not sure this change has the intended effect.

@technillogue Would it make more sense to update `validate_input_type` to handle `Optional` in addition to `Union`?

python/cog/predictor.py
```diff
@@ -531,22 +532,26 @@ class TrainingOutput(BaseModel):
...
-    if module == "builtins":
-        return t.__qualname__
-    elif module.split(".")[0] == "cog":
-        module = "cog"
 
-    try:
-        return module + "." + t.__qualname__
-    except AttributeError:
-        return str(t)
+    if hasattr(t, "__module__"):
```
mattt, line 542: I spun off this fix as a separate PR: https://github.com/replicate/cog/pull/1810

Issue #1381, opened 2023-11-17
Unable to install a git-hosted Python package in cog.yaml


zeke:
I'm getting this error when trying to `cog build` using a git-hosted dependency in my `cog.yaml` file:

```
#0 2.884   ERROR: Error [Errno 2] No such file or directory: 'git' while executing command git version                                                        
#0 2.886 ERROR: Cannot find command 'git' - do you have 'git' installed and in your PATH?
```

---

```
$ cog --version
cog version 0.8.6 (built 2023-08-07T23:02:15+0000)
```

```
$ cat cog.yaml
build:
  gpu: false
  python_packages:
    - "git+https://github.com/m-bain/whisperX.git"

predict: "predict.py:Predictor"
```

```
$ cog build   
Building Docker image from environment in cog.yaml as cog-video-transcriber...
[+] Building 38.1s (14/16)                                                                                                                                    
 => [internal] load build definition from Dockerfile                                                                                                     0.0s
 => => transferring dockerfile: 1.07kB                                                                                                                   0.0s
 => [internal] load .dockerignore                                                                                                                        0.0s
 => => transferring context: 35B                                                                                                                         0.0s
 => resolve image config for docker.io/docker/dockerfile:1.4                                                                                             1.1s
 => CACHED docker-image://docker.io/docker/dockerfile:1.4@sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c4e84ced81e24ef1a0dbc                        0.0s
 => [internal] load .dockerignore                                                                                                                        0.0s
 => [internal] load build definition from Dockerfile                                                                                                     0.0s
 => [internal] load metadata for docker.io/library/python:3.8-slim                                                                                       0.8s
 => [internal] load build context                                                                                                                        0.0s
 => => transferring context: 1.66MB                                                                                                                      0.0s
 => [stage-0 1/8] FROM docker.io/library/python:3.8-slim@sha256:9187d27fd8f222a181292f24f8e7d6b22419d46bd9cc4506adbdf2dcfae68a56                         0.0s
 => CACHED [stage-0 2/8] RUN --mount=type=cache,target=/var/cache/apt set -eux; apt-get update -qq; apt-get install -qqy --no-install-recommends curl;   0.0s
 => [stage-0 3/8] COPY .cog/tmp/build2377246762/cog-0.0.1.dev-py3-none-any.whl /tmp/cog-0.0.1.dev-py3-none-any.whl                                       0.0s
[...]

PR #1383, opened 2023-11-17
document how to install git-hosted python packages

zeke:
This PR adds some documention for people who are using git-based Python dependencies. This used to work, but we switched to using a slimmer image that doesn't include git in https://github.com/replicate/cog/pull/1214

Resolves https://github.com/replicate/cog/issues/1381

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -11,7 +11,7 @@ build:
     - pytorch==2.0.1
   system_packages:
     - "ffmpeg"
-    - "libavcodec-dev"
+    - "git"
 predict: "predict.py:Predictor"
 ```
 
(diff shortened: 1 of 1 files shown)


technillogue:
I don't think this is necessary in latest versions of cog

yondonfu:
@technillogue

> I don't think this is necessary in latest versions of cog

FYI I'm on cog v0.9.0-beta11 and I seem to still be running into the issue described in https://github.com/replicate/cog/issues/1381. In my case, I have a `git clone` command under the `run` key in my `cog.yaml` and `cog build` fails when it reaches that line complaining that git could not be found. I added `git` under `system_packages` to fix the issue.

See `cog --version` output below:

```
cog version 0.9.0-beta11 (built 2023-11-28T17:38:26Z)
```

zeke:
Thanks for the heads-up, @yondonfu. Maybe we should just land this documentation as it does apply to some versions of Cog that are out there in the wild, apparently including the latest versions.

PR #1389, opened 2023-11-20
cog predict: check output path writability before predict/train

yorickvP:
`cog train` failed after 30 minutes with "weights: is a directory".
I refactored `predictIndividualInputs` so it checks output file writability if the path is known.


mattt (approved):
Thanks, @yorickvP! Sorry to keep you waiting. Rally glad to make this quality of life improvement.

PR #1391, opened 2023-11-20
cog train: bring functionality on par with cog predict

yorickvP:
- Currently, `cog train` always writes to `weights`
- Currently, there's no way to use it with already existing images
- Some models have a `weights` directory

Together, this makes for a terrible experience. This PR adds a way to specify the output directory and use an existing image.

### DIFF
--- a/pkg/cli/train.go
+++ b/pkg/cli/train.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


anotherjesse:
@yorickvP I haven't run this - but the weights are/were base64 datauris inside a json file!

this is extremely painful

PR #1393, opened 2023-11-21
Fix types, run typechecker in CI

### LINKED ISSUE #1360: Support new python union syntax
In cog 0.9.

This works:
```
        text: Union[str, List[str]] = Input(description="Text string to embed"),
```

this doesn't:
```
        text: str | list[str] = Input(description="Text string to embed"),
```

yorickvP:
I fell into a typing hole while setting up pyright-lsp. Nevertheless, I think these are good changes, mostly preventing some annoying crashes (ast_openapi_scha + bytes defaults, cpus being uncountable).

Fixes  #1360

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,20 @@ disallow_untyped_defs = true
 no_implicit_optional = false
 exclude = ["python/tests/"]
 
+[tool.pyright]
+# TODO: remove this and bring the codebase inline with the current default
+strictParameterNoneValue = false
+# legacy behavior, fixed in PEP688
+disableBytesTypePromotions = true
+include = ["python"]
+exclude = ["python/tests"]
+reportMissingParameterType = "error"
+reportUnknownLambdaType = "error"
+reportUnneccesaryIsInstance = "warning"
+reportUnneccesaryComparison = "warning"
+reportUnnecessaryContains = "warning"
+reportUnusedExpression = "warning"
+
 [tool.setuptools]
 package-dir = { "" = "python" }
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ optional-dependencies = { "dev" = [
   "httpx",
   'hypothesis<6.80.0; python_version < "3.8"',
   'hypothesis; python_version >= "3.8"',
-  "mypy",
   'numpy<1.22.0; python_version < "3.8"',
   'numpy; python_version >= "3.8"',
   "pillow",
@@ -48,13 +47,6 @@ dynamic = ["version"]
 
 [tool.setuptools_scm]
 write_to = "python/cog/_version.py"
-
-[tool.mypy]
-plugins = "pydantic.mypy"
-disallow_untyped_defs = true
-# TODO: remove this and bring the codebase inline with the current mypy default
-no_implicit_optional = false
-exclude = ["python/tests/"]
 
 [tool.pyright]
 # TODO: remove this and bring the codebase inline with the current default
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ reportUnknownLambdaType = "error"
 reportUnneccesaryIsInstance = "warning"
 reportUnneccesaryComparison = "warning"
 reportUnnecessaryContains = "warning"
+reportMissingTypeArgument = "error"
 reportUnusedExpression = "warning"
 
 [tool.setuptools]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,9 +57,9 @@ include = ["python"]
 exclude = ["python/tests"]
 reportMissingParameterType = "error"
 reportUnknownLambdaType = "error"
-reportUnneccesaryIsInstance = "warning"
-reportUnneccesaryComparison = "warning"
-reportUnnecessaryContains = "warning"
+reportUnnecessaryIsInstance = "warning"
+reportUnnecessaryComparison = "warning"
+reportUnneesssaryContains = "warning"
 reportMissingTypeArgument = "error"
 reportUnusedExpression = "warning"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ optional-dependencies = { "dev" = [
   'numpy<1.22.0; python_version < "3.8"',
   'numpy; python_version >= "3.8"',
   "pillow",
+  "pyright",
   "pytest",
   "pytest-asyncio",
   "pytest-httpserver",
--- a/python/cog/command/ast_openapi_schema.py
+++ b/python/cog/command/ast_openapi_schema.py
@@ -1,6 +1,8 @@
 import ast
 import json
 import sys
+import types
+import typing
 from pathlib import Path
 
 try:
@@ -320,7 +322,7 @@ def get_value(node: ast.AST) -> "int | float | complex | str | list":
     if isinstance(node, (ast.List, ast.Tuple)):
         return [get_value(e) for e in node.elts]
     if isinstance(node, ast.UnaryOp) and isinstance(node.op, ast.USub):
-            return -get_value(node.operand)
+            return -typing.cast(int | float | complex, get_value(node.operand))
     raise ValueError("Unexpected node type", type(node))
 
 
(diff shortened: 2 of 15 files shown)


technillogue:
hmm https://github.com/replicate/cog/blob/main/pyproject.toml#L50-L87

yorickvP:
`mypy` finds largely the same errors. But `lsp-pyright` is really good. I can optionally add a  `tool.pyright` section with some config.

extra stuff found by mypy (mostly untyped defs):
```
python/cog/command/ast_openapi_schema.py:9: error: Function "unparse" could always be true in boolean context  [truthy-function]
python/cog/command/ast_openapi_schema.py:12: error: Incompatible types in assignment (expression has type "Callable[[object], str]", variable has type "Callable[[AST], str]")  [assignment]
python/cog/types.py:264: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
python/cog/types.py:268: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
python/cog/server/http.py:77: error: Unsupported dynamic base class "schema.PredictionRequest.with_types"  [misc]
Found 5 errors in 3 files (checked 23 source files)
```

pyproject.toml
```diff
@@ -54,6 +55,21 @@ disallow_untyped_defs = true
 no_implicit_optional = false
 exclude = ["python/tests/"]
 
+[tool.pyright]
```
erbridge, line 58: Should we pick one of pyright or mypy and drop the other?

technillogue: would agree, would kinda lean towards sticking to mypy rather than making changes. pylsp-mypy exists though I assume it's slower

python/cog/command/ast_openapi_schema.py
```diff
@@ -307,8 +309,25 @@ def find(obj: ast.AST, name: str) -> ast.AST:
     """Find a particular named node in a tree"""
     return next(node for node in ast.walk(obj) if getattr(node, "name", "") == name)
 
+if typing.TYPE_CHECKING:
+    AstVal: "typing.TypeAlias" = int | float | complex | str | list["AstVal"] | bytes | None
+    AstValNoBytes: "typing.TypeAlias" = int | float | str | list["AstValNoBytes"]
+    JSONObject: "typing.TypeAlias" = int | float | str | list["JSONObject"] | "JSONDict" | None
+    JSONDict: "typing.TypeAlias" = dict[str, "JSONObject"]
+
```
technillogue, line 317: I think this breaks on 3.8

python/cog/command/ast_openapi_schema.py
```diff
@@ -307,8 +309,25 @@ def find(obj: ast.AST, name: str) -> ast.AST:
...
+
+def toSerializable(val: "AstVal") -> "JSONObject":
```
technillogue, line 319: it's python, no camelcase

python/cog/command/ast_openapi_schema.py
technillogue: also, wow, a true JSONDict type, intense, I had always given up on doing recursive types like this in the past

python/cog/server/http.py
```diff
@@ -74,16 +86,20 @@ def create_app(
     InputType = get_input_type(predictor)
     OutputType = get_output_type(predictor)
 
-    PredictionRequest = schema.PredictionRequest.with_types(input_type=InputType)
+    class PredictionRequest(schema.PredictionRequest.with_types(input_type=InputType)):
+        pass
```
technillogue, line 90: why?

python/cog/server/http.py
```diff
@@ -74,16 +86,20 @@ def create_app(
...
     http_semaphore = asyncio.Semaphore(threads)
 
-    def limited(f: Callable) -> Callable:
+    if TYPE_CHECKING:
+        P = ParamSpec("P")
+        T = TypeVar("T")
+    def limited(f: "Callable[P, Awaitable[T]]") -> "Callable[P, Awaitable[T]]":
         @functools.wraps(f)
-        async def wrapped(*args: Any, **kwargs: Any) -> Any:
+        async def wrapped(*args: "P.args", **kwargs: "P.kwargs") -> "T":
```
technillogue, line 102: :pray:

python/cog/server/http.py
```diff
@@ -359,12 +375,15 @@ def _signal_set_event(signum: Any, frame: Any) -> None:
...
-    threads = args.threads
+    threads: Optional[int] = args.threads
     if threads is None:
         if config.get("build", {}).get("gpu", False):
             threads = 1
         else:
             threads = os.cpu_count()
+            if threads is None:
+                log.warn("Unable to determine cpu count, defaulting to 1 thread")
+                threads = 1
```
technillogue, line 386: in general os.sched_getaffinity(0) would be better for this but it doesn't actually matter in the slightest because this isn't used in practice and doesn't allow any degree of concurrency

python/cog/server/http.py
yorickvP: tldr this allows typingchecking PredictionRequest usage by pyright (still not mypy, though)

nickstenning:
Love it!

I'd like to request that if we go all in on `pyright` (which I fully support) then let's also get rid of all `mypy` references in the codebase (`pyproject.toml`/`Makefile`/etc.)

nickstenning (approved):
Modulo failing tests (I think we'll have to use "old" ways of defining union types for a while to support <3.10) this seems great, particularly the bit where we typecheck in CI!

pyproject.toml
```diff
@@ -55,6 +56,21 @@ disallow_untyped_defs = true
...
+# TODO: remove this and bring the codebase inline with the current default
+strictParameterNoneValue = false
+# legacy behavior, fixed in PEP688
+disableBytesTypePromotions = true
+include = ["python"]
+exclude = ["python/tests"]
+reportMissingParameterType = "error"
+reportUnknownLambdaType = "error"
+reportUnneccesaryIsInstance = "warning"
+reportUnneccesaryComparison = "warning"
```
nickstenning, line 69:
```suggestion
reportUnnecessaryIsInstance = "warning"
reportUnnecessaryComparison = "warning"
```
?

pyproject.toml
technillogue: ...I've had check-untyped-defs on the entire time I've used mypy. but I have to agree that this is fairly compelling even if the javascript stuff is weird.

python/cog/server/http.py
```diff
@@ -47,7 +59,7 @@ class Health(Enum):
 
 def create_app(
     config: Dict[str, Any],
-    shutdown_event: Optional[threading.Event],
+    shutdown_event: Optional[asyncio.Event],
```
technillogue, line 62:
```suggestion
    shutdown_event: Optional[threading.Event],
```

python/cog/server/http.py
```diff
@@ -312,7 +328,7 @@ def signal_ignore(signum: Any, frame: Any) -> None:
     log.warn("Got a signal to exit, ignoring it...", signal=signal.Signals(signum).name)
 
 
-def signal_set_event(event: threading.Event) -> Callable:
+def signal_set_event(event: threading.Event | asyncio.Event) -> Callable[[Any, Any], None]:
```
technillogue, line 331:
```suggestion
def signal_set_event(event: threading.Event) -> Callable[[Any, Any], None]:
```

python/cog/server/http.py
```diff
@@ -361,14 +377,14 @@ def _signal_set_event(signum: Any, frame: Any) -> None:
...
+    threads: Optional[int] = args.threads
     if threads is None:
         if config.get("build", {}).get("gpu", False):
             threads = 1
         else:
-            threads = os.cpu_count()
+            threads = max(1, len(os.sched_getaffinity(0)))
 
-    shutdown_event = threading.Event()
+    shutdown_event = asyncio.Event()
```
technillogue, line 387:
```suggestion
    shutdown_event = threading.Event()
```

python/cog/server/http.py
```diff
@@ -400,7 +416,7 @@ def _signal_set_event(signum: Any, frame: Any) -> None:
     s.start()
 
     try:
-        shutdown_event.wait()
+        asyncio.run(shutdown_event.wait())
```
technillogue, line 419:
```suggestion
        shutdown_event.wait()
```

unless this was thoroughly tested

technillogue (commented):
this PR probably shouldn't change async functionality

pyproject.toml
yorickvP: Nice catch!

pyproject.toml
mattt: @yorickvP Thanks for the nudge. I just gave Pyright a look, and found it to be an improvement over mypy. I'd be very happy for us to adopt it across the board. (To that end, I just opened [this PR](https://github.com/replicate/replicate-python/pull/206) in our Python client)

mattt (approved):

technillogue (approved):

PR #1396, opened 2023-11-21
update compatibility matrices: support cuda 12.2

technillogue:


yorickvP:
This removes the suffixes from torch 2.0.0, 2.0.1, 2.1.0. Not sure why.
There's also a fun hack here https://github.com/replicate/cog/blob/main/pkg/config/compatibility.go#L359-L367

technillogue:
hm, I am not sure what to do with that, I just ran the compatgen script. getting this out would probably be helpful for H100 models

yorickvP:
**Suffix thing is fine:**
Torch 2.0.0, 2.0.1, 2.1.0 now go through the `parsePreviousTorchVersionsCode` codepath, which parses https://pytorch.org/get-started/previous-versions/ . Looks like they switched to having `pip` look up the compatible versions in https://github.com/pytorch/pytorch.github.io/commit/b6cbb189e221b0d3ca6a8b378c2be28392a7af04, which should work.

(side-note: it would be way nicer to parse https://conda.anaconda.org/pytorch/linux-64/repodata.json instead, but that might require some constraint solving)

**Failing test**
This switches the default torch to 2.1.1 and the default cuda to 12.2. Do coreweave and GCP support that for older GPUs? If not, we should probably override it back to 2.1.1+cu118 or something.

technillogue:
we should probably keep 11.8 as the default

yorickvP:
Okay, set the default to 11.8

yorickvP (approved):

mattt (approved):

Issue #1349, opened 2023-10-24
Document `cog debug` command


zeke:
Cog has a `debug` command that outputs the Dockerfile for the current cog model, but it's not documented.

```
cog debug > Dockerfile
```

It's not available in help:

```
$ cog --help
Containers for machine learning.
...
Available Commands:
  build       Build an image from cog.yaml
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  init        Configure your project for use with Cog
  login       Log in to Replicate Docker registry
  predict     Run a prediction
  push        Build and push model in current directory to a Docker registry
  run         Run a command inside a Docker environment
```

This command is mentioned in passing in the [getting started guide](https://github.com/replicate/cog/blob/main/docs/getting-started.md#build-an-image), but is not present in any reference documentation.

See also:

- https://github.com/replicate/cog/issues/433

PR #1398, opened 2023-11-21
[#1349] Show information about `cog debug` in output of `cog help`

vicfryzel:
Minor; one-liner: Unhide CLI documentation about `cog debug` when running `cog help|-h`.

### DIFF
--- a/pkg/cli/debug.go
+++ b/pkg/cli/debug.go
@@ -15,9 +15,9 @@ var imageName string
 
 func newDebugCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:    "debug",
-		Short:  "Generate a Dockerfile from " + global.ConfigFilename,
-		RunE:   cmdDockerfile,
+		Use:   "debug",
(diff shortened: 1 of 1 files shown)


vicfryzel:
🤦 fix ETA 3-4 hours when I get to a workstation

bfirsh:
Hello Vic! 👋

I think the original concern was people would start relying on this in build pipelines and things, but we don’t make any guarantees to keep dockerfiles in a particular format.

I guess "debug" makes it reasonably clear that this is just a debugging tool. No strong feelings...

vicfryzel:
Hi! I get the concern about making an implicit guarantee of dockerfile formats. My initial view was that all commands should be clearly and openly documented, or have a verbose flag show it as part of the help command. This way, devs find what they need easily and officially.

That said, I also don't feel strongly. Not listing this command underlines that we don't guarantee its consistency.

I'm going to close this, with a decision to leave `cog debug` hidden from `cog help` for now. If anyone does feel strongly, please join in 😁 Thanks Ben

PR #1404, opened 2023-11-29
use a more realistic registry URL in cog push usage

zeke:
It's nice that Cog can push to any Docker registry, so it makes sense that this could be "any URL", but most users will be pushing to Replicate's model registry. This change updates the usage to give more of a hint of what kind of URL they should be constructing.


evilstreak (approved):
Looks good! The commit just needs signing.

PR #1405, opened 2023-11-29
Mux prediction events

technillogue:
A critical part of concurrent predictions is multiplexing several prediction outputs over the same pipe. This takes a stab at that. Once this is done, we might be able to drop some parts of runner.

We tag each _PublicEventType with a prediction id, introduce a `Mux`, and have a `_read_events` task responsible for reading events from the pipe and writing them to the mux. The mux adds it to the right queue, and then the places that previously called `_wait` instead call `Mux.read`. 

We also add a semaphore and keep track of predictions in flight. READY is renamed to IDLE, but that may need to be reworked further. 

---

Some challenges
* contextvar to tag logs that were emitted from inside a predict() 
  * however, logs emitted from cross-prediction stuff (like the actual batching code) have to be discarded to not leak information
* aioprocessing uses a threadpool and "does _not_ re-implement multiprocessing using asynchronous I/O". it hangs at shutdown. it's still useful for getting the rest of the code in shape for now. 
  * aioprocessing itself, especially the part we use, is extremely small (730 loc) so we could vendor/fix it if we wanted to. 
  * https://github.com/kchmck/aiopipe/tree/master is closest to what we would need, but looks a little awkward. 
  * I also have an example that uses loop.connect_read_pipe correctly but it's a little wordy and would take some hacking to be suitable for duplex use
* I don't know what to do with the hypothesis tests

---
[x] mux events
[x] doesn't deadlock
[x] hypothesis tests mostly pass
[ ] serious pipe implementation (future PR?)
[ ] cancellation
[x] READY /  PROCESSING semaphore 
[~] route predict logs to prediction if only one prediction is running

### DIFF
--- a/python/cog/server/eventtypes.py
+++ b/python/cog/server/eventtypes.py
@@ -1,3 +1,4 @@
+import secrets
 from typing import Any, Dict
 
 from attrs import define, field, validators
@@ -8,6 +9,7 @@ from attrs import define, field, validators
 @define
 class PredictionInput:
     payload: Dict[str, Any]
+    id: str = field(factory=lambda: secrets.token_hex(4))
 
 
 @define
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -196,8 +196,11 @@ async def race(
 # functionally this is the exact same thing as aioprocessing but 0.1% the code
 # however it's still worse than just using actual asynchronous io
 class AsyncPipe(Generic[X]):
-    def __init__(self, conn: Connection) -> None:
+    def __init__(
+        self, conn: Connection, alive: Callable[[], bool] = lambda: True
+    ) -> None:
         self.conn = conn
+        self.alive = alive
         self.exiting = threading.Event()
         self.executor = concurrent.futures.ThreadPoolExecutor(1)
 
@@ -216,8 +219,11 @@ class AsyncPipe(Generic[X]):
     def _recv(self) -> Optional[X]:
         # this ugly mess could easily be avoided with loop.connect_read_pipe
         # even loop.add_reader would help but we don't want to mess with a thread-local loop
-        while not self.exiting.is_set():
+        while not self.exiting.is_set() and not self.conn.closed and self.alive():
             if self.conn.poll(0.01):
+                if self.conn.closed or not self.alive():
+                    print("caught conn closed or unalive")
+                    return
                 return self.conn.recv()
         return None
 
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -3,17 +3,20 @@ import concurrent.futures
 import io
 import os
 import selectors
+import sys
 import threading
 import uuid
 from multiprocessing.connection import Connection
 from typing import (
     Any,
     Callable,
+    Coroutine,
     Generic,
     Optional,
     Sequence,
     TextIO,
     TypeVar,
+    Union,
 )
 
 
@@ -160,7 +163,35 @@ class StreamRedirector(threading.Thread):
                         self.drain_event.set()
                         drain_tokens_seen = 0
 
+
 X = TypeVar("X")
+Y = TypeVar("Y")
+
+
+async def race(
+    x: Coroutine[None, None, X],
+    y: Coroutine[None, None, Y],
+    timeout: Optional[float] = None,
+) -> Union[X, Y]:
+    tasks = [asyncio.create_task(x), asyncio.create_task(y)]
+    wait = asyncio.wait(tasks, timeout=timeout, return_when=asyncio.FIRST_COMPLETED)
+    done, pending = await wait
+    for task in pending:
+        task.cancel()
+    if not done:
+        raise TimeoutError
+    # done is an unordered set but we want to preserve original order
+    result_task, *others = (t for t in tasks if t in done)
+    # during shutdown, some of the other completed tasks might be an error
+    # cancel them instead of handling the error to avoid the warning
+    # "Task exception was never retrieved"
+    for task in others:
+        msg = "was completed at the same time as another selected task, canceling"
+        # FIXME: ues a logger?
+        print(task, msg, file=sys.stderr)
+        task.cancel()
+    return result_task.result()
+
 
 # functionally this is the exact same thing as aioprocessing but 0.1% the code
 # however it's still worse than just using actual asynchronous io
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -206,7 +206,7 @@ class AsyncPipe(Generic[X]):
 
     def shutdown(self) -> None:
         self.exiting.set()
-        self.executor.shutdown(wait=False)
+        self.executor.shutdown(wait=True)
         # if we ever need cancel_futures (introduced 3.9), we can copy it in from
         # https://github.com/python/cpython/blob/3.11/Lib/concurrent/futures/thread.py#L216-L235
 
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -225,3 +225,8 @@ class AsyncPipe(Generic[X]):
         loop = asyncio.get_running_loop()
         # believe it or not this can still deadlock!
         return await loop.run_in_executor(self.executor, self._recv)
+
+    async def coro_recv_with_exit(self, exit: asyncio.Event) -> Optional[X]:
+        result = await race(self.coro_recv(), exit.wait())
+        if result is not True:  # wait() would return True
+            return result
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -157,7 +157,7 @@ class Worker:
         return self._state not in {WorkerState.PROCESSING, WorkerState.IDLE}
 
     @contextlib.asynccontextmanager
-    async def prediction_ctx(self, input: PredictionInput) -> AsyncIterator[None]:
+    async def _prediction_ctx(self, input: PredictionInput) -> AsyncIterator[None]:
         async with self._semaphore:
             self._predictions_in_flight.add(input.id)  # idempotent ig
             self._state = self.state_from_predictions_in_flight()
@@ -184,7 +184,7 @@ class Worker:
         self._state = self.state_from_predictions_in_flight()
 
         async def inner() -> AsyncIterator[_PublicEventType]:
-            async with self.prediction_ctx(input):
+            async with self._prediction_ctx(input):
                 self._events.send(input)
                 async for event in self._mux.read(input.id, poll=poll):
                     yield event
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -46,7 +46,7 @@ _PublicEventType = Union[Done, Heartbeat, Log, PredictionOutput, PredictionOutpu
 class WorkerState(Enum):
     NEW = auto()
     STARTING = auto()
-    READY = auto()
+    IDLE = auto()
     PROCESSING = auto()
     DEFUNCT = auto()
 
@@ -87,9 +87,13 @@ class Mux:
 
 
 class Worker:
-    def __init__(self, predictor_ref: str, tee_output: bool = True) -> None:
+    def __init__(
+        self, predictor_ref: str, tee_output: bool = True, concurrent: int = 1
+    ) -> None:
         self._state = WorkerState.NEW
         self._allow_cancel = False
+        self._semaphore = asyncio.Semaphore(concurrent)
+        self._concurrent = concurrent
 
         # A pipe with which to communicate with the child worker.
         events, child_events = _spawn.Pipe()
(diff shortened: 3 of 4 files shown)


python/cog/server/helpers.py
```diff
@@ -148,3 +162,98 @@ def run(self) -> None:
...
+    ...
+
+
+@overload
+async def select(*awaitables: Coro, timeout: Optional[float] = None) -> Any:
+    ...
+
+
+async def select(
+    x: Coro, y: Coro, *awaitables: Coro, timeout: Optional[float] = None
```
yorickvP, line 188:
```suggestion
    x: Coro, y: Coro, *awaitables: List[Coro], timeout: Optional[float] = None
```
?

technillogue: not how variadics work, the list is implicit (also it would be a tuple and not a list)

technillogue: you can tell because the typechecker passes

technillogue:
outstanding questions:

* how to handle cancellation? it's hard to rely on the behavior of raising exceptions from signal handlers if there's an event loop running -- the exception could be raised in any coroutine or the event loop code instead of specifically inside predict. canceling tasks works, but the asyncio.CancelledError only gets raised on the next await and cannot happen inside blocking C code. my best guess is some combination of a new Cancel event and keeping SIGUSR1. 
* we probably need to have a mapping from prediction_id as used by the cancel endpoint to id as used by worker. why not use the same id for both?
* I'm a little confused why read_setup_events/read_predict_events were separate hypothesis rules that had an argument. I've removed it for now and it works, but I don't understand why it was there in the first place
* how do we want to test this stuff? Can we parameterize hypothesis with a few different predictors?
* even without cancellation this is pretty big PR, I would love suggestions for how to break it up into smaller chunks

python/tests/server/test_worker.py
```diff
@@ -465,42 +498,57 @@ def __init__(self):
...
         time.sleep(sleep)
 
+    @precondition(lambda x: not x.setup_done)
     @rule()
     def setup(self):
+        trace("did setup")
         try:
             self.setup_generator = self.worker.setup()
             self.setup_events = []
+        # this except allows setup to be called many times which seems weird
```
nickstenning, line 517: The point of the Hypothesis test is to ensure that calling the methods in any order "works," where "works" might mean throwing `InvalidStateException` because it's not a reasonable time to call that method. Adding a `precondition` and removing this removes that validation.

python/tests/server/test_worker.py
```diff
@@ -465,42 +498,57 @@ def __init__(self):
...
+        # hypothesis would call setup repeatedly, just now the InvalidStateException
+        # comes up when you actually use the iterator
+
+        # @precondition(lambda x: x.setup_generator)
+        # @rule(n=st.integers(min_value=1, max_value=10))
+        # def read_setup_events(self, n):
         try:
-            for _ in range(n):
-                event = next(self.setup_generator)
+            while 1:  # we want all the events, not just "n" events
```
nickstenning, line 529: Again, this seems confused. The whole point of this test is that this might be called multiple times with different values of `n` with other calls in between, and the test ensures that the `Worker` remains in a consistent state throughout.

python/tests/server/test_worker.py
```diff
@@ -465,42 +498,57 @@ def __init__(self):
...
-        print(self.setup_events)
-        result = _process(self.setup_events)
+        trace(self.setup_events)
+        result = await _process(self.setup_events)
         assert result.stdout == "did setup\n"
         assert result.stderr == ""
         assert result.done == Done()
+        self.setup_done = True
 
+    @precondition(lambda x: x.setup_done)
```
nickstenning, line 548: Again, this isn't right. By adding this precondition you're removing branches from the tree. Creating a predictor and calling `predict` before `setup` is a thing that might happen. Without this precondition we're checking that the thing behaves "correctly" if that occurs -- i.e. it throws `InvalidStateException`.

python/cog/server/worker.py
```diff
@@ -179,6 +306,8 @@ def run(self) -> None:
         ws_stdout.wrap()
         ws_stderr.wrap()
 
+        # using a thread for this can potentially cause a deadlock
+        # however, if we made this async, we might interfere with a user's event loop
```
yorickvP, line 310: My earlier idea was to run `non-async def predict` code in a thread without an event loop. That way we don't break any legacy event loops. That won't make it easier to kill though.

python/tests/server/test_worker.py
technillogue: is there some actual point to testing multiple setup() like that? the state error only happens when you try to consume the stream. if you call setup or predict multiple times without actually consuming the generator there's no reason it should work (it will never get to the assertion)

python/cog/server/worker.py
```diff
@@ -111,43 +229,52 @@ def _assert_state(self, state: WorkerState) -> None:
...
+            try:
+                # this can still be running when the task is destroyed
+                id, event = await self._events.coro_recv_with_exit(self._terminating)
+                trace("recv", event)
+            except asyncio.CancelledError:
+                return
+            if id == "LOG" and "SETUP" in self._mux.outs:
+                id = "SETUP"
+            if id == "LOG" and len(self._predictions_in_flight) == 1:
+                id = list(self._predictions_in_flight)[0]
```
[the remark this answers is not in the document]
technillogue: some logs can have a .source if there's a suitable contextvar, but there are also logs for the entire batch that need to be discarded to avoid leaking information about other users' predictions

python/cog/server/eventtypes.py
```diff
@@ -45,3 +45,7 @@ class Done:
 @define
 class Heartbeat:
     pass
+
+@define
+class FatalError:
+    message: str
```
technillogue, line 51: this is unnecessary

python/cog/server/worker.py
```diff
@@ -44,44 +46,152 @@
...
+                        )
+                    self._state = WorkerState.IDLE
+                    self._allow_cancel = False
+
+        return inner()
+
+    @contextlib.asynccontextmanager
+    async def prediction_ctx(self, input: PredictionInput) -> AsyncIterator[None]:
+        async with self._semaphore:
+            self._state = WorkerState.PROCESSING
```
technillogue, line 158: something we could do here like 

```suggestion
            self._state = WorkerState.PROCESSING
            if self._semaphore._value == 0:
                self._state = WorkerState.BUSY
            else:
                self._state = WorkerState.READY
```

python/cog/server/worker.py
```diff
@@ -269,34 +405,42 @@ def _handle_predict_error(self) -> Iterator[None]:
         finally:
             self._cancelable = False
```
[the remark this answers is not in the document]
technillogue: hm, it isn't, but I think the plan is to fix cancellation Later :tm:

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+                "cannot accept new predictions because shutdown requested"
+            )
         self._state = WorkerState.PROCESSING
-        self._allow_cancel = True
-        self._events.send(PredictionInput(payload=payload))
 
-        return self._wait(poll=poll)
+        async def inner() -> AsyncIterator[_PublicEventType]:
+            input = PredictionInput(payload=payload)
+            async with self.prediction_ctx(input):
```
technillogue, line 195: @nickstenning, supporting hypothesis like this kind of ruins the semantics of the semaphore. do you have any suggestions?

python/cog/server/runner.py
```diff
@@ -66,6 +66,13 @@ def __init__(
         self._shutdown_event = shutdown_event
         self._upload_url = upload_url
 
+    # @contextlib.contextmanager
+    # def ctx(self) -> Iterator["PredictionRunner"]:
+    #     with Worker(predictor_ref=self.predictor_ref) as self._worker:
+    #         yield self
+    #         if self._result:
+    #             self._result.cancel()
+
```
mattt, line 75:
```suggestion
```

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+#         @functools.wraps(asyncgen)
+#         def inner(self: "Worker", *args: P.args) -> AsyncIterator[T]:
+#             if self._state != old:
+#                 raise InvalidStateException(
+#                     f"Invalid operation: state is {self._state} (must be {old})"
+#                 )
+#             self._state = new
+#             return asyncgen(*args)
+
+
```
mattt, line 104:
```suggestion
```

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+    # @contextlib.contextmanager
+    # def ctx(self):
+    #     self._events = AsyncPipe(events)
+    #     with self._events.executor:
+    #         self._child = _ChildWorker(predictor_ref, child_events, tee_output)
+    #         yield self
+    #         if self._child.is_alive():
+    #             self._child.terminate()
+    #             self._child.join()
```
mattt, line 136:
```suggestion
```

mattt (approved):
Amazing to see all tests passing for this. I'm sure we can find ways to clean up / simplify this later, but this is good enough to unblock us for now. Nice work!

python/cog/server/worker.py
nickstenning: We've chatted about this since you left this comment, but just to reiterate, I don't care about "supporting Hypothesis," only about verifying (with Hypothesis or otherwise) that expected things happen if you call the various public methods of `Worker` at any point.

I'm not quite sure what you mean about ruining the semantics of the semaphore? If you want a preemptive check of the state of the semaphore then `sem.locked()` exists, but also as this part of the code is not actually multithreaded I do wonder whether a plain old integer on `Worker` would also work just fine.

python/cog/server/worker.py
```diff
@@ -90,21 +212,26 @@ def terminate(self) -> None:
...
         if self._child.is_alive():
             self._child.terminate()
             self._child.join()
         self._events.shutdown()
+        if self._read_events_task:
+            self._read_events_task.cancel()
 
+    # FIXME: this will need to use a combination
+    # of signals and Cancel events on the pipe
     def cancel(self) -> None:
```
nickstenning, line 205: Does it even make sense for `cancel` to be a method of `Worker` any more?

nickstenning: We discussed this and the answer is "no but that's out of scope for this PR" which I think is fine.

nickstenning (approved):
Generally this is looking pretty good. Some of the stuff around `select` I found a bit confusing, and I've left a few suggestions there, but the meat of the changes to `Worker`/`ChildWorker` look good. I still think we have more to work out w.r.t. `WorkerState`, semaphore, etc.

A more general point is that `Mux` itself could ideally do with some -- at least cursory -- docstrings and tests.

python/cog/server/worker.py
technillogue: I've figured this out to my satisfaction in this PR. #1499 addresses this in further depth and clarifies the distinction between predictions_in_flight (updated as soon as work is accepted) and the semaphore (updated once the work is started). the semaphore is the only part that makes it an async context manager; a sync context manager is sufficient for the eager part of the state management

python/cog/server/helpers.py
```diff
@@ -160,13 +163,56 @@ def run(self) -> None:
...
+        task.cancel()
+    if not done:
+        raise TimeoutError
+    # done is an unorded set but we want to preserve original order
+    result_task, *others = (t for t in tasks if t in done)
+    # during shutdown, some of the other completed tasks might be an error
+    # cancel them instead of handling the error to avoid the warning
+    # "Task exception was never retrieved"
+    for task in others:
+        print(task, "was completed simultaniously as another selected task, canceling")
```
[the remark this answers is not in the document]
technillogue: any guidance on what logger to use? this message should almost never come up

python/cog/server/helpers.py
```diff
@@ -160,13 +163,56 @@ def run(self) -> None:
...
+        task.cancel()
+    return result_task.result()
```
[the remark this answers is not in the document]
technillogue: you're completely right, I'll drop the general case

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+        while 1:
+            try:
+                event = await select(
+                    self.outs[id].get(), self.terminating.wait(), timeout=poll
+                )
+            except TimeoutError:
+                if send_heartbeats:
+                    yield Heartbeat()
+                continue
+            if event is True:
```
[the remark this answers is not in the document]
technillogue: yeah, I could just use asyncio.wait directly instead of race if you think it'd be better. this seems almost adequate and the type checker can easily verify correctness so I would be inclined to leave it as it is, but maybe it's enough a footgun?

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+                    self._state = WorkerState.IDLE
+
+        return inner()
+
+    @contextlib.asynccontextmanager
+    async def prediction_ctx(self, input: PredictionInput) -> AsyncIterator[None]:
+        async with self._semaphore:
+            if self._semaphore._value == 0:
+                # maybe this will fix hypothesis
+                self._state = WorkerState.BUSY
```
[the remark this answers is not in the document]
technillogue: BUSY is when we can't accept more work; PROCESSING and IDLE are when we can; PROCESSING and BUSY are when we can cancel predictions. I've cherry-picked a clearer way to of doing this from #1499

python/cog/server/worker.py
```diff
@@ -44,44 +46,164 @@
...
+            self._allow_cancel = True
+            self._predictions_in_flight.add(input.id)
+            try:
+                yield
+            finally:
+                self._predictions_in_flight.remove(input.id)
+        if self._semaphore._value == self._concurrent:
```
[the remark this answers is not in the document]
technillogue: hmm, I'd be fine with that. there's a new approach with moderately less hoop-jumping. this setup seems more legible to hypothesis without being too ugly, though I don't yet know what the final picture will look like for testing (and testing concurrency). the new approach makes predictions_in_flight: set[str] the source of truth and derives state from that. this seems like it could go either way.

PR #1410, opened 2023-12-06
minimal async worker

technillogue:
since #1405 is gnarly, this is trying to pull out the basic make stuff async changes without changing logic

if the pipe is not async, the event loop grinds to a halt after a prediction is started and doesn't respond to cancellations or idempotent predictions. everything else does work


python/cog/server/http.py
```diff
@@ -195,7 +197,10 @@ async def predict_idempotent(
                     )
                 ]
             )
-
+        if runner.is_busy():
```
nickstenning: Why did this get added? Isn't this already handled by `_predict`?

python/cog/types.py
```diff
@@ -200,11 +200,7 @@ def __repr__(self) -> str:
                 type(self).__name__, id(self), object.__getattribute__(self, "__url__")
             )
         else:
-            return "<{} at 0x{:x} wrapping {!r}>".format(
-                type(self).__name__,
-                id(self),
-                target,
-            )
+            return f"<{type(self).__name__} at 0x{id(self):x} wrapping {target!r}>"
```
nickstenning, line 203: Was this done automatically by a formatter? Either way, it's probably not great to have unrelated changes in a file that isn't otherwise changed by the commit.

python/tests/server/test_worker.py
```diff
@@ -103,16 +115,22 @@ class Result:
...
+
+async def _process(events, swallow_exceptions=False):
     """
     Helper function to collect events generated by Worker during tests.
     """
     result = Result()
     stdout = []
     stderr = []
-
+    if not inspect.isasyncgen(events):
```
nickstenning, line 130: I couldn't find anywhere that `_process` is called with a non-async generator. Do we need this and `make_generator_async`?

python/cog/server/worker.py
```diff
@@ -55,26 +55,29 @@ def __init__(self, predictor_ref: str, tee_output: bool = True) -> None:
...
 
         # A pipe with which to communicate with the child worker.
-        self._events, child_events = _spawn.Pipe()
+        events, child_events = _spawn.Pipe()
+        self._events = AsyncPipe(events)
         self._child = _ChildWorker(predictor_ref, child_events, tee_output)
         self._terminating = False
 
-    def setup(self) -> Iterator[_PublicEventType]:
+    async def setup(self) -> AsyncIterator[_PublicEventType]:
```
nickstenning, line 63:
```suggestion
    def setup(self) -> AsyncIterator[_PublicEventType]:
```

python/cog/server/worker.py
```diff
@@ -55,26 +55,29 @@ def __init__(self, predictor_ref: str, tee_output: bool = True) -> None:
...
         self._assert_state(WorkerState.NEW)
         self._state = WorkerState.STARTING
         self._child.start()
 
-        return self._wait(raise_on_error="Predictor errored during setup")
+        async for e in self._wait(raise_on_error="Predictor errored during setup"):
+            yield e
```
nickstenning, line 69:
```suggestion
        return self._wait(raise_on_error="Predictor errored during setup")
```

python/cog/server/worker.py
```diff
@@ -55,26 +55,29 @@ def __init__(self, predictor_ref: str, tee_output: bool = True) -> None:
...
-    ) -> Iterator[_PublicEventType]:
+    ) -> AsyncIterator[_PublicEventType]:
         self._assert_state(WorkerState.READY)
         self._state = WorkerState.PROCESSING
         self._allow_cancel = True
         self._events.send(PredictionInput(payload=payload))
 
-        return self._wait(poll=poll)
+        async for e in self._wait(poll=poll):
+            yield e
```
nickstenning, line 80:
```suggestion
        return self._wait(poll=poll)
```

python/tests/server/test_worker.py
```diff
@@ -475,57 +510,64 @@ def setup(self):
         try:
             self.setup_generator = self.worker.setup()
             self.setup_events = []
+            # read one event so that we raise InvalidStateException
+            event = self.await_(anext(self.setup_generator))
+            self.setup_events.append(event)
```
nickstenning, line 515: If you apply my suggestions from `worker.py` and make the signature of `setup` and `predict` a non-async function returning an `AsyncIterator` directly, this hoop-jumping isn't necessary.

python/tests/server/test_worker.py
nickstenning: Oh, I found it, in the Hypothesis test. Seems to me like it would be simpler and less surprising to define an `_aprocess` and use that where appropriate:

```python
async def _aprocess(events, swallow_exceptions=False):
    events = []
    async for event in events:
        events.append(event)
    return _process(events, swallow_exceptions=swallow_exceptions)
```

python/tests/server/test_worker.py
```diff
@@ -475,57 +510,64 @@ def setup(self):
...
+        except StopAsyncIteration:
             self.setup_generator = None
+            self.await_(self._check_setup_events())
+        # except InvalidStateException:
+        #     pass
 
-            self._check_setup_events()
-
-    def _check_setup_events(self):
+    async def _check_setup_events(self):
```
nickstenning, line 532: if you leave a non-async `_process` available to yourself (as suggested above) you don't need to do this either.

python/tests/server/test_worker.py
```diff
@@ -552,10 +594,9 @@ def cancel(self, r):
 
     def teardown(self):
         self.worker.shutdown()
-        # cheat a bit to make sure we drain the events pipe
-        list(self.worker._wait())
         # really make sure everything is shut down and cleaned up
         self.worker.terminate()
+        self.await_(asyncio.sleep(0.01))
```
nickstenning, line 599: Might be worth a comment to explain what this is doing.

python/cog/server/http.py
```diff
@@ -195,9 +197,6 @@ async def predict_idempotent(
                     )
                 ]
             )
-
-        # We've already checked that the IDs match, now ensure that an ID is
-        # set on the prediction object
```
nickstenning, line 200: Any particular reason you've removed this comment?

nickstenning: Oh I see it got removed when commit 1 added something, and then commit 2 removed it again but took this comment with it. Maybe rebase all of those changes out?

python/cog/server/helpers.py
```diff
@@ -148,3 +159,42 @@ def run(self) -> None:
                     if drain_tokens_seen >= drain_tokens_needed:
                         self.drain_event.set()
                         drain_tokens_seen = 0
+
+X = TypeVar("X")
+
+# functionally this is the exact same thing as aioprocessing but 0.1% the code
+# however it's still worse than just using actual asynchronous io
```
nickstenning, line 166: What do you mean by "actual asynchronous io"? And if it's better, why don't we do it instead?

python/tests/server/test_worker.py
technillogue: something like that was the first thing I tried, that won't catch exceptions correctly

python/tests/server/test_worker.py
technillogue: they'll need to be async functions eventually, I feel like making sync functions that are actually an async generator needlessly confuses the actual code for the sake of the tests

python/tests/test_json.py
```diff
@@ -10,7 +10,7 @@
 
 def test_make_encodeable_recursively_encodes_tuples():
     result = make_encodeable((np.float32(0.1), np.float32(0.2)))
-    assert type(result[0]) == float
+    assert isinstance(result[0], float)
```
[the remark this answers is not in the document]
technillogue: similarly ruff check --fix

I checked, np.float32 is not a subclass of float

python/tests/server/test_worker.py
```diff
@@ -475,57 +510,64 @@ def setup(self):
...
             for _ in range(n):
-                event = next(self.setup_generator)
+                event = self.await_(anext(self.setup_generator))
                 self.setup_events.append(event)
-        except StopIteration:
+        except StopAsyncIteration:
             self.setup_generator = None
+            self.await_(self._check_setup_events())
+        # except InvalidStateException:
+        #     pass
```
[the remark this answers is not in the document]
technillogue: it shouldn't be if you read the iterator

python/cog/server/http.py
technillogue: mistake

python/cog/server/helpers.py
technillogue: I mean loop.connect_read_pipe/connect_write_pipe, so that we're actually epolling the pipe instead of blocking

it would require setting up two pipes so that it's duplex, pickling, and a little more clarity than I have about how to correctly pass stuff into the child. previously when I did this I just use asyncio.create_subprocess_exec and passed stuff explicitly; the automatic stuff confuses me

python/tests/test_json.py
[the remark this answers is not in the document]
nickstenning: I have no problem with fixing files you're working on, but let's leave otherwise untouched files out of the equation.

python/tests/server/test_worker.py
nickstenning: It's not just for the sake of the tests. There is a meaningful distinction between a function which does some stuff synchronously and then returns an async iterator, and an async generator function.

The concept of a non-async function returning a `Task` is commonplace in C#, for example. Seems fine. If they eventually need to become async functions, they can eventually become async functions. But let's cross that bridge when we need to and not before.

python/tests/server/test_worker.py
technillogue: specifically, if you do it that way, swallow_exceptions will not be respected; if you instead put the try/catch block around the `[e async for e in events]`, then you miss the events leading up to the exception

I would consider moving make_generator_async to the part of hypothesis that passes in a list though

python/tests/test_json.py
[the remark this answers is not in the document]
technillogue: it don't think it would pass CI though

python/tests/server/test_worker.py
technillogue: one of the goals for this PR is to make #1405 smaller by getting the simple changes out of the way. predict necessarily has to be async to use the semaphore. making it async requires relatively straightforward changes to the tests. it's easier to make those changes here

as far as I can tell, doing it like this is still correct, so I don't think postponing hoop-jumping that will be necessary anyway benefits us

technillogue: I guess I would accept

```python
async def predict(self, ...): ...

def predict_with_check(self, ...):
        self._assert_state(WorkerState.READY)
        return self.predict(...)
```

python/tests/test_json.py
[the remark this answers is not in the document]
technillogue: oh, never mind, we don't run linters in CI, we should do that

python/tests/server/test_worker.py
technillogue: oh that would have been extremely helpful to know about swallow_exceptions

I've pushed an attempt to be less surprising that keeps the most common async usecase easy, but I can do it your way if you feel adamant

python/tests/server/test_worker.py
nickstenning: I still don't think `predict` needs to be an async method. It can do whatever state checks it needs to and then return an async iterator. If there's an inner `async def _predict(...):` and that's where the semaphore acquire goes then that's fine.

Making it `async` and then having to do backflips just to write the tests seems silly.

I think it's perfectly reasonable to make considered choices about whether any particular method or function actually needs to be async or not. Keep in mind that as long as nothing blocks it's totally fine to call non-async functions from within an async context, while the reverse is of course not true.

technillogue:
```
cog/server/http.py:187: in healthcheck
    "setup": attrs.asdict(app.state.setup_result),
../../../.local/lib/python3.11/site-packages/attr/_next_gen.py:214: in asdict
    return _asdict(
../../../.local/lib/python3.11/site-packages/attr/_funcs.py:53: in asdict
    attrs = fields(inst.__class__)
E           attr.exceptions.NotAnAttrsClassError: <class 'NoneType'> is not an attrs-decorated class.
```

something changed with setup...

python/tests/server/test_worker.py
```diff
@@ -104,14 +115,17 @@ class Result:
     exception: Optional[Exception] = None
 
 
-def _process(events, swallow_exceptions=False):
+async def _process(events) -> Result:
+    return _sync_process([e async for e in events])
+
+
+def _sync_process(events) -> Result:
```
nickstenning, line 122: Nit and not a blocker: it would be more conventional for the async function to have the name with an `a` or an `async` in.

python/tests/server/test_worker.py
```diff
@@ -515,18 +542,19 @@ def predict(self, name, steps):
...
         try:
             for _ in range(n):
-                event = next(self.predict_generator)
+                event = self.await_(anext(self.predict_generator))
                 self.predict_events.append(event)
-        except StopIteration:
+        except StopAsyncIteration:
             self.predict_generator = None
             self._check_predict_events()
+        except InvalidStateException:
```
nickstenning, line 550: I'm pretty sure this isn't possible, right?

nickstenning (commented):
This is getting really close. Just a few small things, and one big one (the worker Hypothesis test can't ever run `predict` right now).

python/tests/server/test_worker.py
technillogue: the sync variant is used in one place, the async version is used everywhere else, it seems silly

python/tests/server/test_worker.py
```diff
@@ -143,8 +157,7 @@ def _process(events, swallow_exceptions=False):
                 pytest.fail(f"saw unexpected event: {event}")
     except Exception as exc:
         result.exception = exc
-        if not swallow_exceptions:
-            raise
+        raise
```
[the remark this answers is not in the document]
technillogue: oh good

python/tests/server/test_worker.py
```diff
@@ -484,22 +512,21 @@ def setup(self):
...
         assert isinstance(self.setup_events[-1], Done)
 
-        print(self.setup_events)
-        result = _process(self.setup_events)
+        result = _sync_process(self.setup_events)
         assert result.stdout == "did setup\n"
         assert result.stderr == ""
         assert result.done == Done()
 
+    @precondition(lambda x: x.setup_done)
```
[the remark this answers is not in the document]
technillogue: :pray:

python/tests/server/test_worker.py
technillogue: Thank you for detailed discussion on this design choice. I'm happy with this PR as it is, but for the next PR this might be different. To me, having yet another underscore method with the same name for a semaphore _is_ doing backflips in the actual runtime code for the sake of marginally simpler tests, and that seems silly. In practice we're always doing `async for event in worker.predict(...)` and this would add more unnecessary indirection to a codebase that already has a great deal of indirection. 

Python and C# are different languages with different conventions. This doesn't seem pythonic to me, though I will concede that `asyncio.create_task` exists. I spent a little while looking for other examples of this pattern. Claude thought one example might be `httpx.AsyncClient.stream`, but that's actually an async function that starts with a non-async assertion that the protocol is valid.

Again, I'm happy with this PR as is, this is only for the future. I feel like this would be cleaner in the long run, but it's not that important in the grand scheme of things and if you feel really strongly about this I'm ultimately willing to stick to your proposal

python/tests/server/test_worker.py
technillogue: ended up being unnecessary

python/cog/server/helpers.py
```diff
@@ -148,3 +159,37 @@ def run(self) -> None:
...
+
+    def poll(self, timeout: float = 0.0) -> bool:
+        return self.conn.poll(timeout)
+
+    def _recv(self) -> Optional[X]:
+        # this ugly mess could easily be avoided with loop.connect_read_pipe
+        # even loop.add_reader would help but we don't want to mess with a thread-local loop
+        while not self.exiting.is_set():
+            if self.conn.poll(0.01):
+                return self.conn.recv()
```
mattt, line 190: Add an explicit `return` statement here to placate type checker

```suggestion
    def _recv(self) -> Optional[X]:
        # this ugly mess could easily be avoided with loop.connect_read_pipe
        # even loop.add_reader would help but we don't want to mess with a thread-local loop
        while not self.exiting.is_set():
            if self.conn.poll(0.01):
                return self.conn.recv()
        
        return None
```

mattt (approved):
Nice work, @technillogue. Thanks for sticking with this.

I'd be curious to explore `connect_read_pipe` for `AsyncPipe` and returning `Task` instead of `AsyncIterator`, but I think this is enough to go on for now to keep moving forward on asynchronous predictions.

technillogue:
connect_read_pipe for AsyncPipe would be _great_. I've done this before: https://github.com/dryadsystems/pqueue/blob/main/pqueue.py#L136-L148, and it would be better but mp.Pipe does a lot of stuff for you that would be annoying to do on your own. my version has no concerns about duplex or much concerns about pickling. it's be a lot of code to replace this

PR #1419, opened 2023-12-08
fix: linting in python code

dkhokhlov:


python/cog/predictor.py
```diff
@@ -126,9 +125,9 @@ def get_weights_type(setup_function: Callable[[Any], Optional[Awaitable[None]]])
 
 
 def run_prediction(
-    predictor: BasePredictor,
-    inputs: Dict[Any, Any],
-    cleanup_functions: List[Callable[[], None]],
+        predictor: BasePredictor,
+        inputs: Dict[Any, Any],
+        cleanup_functions: List[Callable[[], None]],
```
technillogue, line 130: ? that's not right, is it?

dkhokhlov: sorry, my bad. 
i made ruff happy with ruff fix, but missed pyright.
let me roll back?

PR #1428, opened 2023-12-12
fix: handle early exceptions in model setup

dkhokhlov:
- handle early exceptions during model container setup
- catch exceptions during creation of Pydantic validators
- propagate error message to health checks w/o creation of predictor

Ref: https://github.com/replicate/infra/issues/21

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -99,6 +99,8 @@ def create_app(
     try:
         # TODO: avoid loading predictor code in this process
         predictor = load_predictor_from_ref(predictor_ref)
+        InputType = get_input_type(predictor)
+        OutputType = get_output_type(predictor)
     except Exception as ex:
         app.state.health = Health.SETUP_FAILED
         response = schema.PredictionResponse(input={},
@@ -118,7 +120,7 @@ def create_app(
         async def healthcheck_startup_failed() -> Any:
             return jsonable_encoder(
                 {
-                    "status": app.state.health,
+                    "status": app.state.health.name,
                     "setup": app.state.setup_result_payload,
                 }
             )
@@ -130,9 +132,6 @@ def create_app(
         shutdown_event=shutdown_event,
         upload_url=upload_url,
     )
-
-    InputType = get_input_type(predictor)
-    OutputType = get_output_type(predictor)
 
     class PredictionRequest(schema.PredictionRequest.with_types(input_type=InputType)):
         pass
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -8,6 +8,7 @@ import socket
 import sys
 import textwrap
 import threading
+from datetime import datetime
 from enum import Enum, auto, unique
 from typing import (
     TYPE_CHECKING,
@@ -95,13 +96,40 @@ def create_app(
 
     predictor_ref = get_predictor_ref(config, mode)
 
+    try:
+        # TODO: avoid loading predictor code in this process
+        predictor = load_predictor_from_ref(predictor_ref)
+    except Exception as ex:
+        app.state.health = Health.SETUP_FAILED
+        response = schema.PredictionResponse(input={},
+                                             output=None,
+                                             id=predictor_ref,
+                                             version=None,
+                                             created_at=datetime.now(),
+                                             started_at=datetime.now(),
+                                             completed_at=None,
+                                             logs="",
+                                             error=repr(ex),
+                                             status=schema.Status.FAILED,
+                                             metrics=None)
+        app.state.setup_result_payload = response
+
+        @app.get("/health-check")
+        async def healthcheck_startup_failed() -> Any:
+            return jsonable_encoder(
+                {
+                    "status": app.state.health,
+                    "setup": app.state.setup_result_payload,
+                }
+            )
+
+        return app
+
     runner = PredictionRunner(
         predictor_ref=predictor_ref,
         shutdown_event=shutdown_event,
         upload_url=upload_url,
     )
-    # TODO: avoid loading predictor code in this process
-    predictor = load_predictor_from_ref(predictor_ref)
 
     InputType = get_input_type(predictor)
     OutputType = get_output_type(predictor)
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -8,7 +8,8 @@ import socket
 import sys
 import textwrap
 import threading
-from datetime import datetime
+import traceback
+from datetime import datetime, timezone
 from enum import Enum, auto, unique
 from typing import (
     TYPE_CHECKING,
@@ -93,6 +94,7 @@ def create_app(
     app.state.health = Health.STARTING
     app.state.setup_task = None
     app.state.setup_result = None
+    started_at = datetime.now(tz=timezone.utc)
 
     predictor_ref = get_predictor_ref(config, mode)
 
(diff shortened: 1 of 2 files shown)


python/cog/server/http.py
```diff
@@ -86,213 +86,233 @@ def create_app(
...
+            else:
+                app.state.health = Health.SETUP_FAILED
+
+            app.state.setup_result_payload = result
+
+            # Reset app.state.setup_result so future calls are a no-op
+            app.state.setup_result = None
 
-        app.state.setup_result_payload = result
+    except Exception:
```
nickstenning, line 304: I think we might want to do a bit of refactoring here so we don't have to wrap 300 lines of the file in `try:/except:` 😆

dkhokhlov: uploaded new iteration.

python/cog/server/http.py
```diff
@@ -88,16 +89,42 @@ def create_app(
...
     predictor_ref = get_predictor_ref(config, mode)
 
+    try:
+        # TODO: avoid loading predictor code in this process
+        predictor = load_predictor_from_ref(predictor_ref)
+        InputType = get_input_type(predictor)
+        OutputType = get_output_type(predictor)
+    except Exception as ex:
+        app.state.health = Health.SETUP_FAILED
+        response = schema.PredictionResponse(input={},
```
nickstenning, line 99: I don't think this is a change you introduced, but it seems quite wrong to me that`setup` returns a `PredictionResponse`. It looks like that was introduced [here](https://github.com/replicate/cog/commit/78caaa3e27201cfc87309b7a5f62532e7eb6df72#diff-136cc6506e8aaea24081862f89e1cb778251b7726180fe237c8aaf560ae7ec69R310-R322) and I think I'd like to back that change out.

python/cog/server/http.py
```diff
@@ -88,16 +89,42 @@ def create_app(
...
+                                             output=None,
+                                             id=predictor_ref,
+                                             version=None,
+                                             created_at=datetime.now(),
+                                             started_at=datetime.now(),
+                                             completed_at=None,
+                                             logs="",
+                                             error=repr(ex),
```
nickstenning, line 107: In fact, this is further evidence that returning a `PredictionResponse` from `setup` is a bad idea. `replicate-web` (the ultimate recipient of this payload) [doesn't know anything about the `error` field for the value returned as the "setup result payload", it only knows about `logs`](https://github.com/replicate/replicate-web/blob/baacbe21b83d6cfd9e6f1a850e0f5c73d6d100e8/replicate_web/models/api_serializers.py#L74).

Concretely that means that here we should be adding details on the exception to `logs`, not `error`, but I do think this is terribly confusing and we should not be returning a `PredictionResponse` from setup.

I propose we open a separate PR to revert that change.

python/cog/server/http.py
```diff
@@ -88,16 +89,42 @@ def create_app(
...
+                                             status=schema.Status.FAILED,
+                                             metrics=None)
+        app.state.setup_result_payload = response
+
+        @app.get("/health-check")
+        async def healthcheck_startup_failed() -> Any:
```
nickstenning, line 113: As mentioned in our chat, I think we still need a working `shutdown` endpoint when we fail this way.

nickstenning:
Directionally this looks good, but I think we will at least need a working `shutdown` endpoint before this can be merged. Separately, we need to use the `logs` field, not `error`, here, and I think we should (perhaps in a separate PR) reconsider returning a `PredictionResponse` from the `PredictionRunner.setup` method.

python/cog/server/http.py
dkhokhlov: cc: @yorickvP

python/cog/server/http.py
dkhokhlov: uploaded

python/cog/server/http.py
```diff
@@ -88,16 +89,46 @@ def create_app(
...
+            return jsonable_encoder(
+                {
+                    "status": app.state.health.name,
+                    "setup": app.state.setup_result_payload,
+                }
+            )
+
+        @app.on_event("shutdown")
+        def shutdown_startup_failed() -> None:
```
nickstenning, line 122: Hmm, not quite what I meant. I'll leave a longer comment in the PR explaining this.

nickstenning:
Ok, I think maybe there's a bit of a misunderstanding w.r.t. the latest changes pushed (the shutdown stuff).

When we run Cog in production, we run it with `python -m cog.server.http [...] --await-explicit-shutdown`. The `--await-explicit-shutdown` option means that Cog ignores SIGTERM (from kubernetes) and waits for director to call its `POST /shutdown` endpoint.

That needs to work, even when the model module crashed.

It occurred to me that there is perhaps a simpler approach we can take here that doesn't require changing which HTTP routes are defined in the application. We can just skip providing the input/output types when loading the predictor fails:

```python
    PredictionRequest = schema.PredictionRequest
    PredictionResponse = schema.PredictionResponse

    try:
        # TODO: avoid loading predictor code in this process
        predictor = load_predictor_from_ref(predictor_ref)
    except Exception:
        now = datetime.utcnow()
        app.state.health = Health.SETUP_FAILED
        app.state.setup_result_payload = {
            "status": schema.Status.FAILED,
            "started_at": now,
            "completed_at": now,
            "logs": "Error while loading predictor:\n\n" + traceback.format_exc(),
        }
    else:
        # If we successfully loaded the input/output types, then override Prediction{Request,Response} 
        # with those types filled in:
        PredictionRequest = PredictionRequest.with_types(input_type=InputType)
        PredictionResponse = PredictionResponse.with_types(
            input_type=InputType, 
            output_type=OutputType,
        )
```

and everything else can remain the same.

nickstenning:
I also opened https://github.com/replicate/cog/pull/1433 to address the PredictionResponse type confusion.

python/cog/server/http.py
dkhokhlov: resolved by https://github.com/replicate/cog/pull/1433

dkhokhlov:
> Ok, I think maybe there's a bit of a misunderstanding w.r.t. the latest changes pushed (the shutdown stuff).
> 
> When we run Cog in production, we run it with `python -m cog.server.http [...] --await-explicit-shutdown`. The `--await-explicit-shutdown` option means that Cog ignores SIGTERM (from kubernetes) and waits for director to call its `POST /shutdown` endpoint.
> 
> That needs to work, even when the model module crashed.
> 
> It occurred to me that there is perhaps a simpler approach we can take here that doesn't require changing which HTTP routes are defined in the application. We can just skip providing the input/output types when loading the predictor fails:
> 
> ```python
>     PredictionRequest = schema.PredictionRequest
>     PredictionResponse = schema.PredictionResponse
> 
>     try:
>         # TODO: avoid loading predictor code in this process
>         predictor = load_predictor_from_ref(predictor_ref)
>         InputType = get_input_type(predictor)
>         OutputType = get_output_type(predictor)
>     except Exception:
>         now = datetime.utcnow()
>         app.state.health = Health.SETUP_FAILED
>         app.state.setup_result_payload = {
>             "status": schema.Status.FAILED,
>             "started_at": now,
>             "completed_at": now,
>             "logs": "Error while loading predictor:\n\n" + traceback.format_exc(),
>         }
>     else:
>         # If we successfully loaded the input/output types, then override Prediction{Request,Response} 
>         # with those types filled in:
>         PredictionRequest = PredictionRequest.with_types(input_type=InputType)
>         PredictionResponse = PredictionResponse.with_types(
>             input_type=InputType, 
>             output_type=OutputType,
>         )
> ```
> 
> and everything else can remain the same.

will not work in case when predictor  or InputType  or InputType  are nto created bc of Exception.

nickstenning:
> The goal in exception case is simple - handle only two endpoints - shutdown and healthcheck, w/o runner code complexity involved (there is no runner).

That sounds fine. You just need to add an actual shutdown HTTP handler, not a shutdown event handler!

python/cog/server/http.py
```diff
@@ -92,19 +93,48 @@ def create_app(
...
+        # TODO: avoid loading predictor code in this process
+        predictor = load_predictor_from_ref(predictor_ref)
+        InputType = get_input_type(predictor)
+        OutputType = get_output_type(predictor)
+    except Exception as ex:
+        app.state.health = Health.SETUP_FAILED
+        result = SetupResult(
+            started_at=started_at,
+            completed_at=datetime.now(tz=timezone.utc),
+            logs=f"{repr(ex)}\n",
```
nickstenning, line 110: Could you please use `traceback.format_exc()` as in the example I shared?

python/cog/server/http.py
```diff
@@ -274,13 +304,6 @@ async def cancel(prediction_id: str = Path(..., title="Prediction ID")) -> Any:
         else:
             return JSONResponse({}, status_code=200)
 
-    @app.post("/shutdown")
```
nickstenning, line 277: You can't remove this from here! This will mean that models that _do_ boot successfully don't have a shutdown endpoint.

python/tests/server/test_http_input.py
```diff
@@ -239,10 +243,24 @@ def test_union_integers(client):
...
-        make_client("input_untyped")
+    config = {"predict": _fixture_path("input_untyped")}
+    app = create_app(
+        config=config,
+        shutdown_event=threading.Event(),
+        upload_url="input_untyped",
+    )
+    assert app.state.health == Health.SETUP_FAILED
+    assert app.state.setup_result.status == schema.Status.FAILED
+    assert app.state.setup_result.logs.find("TypeError: No input type provided for parameter") != -1
```
nickstenning, line 254:
```suggestion
    assert "TypeError: No input type provided for parameter" in app.state.setup_result.logs
```

This would be more idiomatic, but no big deal either way.

python/tests/server/test_http_input.py
```diff
@@ -239,10 +243,24 @@ def test_union_integers(client):
...
-        make_client("input_unsupported_type")
+    config = {"predict": _fixture_path("input_unsupported_type")}
+    app = create_app(
+        config=config,
+        shutdown_event=threading.Event(),
+        upload_url="input_untyped",
+    )
+    assert app.state.health == Health.SETUP_FAILED
+    assert app.state.setup_result.status == schema.Status.FAILED
+    assert app.state.setup_result.logs.find("TypeError: Unsupported input type input_unsupported_type.") != -1
```
nickstenning, line 266:
```suggestion
    assert "TypeError: Unsupported input type input_unsupported_type." in app.state.setup_result.logs
```

nickstenning (approved):
I tested this locally with director and it seems to work well. LGTM. (Two small suggestions for the tests.)

PR #1439, opened 2023-12-18
fix: update the version typing_extensions to use latest version 

### LINKED ISSUE #1438: The version of typing_extensions breaks the build
When building locally I'm getting the following error:
```
ImportError: cannot import name dataclass_transform in init pydantic.dataclasses
```

This happens because the we are using `typing_extensions` v4.1.0 and the latest version 4.9.0 does not break.

mbukeRepo:
This PR fixes #1438


yorickvP:
Tests fail, 4.9 doesn't support python 3.7 :(

yorickvP:
Thanks!

PR #1443, opened 2023-12-19
check if setup_result is None

technillogue:
#1433 introduced a race condition if healthcheck happens before setup is done

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -115,12 +115,8 @@ def create_app(
 
         @app.get("/health-check")
         async def healthcheck_startup_failed() -> Any:
-            return jsonable_encoder(
-                {
-                    "status": app.state.health.name,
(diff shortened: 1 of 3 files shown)


python/cog/server/http.py
```diff
@@ -115,12 +115,10 @@ def create_app(
 
         @app.get("/health-check")
         async def healthcheck_startup_failed() -> Any:
-            return jsonable_encoder(
-                {
-                    "status": app.state.health.name,
-                    "setup": attrs.asdict(app.state.setup_result),
-                }
+            setup = (
```
dkhokhlov, line 118: this change not needed.  app.state.setup_result = result is always defined above.

dkhokhlov (changes requested):

dkhokhlov (approved):
lgtm

nickstenning (approved):
Oops. 🤦‍♂️

Thank you!

Issue #1275, opened 2023-08-28
Model schema is invalid: invalid components: schema "Output": extra sibling fields: [title]


stri8ed:
I am attempting to push a new model to Replicate, but it's failing during the schema validation stage.

```
Validating model schema...
ⅹ Model schema is invalid: invalid components: schema "Output": extra sibling fields: [title]
```

The output model used is:

```
class ModelOutput(BaseModel):
    preview: str
    srt_file: Path
    vtt_file: Path
```
Looking at the [cog source](https://github.com/replicate/cog/blob/a6a26b7f7a26bec45c92a8644fed3c454099dd39/python/cog/command/ast_openapi_schema.py#L355), it seems it generates a schema based on the model output, and each field includes a title field:

`"properties":{"preview":{"title":"Preview","type":"string"},"srt_file": etc..`

But the validator is complaining about this field.

What can I do to resolve this?

stri8ed:
> Having the same issue today. An update would be appreciated.

I was able to "solve" it, by downgrading to version 0.8.3

xaksis:
@stri8ed could you state how you downgraded to 0.8.3? I've been trying with brew and it just keeps throwing other errors.

stri8ed:
@xaksis I am on Linux (via Windows WSL), so used curl:

```
sudo curl -o /usr/local/bin/cog -L "https://github.com/replicate/cog/releases/v0.8.3/download/cog_$(uname -s)_$(uname -m)"
sudo chmod +x /usr/local/bin/cog
```

peter65374:
Got same error as below 
Validating model schema...
```
ⅹ Model schema is invalid: invalid components: schema "Output": extra sibling fields: [title]
ⅹ
[...]

Issue #1361, opened 2023-11-03
Output type List[Output] causes cog to generate invalid output schema


yorickvP:
This code:

```python
class Output(BaseModel):
  text: str

class Predictor(BasePredictor):
  def predict(self) -> list[Output]:
    pass
```

confuses cog's `get_output_type` into generating a schema with no `Output`, thereby confusing the replicate.com UI

![image](https://github.com/replicate/cog/assets/647076/aea04015-6482-49c7-a1fe-6f01735befe5)

zeke:
Not sure if it's related but there was a change in Python 3.9 that made capital-L `List` the thing you use for type hinting, rather than `list` which is now just the built-in python `list` function.

If you do this, does it work?

```py
from typing import List 

class Output(BaseModel):
  text: str

class Predictor(BasePredictor):
  def predict(self) -> List[Output]:
    pass
```

yorickvP:
No change. Test with `cog run python3 -m cog.command.openapi_schema`

yorickvP:
Here's another one: 
https://github.com/replicate/cog/pull/1267#issuecomment-1885038556

PR #1444, opened 2023-12-19
Add note for naming `Output` class

iamargentum:
It seems there's some [confusion with custom return objects](https://github.com/replicate/cog/issues/1275). This PR intends to avoid this confusion by adding a not in the docs.

### DIFF
--- a/docs/python.md
+++ b/docs/python.md
@@ -158,7 +158,7 @@ class Predictor(BasePredictor):
         return Output(text="hello", file=io.StringIO("hello"))
 ```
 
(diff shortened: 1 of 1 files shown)


yorickvP:
Ah yes, [I remember](https://github.com/replicate/cog/issues/1361)

iamargentum:
probably unrelated ([second comment in the issue thread that you've linked](https://github.com/replicate/cog/issues/1361#issuecomment-1796954988)), but i also [raised a PR](https://github.com/replicate/cog/pull/1327) for updating the deprecated `list` type usage in one of the examples a while ago

PR #1451, opened 2023-12-22
fix: test fixture does not remove docker image

dkhokhlov:
- need to use force as image is used by container (stopped)


nickstenning:
I don't think this is right. If this had been failing then the test would have been failing in CI.

Issue #1465, opened 2024-01-03
GitHub release artifact names have changed


zeke:
Newer versions of Cog have assets with lowercase filenames, like `cog_darwin_x86_64`

But they used to be mixed case, like `cog_Darwin_arm64`

So now our installation instructions on the README don't work for recent releases, because the output of `uname -s` on macOS is capitalized `Darwin`:

```console
sudo curl -o /usr/local/bin/cog -L "https://github.com/replicate/cog/releases/latest/download/cog_$(uname -s)_$(uname -m)"
sudo chmod +x /usr/local/bin/cog
```

I haven't looked into it yet, but I'm guessing this was an unexpected change from a goreleaser upgrade.

## Affected versions:

- All prerelease versions > 0.8.6
- https://github.com/replicate/cog/releases/tag/v0.8.6 ("latest")
- https://github.com/replicate/cog/releases/tag/v0.8.5
- https://github.com/replicate/cog/releases/tag/v0.8.4
- https://github.com/replicate/cog/releases/tag/v0.8.3
- https://github.com/replicate/cog/releases/tag/v0.8.2
- https://github.com/replicate/cog/releases/tag/v0.8.1
- https://github.com/replicate/cog/releases/tag/v0.8.0

## Last versions before breaking change

- https://github.com/replicate/cog/releases/tag/v0.8.0-beta8 (prerelease)
- https://github.com/replicate/cog/releases/tag/v0.7.2

## To Do

- [x] https://github.com/replicate/cog/pull/1466
- [x] https://github.com/replicate/cog/pull/1467
- [x] Manually rename borked artifacts (ignoring prerelease versions)
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.6
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.5
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.4
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.3
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.2
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.1
  - [x] https://github.com/replicate/cog/releases/tag/v0.8.0
- [x] Publish a [prerelease](https://github.com/replicate/cog/blob/main/CONTRIBUTING.md#publishing-a-prerelease) to verify new config is working: https://github.com/replicate/cog/releases/tag/v0.9.0-beta12

zeke:
[...]

PR #1467, opened 2024-01-04
Capitalize OS name in release artifacts

zeke:
This is a partial fix for https://github.com/replicate/cog/issues/1465

This PR updates the Goreleaser config to capitalize OS names like `Darwin` and `Linux` so they match the output of `uname -s`, which we've long used for the [installation instructions in the README](https://github.com/replicate/cog?tab=readme-ov-file#install).

@caarlos0 does this look right? I did a dry run locally and I see this:

```
$ goreleaser release --snapshot --skip=publish --clean              
...
• archives
  • skip archiving                                 binary=cog name=cog_Darwin_arm64
  • skip archiving                                 binary=cog name=cog_Darwin_x86_64
  • skip archiving                                 binary=cog name=cog_Linux_x86_64
  • skip archiving                                 binary=cog name=cog_Linux_arm64
```

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -18,7 +18,10 @@ builds:
 archives:
   - format: binary
     name_template: >-
-      {{ .ProjectName }}_{{ .Os }}_
+      {{ .ProjectName }}_
+      {{- if eq .Os "darwin" }}Darwin
+      {{- else if eq .Os "linux" }}Linux
+      {{- else }}{{ .Os }}{{ end -}}_
       {{- if eq .Arch "amd64" }}x86_64
       {{- else if eq .Arch "386" }}i386
       {{- else }}{{ .Arch }}{{ end -}}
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -18,7 +18,10 @@ builds:
 archives:
   - format: binary
     name_template: >-
-      {{ .ProjectName }}_{{ .Os }}_
+      {{ .ProjectName }}_
+      {{- if eq .Os "darwin" }}Darwin
+      {{- else if eq .Os "linux" }}Linux
+      {{- else }}{{ .Os }}{{ end -}}_
```
caarlos0, line 24: it's not wrong, but it can be simplified to

```suggestion
      {{- title .Os }}_
```

zeke:
Thanks @caarlos0 🙏🏼

Issue #1452, opened 2023-12-22
Inform user if `cog.yaml` or `predict.py` are missing


joehoover:
`cog.command.openapi_schema` should inform users if `cog.yaml` is missing.

Currently, `cog.command.openapi_schema` silently returns `{}` if `cog.yaml` is missing (see [here](https://github.com/replicate/cog/blob/b28dd14e90a6a5d84b372dbd5cd51dadccfd9bbc/python/cog/command/openapi_schema.py#L20C2-L20C2)).

This is undesirable in situations where `cog.yaml` is in fact missing, but the user _thinks_ that it is present. In such cases, `cog build` will fail on openapi schema validation, but it won't be clear why. 

This just happened to me! 

For various questionable reasons, `cog.yaml` in my project root is a symlink and the target was accidentally excluded in my `.dockerignore`. This led to an interesting situation where `cog run python -m cog.command.openapi_schema` returned the correct schema, but `cog build` failed at schema validation because the schema returned empty.

A warning or just a note that my `cog.yaml` was missing would have saved me quite a bit of detective work.

asingh9530:
@joehoover this is a easy fix by adding a warning, should I raise a PR to fix this ?

zeke:
> should I raise a PR to fix this ?

@asingh9530 yes please!

asingh9530:
@zeke missed notification, added PR for fix.

anish-sahoo:
Closing because the Go config loader now returns explicit ConfigNotFound errors and cog doctor checks file existence.

PR #1470, opened 2024-01-05
Warn if `cog.yaml` or predictor are missing

asingh9530:
What issue this PR fix - #1452


zeke:
Thanks for the PR. It's a little confusing that this warning describes two possible issues. What do you think about handling the two exceptions with different warnings?

Some untested pseudocode:

```python
except ConfigDoesNotExist:
        warnings.warn("no cog.yaml found")
except PredictorNotSet:
        warnings.warn("no predict() method found in Predictor")
```

asingh9530:
@zeke yeah I think that is much better let me make those changes.

zeke:
So... this will print a warning, but it will also blow up because the `pass` was removed, right?

Is that that the intended behavior here, or do we want to print a warning without failing?

asingh9530:
> So... this will print a warning, but it will also blow up because the `pass` was removed, right?

I think even in previous setting it would blow but without any warning or error but in this case it will show warning to user and  then blow. So the user will likely know the root cause.  

> Is that that the intended behavior here, or do we want to print a warning without failing?

if we want to prevent the build to even begin then we have to raise exception instead of warning something like this.
```
    except ConfigDoesNotExist:
        raise ConfigDoesNotExist("no cog.yaml found or present")
    except PredictorNotSet:
        raise PredictorNotSet("no predict method found in Predictor")
```

what do you think ?

zeke (approved):
This looks good to me 👍🏼

aron (approved):

asingh9530:
@zeke @aron can you please approve test workflows.

asingh9530:
@zeke could please check why Integration failed as this PR dosen't contain any changes to the file reflected in tests.

yorickvP:
@asingh9530 the test 'https://github.com/replicate/cog/blob/main/test-integration/test_integration/test_build.py#L9' fails because it tests for a specific error message, which has been changed in this PR. Please update the test to match the new error message.

mattt:
Thanks for your help with this, @asingh9530!

zeke:
Thanks for picking this up @mattt! 🙏🏼

PR #1472, opened 2024-01-08
eliminate double load of predict.py in cog model setup

dkhokhlov:
- create schema types from openapi schema stored in container

Ref: https://github.com/replicate/infra/issues/23


pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
   'typing-compat; python_version < "3.8"',
   "typing_extensions>=4.4.0",
   "uvicorn[standard]>=0.12,<1",
+  "datamodel-code-generator >=0.25"
```
technillogue, line 25: if we're going to use this we'd probably need to vendor it, it's not ideal to add new deps

technillogue:
I think https://github.com/replicate/cog/pull/1056 explored how to generate pydantic models from the AST, I think you could do the same thing from the schema json

technillogue:
we should either vendor in the parts of datamodel-code-generator that we need or use the approach from #1056, it's a bad idea to add dependencies to cog if we can avoid them in genreal, and datamodel-code-generator especially has too many dependencies

pyproject.toml
dkhokhlov: removed

dkhokhlov:
superseded by https://github.com/replicate/cog/pull/1503

PR #1487, opened 2024-01-16
fix: added log message and error exit code when setup fails

dkhokhlov:
- log error message when setup fails
- return error exit code when setup fails


technillogue (approved):
seems fine

PR #1490, opened 2024-01-19
fix: add handling of early setup exceptions on "train" path in cog predict/build

dkhokhlov:
- added handling of early setup exceptions on "train" path in cog predict/build
- skip calling setup() on early setup errors, and optionally initiate shutdown in interactive mode 
- added check for setup() errors in cog build (in openapi_schema.py)

Testing:
- use simple model with predict & train
- change input type in train to a bad type
- run cog predict or build

Expected:
- cog prints bad input type error

Ref:  https://github.com/replicate/infra/issues/25

### DIFF
--- a/python/cog/command/openapi_schema.py
+++ b/python/cog/command/openapi_schema.py
@@ -5,8 +5,9 @@ This prints a JSON object describing the inputs of the model.
 """
 import json
 
-from ..errors import ConfigDoesNotExist, PredictorNotSet
+from ..errors import CogError, ConfigDoesNotExist, PredictorNotSet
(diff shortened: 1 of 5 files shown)


technillogue (commented):
how about like this? https://github.com/replicate/cog/commit/c30e618df181896b0451bd4634538fb07cb990a5

dkhokhlov:
> how about like this? [c30e618](https://github.com/replicate/cog/commit/c30e618df181896b0451bd4634538fb07cb990a5)

right, I can extract common.

mattt (approved):

PR #1495, opened 2024-01-24
Use cheaper runner for CI

erbridge:
16-cores costs $0.064/minute. 8-cores costs $0.032/minute. The default 2-cores costs $0.008/minute.


mattt (approved):

mattt:
Looks like tests are executing in the same amount of time with the smaller runner. Thanks, @erbridge!

PR #1496, opened 2024-01-24
fix: ignore "train" part of schema errors during predict runs

dkhokhlov:
- ignore "train" part of schema errors during predict runs for backward compatibility with existing "bad" models already in use
- expected: cog shall not trigger exception/error  when "train" is present in config but "train.py" is missing.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -36,6 +36,7 @@ from pydantic import ValidationError
 from pydantic.error_wrappers import ErrorWrapper
 
 from .. import schema
+from ..errors import PredictorNotSet
 from ..files import upload_file
 from ..json import upload_files
 from ..logging import setup_logging
@@ -104,6 +105,7 @@ def create_app(
     threads: int = 1,
     upload_url: Optional[str] = None,
     mode: str = "predict",
+    is_build: bool = False
 ) -> MyFastAPI:
     app = MyFastAPI(
         title="Cog",  # TODO: mention model name?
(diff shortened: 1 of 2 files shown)


python/cog/server/http.py
```diff
@@ -104,6 +105,7 @@ def create_app(
     threads: int = 1,
     upload_url: Optional[str] = None,
     mode: str = "predict",
+    is_build: bool = False
```
mattt, line 108: I like the decision to define this behavior in a new argument to `create_app`. We originally discussed defining a new `mode` value, but I see the value of using something that doesn't map to a command-line argument, and therefore risk misuse (e.g. What's the expected behavior of a pod running a model with the build mode?)

python/cog/server/http.py
```diff
@@ -163,48 +165,54 @@ async def wrapped(*args: "P.args", **kwargs: "P.kwargs") -> "T":
 
     if "train" in config:
         try:
+            predictor_ref = get_predictor_ref(config, "train")
```
mattt, line 168:
```suggestion
            trainer_ref = get_predictor_ref(config, "train")
```

python/cog/server/http.py
```diff
@@ -163,48 +165,54 @@ async def wrapped(*args: "P.args", **kwargs: "P.kwargs") -> "T":
...
             # TODO: avoid loading trainer code in this process
-            trainer = load_predictor_from_ref(config["train"])
+            trainer = load_predictor_from_ref(predictor_ref)
```
mattt, line 170:
```suggestion
            trainer = load_predictor_from_ref(trainer_ref)
```

python/cog/server/http.py
mattt: I'm not wild about this particular spelling, but I can't think of anything better. Since it's effectively a private API, it's probably fine. Or else, we can change it later without breaking anyone.

mattt (approved):
Nice work, @dkhokhlov. Thanks for taking this on.

Left a few non-blocking suggestions for renaming `predictor_ref` to `training_ref`. But other than that, I think this is solid.

PR #1498, opened 2024-01-25
format and add formatting to ci

technillogue:
#1496 and others introduced formatting errors. let's add ruff format --check to CI. I'm also open to using black instead (since we have it in pyproject) or removing black altogether


dkhokhlov (approved):
lgtm

PR #1503, opened 2024-01-29
feat: eliminate redundant loads of predict.py and train.py in early setup of cog predict

dkhokhlov:
- store openapi schema file json & py files in .cog subdir during cog build;
- conditionally create Input & Output schema types from openapi schema file;
- conditionally create TrainingInput & TrainingOutput schema types from openapi schema file;

Migration:
- optimization will work for new  models, old models are unaffected

Ref: https://github.com/replicate/infra/issues/23

### DIFF
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -76,7 +76,7 @@ func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, bund
 
 	dockerfile := "FROM " + image + "\n"
 	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
-	const env_path = "/tmp/venv/tools/"
+	env_path := "/tmp/venv/tools/"
 	dockerfile += "RUN python -m venv --symlinks " + env_path + " && " +
 		env_path + "/bin/python -m pip install 'datamodel-code-generator>=0.25' && " +
 		env_path + "/bin/datamodel-codegen --version && " +
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -49,7 +49,7 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 	return cmd.Run()
 }
 
-func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, schemaFile string) error {
+func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, bundledSchemaFile string, bundledSchemaPy string) error {
 	var args []string
 
 	args = append(args,
@@ -75,7 +75,13 @@ func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, sche
 	cmd := exec.Command("docker", args...)
 
 	dockerfile := "FROM " + image + "\n"
-	dockerfile += "COPY " + schemaFile + " .cog"
+	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
+	const env_path = "/tmp/venv/tools/"
+	dockerfile += "RUN python -m venv --symlinks " + env_path + " && " +
+		env_path + "/bin/python -m pip install 'datamodel-code-generator>=0.25' && " +
+		env_path + "/bin/datamodel-codegen --version && " +
+		env_path + "/bin/datamodel-codegen --input-file-type openapi --input " + bundledSchemaFile +
+		" --output " + bundledSchemaPy + " && rm -rf " + env_path
 	cmd.Stdin = strings.NewReader(dockerfile)
 
 	console.Debug("$ " + strings.Join(cmd.Args, " "))
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -49,7 +49,7 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 	return cmd.Run()
 }
 
-func BuildAddLabelsToImage(image string, labels map[string]string) error {
+func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, schemaFile string) error {
 	var args []string
 
 	args = append(args,
@@ -74,7 +74,8 @@ func BuildAddLabelsToImage(image string, labels map[string]string) error {
 	args = append(args, ".")
 	cmd := exec.Command("docker", args...)
 
-	dockerfile := "FROM " + image
+	dockerfile := "FROM " + image + "\n"
+	dockerfile += "COPY " + schemaFile + " .cog"
 	cmd.Stdin = strings.NewReader(dockerfile)
 
 	console.Debug("$ " + strings.Join(cmd.Args, " "))
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -21,6 +21,7 @@ import (
 const dockerignoreBackupPath = ".dockerignore.cog.bak"
 const weightsManifestPath = ".cog/cache/weights_manifest.json"
 const bundledSchemaFile = ".cog/openapi_schema.json"
+const bundledSchemaPy = ".cog/openapi_schema.py"
 
 // Build a Cog model from a config
 //
@@ -28,8 +29,9 @@ const bundledSchemaFile = ".cog/openapi_schema.json"
 func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache, separateWeights bool, useCudaBaseImage string, progressOutput string, schemaFile string, dockerfileFile string) error {
 	console.Infof("Building Docker image from environment in cog.yaml as %s...", imageName)
 
-	// remove bundled schema file that may be left from previous runs
+	// remove bundled schema files that may be left from previous builds
 	_ = os.Remove(bundledSchemaFile)
+	_ = os.Remove(bundledSchemaPy)
 
 	if dockerfileFile != "" {
 		dockerfileContents, err := os.ReadFile(dockerfileFile)
@@ -117,6 +119,7 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 		schemaJSON = data
 	}
 
+	// save open_api schema file
 	err := os.WriteFile(bundledSchemaFile, schemaJSON, 0o644)
 	if err != nil {
 		return fmt.Errorf("Failed to store bundled schema file #{bundledSchemaFile}: %w", err)
@@ -172,7 +175,7 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 		}
 	}
 
-	if err := docker.BuildAddLabelsAndSchemaToImage(imageName, labels, bundledSchemaFile); err != nil {
+	if err := docker.BuildAddLabelsAndSchemaToImage(imageName, labels, bundledSchemaFile, bundledSchemaPy); err != nil {
 		return fmt.Errorf("Failed to add labels to image: %w", err)
 	}
 	return nil
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -20,12 +20,16 @@ import (
 
 const dockerignoreBackupPath = ".dockerignore.cog.bak"
 const weightsManifestPath = ".cog/cache/weights_manifest.json"
+const bundledSchemaFile = ".cog/openapi_schema.json"
 
 // Build a Cog model from a config
 //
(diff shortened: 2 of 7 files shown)


python/cog/schema.py
```diff
@@ -92,3 +104,28 @@ class TrainingRequest(PredictionRequest):
...
     pass
+
+
+def create_schema_model(openapi_schema_path: str) -> ModuleType:
+    with TemporaryDirectory() as temporary_directory_name:
+        temporary_directory = Path(temporary_directory_name)
+        output = Path(temporary_directory / "model.py")
+        bin_path = get_site_packages_bin_path()
+        command = [
+            f"{bin_path}/datamodel-codegen",
```
technillogue, line 115: I'm going to reiterate that we really shouldn't add datamodel-codegen as a dependency

dkhokhlov: @technillogue 
any specific reasons?

technillogue: https://github.com/replicate/cog/pull/1472#issuecomment-1912749735

(https://github.com/replicate/cog/pull/1472#issuecomment-1881744921)

technillogue: it's quite bad to add dependencies if we can possibly avoid it, and datamodel-code-generator in particular has a great deal of dependencies itself, and we don't need it

dkhokhlov: We do not import this dependency. It is used in packaging to make this tool installed automatically as dependency. Which will is equivalent to installing this tool as a separate package. It is very convenient as it is completely transparent for cloud and local installations. meaning - it does not special steps for development env setup or changing cog or cloud image build process.

nickstenning: I think Sylvie's concern here is that naively adding a new dependency is going to result in that dependency being installed at model setup time in production. That means more work and more time on a path that we're already trying very hard to optimize (model boot).

If I'm understanding correctly, the dependency is only actually used during `cog build`. If so, we should extract those dependencies used only during `cog build` into an `optional-dependencies` block for `build`, and install that at the relevant time during build.

dkhokhlov: @nickstenning 
The new py dependency introduced by this PR is "datamodel-code-generator". It provides tool that dynamically generates small py code with Pydantic types from openapi-schema during **cog predict**. The datamodel-code-generator package is added to the image during the **cog build** to model container. 
Summary:
- the dependency is not loaded into python code during **cog predict** 
- the dependency added to image during **cog build**

technillogue: thank you Nick, that's right

there were 132k setup runs in the past 24h, and each of them runs `pip install cog==whatever`, which checks the status of every dependency. many of them are in the critical path for a user's request. making this process even a second longer is expensive and consequential.

I didn't realize this is at build time only. we could potentially doing this by having an extra like `cog[schemagen]` and only use the extra at build time, but that seems tricky. I could also see potentially bundling schemagen into cog-go to run outside the container, but if you're installing it in the container at all (to run it from inside) that still seems like at least a bit of a problem; if nothing else it makes images bigger (e.g. looking at https://github.com/koxudaxi/datamodel-code-generator/blob/main/pyproject.toml#L43-L64, toml and graphql-core are unlikely to be otherwise used and are biggish packages).

most of all, I feel like this isn't necessary in the first place. as I mentioned in https://github.com/replicate/cog/pull/1472#issuecomment-1881744921, https://github.com/replicate/cog/pull/1056 had a very viable approach https://github.com/replicate/cog/pull/1056/files#diff-11f72ba4721073e0f674b63fc3c2da7aa256e5e08a343367b5ebe11a8629797bR239-R287 using the AST and [pydantic.create_model](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation) that I think could be adapted to parse the json schema. that approach wouldn't require adding new dependencies or even changing the build process beyond copying openapi-schema.json into the container

technillogue: for example, we could vendor in or take inspiration from https://github.com/c32168/dyntamic/blob/main/src/dyntamic/factory.py, or maybe even vendor in only the parts of datamodel-code-generator we need like https://github.com/koxudaxi/datamodel-code-generator/blob/main/datamodel_code_generator/parser/jsonschema.py

nickstenning: Yup, just to emphasise the point @technillogue is making -- even if we discount installing the extra dependency in every built image, we also need to think about how this works in production with old images, where adding a dependency is going to result in that dep being installed over and over again, even though it is never used/needed.

dkhokhlov: @nickstenning 
This PR is not imported or loaded into the code, as we use cli tool to avoid that for that specific reason. 
Hypothetical dependency conflicts, because of the package is installed in the same env,  are very unlikely, will be gated and reported at cog build time - not in production cog predict runs. The [datamodel-code-generator](https://github.com/koxudaxi/datamodel-code-generator/tree/main) is well known in community used and sponsored and very well tested, with good test coverage. Has pretty limited dependencies not overlapping with typical ML setups.

nickstenning: > I propose to do that in next PR.

What are you actually proposing doing? I'm not super keen on merging this to main (and thus blocking release) without a clear idea of what we're going to do here.

dkhokhlov: @nickstenning 
> thus blocking release
ok, the proposal is removed then . the work will continue in this PR.
>  I have an idea how to completely remove possibility of conflicts in "installation" dependencies.
Ongoing. The final result:
- no additional 3rd party dependency installed in python env used in cog predict run.

dkhokhlov: uploaded new version:
- removed dependency
- no parsing  in cog predict (faster)

pkg/docker/build.go
```diff
@@ -74,7 +74,10 @@ func BuildAddLabelsToImage(image string, labels map[string]string) error {
 	args = append(args, ".")
 	cmd := exec.Command("docker", args...)
 
-	dockerfile := "FROM " + image
+	dockerfile := "FROM " + image + "\n"
+	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
+	dockerfile += "RUN  /opt/venv/tools/bin/datamodel-codegen --input-file-type openapi --input " + bundledSchemaFile +
```
nickstenning, line 79: I like the approach, but I don't think this needs to happen inside the model and I don't think it should be too hard for us to do it outside.

Here's my suggestion:

1. We start this process with a built cog image. 
2. We extract an OpenAPI schema from that image. 
3. We run a one-shot docker container running [`koxudaxi/datamodel-code-generator`](https://hub.docker.com/r/koxudaxi/datamodel-code-generator) to generate the python file.
4. We copy that python file back into the finalised container, and add the labels.

nickstenning (commented):
I like the overall approach, and I have no problem with using `datamodel-code-generator`, but I think we're missing a trick here as I've alluded to in the comment above.

The key point is that anyone running `cog build` must have a working Docker environment anyway, so we can use that to run arbitrary stuff in containers as part of the build. We don't need to install `datamodel-code-generator` inside the built model image as part of the process.

pkg/docker/build.go
```diff
@@ -74,7 +74,14 @@ func BuildAddLabelsToImage(image string, labels map[string]string) error {
 	args = append(args, ".")
 	cmd := exec.Command("docker", args...)
 
-	dockerfile := "FROM " + image
+	dockerfile := "FROM " + image + "\n"
+	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
+	const env_path = "/tmp/venv/tools/"
```
nickstenning, line 79:
```suggestion
	env_path := "/tmp/venv/tools/"
```

(Inline const in Go doesn't achieve anything and looks very odd.)

pkg/image/build.go
```diff
@@ -20,13 +20,19 @@ import (
 
 const dockerignoreBackupPath = ".dockerignore.cog.bak"
 const weightsManifestPath = ".cog/cache/weights_manifest.json"
+const bundledSchemaFile = ".cog/openapi_schema.json"
+const bundledSchemaPy = ".cog/openapi_schema.py"
```
nickstenning, line 24:
```suggestion
const bundledSchemaPy = ".cog/schema.py"
```

It's not an OpenAPI schema.

python/cog/schema.py
```diff
@@ -1,9 +1,16 @@
...
+import os.path
+import sys
 import typing as t
 from datetime import datetime
 from enum import Enum
+from types import ModuleType
 
 import pydantic
 
+OPENAPI_SCHEMA_PY = ".cog/openapi_schema.py"
```
nickstenning, line 12:
```suggestion
BUNDLED_SCHEMA_PATH = ".cog/schema.py"
```

python/cog/schema.py
```diff
@@ -92,3 +99,14 @@ class TrainingRequest(PredictionRequest):
...
+
+def create_schema_model(openapi_model_py: str) -> ModuleType:
+    module_name = os.path.basename(openapi_model_py).rstrip(".py")
+    spec = importlib.util.spec_from_file_location(module_name, openapi_model_py)
+    assert spec is not None
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    sys.modules[module_name] = module
+    spec.loader.exec_module(module)
+    return module
```
nickstenning, line 112:
```suggestion
def create_schema_model() -> Optional[ModuleType]:
    if not os.path.exists(BUNDLED_SCHEMA_PATH):
        return None
    spec = importlib.util.spec_from_file_location("cog.bundled_schema", openapi_model_py)
    assert spec is not None
    module = importlib.util.module_from_spec(spec)
    assert spec.loader is not None
    sys.modules[module_name] = module
    spec.loader.exec_module(module)
    return module
```

python/cog/server/http.py
```diff
@@ -127,10 +127,16 @@ async def start_shutdown() -> Any:
...
+        # use openapi schema if schema file exists in model dir (cloud/k8s run)
+        if os.path.exists(schema.OPENAPI_SCHEMA_PY):
+            log.info(f"using {schema.OPENAPI_SCHEMA_PY}")
+            schema_model = schema.create_schema_model(schema.OPENAPI_SCHEMA_PY)
+            InputType = schema_model.Input
+            OutputType = schema_model.Output
+        else:
+            predictor = load_predictor_from_ref(predictor_ref)
+            InputType = get_input_type(predictor)
+            OutputType = get_output_type(predictor)
```
nickstenning, line 139:
```suggestion
        # use bundled schema if it exists
        schema_module = schema.create_schema_module()
        if schema_module is not None:
            log.info("using bundled schema")
            InputType = schema_module.Input
            OutputType = schema_module.Output
        else:
            predictor = load_predictor_from_ref(predictor_ref)
            InputType = get_input_type(predictor)
            OutputType = get_output_type(predictor)
```

python/cog/server/http.py
```diff
@@ -166,10 +172,16 @@ async def wrapped(*args: "P.args", **kwargs: "P.kwargs") -> "T":
...
+            # use openapi schema if schema file exists in model dir (cloud/k8s run)
+            if os.path.exists(schema.OPENAPI_SCHEMA_PY):
+                log.info(f"using {schema.OPENAPI_SCHEMA_PY}")
+                schema_model = schema.create_schema_model(schema.OPENAPI_SCHEMA_PY)
+                TrainingInputType = schema_model.TrainingInputType
+                TrainingOutputType = schema_model.TrainingOutputType
+            else:
+                trainer = load_predictor_from_ref(trainer_ref)
+                TrainingInputType = get_training_input_type(trainer)
+                TrainingOutputType = get_training_output_type(trainer)
```
nickstenning, line 184:
```suggestion
            # use bundled schema if it exists
            schema_module = schema.create_schema_module()
            if schema_module is not None:
                log.info("using bundled schema")
                TrainingInputType = schema_module.TrainingInputType
                TrainingOutputType = schema_module.TrainingOutputType
            else:
                trainer = load_predictor_from_ref(trainer_ref)
                TrainingInputType = get_training_input_type(trainer)
                TrainingOutputType = get_training_output_type(trainer)
```

nickstenning (approved):
I've made some suggestions for minor tweaks which I think we should make -- mostly for clarity and encapsulation -- but otherwise this LGTM.

Once you've applied (or at least considered) the suggestions, feel free to merge.

PR #1512, opened 2024-02-02
async release channel

technillogue:
this PR tracks the `async` release channel for the 0.10.0 release


technillogue:
steps to carve up:

* drop race and add connection (easy?)
* runner refactor. I still want to reconsider how radical of a change I need to do to here, I don't know how to proceed
* allow concurrency

Issue #1518, opened 2024-02-07
checksum mismatch for 0.9.4 release


stefanb:
👋 it looks like 0.9.4 release got re-created, raise this issue to confirm it. Thanks!

relates to 
* https://github.com/Homebrew/homebrew-core/pull/162013

This affects
* https://github.com/Homebrew/homebrew-core/pull/161976
* https://github.com/Homebrew/homebrew-core/pull/157782

cc: @mattt, @zeke: can you confirm that retagging was done on 0.9.4 ?

zeke:
I'm not sure. cc @technillogue

chenrui333:
Another way to get away with this is maybe publishing the source tarball into pypi.

<img width="1211" alt="image" src="https://github.com/replicate/cog/assets/1580956/8fb50189-05ec-4616-9ec9-5b16fbc496fd">

PR #1522, opened 2024-02-08
remove unstable git archival config

zeke:
This PR is an attempt to change our git configuration to prevent unintended changes in checksums as we create new releases.

From @Bo98 in https://github.com/Homebrew/homebrew-core/pull/162013#issuecomment-1934175795

> The `%D` format used in https://github.com/replicate/cog/blob/a81c7a4338be7596e64f3c3efe363852d6a4193d/python/.git_archival.txt#L4 is not a stable format and can change depending on `HEAD`.
> 
> `ref-names` looks like it is actually only necessary to have if using a Git older than 2.32 - `describe-name` is used instead on modern Git. I'm not sure if that's a concern for `cog` or not, but removing `ref-names` could be the fix here. An alternative would be to upload your own stable source archive on releases.

For the record, I Have No Idea What I'm Doing here. 🐶 Not sure if this is the right approach, and just trying to unblock our Homebrew releases. Relying on smarter people to weigh in on whether this is the right change. 🙏🏼 

Resolves https://github.com/replicate/cog/issues/1518


stefanb:
@zeke, is it expected that this change will only affect the hashes of the future cog release archives or also the existing ones (eg current last release 0.9.4)?

nickstenning:
I'm not sure I understand what has happened here. I get that the `export-subst` configuration produces output that depends on what `HEAD` points to, but what I don't understand is how the source tarball actually changed?

Did it change because GitHub changed how the tarballs are generated, i.e. they previously were on `main` when they generate the export, but now they check out the tag itself?

nickstenning:
> Is it expected that this change will only affect the hashes of the future cog release archives or also the existing ones (eg current last release 0.9.4)?

As far as I can tell it's the `export-subst` configuration _at the relevant treeish_ that is used during archive, and not the one checked out, so changing this should only change the hashes of future source tarballs.

nickstenning:
Looking into this I'm ever more convinced that this _is_ a change in how GitHub generates the tarballs.

According to [the history of `cog.rb`](https://github.com/Homebrew/homebrew-core/commits/9b8f8cbb4951fab19f3e5449e8f8e486fa453cf1/Formula/c/cog.rb):

1. The tarball for `v0.9.3` [had a SHA256 hash](https://github.com/Homebrew/homebrew-core/commit/54c0bfbecedbd2543f465663601483f56a899c54) of `bb3329b6c7daeef166501883946231a58675770ba624e1450c4be00b549be77a`
2. The tarball for `v0.9.2` [had a SHA256 hash](https://github.com/Homebrew/homebrew-core/commit/7dd5d6a0f109716df643f599b99b50bba1876424) of `853befc8934b44e2b1ae6088aad16cc4ad9d5c22f19ed754fc576c0f64aed21f`.

Neither of those hashes are correct any more:

```
~ $ curl -SsL -o- https://github.com/replicate/cog/archive/refs/tags/v0.9.3.tar.gz | openssl sha256
SHA2-256(stdin)= 5a97356b165a8be41241e0ffc9d769c594166ef8b4d537301f7f87af4d7184b8
~ $ curl -SsL -o- https://github.com/replicate/cog/archive/refs/tags/v0.9.2.tar.gz | openssl sha256
SHA2-256(stdin)= ea97dd086b100bb804c99ddaed8d352a6d276ca6e606f5410138feeaa201c0c0
```

So while the solution here for us is to change our `export-subst` configuration to use a more stable output, the problem here was created by GitHub [yet again](https://github.blog/2023-02-21-update-on-the-future-stability-of-source-code-archives-and-hashes/) changing the data in the automatically generated tarballs, albeit this time in a way that affects fewer people -- namely those who have `export-subst` configurations that are sensitive to (e.g.) Git version.

nickstenning:
I pushed a commit to this branch to move `.git_archival.txt` to the root of the repository. It turns out that the relevant `setuptools_scm` code (to generate the right version for the Python package when installed from an archive tarball) has been broken since https://github.com/replicate/cog/commit/d86fb17240d8ac5bb697c1f24dce56bcc063cbdc.

nickstenning (approved):
LGTM

nickstenning:
I'm going to go ahead and get this merged. Thank you for picking it up @zeke :)

PR #1530, opened 2024-02-12
omnibus actual concurrency and major refactor

technillogue:
this should be reviewable at last. I'm mostly interested in whether the changes are comprehensible/legible and what comments I can add. there's a grab-bag of random changes like a /ready route, predictor.log, cancellation fixes, etc and the core change of moving worker into runner. I'm very open to reconsidering this (e.g. call it worker instead and maybe move some of it into http), but not right now. once this is merged, the plan is to gradually cut small changes from the async branch to merge into main, and review those changes more thoroughly.

<details>
<summary>original description</summary>
I originally tried to split up my work in #1499 and #1508 as "refactor runner + add concurrency" and "fix uploads/downloads" but ended up interleaving these changes. this PR will just be the overall changeset for now, and hopefully as this coalesces more it'll be clear how to carve it up into separate changesets

major points:
* add `concurrency` to cog.yaml
* use httpx for everything except URLFile, pull out all the client code from everywhere else
* completely rework URLPath 
* very dirty hack to unblock us on large file uploads
* merge worker into runner (maybe the other way around would be better?)
</details>


yorickvP (commented):
didn't look at the tests yet

pyproject.toml
```diff
@@ -15,6 +15,8 @@ dependencies = [
   # intentionally loose. perhaps these should be vendored to not collide with user code?
   "attrs>=20.1,<24",
   "fastapi>=0.75.2,<0.99.0",
+  # this version specification is pretty arbitrary, and we may not need http2
+  "httpx[http2]>=0.25.0,<0.27",
```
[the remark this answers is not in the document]
technillogue: what should it be?

pkg/config/validator.go
```diff
@@ -99,6 +99,7 @@ func getDescription(err validationError) string {
 		}
 		return err.child.Description()
 	}
+	fmt.Printf("err.parent.Description(): %s\n", err.parent.Description())
```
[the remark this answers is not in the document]
technillogue: I don't remember why I added it tbh

python/cog/server/clients.py
```diff
@@ -0,0 +1,245 @@
...
+class ClientManager:
+    def __init__(self) -> None:
+        # self.file_url = upload_url
+        self.webhook_client = httpx_webhook_client()
+        self.retry_webhook_client = httpx_retry_client()
+        self.file_client = httpx_file_client()
+        self.download_client = httpx.AsyncClient(follow_redirects=True)
+
+    async def aclose(self) -> None:
+        await self.webhook_client.aclose()
```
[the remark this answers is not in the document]
technillogue: this doesn't actually get called yet anyway and I'm not sure there's a significant difference in what close does

python/cog/server/clients.py
```diff
@@ -0,0 +1,245 @@
...
+        await self.download_client.aclose()
+
+    # webhooks
+
+    async def send_webhook(
+        self, url: str, response: Dict[str, Any], event: WebhookEvent
+    ) -> None:
+        if Status.is_terminal(response["status"]):
+            # For terminal updates, retry persistently
+            await self.retry_webhook_client.post(url, json=response)
```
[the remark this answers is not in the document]
technillogue: how would you carry the request flag through to the transport? also these are only over localhost so it doesn't really matter

python/cog/server/clients.py
```diff
@@ -0,0 +1,245 @@
...
+    def make_webhook_sender(
+        self, url: Optional[str], webhook_events_filter: Collection[WebhookEvent]
+    ) -> WebhookSenderType:
+        throttler = ResponseThrottler(response_interval=_response_interval)
+
+        async def sender(response: Any, event: WebhookEvent) -> None:
+            if url and event in webhook_events_filter:
+                if throttler.should_send_response(response):
+                    await self.send_webhook(url, response, event)
+                    throttler.update_last_sent_response_time()
```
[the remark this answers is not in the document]
technillogue: what, update_last_sent_response_time? I think if send_webhook fails we don't need to update that, but maybe I've missed something

python/cog/server/clients.py
```diff
@@ -0,0 +1,245 @@
...
+        # TODO: if file size is >1MB, show upload throughput
+        resp.raise_for_status()
+
+        # strip any signing gubbins from the URL
+        final_url = urlparse(str(resp.url))._replace(query="").geturl()
+
+        return final_url
+
+    # this previously lived in json.upload_files, but it's clearer here
+    async def upload_files(self, obj: Any, url: Optional[str]) -> Any:
```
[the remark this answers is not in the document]
technillogue: really it should be T -> T

python/cog/server/http.py
```diff
@@ -334,23 +323,15 @@ async def _predict(
...
 
-        response_object = response.dict()
-        response_object["output"] = upload_files(
-            response_object["output"],
-            upload_file=lambda fh: upload_file(fh, request.output_file_prefix),  # type: ignore
-        )
-
-        # FIXME: clean up output files
-        encoded_response = jsonable_encoder(response_object)
+        encoded_response = jsonable_encoder(response.dict())
```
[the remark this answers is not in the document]
technillogue: runner.PredictionEventHandler.set_output/append_output calls _upload_files called upload_files

pyproject.toml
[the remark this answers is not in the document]
mattt: FWIW, Replicate's Python client uses `httpx>=0.21.0,<1`, and I haven't heard about that causing problems for anyone in the 6 months since we switched to that from requests.

technillogue:
this should be reviewable at last. I'm mostly interested in whether the changes are comprehensible/legible and what comments I can add. there's a grab-bag of random changes like a /ready route, predictor.log, cancellation fixes, etc and the core change of moving worker into runner. I'm very open to reconsidering this (e.g. call it worker instead and maybe move some of it into http), but not right now. once this is merged, the plan is to gradually cut small changes from the async branch to merge into main, and review those changes more thoroughly.

pkg/config/config.go
```diff
@@ -50,10 +50,11 @@ type Example struct {
...
 type Config struct {
-	Build   *Build `json:"build" yaml:"build"`
-	Image   string `json:"image,omitempty" yaml:"image"`
-	Predict string `json:"predict,omitempty" yaml:"predict"`
-	Train   string `json:"train,omitempty" yaml:"train"`
+	Build       *Build `json:"build" yaml:"build"`
+	Image       string `json:"image,omitempty" yaml:"image"`
+	Predict     string `json:"predict,omitempty" yaml:"predict"`
+	Train       string `json:"train,omitempty" yaml:"train"`
+	Concurrency int    `json:"concurrency,omitempty" yaml:"concurrency"`
```
nickstenning, line 57: Part of me wonders if this should be an object with a single key (`max`) so that it can be extended in future if needed.

python/cog/server/http.py
```diff
@@ -240,7 +242,21 @@ async def healthcheck() -> Any:
...
             health = app.state.health
         setup = attrs.asdict(app.state.setup_result) if app.state.setup_result else {}
-        return jsonable_encoder({"status": health.name, "setup": setup})
+        activity = runner.activity_info()
+        return jsonable_encoder(
+            {"status": health.name, "setup": setup, "concurrency": activity}
+        )
+
+    @app.get("/ready")
+    async def ready() -> Any:
```
nickstenning, line 251: What does this do that the healthcheck endpoint doesn't?

python/cog/server/http.py
```diff
@@ -326,27 +334,20 @@ async def _predict(
...
+        # # in the previous implementation those would then get encoded as strings
+        # # however the changes to Path and File break this and return the filename instead
+        # try:
+        #     prediction = await async_result
+        #     # we're only doing this to catch validation errors
+        #     response = PredictionResponse(**prediction.dict())
+        #     del response
+        # except ValidationError as e:
+        #     _log_invalid_output(e)
+        #     raise HTTPException(status_code=500, detail=str(e)) from e
```
nickstenning, line 348: What are the implications of commenting this out?

python/cog/server/runner.py
nickstenning, line 1: As mentioned when we spoke, it seems quite confusing to me to take 90% of the code from `Worker` and put it in `PredictionRunner` rather than making `Worker` do what we need.

python/cog/server/worker.py
```diff
@@ -355,26 +178,34 @@ def _loop_sync(self) -> None:
                 break
             if isinstance(ev, PredictionInput):
                 self._predict_sync(ev)
+            elif isinstance(ev, Cancel):
+                pass  # we should have gotten a signal
```
nickstenning, line 182: I don't think I understand this. Are cancel events sent down the pipe or are they communicated via signal? It should be one or the other, not both, and if we're sending them down the pipe but then ignoring them... why?

python/tests/server/test_worker.py
```diff
@@ -335,21 +341,22 @@ async def test_cancel_idempotency():
...
 
-        async for event in w.predict({"sleep": 0.5}, poll=0.01):
+        async for event in w.predict(input1, poll=0.01):
             # We call cancel a WHOLE BUNCH to make sure that we don't propagate
             # any of those cancelations to subsequent predictions, regardless
             # of the internal implementation of exceptions raised inside signal
             # handlers.
             for _ in range(100):
-                w.cancel()
+                w.cancel(input1.id)
```
nickstenning, line 352: Yikes, does this mean we're relying on `Worker.predict` to modify its arguments in order to communicate prediction ID back to the caller? That seems very hacky.

python/tests/server/test_worker.py
nickstenning, line 1: Am I missing something? You deleted the `Worker` class so how come all these tests aren't failing?

pkg/config/config.go
technillogue: that sounds like a pretty reasonable idea, what other keys do you think it might have?

python/cog/server/http.py
technillogue: it's suitable for use as a readiness check with e.g. fly, since healthcheck returns 200 during setup or while predictions are running

python/cog/server/http.py
technillogue: outputs are not validated, no error is raised if you return a string instead of an int. this turned out to be a bottleneck with sufficient throughput in profiling, and also led to a bug mentioned in the comments where Path is converted to str but then validation converts it back to Path and this round-trip no longer works the way it used to

python/cog/server/runner.py
technillogue: unfortunately, I'm not sure how to find the line for all the enter_predict/exit_predict stuff if it's bound by being def () -> AsyncIterator[Event]. I would love more detailed input on this.

python/cog/server/worker.py
technillogue: because the parent doesn't necessarily know if the predictor is async or not, we send both the event and the signal. sync predictors ignore the event, async predictors ignore the signal.

python/tests/server/test_http_output.py
```diff
@@ -7,11 +7,11 @@
 
 from .conftest import uses_predictor, uses_predictor_with_client_options
 
-
-@uses_predictor("output_wrong_type")
-def test_return_wrong_type(client):
-    resp = client.post("/predictions")
-    assert resp.status_code == 500
+# it's not the worst idea to validate outputs but it's slow and not required
```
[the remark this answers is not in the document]
technillogue: see previous comment, this is also related to File(upload(File(x))) != x

python/tests/server/test_worker.py
technillogue: no? the prediction ID must be fixed before it makes it to Worker.predict. it is autogenerated if not set on PredictionRequest and then you're supposed to PredictionInput.from_request(request...)

python/tests/server/test_worker.py
technillogue: it's due to pytest.skip(allow_module_level=True)

Issue #1216, opened 2023-07-19
Switch to Pydantic 2 and `Annotated`


bfirsh:
Pydantic 2 uses [annotated types](https://github.com/annotated-types/annotated-types) to annotate arguments instead of the gross magical default values.

Would look something like...

```python
def predict(input: Annotated[File, Input(...)] -> File:
    # ...
```

... or something.

It also means we can annotate outputs.

jeremyadamsfisher:
Bumping this, also running into this issue since I wrote my configuration validation with the pydantic 2 syntax -- need to refactor to be compatible with cog

sCarita:
bumping this. we need to use an underlaying package on our image that depends on pydantic v2.

Issue #1186, opened 2023-07-07
Update to Pydantic 2 and FastAPI 0.100.0


mattt:
From https://fastapi.tiangolo.com/release-notes/#01000:

> Pydantic version 2 has the core re-written in Rust and includes a lot of improvements and features... In some cases, for pure data validation and processing, you can get performance improvements of 20x or more.

nickstenning:
Cross-linking this comment from the dependabot PR so we don't lose it: https://github.com/replicate/cog/pull/1187#issuecomment-1631174754

mattt:
Related to #1216

mattkindy:
This will prevent use of packages (e.g. https://github.com/outlines-dev/outlines ) that have dependencies on pydantic and have upgraded to v2 in their most recent updates.

sinopec:
really need it ,specially when some  llm inference engine requires pydantic 2....
annoying conflicts.....

dkhokhlov:
> really need it ,specially when some llm inference engine requires pydantic 2.... annoying conflicts.....
@sinopec - could you give model examples for testing?

jd7h:
I deployed a model today that required pydantic==2.5.3. While we're waiting for the update to newer versions of pydantic and fastAPI, here are some possible workarounds. Choose one depending on the speed you require and how tightly integrated your code is with the incompatible versions of pydantic and fastAPI:

## Supplemental venv
Create a virtual environment with venv to supplement the packages installed from `cog.yaml`
```
$ cog run bash
# python -m venv --system-site-packages venv_pydantic
# source venv_pydantic/bin/activate
# pip install pydantic==2.5.3
```
predict.py:
```
import subprocess
subprocess.check_call(["venv_pydantic/bin/python", "this_script_requires_pydantic2"])
```
Drawback: You can't easily pre-load your modelweights in `setup()`.

## Poetry
2. in cog.yaml, git clone your code and install it with poetry
cog.yaml:
```
python_packages:
  - poetry

run:
  - git clone myrepo.git && cd myrepo && poetry install
```
in predict.py:
```
import subprocess
subprocess.check_call("cd /myrepo && python -m poetry run python -m script.py".split())
```
Drawback: You can't easily pre-load your modelweights in `setup()`.

## Monkeypatching
3. You can Mock() or otherwise monkeypatch parts of incompatible libraries. 
predict.py:
```
from unittest.mock import Mock

sys.modules['pydantic.computed_field'] = Mock()
[...]

PR #1539, opened 2024-02-19
Switch kin-openapi to libopenapi

yorickvP:
Required for #1482, #1186, #1216


yorickvP:
Sadly, libopenapi doesn't discover the invalidness of https://github.com/replicate/cog/blob/main/test-integration/test_integration/fixtures/invalid-int-project/predict.py

yorickvP:
https://github.com/pb33f/libopenapi-validator/issues/53

andreasjansson:
I think it's okay to delete that test case. It's a pretty rare case, and wouldn't be hard to upstream a change to libopenapi to handle that case.

andreasjansson:
@yorickvP could you describe in a bit more depth why we need to switch to libopenapi? It's not obvious to me.

yorickvP:
This comment contains a summary: https://github.com/replicate/cog/issues/1186#issuecomment-2011832395

We could try to get away with kin-openapi, that would just require fixing the test for the version in https://github.com/replicate/cog/pull/1552

zeke:
cc @nickstenning who's got a fair bit of experience working with kin-openapi

nickstenning:
We use `kin-openapi` in the `api` repository to process the schemas generated by cog models, and I would probably want to see us using the same code both here and in production.

What is actually wrong with `kin-openapi`?

yorickvP:
> What is actually wrong with `kin-openapi`?

No OpenAPI v3.1 support: https://github.com/getkin/kin-openapi/issues/230#issuecomment-1002659224

However, it seems we happen not to use any of the unsupported 3.1 features. Would be good to check somehow.

nickstenning:
Our own API's spec is 3.1, and we use `kin-openapi` to validate it in the `api` codebase. Is the lack of official 3.1 support actually causing us any problems?

yorickvP:
Here's a list of changes: https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0 . My concern here is that these schemas are generated from user input by pedantic, so we don't necessarily control it when they break.
The thing we should mainly test is `exclusiveMinimum` (pydantic's `gte`).

mattt:
Closing in favor of #1687, which rewrites the OpenAPI 3.1 spec version emitted by FastAPI to OpenAPI 3.0.

PR #1546, opened 2024-02-24
Define `Secret` type

mattt:
This PR defines a new `Secret` type in Cog's Python package, which model authors can use to signify that an input holds sensitive information (like a password or API token).

The `Secret` type inherits from [Pydantic's `SecretStr`](https://docs.pydantic.dev/1.10/usage/types/#secret-types):

> You can use the `SecretStr` and the `SecretBytes` data types for storing sensitive information that you do not want to be visible in logging or tracebacks. `SecretStr` and `SecretBytes` can be initialized idempotently or by using `str` or `bytes` literals respectively. The `SecretStr` and `SecretBytes` will be formatted as either `'**********'` or `''` on conversion to json. 

```python
from cog import BasePredictor, Secret


    def predict(self, secret: Secret) -> None:
        # Prints '**********'
        print(secret)        

        # Use get_secret_value method to see the secret's content.
        print(secret.get_secret_value())

```

A predictor's `Secret` inputs are represented in OpenAPI with the following schema:

```json
{
  "type": "string",
  "format": "password",
  "x-cog-secret": true,
}
```

Replicate will treat secret inputs differently throughout its system. For example, a prediction on Replicate will redact secret values in its JSON representation and log lines.

### DIFF
--- a/python/cog/__init__.py
+++ b/python/cog/__init__.py
@@ -1,7 +1,7 @@
 from pydantic import BaseModel
 
 from .predictor import BasePredictor
-from .types import ConcatenateIterator, File, Input, Path
+from .types import ConcatenateIterator, File, Input, Path, Secret
 
 try:
     from ._version import __version__
@@ -17,4 +17,5 @@ __all__ = [
     "File",
     "Input",
     "Path",
+    "Secret",
 ]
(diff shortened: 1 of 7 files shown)


technillogue:
what I really like about this is that it gives us to flexibility to potentially only redact and not store prediction secrets

technillogue (approved):

Issue #1431, opened 2023-12-13
Deprecate python 3.7


technillogue:
Python 3.7 stopped receiving security support in [June](https://endoflife.date/python). We should release a version of cog that deprecates 3.7 (and supports 3.12), then stop letting people push new 3.7 models. Ideally we should push people to use 3.11 and not just 3.8 so that things get faster.

technillogue:
https://github.com/replicate/cog/issues/1481

nickstenning:
We can't drop support for Python 3.7 in one go, as we have models running in production that have Python 3.7. 

As @technillogue alludes to above, we first need to cut a cog point release that spits out deprecation warnings (likely to the prediction logs to maximise the chance that people see them!), then we need to prevent new models with 3.7 from being pushed.

Depending on the scale of the problem, we may also need to put some banners in people's accounts on the website.

mattt:
Is the problem that models running Python 3.7 will fail when we try to install the latest version of Cog at startup? If so, could we skip that step? Or else change the logic to install the latest 3.7 compatible version?

nickstenning:
The entire cluster runs the same version of Cog. I don't think we want to change that right now.

technillogue:
the shortest-term change is to just stop letting people push new models with 3.7. adding deprecation warnings in prediction logs sounds great. after a few months it's entirely possible we will no longer be breaking anything important that's still being used

tempusfrangit:
We no longer support python 3.10 as of the upcoming 0.17.0 release.

PR #1550, opened 2024-02-25
feat: deprecate python 3.7

dkhokhlov:
- do not allow python 3.7 in models
- added python 3.12 to CI verification
- added test:  test_integration/test_build.py::test_python_37_depricated

Ref: issue #1431

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     defaults:
       run:
         shell: bash
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,7 @@ lint-go:
 lint-python:
 	$(RUFF) python/cog
 	$(RUFF) format --check python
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [{ name = "Replicate", email = "team@replicate.com" }]
 license.file = "LICENSE"
 urls."Source" = "https://github.com/replicate/cog"
 
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
technillogue, line 13: this we can't do yet

python/cog/types.py
```diff
@@ -124,9 +124,8 @@ def convert(self) -> Path:
 
     def unlink(self, missing_ok: bool = False) -> None:
         if self._path:
-            # TODO: use unlink(missing_ok=...) when we drop Python 3.7 support.
             try:
-                self._path.unlink()
+                self._path.unlink(missing_ok=missing_ok)
```
technillogue, line 128: this is a strange piece of code but also can't actually be dropped yet

technillogue (changes requested):
otherwise looks solid

dkhokhlov:
@technillogue 
It's interesting to note that Python 3.12 runs tests slower. I had to increase the deadline from 1000 to 1500 in the CI profile only.

technillogue (approved):

PR #1553, opened 2024-02-29
feat: new approach for fast validation - fast cold boot

dkhokhlov:
- fast cold boot
- new approach for fast validation (elimination of redundant heavy imports of predict/train py files into cog at runtime)
- generates stripped down (slim) versions of predict/train py files using dynamic Python code introspection
- Pydantic model creation is based on Python function signatures in slim versions of predict/train py files, and not using openapi schema as in previous attempt in #1503
- now 100% compatible with existing and future cog types and existing code
- removed schema py generation and bundling during cog build
- supports existing models (no need to rebuild) as openapi schema is not used anymore 
- will fallback to slow loader in cases when predict/train function signatures refer to slow imports, like dynamically calculated  defaults. 

ref: #1503


technillogue:
wow this is a lot better! thank you!

technillogue (commented):
this is exactly on the right track. I think my only concern is output classes. there are some tests in https://github.com/replicate/cog/blob/main/python/tests/server/test_http.py#L206-L214 checking that the static schema matches the slow path schema, maybe that would be good. otherwise if this works I would be _delighted_ to merge this

dkhokhlov:
> this is exactly on the right track. I think my only concern is output classes. there are some tests in https://github.com/replicate/cog/blob/main/python/tests/server/test_http.py#L206-L214 checking that the static schema matches the slow path schema, maybe that would be good. otherwise if this works I would be _delighted_ to merge this

Custom Output  classes are supported. There new test for Output class in train  path. 
https://github.com/replicate/cog/pull/1553/files#diff-07c3ff16d436a16bf4d078d79ee50429826a5410896564edc1569b210a62aee5R19
Let me know if you are thinking about different test.

technillogue:
some examples:


https://github.com/replicate/cog-sdxl/blob/main/predict.py#L255-L291

https://github.com/replicate/cog-llama-template/blob/main/predict.py#L146

https://github.com/chenxwh/insanely-fast-whisper/blob/main/predict.py#L90

dkhokhlov:
> some examples:
> 
> https://github.com/replicate/cog-sdxl/blob/main/predict.py#L255-L291
> 
> https://github.com/replicate/cog-llama-template/blob/main/predict.py#L146
> 
> https://github.com/chenxwh/insanely-fast-whisper/blob/main/predict.py#L90

These cases use dynamic defaults from imports that are not in [COG_IMPORT_MODULES](https://github.com/replicate/cog/pull/1553/files#diff-ff38d309317b1ce58085e3244fcec48906907a427c90c035812a10678fbac7a9R5) list, so they will fall back to slow loader with this log message:
```
 "[predict] fast loader failed: name 'LANGUAGES' is not defined"
 "[predict] falling back to slow loader"
```

python/cog/server/http.py
```diff
@@ -127,16 +127,9 @@ async def start_shutdown() -> Any:
...
-        schema_module = None  # schema.create_schema_module() - breaks complex types, always use slow path for now
-        if schema_module is not None:
-            log.info("using bundled schema")
-            InputType = schema_module.Input
-            OutputType = schema_module.Output
-        else:
-            predictor = load_predictor_from_ref(predictor_ref)
-            InputType = get_input_type(predictor)
-            OutputType = get_output_type(predictor)
+        predictor = load_slim_predictor_from_ref(predictor_ref)
```
nickstenning, line 130: Do we not want to fall back to the old way if this blows up?

python/tests/server/test_code_xfroms.py
nickstenning, line 1: The file name has a typo.

python/tests/server/test_code_xfroms.py
```diff
@@ -0,0 +1,77 @@
+import os
+import pytest
+import sys
+import uuid
+import cog.code_xforms as code_xforms
+
+g_module_dir = os.path.dirname(os.path.abspath(__file__))
+
+
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires python3.9 or higher")
```
nickstenning, line 10: What requires Python 3.9 or higher? The test or the code? If it's the code then we definitely need a fallback mechanism because we have lots of models in production still running 3.8, and some running 3.7.

python/cog/server/http.py
nickstenning: Ahhh, I see that this falls back itself to a full load.

python/tests/server/test_code_xfroms.py
dkhokhlov: code.

nickstenning:
Honestly this approach makes me a bit nervous, but I also see that it has certain advantages. How hard would it be to get it tested against all the predictors already in this repository as fixtures?

python/cog/server/http.py
dkhokhlov: the fallback is inside load_slim_predictor_from_ref with try/catch. the load_full_predictor_from_file is the old(slow) way.

dkhokhlov:
@nickstenning 
I instrumented existing integration tests to check for fallbacks. and after that found a bug in train function-only use case.

test-integration/test_integration/test_predict.py
```diff
@@ -28,6 +30,7 @@ def test_predict_takes_int_inputs_and_returns_ints_to_stdout():
     )
     # stdout should be clean without any log messages so it can be piped to other commands
     assert result.stdout == b"4\n"
+    assert "falling back to slow loader" not in str(result.stderr)
```
nickstenning, line 33: Hmmm, this feels like a very brittle way to test that we're using the "fast loader". I'm also not really sure about us emitting these logs into the setup output that aren't going to mean anything to the people that wrote the models.

I don't think this is blocking but is there perhaps a separate set of standalone tests that look a bit less like integration tests that are reliant on log output?

I was thinking something like:

```python
@pytest.mark.parametrize("fixture", ALL_FIXTURES)
def test_slim_loader_generates_correct_model(fixture):
    model_slim = load_slim(fixture)
    model_full = load_full(fixture)

    assert model_slim.model_dump() == model_full.model_dump()
```

dkhokhlov: @nickstenning 
uploaded 
- disable emitting fast loader logs into the setup 
- added test that compares fast vs slow predict signatures. it goes over all fixtures in server tests.

PR #1554, opened 2024-03-01
feat: upgrade to Pydantic v2

dkhokhlov:
Ref: https://github.com/replicate/cog/issues/1186


dkhokhlov:
needs different approach

PR #1559, opened 2024-03-05
allow setup(self, weights: str)

technillogue:
the behavior of cog.Path and cog.File as setup weights argument is inconsistent with how it works in predict. allowing str is a way out until that's fixed properly

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -90,7 +90,7 @@ def run_setup(predictor: BasePredictor) -> None:
             weights = weights_url
         else:
             raise ValueError(
-                f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
+                f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File, Path and str are supported"
             )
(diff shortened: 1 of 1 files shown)


python/cog/predictor.py
```diff
@@ -85,6 +85,9 @@ def run_setup(predictor: BasePredictor) -> None:
         elif weights_type == CogPath:
             # TODO: So this can be a url. evil!
             weights = cast(CogPath, CogPath.validate(weights_url))
+        # allow people to download weights themselves
+        elif weights_type == str:
+            weights = weights_url
         else:
             raise ValueError(
                 f"Predictor.setup() has an argument 'weights' of type {weights_type}, but only File and Path are supported"
```
anotherjesse, line 93: should we add `str` to the error message :shrug:

anotherjesse:
lgtm except the test is failing

technillogue:
something changed in one our deps and now the test is flaking idk

technillogue:
I reran the tests a few times and it worked. I'll debug why tests are flaking in a different PR.

PR #1561, opened 2024-03-06
feat: support for list in inputs

dkhokhlov:
- added support for *list[File]* and *list[Path]*
- only in cog, still need to [add list support in web UI](https://linear.app/replicate/issue/PRO-1153/add-support-for-list[file]-input-to-web-ui)
- updated docs/python.md

```
$ echo test1 > 1.txt
$ echo test2 > 2.txt
$ cat predict.py
from cog import BasePredictor, Path

   def predict(self, paths: list[Path]) -> str:
       output_parts = [] 
       for path in paths:
           with open(path) as f:
             output_parts.append(f.read())
       return "".join(output_parts)

$ cog predict -i paths=@1.txt -i paths=@2.txt

Running prediction...
test1

test2
```
- Note the repeated inputs with the same name "paths" which constitute the list

Ref: 
- https://github.com/replicate/cog/issues/527
- https://linear.app/replicate/issue/PRO-1153/add-support-for-list[file]-input-to-web-ui


python/cog/command/openapi_schema.py
```diff
@@ -3,6 +3,7 @@
 
 This prints a JSON object describing the inputs of the model.
 """
+
```
nickstenning, line 6:
```suggestion
```

python/cog/schema.py
```diff
@@ -50,9 +50,9 @@ class PredictionRequest(PredictionBaseModel):
     output_file_prefix: t.Optional[str]
 
     webhook: t.Optional[pydantic.AnyHttpUrl]
-    webhook_events_filter: t.Optional[
-        t.List[WebhookEvent]
-    ] = WebhookEvent.default_events()
+    webhook_events_filter: t.Optional[t.List[WebhookEvent]] = (
```
nickstenning, line 53: Changes like this suggest that you don't have your editor set up to do autoformatting the same as `script/format`. Would you look into that?

pkg/predict/input.go
```diff
@@ -1,20 +1,23 @@
...
 
+	"github.com/mitchellh/go-homedir"
 	"github.com/replicate/cog/pkg/util/console"
-	"github.com/replicate/cog/pkg/util/mime"
 )
 
 type Input struct {
 	String *string
 	File   *string
+	Array  *[]interface{}
```
nickstenning, line 20: Two small points: 

1. it's nice to use `any` rather than `interface{}` these days
2. this doesn't need to be a pointer -- slices are already pointer types

pkg/predict/input.go
```diff
@@ -55,19 +59,54 @@ func NewInputsWithBaseDir(keyVals map[string]string, baseDir string) Inputs {
...
+					if err != nil {
+						return keyVals, err
+					}
+					dataURLs = append(dataURLs, dataURL)
+				} else if ok {
+					// Directly use the string if it's not a file path
+					dataURLs = append(dataURLs, str)
+				}
+			}
+			keyVals[key] = dataURLs
```
nickstenning, line 91: It's not a big deal, but it would be more conventional when you know the length of the slice beforehand to do something like

```suggestion
			dataURLs := make([]string, len(input.Array))
			for i, elem := range input.Array {
				if str, ok := elem.(string); ok && strings.HasPrefix(str, "@") {
					filePath := str[1:] // Remove '@' prefix
					dataURL, err := fileToDataURL(filePath)
					if err != nil {
						return keyVals, err
					}
					dataURLs[i] = dataURL
				} else if ok {
					// Directly use the string if it's not a file path
					dataURLs[i] = str
				}
			}
			keyVals[key] = dataURLs
```

pkg/predict/input.go
```diff
@@ -25,15 +28,16 @@ func NewInputs(keyVals map[string]string) Inputs {
...
 			val = val[1:]
-			expandedVal, err := homedir.Expand(val)
-			if err != nil {
-				// FIXME: handle this better?
-				console.Warnf("Error expanding homedir: %s", err)
+			input[key] = Input{File: &val}
+		} else if strings.HasPrefix(val, "[") && strings.HasSuffix(val, "]") {
+			// Handle array of strings
+			var arr []interface{}
+			if err := json.Unmarshal([]byte(val), &arr); err != nil {
```
nickstenning, line 35: Is JSON unmarshalling right here? That means the strings have to be quoted, which is inconsistent with the other types.

For example, I can do

```
cog predict hello-world -i text=world
```

or

```
cog predict resnet -i image=@input.jpg
```

but with this implementation, the following will throw an error

```
cog predict mymodel -i names=[foo,bar]
```

I think there are two possibilities here. Either:

1. do the work to allow individual strings to be either quoted or unquoted, so that they're consistent with the rest of the interface, or 
2. change approach to do arrays by just repeating inputs, e.g. `cog predict ... -i names=one -i names=two`

nickstenning (commented):
I haven't finished going through this yet, but I figure I should release the comments I've made already to get started :)

pkg/predict/input.go
dkhokhlov: I chose repeated inputs. Good idea. Thanks.

python/cog/command/openapi_schema.py
dkhokhlov: this change required by: python -m ruff format --check python
not changing.

python/cog/schema.py
dkhokhlov: this change is required by one of linters in our make liner: 
`python -m ruff python/cog`
and reformatted by: 
`python -m ruff format python/cog/schema.py`

dkhokhlov:
> This is awesome, @dkhokhlov!
> 
> Can you take a shot at documenting this in docs/python.md ?
> 
> Let me know if you want help with that.
@zeke 
added List type section in docs/python.md with cog cli example:
https://github.com/replicate/cog/pull/1561/files#diff-3576f8388f597edf93166fb888bc0c6c7ebca8aba465d941c17b3c1804b99cab

PR #1574, opened 2024-03-12
replace requests with httpx and factor out clients

technillogue:
this is a step towards merging #1530 into the async release channel. it's a repeat of #1508 without actually adding concurrency or the less runner stuff from #1499. hopefully the remainder of #1530 is easier to review once this is done.

as I recall, the main reason we couldn't release async runner was that cog.Path/File using requests would block the event loop and prevent work from advancing. because of that, these are probably the first changes that need to be released into mainline cog to proceed.

* input downloads, output uploads, and webhooks are now handled by ClientManager, which persists for the lifetime of runner, allowing us to reuse connections, which may significantly help with large uploads. 
* although I was originally going to drop output_file_prefix, it's not actually hard to maintain. the behavior is changed now and objects are uploaded as soon as they're outputted rather than after the prediction is completed.
* there's a kind of ugly hack with uploading an empty body to get the redirect instead of making api time out from trying to upload an 140GB file. that can be fixed by implemented an MPU endpoint and/or a "fetch upload url" endpoint.
* the behavior of the non-indempotent endpoint is changed; the id is now randomly generated if it's not provided in the body. this isn't strictly required for this change alone, but is hard to carve out.
* the behavior of Path is changed significantly. see https://www.notion.so/replicate/Cog-Setup-Path-Problem-2fc41d40bcaf47579ccd8b2f4c71ee24


.github/workflows/ci.yaml
```diff
@@ -53,7 +53,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        # python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
```
mattt, line 56:
```suggestion
```

python/cog/server/clients.py
```diff
@@ -0,0 +1,258 @@
...
+def webhook_headers() -> "dict[str, str]":
+    headers = {"user-agent": _user_agent}
+    auth_token = os.environ.get("WEBHOOK_AUTH_TOKEN")
+    if auth_token:
+        headers["authorization"] = "Bearer " + auth_token
+    return headers
+
+
+def httpx_webhook_client() -> httpx.AsyncClient:
+    return httpx.AsyncClient(headers=webhook_headers(), follow_redirects=True)
```
mattt, line 56: Looks like we need to opt-in to enable HTTP/2

```suggestion
    return httpx.AsyncClient(headers=webhook_headers(), follow_redirects=True, http2=True)
```

python/cog/server/clients.py
```diff
@@ -0,0 +1,258 @@
...
+    )
+    # set connect timeout to slightly more than a multiple of 3 to avoid
+    # aligning perfectly with TCP retransmission timer
+    # requests has no write timeout, keep that
+    # httpx default for pool is 5, use that
+    timeout = httpx.Timeout(connect=10, read=15, write=None, pool=5)
+    return httpx.AsyncClient(
+        transport=transport,
+        follow_redirects=True,
+        timeout=timeout,
```
mattt, line 91:
```suggestion
        timeout=timeout,
        http2=True,
```

python/cog/server/clients.py
```diff
@@ -0,0 +1,258 @@
...
+# the same semantics of holding a client for the lifetime of runner
+# also, both are used by PredictionEventHandler
+
+
+class ClientManager:
+    def __init__(self) -> None:
+        self.webhook_client = httpx_webhook_client()
+        self.retry_webhook_client = httpx_retry_client()
+        self.file_client = httpx_file_client()
+        self.download_client = httpx.AsyncClient(follow_redirects=True)
```
mattt, line 106: Extract this into a helper method like the others?

Also, initialize with `http2=True`

python/tests/server/test_webhook.py
```diff
@@ -1,9 +1,12 @@
+import pytest
 import requests
 import responses
 from cog.schema import WebhookEvent
-from cog.server.webhook import webhook_caller, webhook_caller_filtered
+
+#from cog.server.webhook import webhook_caller, webhook_caller_filtered
```
mattt, line 6:
```suggestion
```

python/cog/types.py
```diff
@@ -131,17 +155,44 @@ def unlink(self, missing_ok: bool = False) -> None:
...
+# except... that doesn't really work with httpx?
+
 
 class URLFile(io.IOBase):
     """
     URLFile is a proxy object for a :class:`urllib3.response.HTTPResponse`
     object that is created lazily. It's a file-like object constructed from a
     URL that can survive pickling/unpickling.
+
+    This is the only place Cog uses requests
```
mattt, line 195: What would it take to get rid of requests outright?

Would something like this work?

```python
@property
def __wrapped__(self) -> Any:
    try:
        return object.__getattribute__(self, "__target__")
    except AttributeError:
        url = object.__getattribute__(self, "__url__")
        with httpx.stream("GET", url) as resp:
            resp.raise_for_status()
            resp.raw.decode_content = True
            object.__setattr__(self, "__target__", resp.raw)
            return resp.raw
```

test-integration/test_integration/test_run.py
```diff
@@ -24,7 +24,7 @@ def test_run_with_secret(tmpdir_factory):
     with open(tmpdir / "cog.yaml", "w") as f:
         cog_yaml = """
 build:
-  python_version: "3.8"
+  python_version: "3.9"
```
mattt, line 27: Any reason why this is targeting 3.9 instead of 3.8?

python/cog/types.py
technillogue: I think that would work, I just didn't bother yet because nobody anywhere uses File and this solution is still a little unsatisfying (it would block the event loop). I also suspect requests is always installed because pip depends on it.

mattt: > requests is always installed because pip depends on it 

🤨 

Maybe that's no longer the case? According to [this](https://libraries.io/pypi/pip), pip has no external dependencies.

Not urgent or blocking, but it'd be nice to make a clean break.

technillogue: oh yeah they vendored it in a while ago, I forgot. and I'm mistaken, python:3.11-slim doesn't have requests by default. 

I'd still lean towards leaving it obviously broken rather then seeming okay but actually blocking the event loop. good ways to do this include subprocesses with a pipe, some kind of thread nonsense or a hypothetical pget-py binding

python/cog/server/clients.py
technillogue: I will note there is likely no benefit to using http2 for director, it only really matters for file downloads and maybe uploads from the internet

python/cog/server/clients.py
technillogue: :woman_shrugging: it's one line, all the other ones have more involved configuration

PR #1575, opened 2024-03-13
Increase timeout of flakey test to 1.5s

technillogue:
this PR attempts to debug why hypothesis tests are flaking

### DIFF
--- a/python/tests/server/test_worker.py
+++ b/python/tests/server/test_worker.py
@@ -22,7 +22,7 @@ from hypothesis.stateful import (
 )
 
 # Set a longer deadline on CI as the instances are a bit slower.
-settings.register_profile("ci", max_examples=100, deadline=1000)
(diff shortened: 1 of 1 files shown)


dkhokhlov:
> hypothesis.errors.DeadlineExceeded: Test took 1004.39ms, which exceeds the deadline of 1000.00ms
need to bump deadline in CI profile?

technillogue:
I think it's always the exit_in_predict case, did something make that actually slower in a way we should fix directly?

yorickvP:
I ran it locally and it failed the deadline in some other tests. Re-running them seems to have no issues, though. I vote to increase the deadline to 1.5s and think about it again if this re-surfaces.

yorickvP:
LGTM. Can you cherry pick out the fake commit?

PR #1576, opened 2024-03-13
Better formatting for `pget` instructions in `setup()` docs for Python

cuuupid:
The bulleted list doesn't render well at the moment:

image

Adding enters before the blocks to ƒix that!


zeke (approved):
Good catch. 👍🏼

PR #1581, opened 2024-03-14
Include options for compatgen targets in Usage help

jd7h:
Make the help message of compatgen clearer by explicitly stating the valid options for <target>.

### DIFF
--- a/tools/compatgen/main.go
+++ b/tools/compatgen/main.go
@@ -14,7 +14,7 @@ func main() {
 	var output string
 
 	var rootCmd = &cobra.Command{
-		Use:   "compatgen <target>",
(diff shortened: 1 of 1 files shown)


zeke (approved):
I wasn't familiar with this `{foo|bar|baz}` syntax for describing allowable values in an argument, so I went on a [ChatGPT odyssey](https://chat.openai.com/share/623fbbdf-40e0-4165-b4f7-08056b066e2b) to understand it.

Looks like this is legit! 👍🏼 

Another approach would be to leave it as `compatgen <target>` and expound on the values and their meanings using some combination of `Short`, `Long`, and `Example`.

PR #1582, opened 2024-03-14
feat: deprecate python 3.7 #2

dkhokhlov:
- previous deprecation of Ptyhon 3.7 in models did not have actual check (https://github.com/replicate/cog/pull/1550)
- now the model python minimum version is checked in `cog build` only
- this new approach is compatible with existing already built and/or uploaded models
- fixed integration test
- keep running CI tests on python 3.7

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,7 +50,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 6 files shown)


yorickvP (approved):
Should we also add a check in registry-proxy to stop people from pushing 3.7 models using old cogs? https://github.com/replicate/registry-proxy/blob/main/pkg/server/proxy.go#L138

dkhokhlov:
> to stop people from pushing 3.7 models using old cogs

I do not think it is good idea to do it now. better to leave old cog fallback for now in case user is stuck with 3.7 and need time to port.

yorickvP:
Replicate injects newer cog versions into old images by calling `pip install cog==0.9.4`. Since we no longer test new cog releases on 3.7, we should stop doing this for python3.7 images (or they will stop working).

@technillogue Should we create a 0.9.x branch that we can make releases out of if we ever want do to cog updates that apply to already existing py3.7 images?

technillogue:
we should just keep the 3.7 tests then

PR #1585, opened 2024-03-19
Cog build instructions and cog dockerfile generation

ghost:
Use datetime instead of random for dockerfile generation to distinguish older .cog/tmp/buildX from newer .cog/tmp/buildY

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,8 +132,7 @@ sudo chmod +x /usr/local/bin/cog
 Alternatively, you can build Cog from source and install it with these commands:
 
 ```console
-make
-sudo make install
+make cog
 ```
 
 ## Upgrade
--- a/README.md
+++ b/README.md
@@ -132,7 +132,8 @@ sudo chmod +x /usr/local/bin/cog
 Alternatively, you can build Cog from source and install it with these commands:
 
 ```console
-make cog
+make
+sudo make install
 ```
 
 ## Upgrade
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/replicate/cog/pkg/config"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -132,8 +132,7 @@ sudo chmod +x /usr/local/bin/cog
 Alternatively, you can build Cog from source and install it with these commands:
 
 ```console
-make
-sudo make install
```
dkhokhlov, line 136: why this line removed?

dkhokhlov: make   -- is the same as `make cog`
sudo make install -- just install cog binary to /usr/local/bin., it is static single file, no go involved.

pkg/dockerfile/generator.go
```diff
@@ -66,8 +67,9 @@ func NewGenerator(config *config.Config, dir string) (*Generator, error) {
 	if err := os.MkdirAll(rootTmp, 0o755); err != nil {
 		return nil, err
 	}
-	// tmpDir ends up being something like dir/.cog/tmp/build123456789
-	tmpDir, err := os.MkdirTemp(rootTmp, "build")
+	// tmpDir ends up being something like dir/.cog/tmp/build20230731
+	now := time.Now().Format("20060102150405")
+	tmpDir, err := os.MkdirTemp(rootTmp, "build"+now)
```
mattt, line 72: The temporary directory should be different each time the command is run, so let's use a more granular timestamp.

```suggestion
	// tmpDir ends up being something like dir/.cog/tmp/build20240319021234.000000
	now := time.Now().Format("20060102150405.000000")
	tmpDir, err := os.MkdirTemp(rootTmp, "build"+now)
```

README.md
mattt: @dsingal0 Let's revert this part for now, since it's not related to the change you're proposing to use meaningful timestamps for temporary build directories. If you're getting unexpected errors when running the current commands, please open an issue and share the error you're seeing.

mattt:
Closing in favor of #1756

PR #1589, opened 2024-03-20
Add workflow to build cog-base images

tempusfrangit:
This PR adds a manual dispatch workflow that will build the cog base images. The dockerfiles and imagenames are defined in `cog-bases/matrix.json`.

The intent is to make it easy to build/add and support a variety of base images that cog can rely on for building more consistent and reusable images. This will improve cache hit when run in environments with a node cache of images.

### DIFF
--- a/.github/workflows/build-bases.yaml
+++ b/.github/workflows/build-bases.yaml
@@ -1,0 +1,64 @@
+name: Build Base Images
+
+on:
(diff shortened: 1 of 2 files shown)


tempusfrangit:
Note: the matrix.json is not currently usable. We need to populate this before merging/using this workflow.

PR #1592, opened 2024-03-24
optimize webhook serialization

technillogue:
this addresses some bottlenecks in models that output tokens fast

- encode webhooks only after knowing they will be sent
- bail out of upload type checks early for strs


technillogue:
this is missing a few things to get over the finish line, I'm not entirely sure how the mocking works, but should otherwise good to go

mattt (approved):
Took some doing to get the tests sorted, but this looks like a nice optimization. Thanks, @technillogue!

PR #1605, opened 2024-04-02
Cog base images

andreasjansson:


bfirsh:
Wahey. A couple of things:

- I presume that this will be the default at some point? Is there any reason why it shouldn't be the default now so other people building models for Replicate will get speedups automatically?
- The `baseimage` command feels like an internal thing. Could we make it hidden so it doesn't confuse users and so we don't have to worry about the design?

mattt (approved):

Issue #1586, opened 2024-03-19
cog issue when push model to replicate


aros0918:
I have an ai model. and I am trying to push the model by using cog.yaml and predict.py.
but at the end of the process, I got this error message.
They always ask me pydantic module error, but I have already installed that on the docker.
what is the problem?
plz help me

Traceback (most recent call last):
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/cog/command/openapi_schema.py", line 25, in <module>
    raise CogError(app.state.setup_result.logs)
cog.errors.CogError: Error while loading predictor:

Traceback (most recent call last):
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/cog/server/http.py", line 137, in create_app
    predictor = load_predictor_from_ref(predictor_ref)
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/cog/predictor.py", line 187, in load_predictor_from_ref
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/src/predict.py", line 1, in <module>
    import gradio as gr
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/gradio/__init__.py", line 3, in <module>
    import gradio._simple_templates
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/gradio/_simple_templates/__init__.py", line 1, in <module>
    from .simpledropdown import SimpleDropdown
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/gradio/_simple_templates/simpledropdown.py", line 6, in <module>
    from gradio.components.base import FormComponent
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/gradio/components/__init__.py", line 1, in <module>
    from gradio.components.annotated_image import AnnotatedImage
  File "/root/.pyenv/versions/3.10.13/lib/python3.10/site-packages/gradio/components/annotated_image.py", line 13, in <module>
    from gradio import processing_utils, utils
[...]

Issue #1562, opened 2024-03-06
Cog dependencies silently overrule cog.yaml dependencies


jd7h:
My cog.yaml specifies pydantic 2.5.3. My `cog build` command terminates succesfully. However, during `cog build`, cog's dependencies (pydantic < 2) have silently overruled cog.yaml. Because this is not shown clearly, this discrepancy between the cog.yaml and the actual environment is very hard to debug. 

Minimum example:

cog.yaml:
```
build:
  gpu: false
  python_version: "3.11"
  python_packages:
      - "pydantic==2.5.3"
```

Steps to reproduce: 
```
cog build
cog run bash

# in cog container bash: 
root@51b77719dae8:/src# pip freeze
annotated-types==0.6.0
anyio==4.3.0
attrs==23.2.0
certifi==2024.2.2
charset-normalizer==3.3.2
click==8.1.7
cog @ file:///tmp/cog-0.0.1.dev-py3-none-any.whl#sha256=41afe718f6fc8f969c35eb479a31948699b4c7f33c755410f617542291e8cd90
fastapi==0.98.0
h11==0.14.0
httptools==0.6.1
idna==3.6
pydantic==1.10.14
pydantic_core==2.14.6
python-dotenv==1.0.1
PyYAML==6.0.1
requests==2.31.0
sniffio==1.3.1
starlette==0.27.0
structlog==24.1.0
typing_extensions==4.10.0
urllib3==2.2.1
uvicorn==0.27.1
uvloop==0.19.0
watchfiles==0.21.0
websockets==12.0
```

Pydantic has version 1.10.14, and not the version we asked for in cog.yaml

This issue is slightly related to issue #1186.

woodybury:
@andreasjansson any update on the vendor python packages? I'm having same issue where my model requires pydantic > 2. Or any workarounds?

micos7:
Same issue

Issue #1384, opened 2023-11-19
Compatibility issues with Cog Pydantic < 2.*


sCarita:
Hello everyone,

I am building an image that needs `segments-ai` python package, which requires pydantic to be above version `2.*`. The problem is that if I install the pydantic python package above version 2, I get an error during the "Validating model schema..." phase.

```
Validating model schema...

Traceback (most recent call last):
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/command/openapi_schema.py", line 10, in <module>
    from ..server.http import create_app
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/server/http.py", line 16, in <module>
    from fastapi import Body, FastAPI, Header, HTTPException, Path, Response
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/fastapi/__init__.py", line 7, in <module>
    from .applications import FastAPI as FastAPI
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/fastapi/applications.py", line 16, in <module>
    from fastapi import routing
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/fastapi/routing.py", line 22, in <module>
    from fastapi import params
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/fastapi/params.py", line 4, in <module>
    from pydantic.fields import FieldInfo, Undefined
ImportError: cannot import name 'Undefined' from 'pydantic.fields' (/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/pydantic/fields.py)
```

It seems that the fastapi version used by cog is expecting pydantic to be below version 2, where "Undefined" object exists...

My current cog version is: `cog version 0.8.6 (built 2023-08-07T21:51:56Z)`

Any ideas on how to workaround this are gladly accepted.

Thank you for your time.

sCarita:
I understood we can control the fastapi version on the `cog.yaml` via `python_packages` key. So I did update, but now I get a new error which seems to be related with the internals of cog using pydantic related options.
```
Traceback (most recent call last):
[...]

PR #1615, opened 2024-04-15
Vendoring (take 2)

andreasjansson:
This PR vendors all Python packages using the [vendoring](https://pypi.org/project/vendoring/) library.

Vendoring allows users to install packages that rely on versions of libraries that are in conflict with Cog's dependencies (e.g. Pydantic, see https://github.com/replicate/cog/issues/1562, https://github.com/replicate/cog/issues/1384, https://github.com/replicate/cog/issues/1186, https://github.com/replicate/cog/issues/1586, https://github.com/replicate/cog/issues/1336, https://github.com/replicate/cog/issues/785).

Vendored packages are gitignored and are sync'd by the default `make` rule.

Closes https://github.com/replicate/cog/issues/409

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,8 +35,7 @@ jobs:
         with:
           python-version: 3.11
       - name: Install Python dependencies
-        run: |
-          python -m pip install '.[dev]'
+        run: make install-python
       - name: Build
         run: make cog
       - name: Lint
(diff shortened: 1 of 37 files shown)


nickstenning:
Had a quick look at this and I think it could work. A couple of thoughts.

1. I don't think we actually need to check in the files, do we? We have everything pinned in `vendor.txt`, so maybe let's have `vendoring sync` be part of the build process for the wheel. This will also help us ensure that we've nailed down any platform-specific stuff like cpython modules. As it is you've got a bunch of Darwin `.so`s checked in.
2. You probably want to exclude everything in `bin/` too. The shebang lines will be wrong and we don't ever need to use any of the deps as scripts.

andreasjansson:
Great points @nickstenning. I've implemented your suggestions, much better!

mattt:
Closing in favor of #1687. I think we're close, but if we can't manage to get that working, this is probably the next best alternative.

Issue #1620, opened 2024-04-15
Cog push not overriding latest


jlalmes:
I cannot push a new version to my model.

The logs show that the `cog push` command succeeded, but the model is never updated.

```
...
latest: digest: sha256:c9882636155e96ce5d87aed60b137485d472cf092a5fa60dff027720c3f21c53 size: 4954
Image 'r8.im/xxx/yyy' pushed

Run your model on Replicate:
    https://replicate.com/xxx/yyy
```

The same hash (`c9882636155e96ce5d87aed60b137485d472cf092a5fa60dff027720c3f21c53`) is shown each time I rerun the command & this matches the most recent version on the Replicate UI. 

<kbd>
<img width="790" alt="Screenshot 2024-04-15 at 19 53 18" src="https://github.com/replicate/cog/assets/69924001/8b23e45e-3186-49ac-a9a3-1761f118b960">
</kbd>

jlalmes:
Additional info:

Just tried creating a new model `xxx/yyy2` and pushing to that, and it failed with the following message:

```
Unable to find image 'r8.im/xxx/yyy2:latest' locally
docker: Error response from daemon: manifest for r8.im/xxx/yyy2:latest not found: manifest unknown: Failed to fetch "latest".
See 'docker run --help'.

ⅹ Failed to get type signature: exit status 125
Error: Process completed with exit code 1.
```


Yes, I am sure that I have created the model in the replicate UI under the correct account before pushing :)

App3x:
I have quite similar issue, but hash has changed on the platform but the running code from previous version.

hamzaraouzi:
I'm having  exactly the same issue, I tried the second approach and I'm  getting the same error.

App3x:
updated docker to lastest and `docker buildx prune -f` and solved

zeke:
Thanks for reporting, @jlalmes

https://github.com/replicate/cog/pull/1621 should address this. We'll release new versions of Cog and replicate/setup-cog with this fix this week.

anish-sahoo:
Closing because the push path now exports a fresh image before pushing, so latest is updated.

PR #1621, opened 2024-04-17
Always pass `--load` flag to `docker buildx build` command

mattt:
Related to https://github.com/replicate/cog/issues/1620

Some customers have observed the [`replicate/setup-cog` action](https://github.com/replicate/setup-cog) failing to push new versions of images to Replicate.

In logs for such failing GitHub Actions runs, the following is printed:

> No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load

We currently _do_ pass `--load`, but only when building on macOS:

https://github.com/replicate/cog/blob/03216b69e2f0aa2b93fe42e00f752fae5b576a78/pkg/docker/build.go#L25

I believe this code was leftover from when `cog build` didn't always imply `docker buildx build`, and BuildKit was used only when building `linux/amd64` on `linux/arm64/v8`.

From https://docs.docker.com/reference/cli/docker/buildx/create/#docker-driver:

> `docker driver` - Uses the builder that is built into the Docker daemon. With this driver, the --load flag is implied by default on buildx build. However, building multi-platform images or exporting cache is not currently supported.

> `docker-container driver` - Uses a BuildKit container that will be spawned via Docker. With this driver, both building multi-platform images and exporting cache are supported.

This PR adds the `--load` flag to the `docker buildx build` command invocation for all code paths. Since that's the implied default for the `docker` driver, this change shouldn't affect anyone getting the correct behavior.

### DIFF
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -17,12 +17,12 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 	var args []string
 
 	args = append(args,
-		"buildx", "build",
+		"buildx", "build", "--load",
 	)
 
 	if util.IsAppleSiliconMac(runtime.GOOS, runtime.GOARCH) {
 		// Fixes "WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested"
-		args = append(args, "--platform", "linux/amd64", "--load")
+		args = append(args, "--platform", "linux/amd64")
 	}
 
 	for _, secret := range secrets {


nickstenning (approved):

jlalmes:
RE. https://github.com/replicate/cog/issues/1620 - I am running `cog build` & `cog push` on a GitHub action (`ubuntu-latest-8-cores`), so don't believe this PR solves the issue 🤔

mattt:
@jlalmes Correct. These changes haven't been released yet, so this won't work until they are. Working to get those out tomorrow.

joe-aivatarz:
I updated cog to v0.9.6 and still getting this error on ubuntu-22.04 using replicate/setup-cog github action:
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Validating model schema...
Unable to find image 'r8.im/joe-aivatarz/comfyui-runner:latest' locally
docker: Error response from daemon: manifest for r8.im/joe-aivatarz/comfyui-runner:latest not found: manifest unknown: Failed to fetch "latest".

zeke:
@joe-aivatarz this PR has not been included in a Cog release yet, but we're working on it.

PR #1622, opened 2024-04-17
Fix Pydantic v1 incompatibility

mattt:
Related to https://github.com/replicate/cog/issues/1384

PR #1625, opened 2024-04-18
use pydantic v2

technillogue:


mattt:
My understanding is that this will cause incompatibilities with anyone using Pydantic 1, which I reckon is many if not most models.

Any objections to porting this to use the same "try import v2 shim for v1, fallback to actual v1" approach taken by #1622?

technillogue:
yeah that's sensible, this is just to unblock testing

pyproject.toml
```diff
@@ -14,10 +14,10 @@ requires-python = ">=3.8"
 dependencies = [
   # intentionally loose. perhaps these should be vendored to not collide with user code?
   "attrs>=20.1,<24",
-  "fastapi>=0.75.2,<0.99.0",
+  "fastapi>=0.75.2",
```
mattt, line 17: Bump minimum requirement to first version of FastAPI supporting Pydantic 2

```suggestion
  "fastapi>=0.100.0",
```

technillogue:
wait so is there some benefit to doing this over #1622 ?

Issue #1635, opened 2024-04-29
GitHub Actions Runner Error: Failed to Load Docker Image Metadata During cog Push Command


tsui66:
### Issue Description
When running `cog push r8.im/aicapcut/${{ inputs.model_name }}` from a GitHub Actions runner using Docker buildx, the following error occurs:

### Error Message
```plaintext
#2 [internal] load metadata for r8.im/aicapcut/anima-pencil-v310:latest
#2 ERROR: r8.im/aicapcut/anima-pencil-v310:latest: not found
------
 > [internal] load metadata for r8.im/aicapcut/anima-pencil-v310:latest:
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:1
--------------------
   1 | >>> FROM r8.im/aicapcut/anima-pencil-v310
--------------------
ERROR: failed to solve: r8.im/aicapcut/anima-pencil-v310: failed to resolve source metadata for r8.im/aicapcut/anima-pencil-v310

zeke:
Duplicate of https://github.com/replicate/cog/issues/1631

zeke:
This is fixed in https://github.com/replicate/cog/releases/tag/v0.9.8

Issue #1631, opened 2024-04-22
`No output specified with docker-container driver` error when using Docker buildx


zeke:
When running `cog build` or `cog push` from a GitHub Actions runner using Docker buildx, the following error occurs:

```
#3 [auth] library/cog-hello-world-cog:pull token for registry-1.docker.io
#3 DONE 0.0s

#2 [internal] load metadata for docker.io/library/cog-hello-world-cog:latest
#2 ERROR: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
------
 > [internal] load metadata for docker.io/library/cog-hello-world-cog:latest:
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:1
--------------------
   1 | >>> FROM cog-hello-world-cog
   2 |     COPY .cog/openapi_schema.json .cog
   3 |     
--------------------
ERROR: failed to solve: cog-hello-world-cog: failed to resolve source metadata for docker.io/library/cog-hello-world-cog:latest: pull access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed
```

Related issues: 

- https://github.com/replicate/cog/issues/300
- https://github.com/replicate/cog/issues/1026
- https://github.com/replicate/cog/issues/1105
- https://github.com/replicate/cog/issues/1620

PR https://github.com/replicate/cog/pull/1621 was intended to fix this issue, but I think I have an [Actions Run repro](https://github.com/zeke/hello-world-cog/actions/runs/8792119313/job/24127668946) here that shows the issue persists.

cc @mattt

jinchanz:
https://github.com/replicate/cog/blob/fae2ac3d4226a2706dbbff32938c257bc9024c1d/pkg/docker/build.go#L18

v0.9.6 release may not include the change in PR https://github.com/replicate/cog/pull/1621

zeke:
> https://github.com/replicate/cog/blob/fae2ac3d4226a2706dbbff32938c257bc9024c1d/pkg/docker/build.go#L18
> 
> v0.9.6 release may not include the change in PR #1621

That actions run is actually building Cog from source, rather than using a published release of Cog.

[...]

PR #1638, opened 2024-04-30
fix: 🐛 Failed to Load Docker Image Metadata in Github Action

EduarteXD:
Related to #1635 and #1631

PR #1621 attempts to address this issue, but passing the ```--load``` parameter in each build only partially solves the problem.

### DIFF
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -71,7 +71,7 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
 	return cmd.Run()
 }
 
-func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, bundledSchemaFile string, bundledSchemaPy string) error {
+func BuildAddLabelsAndSchemaToImage(dir, image string, labels map[string]string, bundledSchemaFile string, bundledSchemaPy string) error {
 	var args []string
 
 	args = append(args,
@@ -95,6 +95,7 @@ func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, bund
 	// We're not using context, but Docker requires we pass a context
 	args = append(args, ".")
 	cmd := exec.Command("docker", args...)
+	cmd.Dir = dir
 
 	dockerfile := "FROM " + image + "\n"
 	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
--- a/pkg/docker/build.go
+++ b/pkg/docker/build.go
@@ -75,12 +75,12 @@ func BuildAddLabelsAndSchemaToImage(dir, image string, labels map[string]string,
 	var args []string
 
 	args = append(args,
-		"buildx", "build",
+		"buildx", "build", "--load",
 	)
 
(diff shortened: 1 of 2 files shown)


pkg/docker/build.go
```diff
@@ -95,6 +95,7 @@ func BuildAddLabelsAndSchemaToImage(image string, labels map[string]string, bund
 	// We're not using context, but Docker requires we pass a context
 	args = append(args, ".")
 	cmd := exec.Command("docker", args...)
+	cmd.Dir = dir
```
zeke, line 98: Why is dir needed here?

pkg/docker/build.go
```diff
@@ -71,16 +71,16 @@ func Build(dir, dockerfile, imageName string, secrets []string, noCache bool, pr
...
 
 	args = append(args,
-		"buildx", "build",
+		"buildx", "build", "--load",
 	)
 
 	if util.IsAppleSiliconMac(runtime.GOOS, runtime.GOARCH) {
 		// Fixes "WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested"
-		args = append(args, "--platform", "linux/amd64", "--load")
+		args = append(args, "--platform", "linux/amd64")
```
zeke, line 83: Why is `--load` not needed here?

EduarteXD:
I have only tested it for [cog-sdxl](https://github.com/replicate/cog-sdxl/) and it resolved my issue with pushing the sdxl model while running the ```cog push``` command.

[0_build.txt](https://github.com/replicate/cog/files/15252997/0_build.txt)

And before this, we were encountering the error below:

```
#2 [internal] load metadata for r8.im/aicapcut/anima-pencil-v310:latest
#2 ERROR: r8.im/aicapcut/anima-pencil-v310:latest: not found
------
 > [internal] load metadata for r8.im/aicapcut/anima-pencil-v310:latest:
------
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
Dockerfile:1
--------------------
   1 | >>> FROM r8.im/aicapcut/anima-pencil-v310
--------------------
ERROR: failed to solve: r8.im/aicapcut/anima-pencil-v310: failed to resolve source metadata for r8.im/aicapcut/anima-pencil-v310
```

pkg/docker/build.go
EduarteXD: the parameter ``` --load ``` has already been appended at line 78

pkg/docker/build.go
EduarteXD: it just resolved my issue...

zeke: Just to be sure I understand you correctly: Which issue are you referring to? Was it related to the `--load` issue, or something else?

Theodotus1243:
@zeke, I have the same issue
Was it fixed somewhere?

zeke:
@Theodotus1243 this is fixed in https://github.com/replicate/cog/releases/tag/v0.9.8

PR #1643, opened 2024-05-04
time_share metric

technillogue:
it might be helpful to account for the cost of individual predictions in a batch, including the changes in batch size over the course of the prediction. this would add a metric representing that. it can also be used to determine the average batch size (`batch_size = predict_time / time_share`).

if there are two predictions that take 1 second, they would both cost 0.5s. if there's one prediction that takes 3 seconds and another one that takes 1 second (overlapping with the first prediction), the first will cost 2.5 (1 + 1/2 + 1) seconds and the second will cost 0.5 (1/2) seconds.

```
AAA***
*BBBB*
**CC**
***DDD
```
job A is running for the first 3 seconds, job B is running from second 2 to 5, job C runs seconds 3–4, job D runs seconds 4–6. suppose each second costs $1. A costs $1 + $.5 + $.3, B costs 0.5 + 0.3 + 0.3 + 0.5, C costs 0.3 + 0.3, etc.

### DIFF
--- a/python/cog/schema.py
+++ b/python/cog/schema.py
@@ -79,7 +79,7 @@ class PredictionResponse(PredictionBaseModel):
     error: t.Optional[str]
     status: t.Optional[Status]
 
-    metrics: t.Optional[t.Dict[str, t.Any]]
+    metrics: t.Dict[str, t.Any] = pydantic.Field(default_factory=dict)
 
     @classmethod
     def with_types(cls, input_type: t.Type[t.Any], output_type: t.Type[t.Any]) -> t.Any:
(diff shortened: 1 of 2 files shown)


zeke:
I asked Cursor to help me understand this.

> In the context of this code, "time share" refers to the amount of time a particular prediction task occupied the worker process.
> 
> So in summary, "time share" here represents the portion of the worker's processing time that was dedicated to each individual prediction task. It's a way to measure and report how the worker's time is being allocated among concurrent predictions.

zeke:
Wonder if there's a more descriptive name than `time_share` that encapsulates what it is. 🤔

PR #1649, opened 2024-05-06
function to emit metrics

technillogue:
we need the model code to emit accurate input token counts using the correct prompt formatting and tokenizer for the model in question. after some internal discussion `from cog import emit_metric` seems similar to the best approach for this. presently these metrics are ignored for non-official models and shouldn't really matter anywhere else, so it shouldn't be a concern with billing. this particular approach is janky but should unblock us for now. 

usage:
```python
     async def predict(self, ...) -> ...:
        ...
        formatted_prompt = prompt_template.format(prompt=prompt, system_prompt=system_prompt)
        input_token_count = len(self.tokenizer(formatted_prompt )["input_ids"])
        cog.emit_metric(name="input_token_count", value=input_token_count)
        ...
```

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,14 @@
+# Metrics
+
+Prediction objects have a `metrics` field. This normally includes `predict_time` and `total_time`. Official language models have metrics like `input_token_count`, `output_token_count`, `tokens_per_second`, and `time_to_first_token`. Currently, custom metrics from Cog are ignored when running on Replicate. Official Replicate-published models are the only exception to this. When running outside of Replicate, you can emit custom metrics like this:
+
+
+```python
+import cog
+from cog import BasePredictor, Path
+
+class Predictor(BasePredictor):
+    def predict(self, width: int, height: int) -> Path:
+        """Run a single prediction on the model"""
+        cog.emit_metric(name="pixel_count", value=width * height)
+```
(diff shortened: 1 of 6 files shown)


python/cog/server/worker.py
```diff
@@ -85,6 +86,17 @@ async def read(
             raise self.fatal
 
 
+# janky mutable container for a single eventual ChildWorker
+worker_reference: "dict[None, _ChildWorker]" = {}
```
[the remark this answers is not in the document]
technillogue: I don't see why, it doesn't care about threads or contexts or anything, this is pretty much just a hack to avoid `global`

zeke (changes requested):
I can't speak to the technical implementation of this, but I think it needs docs.

I would suggest creating a `docs/metrics.md` file to document why these exists, and the API for sending metrics. Maybe also some recommendations like "Use snake case for metric names" or "Put an x_ prefix on your custom metrics names" or similar.

docs/metrics.md
```diff
@@ -0,0 +1,14 @@
+# Metrics
+
+Prediction objects have a `metrics` field. This normally includes `predict_time` and `total_time`. Official language models have metrics like `input_token_count`, `output_token_count`, `tokens_per_second`, and `time_to_first_token`. Currently, custom metrics from Cog are ignored when running on Replicate. Official Replicate-published models are the only exception to this. When running outside of Cog, you can emit custom metrics like this:
```
zeke, line 3: What does "When running outside of Cog" mean? Should this be "outside of Replicate"?

technillogue: fixed; we'll also have another chance to review this when merging into main

PR #1662, opened 2024-05-13
truly support pydantic v2 without shims

technillogue:
to avoid vendoring or forking fastapi, we need to actually use the v2 pydantic models. this aims to detect which pydantic version is used and reimplement cog types for pydantic 2


python/cog/predictor.py
```diff
@@ -253,10 +249,14 @@ def load_predictor_from_ref(ref: str) -> BasePredictor:
 # Base class for inputs, constructed dynamically in get_input_type().
 # (This can't be a docstring or it gets passed through to the schema.)
 class BaseInput(BaseModel):
-    class Config:
-        # When using `choices`, the type is converted into an enum to validate
-        # But, after validation, we want to pass the actual value to predict(), not the enum object
-        use_enum_values = True
+    if PYDANTIC_V2:
+        model_config = ConfigDict(use_enum_values=True)
```
mattt, line 253: Move `ConfigDict` import statement here

python/cog/predictor.py
```diff
@@ -23,28 +23,24 @@
...
-
 try:
     from typing import get_args, get_origin
 except ImportError:  # Python < 3.8
     from typing_compat import get_args, get_origin  # type: ignore
 
+import pydantic
 import yaml
-from pydantic import BaseModel, Field, create_model
+from pydantic import BaseModel, ConfigDict, Field, create_model
```
mattt, line 33:
```suggestion
from pydantic import BaseModel, Field, create_model
```

Move `ConfigDict` import statement under this conditional

python/cog/predictor.py
```diff
@@ -306,21 +306,26 @@ def get_input_create_model_kwargs(signature: inspect.Signature) -> Dict[str, Any
...
+            if not isinstance(parameter.default, FieldInfo):
+                default = Input(default=parameter.default)
+            else:
+                default = parameter.default
 
         # Fields aren't ordered, so use this pattern to ensure defined order
         # https://github.com/go-openapi/spec/pull/116
-        default.extra["x-order"] = order
+        if PYDANTIC_V2:
+            extra = default.json_schema_extra = {}
```
mattt, line 318: 🤨 

```suggestion
            extra = {}
```

python/cog/schema.py
```diff
@@ -38,18 +38,18 @@ def default_events(cls) -> t.List["WebhookEvent"]:
         return [cls.START, cls.OUTPUT, cls.LOGS, cls.COMPLETED]
 
 
-class PredictionBaseModel(pydantic.BaseModel, extra=pydantic.Extra.allow):
+class PredictionBaseModel(pydantic.BaseModel, extra="allow"):
```
mattt, line 41: Does Pydantic 1 support string literal? Or should we conditionalize?

```suggestion
class PredictionBaseModel(pydantic.BaseModel, extra="allow" if PYDANTIC_V2 else pydantic.Extra.allow):
```

python/cog/predictor.py
technillogue: no, then it wouldn't be set on the default field. the problem is that default.json_schema_extra is None by default, not {} like default.extra in v1

python/cog/schema.py
technillogue: yup: https://docs.pydantic.dev/1.10/usage/model_config/#:~:text=Accepts%20the%20string%20values%20of%20%27ignore%27%2C%20%27allow%27%2C%20or%20%27forbid%27

yorickvP:
The annotation is hopeless, `typing.Annotated` was added in python 3.9 and also doesn't seem to like being `isinstance`'d.

Tests in the commit before that are mainly failing because of openapi schema changes in 3.1.0: `{'openapi': '3.1.0'} != {'openapi': '3.0.2'}`, and also `ge` vs `limit_value`

mattt:
Closing in favor of #1687

It's not all the way there yet, but it's the farthest along we've got so far. So let's focus on that for now.

PR #1664, opened 2024-05-14
document pinning python package installations to commit SHAs

zeke:
This PR adds some docs about how to pin to a specific git commit when installing Python packages directly from a GitHub repo.

cc @fofr @lucataco

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -65,8 +65,20 @@ build:
   system_packages:
     - "git"
   python_packages:
-    - "git+https://github.com/m-bain/whisperX.git"
+    - "git+https://github.com/huggingface/transformers"
 ```
+
+You can also pin Python package installations to a specific git commit:
+
+```yaml
+build:
+  system_packages:
+    - "git"
+  python_packages:
+    - "git+https://github.com/huggingface/transformers@2d1602a"
+```
+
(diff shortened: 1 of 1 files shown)


docs/yaml.md
```diff
@@ -68,6 +68,19 @@ build:
...
 ```
 
+You can also pin Python package installations to a specific git commit:
+
+```yaml
+build:
+  system_packages:
+    - "git"
+  python_packages:
+    - "git+https://github.com/m-bain/whisperX.git@f2da2f858e99e4211fe4f64b5f2938b007827e17"
```
fofr, line 78: This example format didn't work for me, I think it needs to be in the format:

`- cog-model-helpers @ git+https://github.com/fofr/cog-model-helpers.git@19b0a2a2ad9f4e464c3573717259915f8250e594`

zeke: Oh i see. I didn't test it. Is that for any git-based installation? Or just when you're specifying the SHA? (In other words, is the existing doc here also maybe wrong?)

fofr: The existing pattern still works, but you cannot append a SHA to it

zeke: Updated!

bfirsh:
I'd love to switch entirely to `python_requirements` then it's effectively self-documenting.

PR #1667, opened 2024-05-16
Include prediction id upload request

aron:
This PR introduces two small changes to the file upload interface.

1. We now allow downstream services to include the destination of the asset in a `Location` header, rather than assuming that it's the same as the final upload url (either the one passed via `--upload-url` or the result of a 307 redirect response.
2. We now include the `X-Prediction-Id` header in upload request, this allows the downstream client to potentially do configuration/routing based on the prediction ID. This ID should be considered unsafe and needs to be validated by the downstream service.

### DIFF
--- a/python/cog/files.py
+++ b/python/cog/files.py
@@ -39,7 +39,7 @@ def guess_filename(obj: io.IOBase) -> str:
 
 
 def put_file_to_signed_endpoint(
-    fh: io.IOBase, endpoint: str, client: requests.Session
+    fh: io.IOBase, endpoint: str, client: requests.Session, prediction_id: str | None
 ) -> str:
     fh.seek(0)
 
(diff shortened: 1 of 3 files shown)


python/cog/files.py
```diff
@@ -54,15 +54,22 @@ def put_file_to_signed_endpoint(
     resp = client.put(
         ensure_trailing_slash(endpoint) + filename,
         fh,  # type: ignore
-        headers={"Content-type": content_type},
+        headers={
+            "Content-Type": content_type,
+            "X-Prediction-ID": prediction_id,
+        },
```
evilstreak, line 60: We probably don't want to set `X-Prediction-ID` if `prediction_id` is `None`.

```suggestion
        headers={"Content-Type": content_type}

        if prediction_id is not None:
            headers["X-Prediction-ID"] = prediction_id
```

evilstreak (approved):
One nit, otherwise it looks good to me

mattt (approved):

PR #1671, opened 2024-05-17
update `cog init` to include a GitHub Actions workflow file

zeke:
This PR adds a "Push to Replicate" GitHub Actions workflow file to the `cog init` template, so users will have a smoother path to continuous delivery.

The workflow is configured to start users off with a manual "repository dispatch" so pushes are explicit and intentional. It also has some commented-out settings to support a more continuous workflow once they've kicked the tires, plus some pointers for common issues like running out of disk space.

```
$ ~/go/bin/cog init     

✅ Created /Users/z/Desktop/cog-init-test/cog.yaml
✅ Created /Users/z/Desktop/cog-init-test/predict.py
✅ Created /Users/z/Desktop/cog-init-test/.dockerignore
✅ Created /Users/z/Desktop/cog-init-test/.github/workflows/push.yaml

```

### DIFF
--- a/pkg/cli/init-templates/.github/workflows/push.yaml
+++ b/pkg/cli/init-templates/.github/workflows/push.yaml
@@ -6,8 +6,7 @@ on:
   workflow_dispatch:
     inputs:
       model_name:
-        description: 'Enter the model name, like "alice/bunny-detector"'
-        required: true
+        description: 'Enter the model name, like "alice/bunny-detector". If unset, this will default to the value of `image` in cog.yaml.'
   # # Uncomment these lines to trigger the workflow on every push to the main branch
   # push:
   #   branches:
@@ -42,9 +41,14 @@ jobs:
           # the action will authenticate with Replicate automatically so you
           # can push your model
           token: ${{ secrets.REPLICATE_API_TOKEN }}
-        
-      # If you run the action manually using workflow dispatch, you'll enter
-      # the model name manually. If the workflow is run automatically (say on
-      # a push to main), the model name will be derived from Repsoitory Variables.
+
+      # If you trigger the workflow manually, you can specify the model name.
+      # If you leave it blank (or if the workflow is triggered by a push), the 
+      # model name will be derived from the `image` value in cog.yaml.
       - name: Push to Replicate
-        run: cog push r8.im/${{ inputs.model_name || vars.DEFAULT_MODEL_NAME }}
+        run: |
+          if [ -n "${{ inputs.model_name }}" ]; then
+            cog push r8.im/${{ inputs.model_name }}
+          else
+            cog push
+          fi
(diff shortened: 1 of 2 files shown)


yorickvP:
The test is failing because of formatting.

pkg/cli/init-templates/.github/workflows/push.yaml
```diff
@@ -0,0 +1,50 @@
...
+          # If you set REPLICATE_API_TOKEN in your GitHub repository secrets,
+          # the action will authenticate with Replicate automatically so you
+          # can push your model
+          token: ${{ secrets.REPLICATE_API_TOKEN }}
+        
+      # If you run the action manually using workflow dispatch, you'll enter
+      # the model name manually. If the workflow is run automatically (say on
+      # a push to main), the model name will be derived from Repsoitory Variables.
+      - name: Push to Replicate
+        run: cog push r8.im/${{ inputs.model_name || vars.DEFAULT_MODEL_NAME }}
```
yorickvP, line 50: if inputs.model_name is required, would this ever hit the `vars.DEFAULT_MODEL_NAME` branch?

pkg/cli/init-templates/.github/workflows/push.yaml
```diff
@@ -0,0 +1,50 @@
+name: Push to Replicate
+
+on:
+  # Workflow dispatch allows you to manually trigger the workflow from GitHub.com
+  # Go to your repo, click "Actions", click "Push to Replicate", click "Run workflow"
+  workflow_dispatch:
+    inputs:
+      model_name:
```
yorickvP, line 8: Some models put this in `cog.yaml`. I'm not sure what kind of error `cog push` would generate if no model name is specified anywhere, though.

pkg/cli/init-templates/.github/workflows/push.yaml
zeke: Good point. I'm simplifying this by making the input optional, getting rid of DEFAULT_MODEL_NAME, and instead falling back to `cog push` if the input is not set.

pkg/cli/init-templates/.github/workflows/push.yaml
zeke: Luckily Cog catches this case right away before proceeding:

```
$ cog push
ⅹ To push images, you must either set the 'image' option in cog.yaml or pass an image name as an argument. For example, 'cog push r8.im/your-username/hotdog-detector'
```

pkg/cli/init-templates/.github/workflows/push.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+    # If your model is large, the default GitHub Actions runner may not 
+    # have enough disk space. If you need more space you can set up a 
+    # bigger runner on GitHub.
+    runs-on: ubuntu-latest
+
+    steps:
+      # This action cleans up disk space to make more room for your
+      # model code, weights, etc.
+      - name: Free disk pace
```
technillogue, line 27:
```suggestion
      - name: Free disk space
```

technillogue (approved):

Issue #1668, opened 2024-05-16
Concurrency configuration


technillogue:
currently in the async branch, cog.yaml is changed to have a top-level concurrency

```yaml
predict: "predict.py:Predictor"
concurrency: 64
```

however, only admins can change the target_metric. we eventually want users to be able to deploy batched models with the correct scaling profile. the new proposed design would be like this:

```yaml
predict: "predict.py:Predictor"
concurrency:
  max: 64
  default_target: 16
```

the schema would include restrictions that `1 < default_target < max`. this approach has a few upsides:

* it's really easy to parse from go and python alike
* it can be modified in e.g. github actions without installing anything
  * you can change both director and cog at the same time with a tool like yolo
  * otherwise a static tool would need to rewrite both python and the schema 
* it doesn't require creative reinterpretations of openapi schema semantics

it also has a downside in that it "feels" like a property of the prediction code.

when discussing this, it's worth keeping in mind a few use cases:
* triton is configured with a specific batch size when the engine is built, configured from checked-in yaml files. exceeding the batch size causes errors; being under the batch size is suboptimal because memory is allocated based on the same configuration
* vllm has no such configuration, can run on different hardware, and can support different batch sizes depending on that hardware
* whisper and other non-language models can be batched with their own concerns about the relationship between model code, build options, and runtime options like hardware

it's also possible we might want something like "max number of loaded finetunes" to be configurable as well

PR #1672, opened 2024-05-17
allow setting both max and target concurrency in cog.yaml

technillogue:
we need users to be able to set target metric. after discussion in #1668, this seems like the simplest approach for now and could be reconsidered later.


technillogue:
I'm going to just merge this for now and we can revisit it later

PR #1674, opened 2024-05-20
Rename `DefaultTarget` to `Target`

mattt:
Follow-up to #1672

Both max and target concurrency levels can be overridden by web, so in effect both are default values.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -45,8 +45,8 @@ type Build struct {
 }
 
 type Concurrency struct {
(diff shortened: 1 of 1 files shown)


technillogue:
max concurrency can't overriden in web though

mattt:
How about the value sent to director from cluster? I can imagine situations where it'd be nice to lower an ambitious max value for a model. For example, when running on a different hardware.

technillogue:
there are some occasional models that could potentially run on different hardware (not triton), but what value does director send to cluster...?

PR #1680, opened 2024-05-21
[async] Include prediction id upload request

technillogue:
httpx version of #1667 

> We now allow downstream services to include the destination of the asset in a Location header, rather than assuming that it's the same as the final upload url (either the one passed via --upload-url or the result of a 307 redirect response.
> We now include the X-Prediction-Id header in upload request, this allows the downstream client to potentially do configuration/routing based on the prediction ID. This ID should be considered unsafe and needs to be validated by the downstream service.

### DIFF
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -150,7 +150,9 @@ class ClientManager:
 
     # files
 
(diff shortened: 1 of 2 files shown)


aron:
I opened #1788 in place of this one. Not because there was anything wrong with this code, I just found it useful to re-implement the PR to get my head around the async branch.

PR #1683, opened 2024-05-21
predict_time_share needs to be set before sending the completed webhook

technillogue:
#1643 adding a predict_time_share metric but it was only set on the prediction after the completed webhook was sent


technillogue:
on further reflection, this should be moved to director, but that doesn't have to happen right this second necessarily

PR #1684, opened 2024-05-22
add a Pull Request template

zeke:
This PR adds a [pull request template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) to help remind contributors to describe the changes they're making, and include tests and documentation as appropriate.

This will show up as the default text in the textarea when opening a PR. Authors can keep it and fill it out, or delete it and start fresh.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,4 @@
+<!-- Briefly describe this pull request. What's good about it? -->
+
+- [ ] Tests
+- [ ] Docs


mattt:
@zeke I haven't found our volume or quality of PRs to cause any problems so far. PR templates can be nice for enforcing rules and process, but I'm not sure we need something like that right now.

zeke:
I disagree. Every PR we land that adds a feature without accompanying docs becomes a liability.

I think we'll get better and more accurate docs if they are contributed by the authors of the features, at the time of implementation. Furthermore, writing docs gives authors an extra opportunity to consider the shape of the feature they're proposing, test it, and evaluate whether it actually works and makes sense.

erbridge:
I agree with the sentiment of a pull request template, but I don't think this one has enough guidance or structure as to what to do with it. By my naive reading, the comment seems to apply to the check list, but in a way that doesn't make sense (am I supposed to write tests and docs that describe the pull request?).

mattt:
I agree with @erbridge. I think a pull request template can be helpful, but I don't think this particular implementation would add much.

Happy to revisit this once we've identified a specific need for PR template, or have more consensus around what kind of process we want to have for this project.

PR #1685, opened 2024-05-22
drop default_target

technillogue:


mattt:
@technillogue Interesting! Can you share some context about this? Is it to support fractional values?

technillogue:
yeah just because the default is supposed to be 1.05

PR #1688, opened 2024-05-24
fixing install logic to reenable no cuda base image

daanelson:
As of cog v0.9.7, setting the `--use-cuda-base-image=false` flag fails b/c we changed the logic below around copying python packages to the staging images. This fixes that logic for the case when we have that flag set and aren't using cog base images.


technillogue:
just so I understand, it was trying to copy to pyenv when doing --use-cuda-base-image=false?

technillogue (approved):

PR #1689, opened 2024-05-24
Parse []BaseImageConfiguration from a JSON

nevillelyh:


mattt:
https://github.com/replicate/cog/pull/1691 should fix the failing lint job in CI. Re-running after rebasing should resolve that flakey test timeout.

mattt:
I believe this is superseded by https://github.com/replicate/cog/pull/1712. 

@nevillelyh If that's not the case, could you please reopen and handle rebasing?

PR #1690, opened 2024-05-24
Improve CUDA compatibility checks

nevillelyh:
- Check version major/minor instead of exact match
- Suggest compatible version if available


mattt:
@nevillelyh Rebased on top of #1691 to fix CI failure

pkg/config/config.go
```diff
@@ -461,9 +464,12 @@ Compatible cuDNN version is: %s`,
 				return err
 			}
 			console.Debugf("Setting CUDA to version %s from Torch version", c.Build.CUDA)
-		} else if !slices.ContainsString(torchCUDAs, c.Build.CUDA) {
+		} else if len(slices.FilterString(torchCUDAs, func(torchCUDA string) bool { return version.EqualMinor(torchCUDA, c.Build.CUDA) })) == 0 {
```
mattt, line 467: I had trouble understanding this conditional. Could we make this an `else` and break this up over a couple lines?

nevillelyh: Fixed!

zeke:
Heads up @marcusmartins in case you're working on something related to this in https://github.com/replicate/cog/pull/1699

> Check that the combination of cuda, python and pytorch is valid before attempting to run/build the image

PR #1695, opened 2024-05-28
Forward on trace context to upstream services

aron:
When the `OTEL_SERVICE_NAME` environment variable is provided, we use the Open Telemetry `requests` library to instrument the `requests.Session` to ensure trace context is forwarded on to upstream services for functionality like webhooks and uploads.

As we only want the internal sessions used by cog to be instrumented rather than the global session available via `request.get` etc, we do some awkward monkey patching in the new `telemetry.requests_session()` method.

Tests have been added to verify the telemetry is added to the correct clients and not the globals.

> [!NOTE]
> Currently there is no support for configuring any exporters, and no telemetry will be collected for cog. This PR is primarily about ensuring that the trace context is properly propagated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,9 @@ dependencies = [
   # intentionally loose. perhaps these should be vendored to not collide with user code?
   "attrs>=20.1,<24",
   "fastapi>=0.75.2,<0.99.0",
+  "opentelemetry-api",
+  "opentelemetry-sdk",
+  "opentelemetry-instrumentation-requests",
   "pydantic>=1.9,<2",
   "PyYAML",
   "requests>=2,<3",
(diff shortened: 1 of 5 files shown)


aron:
@nickstenning I'm looking at this and wondering for our needs whether it would be simpler to capture the value of the `traceparent` header in the prediction & training HTTP handlers and just manually forward that on in the upload and webhook handlers, it would require a little more code but the intent would be clearer and more straightforward than using the otel library.

nickstenning:
> whether it would be simpler to capture the value of the `traceparent` header in the prediction

Yeah, looking at this I think you're probably right. It would be `traceparent` and `tracestate` headers: see https://www.w3.org/TR/trace-context/.

aron:
Closing in favor of https://github.com/replicate/cog/pull/1698

PR #1698, opened 2024-05-29
Propagate trace context to webhook and upload requests

aron:
If the request to /predict contains headers `traceparent` and `tracestate` defined by w3c Trace Context[^1] then these headers are forwarded on to the webhook and upload calls.

This allows observability systems to link requests passing through cog.

[^1]: https://www.w3.org/TR/trace-context/

### DIFF
--- a/python/cog/server/telemetry.py
+++ b/python/cog/server/telemetry.py
@@ -1,11 +1,16 @@
-from typing import TypedDict, Union, Optional
 from contextlib import contextmanager
 from contextvars import ContextVar
+from typing import Generator, Optional
+
+# TypedDict was added in 3.8
+from typing_extensions import TypedDict
+
 
 # See: https://www.w3.org/TR/trace-context/
-TraceContext = TypedDict(
-    "TraceContext", {"traceparent": Optional[str], "tracestate": Optional[str]}
-)
+class TraceContext(TypedDict, total=False):
+    traceparent: str
+    tracestate: str
+
 
 TRACE_CONTEXT: ContextVar[Optional[TraceContext]] = ContextVar(
     "trace_context", default=None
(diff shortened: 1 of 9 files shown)


nickstenning (commented):
I think this is ok, but this seems like a clear-cut case where we could use a [`ContextVar`](https://docs.python.org/3/library/contextvars.html) and a couple of helper functions (in, say, a `telemetry` module) to make this a lot less messy.

The idea would be something like:

```python
# telemetry.py
from contextlib import contextmanager
from contextvars import ContextVar

TRACE_CONTEXT = ContextVar("trace_context", default=None)

def current_trace_context():
    return TRACE_CONTEXT.get()

@contextmanager
def trace_context(ctx):
    t = TRACE_CONTEXT.set(ctx)
    try:
        yield
    finally:
        TRACE_CONTEXT.reset(t)

# http.py
async def predict(
    request: PredictionRequest = Body(default=None),
    prefer: Union[str, None] = Header(default=None),
    traceparent: Union[str, None] = Header(default=None),
    tracestate: Union[str, None] = Header(default=None),
) -> Any:  # type: ignore
    ...
    ctx = {}
    if traceparent:
        ctx['traceparent'] = traceparent
    if tracestate:
        ctx['tracestate'] = tracestate

    if ctx:
        with telemetry.trace_context(ctx):
            return predict(...)
    
    return predict(...)
```

aron:
Okay, nice, I'll refactor the code.

python/cog/server/telemetry.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def make_trace_context(
+    traceparent: Union[str, None] = None, tracestate: Union[str, None] = None
+) -> TraceContext:
+    """
+    Creates a trace context dictionary from the given traceparent and tracestate
+    headers. This is used to pass the trace context between services.
+    """
+    return {"traceparent": traceparent, "tracestate": tracestate}
```
nickstenning, line 22: I'm not sure you want to set these in the object if they're none. As written you'll be setting headers later to the literal string `"None"`.

nickstenning (approved):
Generally LGTM other than the handling of absent/`None` values. The branch diff here also introduces a bunch of spurious whitespace and formatting changes which it would maybe be nice to rebase out, but that's not a huge deal.

python/cog/server/telemetry.py
aron: Hmm, good catch, I was getting frustrated with the typechecker. But I see `TypedDict` takes a `total` keyword that does what I need.

aron:
Hmm, @nickstenning the ContextVar doesn't seem to propagate across the threads, does this need to be handled manually?

aron:
I've now rebased on main and run the `python` directory through ruff:

     python -m ruff check python --fix

I'm leaving that as the canonical whitespace.

PR #1699, opened 2024-05-29
Improve cog base image handling

marcusmartins:
In preparation for making `--use-cog-base-image` flag the default behavior, we are adding better
handling for invalid configuration.

* Check that the combination of cuda, python and pytorch is valid before attempting to run/build the image
* Check that we are not missing flags that don't work together

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -25,10 +25,11 @@ var buildUseCogBaseImage bool
 
 func newBuildCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "build",
-		Short: "Build an image from cog.yaml",
-		Args:  cobra.NoArgs,
-		RunE:  buildCommand,
+		Use:     "build",
+		Short:   "Build an image from cog.yaml",
+		Args:    cobra.NoArgs,
+		RunE:    buildCommand,
+		PreRunE: checkMutuallyExclusiveFlags,
 	}
 	addBuildProgressOutputFlag(cmd)
 	addSecretsFlag(cmd)
@@ -55,11 +56,6 @@ func buildCommand(cmd *cobra.Command, args []string) error {
 	}
 	if imageName == "" {
 		imageName = config.DockerImageName(projectDir)
-	}
-
-	err = checkMutuallyExclusiveFlags(cmd)
-	if err != nil {
-		return err
 	}
 
 	err = config.ValidateModelPythonVersion(cfg.Build.PythonVersion)
@@ -122,7 +118,7 @@ func addBuildTimestampFlag(cmd *cobra.Command) {
 	_ = cmd.Flags().MarkHidden("timestamp")
 }
 
-func checkMutuallyExclusiveFlags(cmd *cobra.Command) error {
+func checkMutuallyExclusiveFlags(cmd *cobra.Command, args []string) error {
 	flags := []string{"use-cog-base-image", "use-cuda-base-image", "dockerfile"}
 	var flagsSet []string
 	for _, flag := range flags {
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -1,7 +1,9 @@
 package cli
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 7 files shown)


pkg/dockerfile/generator.go
```diff
@@ -256,7 +256,12 @@ func (g *Generator) baseImage() (string, error) {
 			return "", err
 		}
 		torchVersion, _ := g.Config.TorchVersion()
-		baseImage := BaseImageName(cudaVersion, pythonVersion, torchVersion)
+		// validate that we base image configuration exists
```
zeke, line 259:
```suggestion
		// validate that the base image configuration exists
```

pkg/cli/build.go
```diff
@@ -108,14 +115,23 @@ func addDockerfileFlag(cmd *cobra.Command) {
...
+func checkMutuallyExclusiveFlags(cmd *cobra.Command) error {
+	flags := []string{"use-cog-base-image", "use-cuda-base-image", "dockerfile"}
+	var flagsSet []string
+	for _, flag := range flags {
+		if cmd.Flag(flag).Changed {
+			flagsSet = append(flagsSet, "--"+flag)
+		}
+	}
+	if len(flagsSet) > 1 {
+		return fmt.Errorf("the flags %s are mutually exclusive, you can only set one of them", strings.Join(flagsSet, " and "))
```
zeke, line 134:
```suggestion
		return fmt.Errorf("The flags %s are mutually exclusive: you can only set one of them.", strings.Join(flagsSet, " and "))
```

Little bit of formatting to make it match other `fmt.Errorf` invocations.

zeke (commented):
I'm not the right person to review the code itself, but it makes sense to me.

Does this PR need any user-facing docs?

zeke:
> Not sure why the tests are failing with a timeout

I thought it might be a flake but I reran it and it still failed

marcusmartins:
Yes, the tests are flaky. We have merged a deadline change that I hope will fix the test.

I ran into usability issues with base images that I want to fix before merging this PR.

marcusmartins:
@zeke Yes, we will need the documentation, but I was thinking waiting until we're ready to make the behavior default. What do you think?

pkg/cli/build.go
```diff
@@ -108,14 +115,23 @@ func addDockerfileFlag(cmd *cobra.Command) {
...
-	})
 }
 
 func addBuildTimestampFlag(cmd *cobra.Command) {
 	cmd.Flags().Int64Var(&config.BuildSourceEpochTimestamp, "timestamp", -1, "Number of seconds sing Epoch to use for the build timestamp; this rewrites the timestamp of each layer. Useful for reproducibility. (`-1` to disable timestamp rewrites)")
 	_ = cmd.Flags().MarkHidden("timestamp")
 }
+
+func checkMutuallyExclusiveFlags(cmd *cobra.Command) error {
+	flags := []string{"use-cog-base-image", "use-cuda-base-image", "dockerfile"}
```
timfallmk, line 126: Can these be programmatically determined? I'm ok with them being hardcoded for a first pass as we go, but it might be helpful to break this out into separate logic

marcusmartins: Can you give an example of what you are thinking in terms of improvement?

zeke:
> @zeke Yes, we will need the documentation, but I was thinking of waiting until we're ready to make the behavior default. What do you think?

Yeah that's fine. Just wanna make sure we don't forget about docs down the line.

jessie-young (approved):

marcusmartins:
Depends on https://github.com/replicate/cog/pull/1715

pkg/cli/build.go
marcusmartins: Not sure if that would add much here as those commands are only part of this package and relationship between the flags (what makes them mutually exclusive) is very specific to the code here.

tempusfrangit (approved):
LGTM, this needs an additional commit in the builder repo so we can build the new base images.

pkg/cli/run.go
```diff
@@ -54,6 +54,11 @@ func run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	err = checkMutuallyExclusiveFlags(cmd)
```
tempusfrangit, line 57: this could also be done in PreRunE but it's fine here.

marcusmartins: Done! Thank you!

PR #1700, opened 2024-05-30
Increase Hypothesis test deadlines

marcusmartins:
Recently we have been hitting the timings deadline in the python integration tests.

I didn't investigate if the regression is expected or not.


mattt (approved):

yorickvP:
These were recently set to 1500 after they failed at 1000. Maybe something's up.

marcusmartins:
thanks @yorickvP I will investigate

PR #1704, opened 2024-05-31
Quick fix for Apple Silicon M1 support in `run`

timfallmk:
This commit fixes the `run` command for Apple Silicon M1 support. The `run` command now uses the `docker run` command with the `--platform` flag set to `linux/amd64` to ensure that the container runs on the correct architecture. This behavior matches the `build` command, which also uses the `--platform` flag to ensure that the image is built for the correct architecture.

### DIFF
--- a/pkg/cli/run.go
+++ b/pkg/cli/run.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"runtime"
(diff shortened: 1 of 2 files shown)


mattt (approved):
Thanks, @timfallmk!

PR #1715, opened 2024-06-04
Set a minimum CUDA version for Tensorflow compatibility

marcusmartins:
Previously, we fetched the TensorFlow compatibility matrix directly from the TensorFlow release page, including all versions without checking for a minimum CUDA version. This led to conflicts with other code paths that expect a smaller set of CUDA versions (e.g., `cuda_base_images.json`).

This change introduces a minimum CUDA version requirement when generating the TensorFlow compatibility matrix. By doing so, we ensure compatibility and consistency across the codebase, preventing conflicts with other components that rely on specific CUDA versions.


tools/compatgen/internal/tensorflow.go
```diff
@@ -12,6 +13,7 @@ import (
 
 func FetchTensorFlowCompatibilityMatrix() ([]config.TFCompatibility, error) {
 	url := "https://www.tensorflow.org/install/source"
+	min_cuda_version := "10.1"
```
jessie-young, line 16: Should this be defined as a const at the package level outside of the functions? 

How do you determine the min CUDA version and is that something that will need to be updated on a semi-frequent basis?

mattt: How about making minimum CUDA version a parameter of this method, and then defining a constant at the call site?

marcusmartins: yeah, those are good points. I will add the min CUDA version to the compatibility package.

andreasjansson: Shall we say 11.0 is the minimum? It was released in 2020, and I doubt people will build new images with CUDA<11

andreasjansson (approved):
This is great, just had a comment on which minimum CUDA version we're pinning to.

marcusmartins:
@mattt @jessie-young can you take another look? 

See https://github.com/replicate/cog/pull/1715/commits/ab290506998a3f013a664d0f9de2132c2ecfb22c for more details about my approach

tools/compatgen/internal/tensorflow.go
```diff
@@ -45,7 +51,7 @@ func FetchTensorFlowCompatibilityMatrix() ([]config.TFCompatibility, error) {
 	}
 
 	// sanity check
-	if len(compats) < 21 {
+	if len(compats) < 12 {
```
marcusmartins, line 54: With min_cuda_version, I am not sure what is best way to preserve the intention of this check so I am leaving as the floor value.

tools/compatgen/internal/tensorflow.go
```diff
@@ -33,6 +35,10 @@ func FetchTensorFlowCompatibilityMatrix() ([]config.TFCompatibility, error) {
 		cuDNN := cells[4].Text()
 		cuda := cells[5].Text()
 
+		if !version.Greater(cuda, min_cuda_version) {
```
marcusmartins, line 38: Oops, this is wrong.

pkg/config/compatibility.go
```diff
@@ -85,7 +85,7 @@ func (i *CUDABaseImage) ImageTag() string {
 var cudaBaseImagesData []byte
 var CUDABaseImages []CUDABaseImage
 
-//go:generate go run ../../tools/compatgen/main.go tensorflow -o tf_compatibility_matrix.json
+//go:generate go run ../../tools/compatgen/main.go tensorflow --min-cuda-version 11 -o tf_compatibility_matrix.json
```
marcusmartins, line 88: This is a bit magical but AFAIK there's no good way to pass a variable from the package to `go generate`

tools/compatgen/main.go
```diff
@@ -12,11 +12,12 @@ import (
 
 func main() {
 	var output string
+	var minCudaVersion string
 
 	var rootCmd = &cobra.Command{
 		Use:   "compatgen {cuda|torch|tensorflow}",
 		Short: "Generate compatibility matrix for Cog base images",
-		Args:  cobra.ExactArgs(1),
+		Args:  cobra.MinimumNArgs(1),
```
mattt, line 20: The way Cobra handles CLIs, flags aren't considered arguments. They're parsed out and captured before the remaining args are passed. 


```suggestion
		Args:  cobra.ExactArgs(1),
```

tools/compatgen/main.go
```diff
@@ -30,7 +31,10 @@ func main() {
 					console.Fatalf("Failed to fetch CUDA base image tags: %s", err)
 				}
 			case "tensorflow":
-				v, err = internal.FetchTensorFlowCompatibilityMatrix()
+				if minCudaVersion == "" {
```
mattt, line 34: This effectively makes a `--min-cuda-version` flag required. Instead, why don't we:

- Define a package-level `const defaultMinCudaVersion` set to 11.0 (or whatever). Add a comment with relevant date / link to provide context when that code is updated.
- Set `defaultMinCudaVersion` as the third argument to `StringVar(&minCudaVersion, ...` to set it as the default.
- Closer to the top of `RunE`, parse/validate `minCudaVersion` and return an error if it's an invalid version string.
- Later / Bonus: Update `FetchTorchCompatibilityMatrix` to take `minCudaVersion` argument.

tools/compatgen/internal/tensorflow.go
```diff
@@ -45,7 +51,7 @@ func FetchTensorFlowCompatibilityMatrix() ([]config.TFCompatibility, error) {
 	}
 
 	// sanity check
-	if len(compats) < 21 {
+	if len(compats) < 12 {
 		return nil, fmt.Errorf("Tensorflow compatibility matrix only had %d rows, has the html changed?", len(compats))
 	}
```
mattt, line 56: Since this is now parameterized, this should no longer return an error. We should either: 

1. Remove / replace it with a warning later on if no solutions are found, or
2. Print an error instead of returning an error

pkg/config/compatibility.go
mattt: Setting a default value for that flag should do the trick. See https://github.com/replicate/cog/pull/1715/files#r1628418365.

tools/compatgen/main.go
marcusmartins: I was considering setting an minimum version here but I didn't want to leak the concern of what minimum version is acceptable from the `github.com/replicate/cog/pkg/config` package. But as there's a lot of coupling between the code here and `github.com/replicate/cog/pkg/config`, I am reconsidering this change and wondering if making this configurable via a flag is worth the expansion in the footprint.

I am proposing that we keep the FetchTensorFlowCompatibilityMatrix parameterized but just set it from a constant here.

timfallmk (approved):

Issue #1719, opened 2024-06-06
Health check endpoint unresponsive during predictions


ggilder:
Hi, I'm trying to use cog to build images I can run in a kubernetes cluster. I'm seeing an issue where I can't query the state of each pod (e.g. using the `/health-check` endpoint that cog defines) while a prediction is running.

I have a simple demonstration here: https://github.com/ggilder/cog-health-check-test

You can run the above on Docker Desktop. It's basically the cog "hello world" example with a sleep added to simulate a long-running prediction. The test script hits the prediction endpoint and then attempts to hit the health check endpoint, but the server waits for the prediction to complete before handling the health check, which doesn't seem to match the intended behavior (since e.g. the [health state](https://github.com/replicate/cog/blob/56615a7ed1142ef5abd6b0aa4133d2c0c2b972cf/python/cog/server/http.py#L64) is defined to allow a "busy" state).

Am I missing something that would allow the server to respond to health checks during predictions?

ggilder:
Perhaps the [prediction endpoint](https://github.com/replicate/cog/blob/56615a7ed1142ef5abd6b0aa4133d2c0c2b972cf/python/cog/server/http.py#L274) should be declared with `def` rather than `async def` so that it doesn't block the main thread? (per [FastAPI docs](https://fastapi.tiangolo.com/async/#path-operation-functions))

mattt:
Hi @ggilder. Thank you so much for taking the time to put together that demo and contributing a possible fix with #1720.

What you're observing is a result of the asynchronous prediction handling introduced by https://github.com/replicate/cog/pull/870 interacting with a more recent change in #1355.

In your demo, adding `-H "Prefer: respond-async"` to the original `curl` command allows the healthcheck to complete. But I'm not sure how expected / obvious / useful that behavior is. 

And as @nickstenning points out in that PR:

> There's a `--threads` argument to `cog.server.http` and I believe that it functions as a concurrency control for the synchronous prediction create endpoint.

@technillogue You have more context about this change and async Python in general. Do you think we should reconsider #1355? Or is that a prerequisite for the `async` branch we're working to merge?

ggilder:
I think there's definitely a use case here for running (synchronous) predictions on multiple VMs and check each one's readiness for additional work. The `Prefer` header is interesting but requiring use of webhooks isn't necessarily ideal.

[...]

PR #1720, opened 2024-06-06
Run predictions off main thread to avoid blocking health check

ggilder:
Fixes https://github.com/replicate/cog/issues/1719

Defining the prediction endpoints with `async def` runs them on the main thread per [FastAPI docs](https://fastapi.tiangolo.com/async/#path-operation-functions), which is problematic because it blocks the server from responding to the health check endpoint. Converting these to `def` allows health checks to run and fixes the problem I described in the above issue.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -271,7 +271,7 @@ def create_app(
         response_model=PredictionResponse,
         response_model_exclude_unset=True,
     )
-    async def predict(
+    def predict(
         request: PredictionRequest = Body(default=None),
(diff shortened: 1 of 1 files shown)


ggilder:
One side effect of this, which may or may not be desirable depending on your perspective, is that prediction requests to an instance that is currently running a prediction now fail with status code 409 and a “currently running a prediction” message, rather than essentially being queued up by uvicorn. I _think_ this is generally desirable since retrying the request could succeed (e.g. in a situation where multiple instances are available behind a load balancer).

PR #1721, opened 2024-06-06
Validate minor version for CUDA and Python

marcusmartins:
When the user specifies a CUDA or Python version in the cog.yaml, we now ensure that at least the minor version is set. This validation ensures that our compatibility version logic works properly.

Additionally, this commit adds logic to the CUDA version selection process, enabling it to pick the latest versions of both CUDA and Ubuntu that satisfy the major.minor constraints.

It also fixes a bug in the config parsing that caused a panic if the user defined only the major version of Python.

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -25,13 +25,13 @@ This stanza describes how to build the Docker image your model runs in. It conta
 
 ### `cuda`
 
-Cog automatically picks the correct version of CUDA to install, but this lets you override it for whatever reason.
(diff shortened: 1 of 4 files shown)


docs/yaml.md
```diff
@@ -25,13 +25,13 @@ This stanza describes how to build the Docker image your model runs in. It conta
 
 ### `cuda`
 
-Cog automatically picks the correct version of CUDA to install, but this lets you override it for whatever reason.
+Cog automatically picks the correct version of CUDA to install, but this lets you override it for whatever reason by specifying the minor (`11.8`) or patch (`11.8.0`) version of CUDA to use.
```
zeke, line 28: 👌🏼

zeke (approved):

mattt (approved):

PR #1747, opened 2024-06-18
Add label for base image's last layer's diff id to cog image

jessie-young:
Add a Docker label to the cog image so that if a cog base image was used, we add labels for the 1) diff id of the base image's last layer and 2) number of layers in the base image. Note: we are using the diff id, not layer digest, here so that we can inspect the image's config instead of having to query the remote registry to inspect the manifest.  

This allows the image conversion pipeline to find where the the base image layers end and where the user/application layers begin, so that it can separate the two.

Note that this change doesn't include any modifications to the tests - submitting this for initial review and will update the corresponding tests if this initial pass looks good.


pkg/image/build.go
```diff
@@ -156,6 +167,26 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 		global.LabelNamespace + "has_init": "true",
 	}
 
+	if isUsingCogBaseImage {
+		labels[global.LabelNamespace+"using_cog_base_image"] = "true"
```
jessie-young, line 171: nit, note to self: fix spacing

pkg/image/build.go
```diff
@@ -156,6 +167,38 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
...
+		// get the last layer of the cog base image so that when we look at the built cog image,
+		// we know where the base image ends
+		// pull the base image, as we'll need to pull it anyway to build the cog image
+		// Q: are we ok with the cog base image being pulled to the user's machine? can fix this later by implementing 
+		// a manifest inspect which doesn't require a pull
+
+		err := docker.Pull(cogBaseImageName)
+		if err != nil {
+			return fmt.Errorf("Failed to pull cog base image: %w", err)
+		}
```
mattt, line 183: Could we avoid pulling by using something like [go-containerregistry](https://github.com/google/go-containerregistry/blob/main/README.md) to inspect the manifest directly?

pkg/image/build.go
```diff
@@ -156,6 +167,38 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
...
+
+		if cogBaseImage.RootFS.Layers == nil || len(cogBaseImage.RootFS.Layers) == 0 {
+			return fmt.Errorf("Cog base image has no layers or RootFS is nil: %s", cogBaseImageName)
+		}
+
+		lastLayerIndex := len(cogBaseImage.RootFS.Layers) - 1
+		lastLayer := cogBaseImage.RootFS.Layers[lastLayerIndex]
+		console.Debugf("Last layer of the cog base image: %s", lastLayer) // prints the sha
+
+		labels[global.LabelNamespace + "cog-base-image-last-layer-diff-id"] = lastLayer
```
mattt, line 198: Is this a better name? Or is there some significance to it being a diff.

```suggestion
		labels[global.LabelNamespace + "cog-base-image-last-layer-sha"] = lastLayer
```

pkg/image/build.go
```diff
@@ -156,6 +167,38 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
...
+		labels[global.LabelNamespace + "cog-base-image-last-layer-idx"] = fmt.Sprintf("%d", lastLayerIndex)
```
mattt, line 199: Do we need this information if we have the layer identifier itself?

```suggestion
```

pkg/image/build.go
```diff
@@ -33,6 +33,9 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 	_ = os.Remove(bundledSchemaFile)
 	_ = os.Remove(bundledSchemaPy)
 
+	var isUsingCogBaseImage bool = false
```
mattt, line 36:
```suggestion
```

This is implied by `cogBaseImageName != ""`, so we can drop this variable

pkg/image/build.go
```diff
@@ -156,6 +167,38 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 		global.LabelNamespace + "has_init": "true",
 	}
 
+	if isUsingCogBaseImage {
+		labels[global.LabelNamespace + "using_cog_base_image"] = "true"
+		labels[global.LabelNamespace + "cog-base-image-name"] = cogBaseImageName
```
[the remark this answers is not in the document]
jessie-young: Yes - we can drop the use_cog_base_image. The convention for labels according to the OCI spec is using hyphens; however, I see that the other preexisting labels are using underscores, and I don't want to touch those and break anything downstream, so will keep the new labels as using hyphens.

pkg/image/build.go
jessie-young: The fetcher or inspect methods that are part of `crane` there looks promising - let me look into using that to prevent a base image pull.

pkg/image/build.go
tempusfrangit: Is there a reason we're doing underscores here and dashes the next line down?

tempusfrangit (commented):
The label looks fine to me. I defer to @mattt for the other comments. 

One question in-line.

pkg/image/build.go
mattt: See also https://github.com/replicate/cog/pull/1747#discussion_r1649278794

pkg/image/build.go
jessie-young: Added a TODO - since this requires a more more involved change involving getting hooked up with creds to use an http client to call the remote registry, will do this in a follow-up PR in the interest of getting this merged.

jessie-young: We can use the `remote` package to get a new instance of an Image, and then get the manifest via `img.Manifest()`

pkg/image/build.go
jessie-young: Yes, it's a diff-id of a layer, rather than the digest of the layer - but it's still a sha. Switched to sha for more conciseness.

pkg/image/build.go
jessie-young: Yes - this is an optimization for the image conversion pipeline so that rather than looking through the shas until we find a match, we can directly start looking for non-base layers starting at cog-base-image-last-layer-idx + 1. We double check things - we check the layer at the index, and check that its value matches the expected sha, in the pipeline, just in case the image somehow got tampered with or an extra layer was added.

PR #1751, opened 2024-06-18
add necessary headers to curl

Kettukaa:
The example curl to `/predict` does not include the necessary headers shown in [https://github.com/replicate/cog/blob/main/docs/http.md#post-predictions-synchronous](https://github.com/replicate/cog/blob/main/docs/http.md#post-predictions-synchronous)

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -27,6 +27,7 @@ Port 5000 is now serving the API:
 To run a prediction on the model, call the `/predictions` endpoint, passing input in the format expected by your model:
 
     curl http://localhost:5000/predictions -X POST \
+        --header "Content-Type:application/json" \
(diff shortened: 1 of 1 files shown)


docs/deploy.md
```diff
@@ -27,6 +27,7 @@ Port 5000 is now serving the API:
 To run a prediction on the model, call the `/predictions` endpoint, passing input in the format expected by your model:
 
     curl http://localhost:5000/predictions -X POST \
+        --header "Content-Type:application/json" \
```
mattt, line 30:
```suggestion
        --header "Content-Type: application/json" \
```

mattt (approved):
Thanks, @Kettukaa!

PR #1734, opened 2024-06-11
log traceback properly

technillogue:
when an error happens, we log the traceback. this goes through the normal logging mechanism, which drops the log if there's more than one prediction. we can use `_log(traceback.format_exc())` instead. also fix some minor formatting things.

PR #1752, opened 2024-06-19
async but refactored

technillogue:
this is like #1512, but I squashed and reordered some commits and then rebased onto main

conceptually we could split it up like:
* the commits up to Mux prediction events  (async runner, async predict, minimal async worker). this is uncertain because much but not all of it is fully dropped later
* httpx
* AsyncConnection, optimizations, tweak names
* omnibus actual concurrency
* remaining async features (emit_metric, time_share_metric, log traceback, batch_size metric)
however, I don't think we can actually release anything before "actual concurrency", the intermediate commits would likely be broken
we could alternatively write new code starting with the httpx refactor like you said, or otherwise carve out changes from that branch as new PRs (rather than rebasing the existing commits)

- **run CI for this branch the same way as for main**
- **async runner (#1352)**
- **support async predict functions  (#1350)**
- **AsyncConcatenateIterator**
- **create event loop before predictor setup (#1366)**
- **minimal async worker (#1410)**
- **Mux prediction events (#1405)**
- **replace requests with httpx and factor out clients (#1574, #1707, #1714)**
- **implement mp.Connection with async streams (#1640)**
- **optimize webhook serialization and logging (#1651)**
- **tweak names and style**
- **omnibus actual concurrency and major refactor (#1530)**
- **function to emit metrics (#1649)**
- **predict_time_share metric (#1643, #1683)**
- **Backport `Secret` type to async branch (#1706)**
- **log traceback properly (#1734)**


mattt:
Closing in favor of #1813

PR #1754, opened 2024-06-20
Document how Cog raises `CancelationException` in `predict` when a prediction is canceled

mattt:


nickstenning:
We should say something along the lines of:

> Your `predict` method MAY catch `CancelationException` in order to perform any necessary cleanup, but that cleanup should take as little time as possible (no more than a few seconds at most) and you MUST reraise the exception with a bare `raise`. If you fail to reraise the exception the container may be terminated.

I suspect a full example of what such a cleanup looks like would be helpful.

In addition, I think we probably want to move `CancelationException` to a public part of cog rather than `cog.server.exceptions`.

mattt:
Thanks, @nickstenning. Added that information with some example code.

docs/http.md
```diff
@@ -241,12 +241,36 @@ Prefer: respond-async
...
+or synchronously, without the `Prefer: respond-async` header.
+
+If no prediction exists with the provided `id`,
+then the server responds with status `404 Not Found`.
+Otherwise, the server responds with `200 OK`.
+
+When a prediction is canceled,
+Cog raises `cog.server.exceptions.CancelationException`
+in the model's `predict` function.
+This exception may be caught by the model to perform necessary cleanup.
```
mattt, line 261: Passive voice is not something liked by me, however...

I originally had "You", but thought that was ambiguous since the rest of this document is framed with explicit client and server roles.

PR #1558, opened 2024-03-05
Install Script

chigozienri:
Proposal for install.sh to reduce friction when installing cog

Run as follows:
```console
# via curl
sh -c "$(curl -fsSL https://raw.githubusercontent.com/replicate/cog/main/tools/install.sh)"
# or via wget:
sh -c "$(wget -qO- https://raw.githubusercontent.com/replicate/cog/main/tools/install.sh)"
# or via fetch:
sh -c "$(fetch -o - https://raw.githubusercontent.com/replicate/cog/main/tools/install.sh)"
```

See Slack comments here https://replicatehq.slack.com/archives/C05600FDYTE/p1709142996116999

PR #1757, opened 2024-06-20
Update install script instructions

mattt:
Follow up to #1558 

Trying this out myself, I noticed that it wasn't being run interactively. To make that work, you need to redirect the source directly into a `sh` process. Piped execution doesn't support interactivity.

```console
$ curl -fsSL https://cog.run/install.sh | sh
On OS X, it is recommended to install cog using Homebrew instead:
`brew install cog`
Do you want to continue with this installation anyway?
```

This PR updates to use process substitution syntax. As discussed in https://github.com/fish-shell/fish-shell/issues/5181, fish (my preferred shell) uses a different syntax, so I added special instructions for that. I also removed instructions for `fetch`, because I think `wget` and `curl` cover the vast majority of use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,13 +124,16 @@ brew install cog
 You can also download and install the latest release using our 
 [install script](https://cog.run/install):
 
-```
-# via curl
-curl -fsSL https://cog.run/install | sh
-# or via wget:
-wget -qO- https://cog.run/install | sh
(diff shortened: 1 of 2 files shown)


mattt:
Also, I'm hard-wired to notice "OS X", so I updated that to "macOS".

Issue #982, opened 2023-04-05
Unpickling errors


akshhack:
Our COG fails on very peculiar issue related to unpickling which isn't very clear. Wondering if anybody has faced this and what the resolution was?

```
jsonPayload: {
exception: "Traceback (most recent call last):
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/server/runner.py", line 113, in handle_error
    raise error
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/multiprocessing/pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/server/runner.py", line 334, in predict
    return _predict(
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/server/runner.py", line 370, in _predict
    for event in worker.predict(input_dict, poll=0.1):
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/site-packages/cog/server/worker.py", line 118, in _wait
    ev = self._events.recv()
  File "/root/.pyenv/versions/3.10.10/lib/python3.10/multiprocessing/connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
_pickle.UnpicklingError: invalid load key, '\x00'."
logger: "cog.server.runner"
message: "caught exception while running prediction"
timestamp: "2023-04-05T08:19:51.798365Z"
}
```

For context, we're calling a command line utility from subprocess.run() inside the COG. Wondering if it could cause issues. Furthermore, this issue is intermittent... so very sneaky to debug. Thanks!

ivanprado:
I think the problem in my case is the 32MiB limit of the [Connection in multiprocessing](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.connection.Connection.send). When the response is too big, more than 32MiB, it just fails. I don't see an easy solution.

Issue #1446, opened 2023-12-20
UnpicklingError: invalid load key, '\x00'


technillogue:
Large outputs such as embedding batches will throw errors like this

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/cog/server/runner.py", line 338, in predict
    return _predict(
  File "/usr/local/lib/python3.10/site-packages/cog/server/runner.py", line 374, in _predict
    for event in worker.predict(input_dict, poll=0.1):
  File "/usr/local/lib/python3.10/site-packages/cog/server/worker.py", line 118, in _wait
    ev = self._events.recv()
  File "/usr/local/lib/python3.10/multiprocessing/connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
_pickle.UnpicklingError: invalid load key, '\x00'.
```

This is because the object is larger than connection socket buffer, so the unpickler tries to parse a null byte, which is not a valid opcode, leading to this somewhat hard-to-google error caused by cog. 

The workaround for this is to use a smaller batch size or return objects as files; this is often unrelated to corrupted weights or pget. Returning large objects not as a file is discouraged even when this is fixed in #1316. I'm putting this here so the workaround is findable-ish in the meantime.

related: #1308, #982, and many requests on discord

Issue #1308, opened 2023-09-19
invalid load key


zhangvia:
```json
{"error": "invalid load key, '\\x00'.", "prediction_id": "d3dfa92e-40c9-4866-b854-1437dd8b264a", "logger": "cog.server.runner", "timestamp": "2023-09-19T02:31:29.259351Z", "severity": "INFO", "message": "prediction failed"}
{"logger": "cog.server.runner", "timestamp": "2023-09-19T02:31:29.362471Z", "exception": "Traceback (most recent call last):\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/site-packages/cog/server/runner.py\", line 113, in handle_error\n    raise error\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/multiprocessing/pool.py\", line 125, in worker\n    result = (True, func(*args, **kwds))\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/site-packages/cog/server/runner.py\", line 334, in predict\n    return _predict(\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/site-packages/cog/server/runner.py\", line 370, in _predict\n    for event in worker.predict(input_dict, poll=0.1):\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/site-packages/cog/server/worker.py\", line 118, in _wait\n    ev = self._events.recv()\n  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/multiprocessing/connection.py\", line 251, in recv\n    return _ForkingPickler.loads(buf.getbuffer())\n_pickle.UnpicklingError: invalid load key, '\\x00'.", "severity": "ERROR", "message": "caught exception while running prediction"}
```
what does this error means? i think the error is not due to my predict code, but why is this happen?

mattt:
Reformatting for readability:

```
Traceback (most recent call last):
  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/site-packages/cog/server/runner.py\", line 113, in handle_error
    raise error
  File \"/media/74nvme/software/miniconda3/envs/goods_inpaint_dev/lib/python3.8/multiprocessing/pool.py\", line 125, in worker
    result = (True, func(*args, **kwds))
[...]

PR #1758, opened 2024-06-21
fix: StreamRedirector may cause data corruption due to race condition

masolin:
Fix: #982 #1308 #1446 

This bug also causes another error pattern which hasn't been reported by other yet.

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 401896: invalid start byte

  File "cog/server/runner.py", line 335, in predict
    return _predict(
  File "cog/server/runner.py", line 371, in _predict
    for event in worker.predict(input_dict, poll=0.1):
  File "cog/server/worker.py", line 111, in _wait
    ev = self._events.recv()
  File "multiprocessing/connection.py", line 251, in recv
    return _ForkingPickler.loads(buf.getbuffer())
```

**Root cause**

`StreamRedirector` and `_ChildWorker` use the same pipe, `_events`, and they run in different threads, so there is a chance that `StreamRedirector` and `_ChildWorker` are writing to the `_events` at the same time. This causes `Worker`, the receiver of the pipe, fails to unpickle due to data corruption.

This can also explains why the error happens more often when the prediction output is big. Big prediction output takes longer time to send, so it's more likely `StreamRedirector` happens to send while prediction output is sending.

**FIx**

Add a lock to force only one of `StreamRedirector` and `_ChildWorker` can send data to the pipe, `_events`, at the same time.

@allcontributors please add @masolin for code


mattt:
Hi @masolin. Thanks so much for opening this PR. I just approved a CI run and will take a closer look later today.

python/cog/server/worker.py
```diff
@@ -221,13 +225,14 @@ def _predict(self, payload: Dict[str, Any]) -> None:
...
-                    self._events.send(PredictionOutputType(multi=False))
-                    self._events.send(PredictionOutput(payload=make_encodeable(result)))
+                with self._events_lock:
+                    if isinstance(result, types.GeneratorType):
+                        self._events.send(PredictionOutputType(multi=True))
+                        for r in result:
+                            self._events.send(PredictionOutput(payload=make_encodeable(r)))
+                    else:
+                        self._events.send(PredictionOutputType(multi=False))
+                        self._events.send(PredictionOutput(payload=make_encodeable(result)))
```
mattt, line 235: Fix formatting error (`python -m ruff format python`)

```suggestion
                with self._events_lock:
                    if isinstance(result, types.GeneratorType):
                        self._events.send(PredictionOutputType(multi=True))
                        for r in result:
                            self._events.send(
                                PredictionOutput(payload=make_encodeable(r))
                            )
                    else:
                        self._events.send(PredictionOutputType(multi=False))
                        self._events.send(
                            PredictionOutput(payload=make_encodeable(result))
                        )
```

mattt:
@masolin Thanks again for your PR. One of the CI failures was a formatting issue. I tried pushing, but it seems like [your fork doesn't allow maintainers to push](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). So please apply the suggested changes or run `python -m ruff format python` and commit yourself.

python/cog/server/worker.py
masolin: Thanks, I have reformatted the code.

masolin:
Thanks @mattt . I have reformatted the code.
In the previous run, the Python 3.12 test failed due to timeout, should I extend the timeout duration for CI?

python/cog/server/worker.py
```diff
@@ -153,10 +153,13 @@ def __init__(
         self._events = events
         self._tee_output = tee_output
         self._cancelable = False
+        self._events_lock = None
 
         super().__init__()
 
     def run(self) -> None:
+        self._events_lock = multiprocessing.Lock()
+
```
mattt, line 162: I'm seeing a new [CI failure for typechecking](https://github.com/replicate/cog/actions/runs/9638832220/job/26665632298?pr=1758#step:5:22)

> Object of type "None" cannot be used with "with" (reportOptionalContextManager)

I think that could be fixed by doing assignment in `__init__`.

```suggestion
        self._events_lock = multiprocessing.Lock()

        super().__init__()

    def run(self) -> None:
```

mattt: Or else, you could annotate as `Optional` in the initializer. Downside of that approach is that you'd have to add conditionals to check for non-`None` wherever that's used.

```suggestion
        self._events_lock: Optional[multiprocessing.Lock] = None

        super().__init__()

    def run(self) -> None:
        self._events_lock = multiprocessing.Lock()

```

mattt (commented):
Thanks for applying those changes, @masolin. I'm seeing a new CI failure for typechecking. Left some comments with some suggestions on how to fix that.

python/cog/server/worker.py
masolin: Thanks @mattt . I didn't notice that there was a type checker.
I have moved the lock initialization to the `__init__` and specified the lock for spawn, because  `_ChildWorker` is a spawn process. The default `multiprocessing.Lock()` is based on OS and can be `fork` lock, so it may not work if we move the lock initialization to `__init__` for some OS.

mattt:
Thanks again for contributing this fix, @masolin! And thanks for your patience in getting this landed. I created https://github.com/replicate/cog/pull/1773 from this PR branch to work around some weirdness I was seeing with CI. That's now now merged.

My plan is to cut a new release with this sometime this week.

PR #1761, opened 2024-06-24
Rewrite HTTP API documentation

mattt:
I started documenting the HTTP server host parameter added by #1748, but one thing led to another, and I ended up rewriting the [HTTP API docs](https://github.com/replicate/cog/blob/main/docs/http.md).

At a high level, this PR reorganizes the document by extracting discussion of topics like sync/async, idempotency, and webhooks out of individual endpoints and into top-level sections. I also added missing discussion of how file uploads work, and removed information about running the HTTP server which were duplicated in the [deployment](https://github.com/replicate/cog/blob/main/docs/deploy.md) doc.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -51,13 +51,9 @@ call the `/predictions` endpoint,
 passing input in the format expected by your model:
 
 ```console
-curl http://localhost:5000/predictions -X POST \
+curl http://localhost:5001/predictions -X POST \
     --header "Content-Type: application/json" \
     --data '{"input": {"image": "https://.../input.jpg"}}'
-```
-
-To view the API documentation in browser for the model that is running, 
-open [http://localhost:5000/docs](http://localhost:5000/docs).
 
 For more details about the HTTP API, 
 see the [HTTP API reference documentation](http.md).
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -1,38 +1,66 @@
 # Deploy models with Cog
 
-Cog containers are Docker containers that serve an HTTP server for running predictions on your model. You can deploy them anywhere that Docker containers run.
+Cog containers are Docker containers that serve an HTTP server 
+for running predictions on your model. 
+You can deploy them anywhere that Docker containers run.
 
-This guide assumes you have a model packaged with Cog. If you don't, [follow our getting started guide](getting-started-own-model.md), or use [an example model](https://github.com/replicate/cog-examples).
+This guide assumes you have a model packaged with Cog. 
+If you don't, [follow our getting started guide](getting-started-own-model.md), 
+or use [an example model](https://github.com/replicate/cog-examples).
 
 ## Getting started
 
 First, build your model:
 
-    cog build -t my-model
+```console
+cog build -t my-model
+```
 
 Then, start the Docker container:
 
-    docker run -d -p 5000:5000 my-model
+```shell
+# If your model uses a CPU:
+docker run -d -p 5001:5000 my-model
 
-    # If your model uses a GPU:
-    docker run -d -p 5000:5000 --gpus all my-model
+# If your model uses a GPU:
+docker run -d -p 5001:5000 --gpus all my-model
 
-    # If you're on an M1 Mac:
-    docker run -d -p 5000:5000 --platform=linux/amd64 my-model
+# If you're on an M1 Mac:
+docker run -d -p 5001:5000 --platform=linux/amd64 my-model
+```
 
-Port 5000 is now serving the API:
+The server is now running locally on port 5001.
 
-    curl http://localhost:5000
+To view the OpenAPI schema, 
+open [localhost:5001/openapi.json](http://localhost:5001/openapi.json) 
+in your browser 
+or use cURL to make a request:
 
-To run a prediction on the model, call the `/predictions` endpoint, passing input in the format expected by your model:
+```console
+curl http://localhost:5001/openapi.json
+```
 
-    curl http://localhost:5000/predictions -X POST \
-        --header "Content-Type: application/json" \
-        --data '{"input": {"image": "https://.../input.jpg"}}'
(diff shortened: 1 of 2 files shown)


docs/deploy.md
```diff
@@ -1,38 +1,66 @@
 # Deploy models with Cog
 
-Cog containers are Docker containers that serve an HTTP server for running predictions on your model. You can deploy them anywhere that Docker containers run.
+Cog containers are Docker containers that serve an HTTP server
```
evazorro, line 3: Looks like you're using [semantic line breaks](https://sembr.org/)? I like it! My only concern is making sure others know to follow this convention, otherwise you lose the diffing benefits

docs/deploy.md
```diff
@@ -1,38 +1,66 @@
...
 
-    curl http://localhost:5000
+To view the OpenAPI schema, 
+open [localhost:5001/openapi.json](http://localhost:5001/openapi.json) 
+in your browser 
+or use cURL to make a request:
 
-To run a prediction on the model, call the `/predictions` endpoint, passing input in the format expected by your model:
+```console
+curl http://localhost:5001/openapi.json
```
evazorro, line 40: I took the opportunity to run through this doc and test it out. I'm able to see http://localhost:5001/openapi.json just fine, but it looks like port 5000 is not connecting.

```
$ curl http://localhost:5000/predictions -X POST \
    --header "Content-Type: application/json" \
    --data '{"input": "Evazorro"}}'
curl: (7) Failed to connect to localhost port 5000 after 0 ms: Connection refused
```

It may just be user error on my part (or a quirk of using Docker with WSL), but maybe we can debug just to be sure?

docs/http.md
```diff
@@ -1,79 +1,239 @@
 # HTTP API
 
-When a Cog Docker image is run, it serves an HTTP API for making predictions. For more information, take a look at [the documentation for deploying models](deploy.md).
+> [!TIP]
+> For information about how to run the HTTP server, 
+> see [our documentation to deploying models](deploy.md).
+
+When you run a Docker image built by Cog, 
+it serves an HTTP API for making predictions.
```
evazorro, line 8: This would be a larger change (and therefore perhaps best made in a separate PR?) but we've been talking about moving away from using the term "prediction" in favor of more software-engineering-y terms like "run" (context: [style guide](https://www.notion.so/replicate/Style-guide-377cfdbd142540a9b3177734b63d0baf?pvs=4#8a0bfae63f514b2d955cf332cf429cfb)).

So this would become something like "it serves an HTTP API for running the model" ? Again, mostly just bringing it up here for discussion; this doc has so many mentions of prediction that I think if we do want to change the language we'd want to do that in a followup.

docs/http.md
```diff
@@ -1,79 +1,239 @@
...
-  - [`POST /predictions/<prediction_id>/cancel`](#post-predictionsprediction_idcancel)
+- Use synchronous endpoints when you want to wait for the prediction result.
+- Use asynchronous endpoints when you want to start a prediction 
+  and receive updates via webhooks.
+- Use idempotent endpoints when you need to safely retry requests 
+  without creating duplicate predictions.
+
+## Webhooks
+
+The client can provide a `webhook` parameter in the request body
```
evazorro, line 58:
```suggestion
You can provide a `webhook` parameter in the client request body
```

Making it a bit more active

docs/http.md
```diff
@@ -1,79 +1,239 @@
...
+by setting the `Prefer: respond-async` header in their request.
+When provided, the server responds immediately after starting the prediction 
+with `202 Accepted` status and a prediction object in status `processing`.
+
+> [!NOTE]
+> The only supported way to receive updates on the status of predictions
+> started asynchronously is using [webhooks](#webhooks). 
+> Polling for prediction status is not currently supported.
+
+The server also provides a way to create predictions idempotently,
```
evazorro, line 29: I want to make this a bit more user-active, but without losing any technical nuance. What do you think of this?

```suggestion
You can also use certain server endpoints to create predictions idempotently,
```

evazorro (commented):
I like the reorganization! Definitely agree with putting deploy instructions in a separate doc and pulling out info on webhooks / unique IDs / file uploads into their own sections.

Some of the language gets a little dense, so I put in a few suggestions to simplify and make the user action more clear.

docs/deploy.md
mattt: Well, gotta practice what I [preach](https://github.com/sembr/specification?tab=readme-ov-file#about) 😇 

I'd be very happy to adopt this as an internal standard. For now, I'm using the lack of sembr as a signal for what could use a review.

docs/deploy.md
mattt: Fun backstory on this is that [Apple took over port 5000](https://www.reddit.com/r/webdev/comments/qg8yt9/apple_took_over_port_5000_in_the_latest_macos/) by running the AirPlay server that port by default. The first version of Cog rolled out just in time to get bit by that 🥲  

Hence this awkward misdirection where you have to map container 5000 to host 5001.

evazorro: Oof Apple. Ok great, glad I found something.

docs/deploy.md
zeke: Whoa sembr never heard of it but I like it!

docs/deploy.md
mattt: Filed a tracking issue here: #1777

docs/http.md
mattt: Yes, please! Caveat:

You run models, models are run — I agree with that. When you run the model for inference, the method called is `predict` and the record / result of your running the model is a "prediction". I don't think we should start calling that a "run" (if only because of its scatalogical plural form, but also because that's a big API change).

mattt:
Thanks so much for the feedback, @evazorro! Merging this now.

PR #1773, opened 2024-06-26
Fix possible data corruption due to race condition in StreamRedirector

mattt:
Alternative to #1758

Resolves https://github.com/replicate/cog/issues/982 Resolves https://github.com/replicate/cog/issues/1308 Resolves https://github.com/replicate/cog/issues/1446

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -229,10 +229,14 @@ class _ChildWorker(_spawn.Process):  # type: ignore
                     if isinstance(result, types.GeneratorType):
                         self._events.send(PredictionOutputType(multi=True))
                         for r in result:
-                            self._events.send(PredictionOutput(payload=make_encodeable(r)))
+                            self._events.send(
+                                PredictionOutput(payload=make_encodeable(r))
+                            )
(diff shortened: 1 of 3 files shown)


python/cog/server/worker.py
```diff
@@ -221,13 +223,18 @@ def _predict(self, payload: Dict[str, Any]) -> None:
...
-                    self._events.send(PredictionOutputType(multi=True))
-                    for r in result:
-                        self._events.send(PredictionOutput(payload=make_encodeable(r)))
-                else:
-                    self._events.send(PredictionOutputType(multi=False))
-                    self._events.send(PredictionOutput(payload=make_encodeable(result)))
+                with self._events_lock:
+                    if isinstance(result, types.GeneratorType):
+                        self._events.send(PredictionOutputType(multi=True))
+                        for r in result:
```
nickstenning, line 229: This is holding the lock for wayyyyyy too long. If the result is generator we absolutely should not be holding the lock for the entire `for r in result` loop.

I wonder if this is related to the occasional customer report of not getting logs?

PR #1774, opened 2024-06-26
Fix `cog build --use-cog-base-image` with minor and patch versions of Python and Torch

mattt:
Follow-up to #1769 

This PR does the following:

* Adds (previously failing) test coverage for `cog.yaml` files with various minor and patch versions of Torch
* Updates logic for matching base images to pass aforementioned test case
* Strips patch from configured versions when generating `FROM` base image name

Now, when I run `cog build --use-cog-base-image` with either of the following configurations:

**With patch versions**

```yaml
build:
  gpu: true
  cuda: "12.1"
  python_version: "3.11.7"
  python_packages:
    - "torch==2.3.0"
```

**Without patch versions**

```yaml
build:
  gpu: true
  cuda: "12.1"
  python_version: "3.11"
  python_packages:
    - "torch==2.3"
```

...both attempt to pull `FROM r8.im/cog-base:cuda12.1-python3.11-torch2.3` [^1]

[^1]: However, this fails with `ERROR: failed to solve: failed to resolve source metadata for r8.im/cog-base:cuda12.1-python3.11-torch2.3: r8.im/cog-base:cuda12.1-python3.11-torch2.3: not found`

### DIFF
--- a/pkg/cli/baseimage.go
+++ b/pkg/cli/baseimage.go
@@ -139,5 +139,4 @@ func baseImageGeneratorFromFlags() (*dockerfile.BaseImageGenerator, error) {
 		baseImagePythonVersion,
 		baseImageTorchVersion,
 	)
-
 }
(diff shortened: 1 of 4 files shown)


andreasjansson:
This looks good! Should we message the user if they specify a patch version, since it may be different from the version in the base image?

mattt:
@andreasjansson Yes, absolutely. I was planning to implement that as a separate PR. But I can do that here instead.

mattt:
_Wake up babe, new Ruff errors just dropped_ 🫠  https://github.com/replicate/cog/pull/1775

PR #1778, opened 2024-07-01
Enable `--use-cog-base-image` by default

mattt:


pkg/cli/build.go
```diff
@@ -110,7 +110,7 @@ func addDockerfileFlag(cmd *cobra.Command) {
 }
 
 func addUseCogBaseImageFlag(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&buildUseCogBaseImage, "use-cog-base-image", false, "Use pre-built Cog base image for faster cold boots")
+	cmd.Flags().BoolVar(&buildUseCogBaseImage, "use-cog-base-image", true, "Use pre-built Cog base image for faster cold boots")
```
andreasjansson, line 113: Is the correct syntax for disabling this `--use-cog-base-image=false` or `--no-use-cog-base-image`?

andreasjansson (approved):

andreasjansson:
Base images are still not pulled on all clusters, so shall we hold this off until that's fully deployed?

mattt:
> Base images are still not pulled on all clusters, so shall we hold this off until that's fully deployed?

Merging would only affect anyone running source builds from latest main. So I don't think that's a blocker until we cut a new release.

pkg/cli/build.go
mattt: The former. `--no-` prefixes aren't natively supported. See https://github.com/spf13/pflag/issues/214

mattt: That said, we could add a `--no-use-cog-base-image` flag alias

andreasjansson:
Base images have been deployed on most clusters, so I think this is ready to merge.

PR #1785, opened 2024-07-03
fix ci

technillogue:
this is to release the should_exit fix


technillogue:
helpfully fixed in #1799 #1797

PR #1786, opened 2024-07-03
Fix send race

technillogue:
This solves a similar problem as #1758: in some cases, currently emitting a metric or yielding an output can result in an EAGAIN error, and we believe Done events are sometimes dropped. In async land, we shouldn't need a lock; multiple writers should be able to write to a [StreamWriter](https://docs.python.org/3/library/asyncio-stream.html#streamwriter) safely. We also keep a sync Lock when the predictor is not async.

<details><summary>However, I'm still not sure if StreamWriter is thread-safe. </summary>
After some research, StreamWriter.write calls self._transport.write and when StreamWriter is wrapping a unix socket and using the default selector, the _transport is a _SelectorSocketTransport, which calls [self._buffer.extend](https://github.com/python/cpython/blob/3.11/Lib/asyncio/selector_events.py#L1077), and self._buffer is a [bytearray](https://github.com/python/cpython/blob/3.11/Lib/asyncio/selector_events.py#L761). However, on 3.12, [append is called](https://github.com/python/cpython/blob/3.12/Lib/asyncio/selector_events.py#L1091) instead of extend.

After more soul-searching, I _think_ write should be basically thread-safe: the two critical calls, socket.send() and bytearray.append/extend should be represented as a single bytecode instruction and some native code which works on python objects and does not release the GIL. As far as I understand, [python threads are switched "between bytecodes"](https://stackoverflow.com/a/55197452). Even though the individual calls are thread-safe, the overall write method is not quite thread safe: a relevant thread switch could occur between the `if not self._buffer`, `_sock.send`, and `_buffer.extend` lines. However, in that case the race condition would result data could being sent out of order or incorrectly delayed/sent early, but it I think it shouldn't be corrupted, and we don't really care about the ordering of log lines vs outputs.
</details>

I can see two solutions:

1. Ideally, we would probably add an alternate implementation of StreamWriter that uses the main event loop and wraps stderr/stdout in StreamReaders, start a separate task for each stream, and use `await wrapped_stream.read()` so that threads are not necessary. there's a tricky moment where we need to use the threaded StreamRedirector to capture logs while the predictor is being imported, since we don't know if we're going to be async or not. 
2. Alternatively, we could try to use a queue or deque to communicate between threads, so instead of calling stream_write_hook StreamRedirector would do queue.put, the main event loop would do queue.get and not worry about thread safety for the pipe. The problem with this is that asyncio.Queue is not thread-safe and queue.Queue could block the event loop. You could use a busy wait or similar and get_nowait, but that has other downsides.

### DIFF
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -200,8 +200,16 @@ class StreamRedirector(threading.Thread):
 
         We must not call write_hook twice for the same data during the switch.
         """
-        debug("swithc async")
+        debug("switch async, drain")
+        # Drain the streams to ensure all buffered data is processed
+        self.drain()
+
+        # Shut down the thread
+        # we do this before starting a coroutine that will also read from the same fd
+        # so that shutdown can find the terminate tokens correctly
+        self.shutdown()
         self.stream_tasks = []
+        self.is_async = True
 
         for stream in self._streams:
             # Open each stream as a StreamReader
@@ -212,13 +220,6 @@ class StreamRedirector(threading.Thread):
             task = asyncio.create_task(self.process_stream(stream, reader))
             self.stream_tasks.append(task)
 
-        # Drain the streams to ensure all buffered data is processed
-        self.drain()
-
-        # Shut down the thread
-        self.shutdown()
-        self.is_async = True
-
     async def process_stream(
         self, stream: WrappedStream, reader: asyncio.StreamReader
     ) -> None:
@@ -227,20 +228,17 @@ class StreamRedirector(threading.Thread):
         drain_tokens_seen = 0
         should_exit = False
 
-        while not should_exit:
-            line = await reader.readline()
-            debug("process stream got line")
+        async for line in reader:
 
             if not line:
                 break
 
-            line = line.decode().strip()
+            line = line.decode()
+            debug("redirector saw", line)
 
             if not line.endswith("\n"):
                 buffer.write(line)
                 continue
-
-            debug("redirector saw", line)
 
             full_line = buffer.getvalue() + line.strip()
 
@@ -263,3 +261,5 @@ class StreamRedirector(threading.Thread):
                 debug("drain event set")
                 self.drain_event.set()
                 drain_tokens_seen = 0
+            if should_exit:
+                break
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -11,6 +11,8 @@ from typing import (
 )
 
 
+debug = lambda *args, f=open("/tmp/debug", "a"): print(*args, file=f, flush=True)
+
 class WrappedStream:
     def __init__(self, name: str, stream: TextIO) -> None:
         self.name = name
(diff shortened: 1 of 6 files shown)


python/cog/server/helpers.py
```diff
@@ -148,8 +180,98 @@ def run(self) -> None:
...
+                full_line = full_line[: -len(self.terminate_token)]
+                should_exit = True
+
+            if full_line.endswith(self.drain_token):
+                drain_tokens_seen += 1
+                full_line = full_line[: -len(self.drain_token)]
+
+            if full_line:
+                # Call write_hook from the main thread
+                self._write_hook(stream.name, stream.original, full_line + "\n")
```
mattt, line 270: Worth profiling, but precomputed / fixed lengths instead of a string suffix and `len` could be noticeably faster if this is a hot code path.

mattt (commented):
This is looking good! I think if we can get tests passing (integration got stuck on the last commit, and I cancelled after 38 minutes) and clean up the debug statements / comments, this should be ready to go.

python/cog/server/helpers.py
```diff
@@ -148,8 +180,98 @@ def run(self) -> None:
...
+    ) -> None:
+        debug("process_stream", stream.name)
+        buffer = io.StringIO()
+        drain_tokens_seen = 0
+        should_exit = False
+
+        async for line in reader:
+
+            if not line:
+                break
```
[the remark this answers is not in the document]
technillogue: I'm just following exactly what StreamRedirector does here and for the other comment. it could probably be improved though, yeah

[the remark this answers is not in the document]
technillogue: in theory, all of the non-EOF lines returned by `__[a]iter__` should have a trailing newline, I think. I think we are supposed to exit if there's a blank line with no newline or anything

python/cog/server/worker.py
```diff
@@ -155,20 +181,40 @@ def _setup(self) -> None:
...
+                debug("inspect")
                 if inspect.iscoroutinefunction(self._predictor.setup):
-                    self.loop.run_until_complete(run_setup_async(self._predictor))
+
+                    async def setup_async() -> None:
+                        # we prefer to not stop-start the event loop between these calls
+                        await self._async_init()  # this creates tasks
+                        debug("async_init done")
+                        await run_setup_async(self._predictor)
+                        debug("ran setup async")
```
technillogue, line 200: not sure if this is still needed

mattt: Repeating back to confirm my own understanding: We no longer need `setup_async`, because our new approach is to always run (sync) `setup` and then transition to async at the end after we know the signature of `predict`. Right?

technillogue: no, setup may be async -- what I mean here is that 
```suggestion
                    self.loop.run_until_complete(self._async_init())
                    self.loop.run_until_complete(run_setup_async(self._predictor))
```
I added the setup_async function there so that the event loop kept going between the two coros because I thought that was causing problems, but I'm not sure that's necessary

technillogue:
https://replicatehq.slack.com/archives/C0617RF4HHP/p1719896679090459?thread_ts=1719867269.654439&cid=C0617RF4HHP

mattt:
Closing in favor of #1831

PR #1787, opened 2024-07-04
[async] Propagate trace context to webhook and upload requests

aron:
Based on the implementation in #1698.

If the request to /predict contains headers `traceparent` and `tracestate` defined by w3c Trace Context[^1] then these headers are forwarded on to the webhook and upload calls.

This allows observability systems to link requests passing through cog.

[^1]: https://www.w3.org/TR/trace-context/

### DIFF
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -15,6 +15,7 @@ from ..types import Path
 from .eventtypes import PredictionInput
 from .response_throttler import ResponseThrottler
 from .retry_transport import RetryTransport
+from .telemetry import current_trace_context
 
 log = structlog.get_logger(__name__)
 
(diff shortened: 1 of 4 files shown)


python/tests/server/test_http.py
```diff
@@ -560,6 +560,78 @@ def test_asynchronous_prediction_endpoint(client, match):
     assert webhook.call_count == 1
 
 
+# End-to-end test for passing tracing headers on to downstream services.
+@pytest.mark.xfail  # requires respx to pass
```
aron, line 564: I dunno what `respx` is… how would I test that this actually passes?

aron: Okay, worked it out and added a test.

python/cog/server/useragent.py
```diff
@@ -0,0 +1,17 @@
...
+            return version("cog")
+        import pkg_resources
+
+        return pkg_resources.get_distribution("cog").version
+    except Exception:
+        return "unknown"
+
+
+def get_user_agent() -> str:
+    return f"cog-worker/{_get_version()}"
```
technillogue, line 17: why pull this out? in the async branch I've been doing more inlining and less indirection where this makes sense. is get_user_agent/_get_version used in more than one place?

technillogue (commented):
thank you so much for taking a stab at the tests :pray:

python/cog/server/useragent.py
aron: Hmm, in this branch it's not. But these files exist on main, so I tried to keep it consistent. I'm happy to inline it.

python/cog/server/clients.py
```diff
@@ -45,14 +46,26 @@ def _get_version() -> str:
...
     if auth_token:
         headers["authorization"] = "Bearer " + auth_token
+
     return headers
 
 
+async def on_request_trace_context_hook(request: httpx.Request) -> None:
+    ctx = current_trace_context() or {}
+    for key, value in ctx.items():
+        request.headers[key] = str(value)
```
technillogue, line 66:
```suggestion
    request.headers.update(ctx)
```

PR #1788, opened 2024-07-04
[async] Include prediction id upload request

aron:
Based on #1667

This PR introduces two small changes to the file upload interface.

1. We now allow downstream services to include the destination of the asset in a `Location` header, rather than assuming that it's the same as the final upload url (either the one passed via `--upload-url` or the result of a 307 redirect response.

2. We now include the `X-Prediction-Id` header in upload request, this allows the downstream client to potentially do configuration/routing based on the prediction ID. This ID should be considered unsafe and needs to be validated by the downstream service.

### DIFF
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -177,10 +177,11 @@ class ClientManager:
 
     # files
 
-    async def upload_file(self, fh: io.IOBase, url: Optional[str]) -> str:
+    async def upload_file(
+        self, fh: io.IOBase, *, url: Optional[str], prediction_id: Optional[str]
+    ) -> str:
         """put file to signed endpoint"""
         log.debug("upload_file")
-        fh.seek(0)
         # try to guess the filename of the given object
         name = getattr(fh, "name", "file")
         filename = os.path.basename(name) or "file"
@@ -198,17 +199,24 @@ class ClientManager:
         # ensure trailing slash
         url_with_trailing_slash = url if url.endswith("/") else url + "/"
 
-        async def chunk_file_reader() -> AsyncIterator[bytes]:
-            while 1:
-                chunk = fh.read(1024 * 1024)
-                if isinstance(chunk, str):
-                    chunk = chunk.encode("utf-8")
-                if not chunk:
-                    log.info("finished reading file")
-                    break
-                yield chunk
+        class ChunkFileReader:
+            async def __aiter__(self) -> AsyncIterator[bytes]:
+                fh.seek(0)
+                while 1:
+                    chunk = fh.read(1024 * 1024)
+                    if isinstance(chunk, str):
+                        chunk = chunk.encode("utf-8")
+                    if not chunk:
+                        log.info("finished reading file")
+                        break
+                    yield chunk
 
         url = url_with_trailing_slash + filename
+
+        headers = {"Content-Type": content_type}
+        if prediction_id is not None:
+            headers["X-Prediction-ID"] = prediction_id
+
         # this is a somewhat unfortunate hack, but it works
         # and is critical for upload training/quantization outputs
         # if we get multipart uploads working or a separate API route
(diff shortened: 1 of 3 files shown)


technillogue (commented):
this looks basically reasonable to me, though I'm not sure if there's anything I should watch out for. how come CI didn't pass?

aron:
> how come CI didn't pass?

Something off with the go deps, I'll rebase and try again.

aron:
@technillogue fix for CI is here https://github.com/replicate/cog/pull/1797

aron:
@technillogue would you mind reviewing this when you have a moment. I'd like to republish real-esrgan with a cut of a new alpha release that supports the new uploader tomorrow if possible.

python/cog/server/clients.py
```diff
@@ -62,7 +72,7 @@ def webhook_headers() -> "dict[str, str]":
 
 async def on_request_trace_context_hook(request: httpx.Request) -> None:
     ctx = current_trace_context() or {}
-    request.headers.update(ctx)
+    request.headers.update(cast(Mapping[str, str], ctx))
```
aron, line 75: Fixes the linting issue on `async`.

python/cog/server/clients.py
```diff
@@ -90,7 +100,7 @@ def httpx_retry_client() -> httpx.AsyncClient:
 def httpx_file_client() -> httpx.AsyncClient:
     # verify: Union[str, bool, ssl.SSLContext] = True
     transport = RetryTransport(
-        max_attempts=3,
+        max_attempts=1,
```
aron, line 103: Hmm, I think this is a merge conflict, I'll back it out.

python/cog/server/clients.py
```diff
@@ -189,6 +200,7 @@ async def upload_file(self, fh: io.IOBase, url: Optional[str]) -> str:
         url_with_trailing_slash = url if url.endswith("/") else url + "/"
 
         async def chunk_file_reader() -> AsyncIterator[bytes]:
+            fh.seek(0)
```
aron, line 203: This is needed, otherwise the retry will be unable to read the file.

python/cog/server/clients.py
```diff
@@ -188,17 +199,24 @@ async def upload_file(self, fh: io.IOBase, url: Optional[str]) -> str:
...
+            async def __aiter__(self) -> AsyncIterator[bytes]:
+                fh.seek(0)
+                while 1:
+                    chunk = fh.read(1024 * 1024)
+                    if isinstance(chunk, str):
+                        chunk = chunk.encode("utf-8")
+                    if not chunk:
+                        log.info("finished reading file")
+                        break
+                    yield chunk
```
aron, line 212: The previous implementation didn't reset the file handle on subsequent calls which resulted in an `httpx.StreamConsumed` exception. I added a test and implemented this fix which will ensure a new iterator is created for each request.

python/cog/server/clients.py
```diff
@@ -188,17 +199,24 @@ async def upload_file(self, fh: io.IOBase, url: Optional[str]) -> str:
...
-        async def chunk_file_reader() -> AsyncIterator[bytes]:
-            while 1:
-                chunk = fh.read(1024 * 1024)
-                if isinstance(chunk, str):
-                    chunk = chunk.encode("utf-8")
-                if not chunk:
-                    log.info("finished reading file")
-                    break
-                yield chunk
+        class ChunkFileReader:
```
mattt, line 202: What do you think about declaring `ChunkFileReader` as a top-level member?

aron: And passing in the file descriptor? No strong opinion, I'm happy to refactor it.

aron:
@mattt thanks! Looks great!

python/cog/server/clients.py
technillogue: FWIW I believe the function-based implementation would work with a fh.seek(0) at the top of the function, but this is great/cleaner

PR #1794, opened 2024-07-08
`None` return in `predict` function results in panic error

Shubhamai:
If `None` is returned in the `predict` function of `predict.py`, it results it a panic error, with no clear error message to indicate what went wrong.

```
def predict(self) -> str:
        return None
```

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8c3692]
```

The changes now check if none is returned, since `None` return is unintended, display an error message.

```
ⅹ Prediction output is none, return a valid output
```

### DIFF
--- a/pkg/cli/login.go
+++ b/pkg/cli/login.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"golang.org/x/term"
 
 	"github.com/replicate/cog/pkg/docker"
 	"github.com/replicate/cog/pkg/global"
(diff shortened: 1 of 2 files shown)


mattt:
Hi @Shubhamai. Thanks for sharing this!

I started to add an integration test, but noticed that the panic was specific to predictors returning `str`, but not other return types, which go through `json.Marshall`. Based on this, I came up with a narrower change in https://github.com/replicate/cog/pull/1809. The important difference there is that `cog predict` emits a warning to stderr instead of exiting with an error. This more closely matches the behavior for other output types, which aren't checked.

PR #1798, opened 2024-07-11
patch cancel bug: immediately mark predictions as cancelled

technillogue:
unconditionally mark predictions as cancelled without waiting for cancellation to succeed

while I struggle with #1786, we see queue spikes because not responding to cancellation promptly causes the pod to get restarted. this is a dirty hack to pretend like cancellation works immediately. as soon as we fix the race condition (and possibly any issues with task.cancel() behaving differently from signal handlers), we can drop this.

this is implemented by pretending as though the right event was read from the pipe

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -404,6 +404,15 @@ class PredictionRunner:
             self._events.send(Cancel(prediction_id))
             # maybe this should probably check self._semaphore._value == self._concurrent
 
+            # HACK: sometimes, Done events may be dropped due to a race condition with logging
+            # (or possibly because asyncio cancellation is less strict than signal handlers)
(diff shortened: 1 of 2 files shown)


nickstenning:
I'm not sure how I feel about this, fwiw. By doing this we tell director that work has been stopped when it may not have been, and it seems to me there's a real risk that opens the door to errors later on that may be much harder to debug.

If we know for a fact that's not the case for the specific models we're trying to patch, is there a way we can address this in the model? Or only turn it on for those models?

technillogue:
it's definitely not possible to patch it in the model, and I think it should occur for any sufficiently high-throughput model (but we only had one or two of those). we can definitely gate this behind a feature flag, though

mattt (approved):
Thanks for putting that behind the feature flag. Interested to see how this works in practice.

PR #1802, opened 2024-07-12
don't block logs during iterators by fixing the lock

technillogue:
we're currently holding the lock the whole time an iterator is running; it needs to be held separately for each send. instead of doing a bunch of indents, just factor out the locked send into a separate class

although the lock isn't actually shared between processes and the more lightweight threading.Lock could be used, that means it has to be created in run() rather than __init__

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -141,6 +141,19 @@ class Worker:
             )
 
 
+class LockedConn:
+    def __init__(self, conn: Connection) -> None:
+        self.conn = conn
+        self._lock = _spawn.Lock()
+
+    def send(self, obj: Any) -> None:
+        with self._lock:
+            self.conn.send(obj)
+
+    def recv(self) -> Any:
+        return self.conn.recv()
+
+
 class _ChildWorker(_spawn.Process):  # type: ignore
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


nickstenning (approved):

python/cog/server/worker.py
```diff
@@ -141,6 +141,19 @@ def _wait(
...
+    def __init__(self, conn: Connection) -> None:
+        self.conn = conn
+        self._lock = _spawn.Lock()
+
+    def send(self, obj: Any) -> None:
+        with self._lock:
+            self.conn.send(obj)
+
+    def recv(self) -> Any:
+        return self.conn.recv()
```
mattt, line 154: Just confirming, `recv` doesn't require a lock, right?

python/cog/server/worker.py
```diff
@@ -141,6 +141,19 @@ def _wait(
             )
 
 
+class LockedConn:
```
mattt, line 144: Any other methods we need to ensure compatibility with `multiprocessing.Lock`?

python/cog/server/worker.py
technillogue: it's only used from one place at a time and probably doesn't require a lock anyawy

python/cog/server/worker.py
mattt: Sorry, yeah — that's what I meant. Just thinking out loud in case there were any implicit requirements that we should cover.

mattt (approved):

PR #1807, opened 2024-07-16
Fix regression in `--separate-weights`

LagPixelLOL:
Title.


mattt:
Hi @LagPixelLOL. Thanks so much for contributing this PR. I'm sorry for the inconvenience this has caused.

I had a chance to dig into this a bit more, and I think the fix is a bit more involved. I opened up #1811 with passing tests. Could you please try this out and confirm that it resolves the issue for you?

mattt:
Closing in favor of https://github.com/replicate/cog/pull/1811. Thanks again for your help with this, @LagPixelLOL — I really appreciate it 🫶

PR #1813, opened 2024-07-18
Add support for async predictors

mattt:
Rebase of #1752 and #1512

PR #1823, opened 2024-07-24
httpx from scratch

technillogue:
as a prelude to #1813, we would like to cut an intermediate release that can be tested so we can discover bugs in a smaller surface area. here, we're postponing all of the async worker changes to only bring async runner and httpx, without supporting any kind of concurrency. this is slightly artificial but hopefully will reduce how many times we need to roll back #1813 

this is perhaps an opportunity to fix the webhook and files related tests as well

- **async runner (#1352)**
- **replace requests with httpx and factor out clients (#1574, #1707, #1714)**


pyproject.toml
```diff
@@ -10,11 +10,13 @@ authors = [{ name = "Replicate", email = "team@replicate.com" }]
 license.file = "LICENSE"
 urls."Source" = "https://github.com/replicate/cog"
 
-requires-python = ">=3.7"
+requires-python = ">=3.8"
 dependencies = [
   # intentionally loose. perhaps these should be vendored to not collide with user code?
   "attrs>=20.1,<24",
   "fastapi>=0.75.2,<0.99.0",
+  # we may not need http2
```
nickstenning, line 18: We almost certainly don't. What's the cost of including it vs not?

python/cog/logging.py
```diff
@@ -86,4 +86,5 @@ def setup_logging(*, log_level: int = logging.NOTSET) -> None:
 
     # Reconfigure log levels for some overly chatty libraries
     logging.getLogger("uvicorn.access").setLevel(logging.WARNING)
+    # FIXME: no more urllib3(?)
```
nickstenning, line 89: Does `httpx` log to a system logger for normal requests? If not, you can just remove this.

python/cog/predictor.py
```diff
@@ -81,51 +71,69 @@ def predict(self, **kwargs: Any) -> Any:
...
     # No weights need to be passed, so just run setup() without any arguments.
     if weights_type is None:
         predictor.setup()
         return
 
-    weights: Union[io.IOBase, Path, str, None]
-
     weights_url = os.environ.get("COG_WEIGHTS")
-    weights_path = "weights"
+    weights_path = "weights"  # this is the source of a bug isn't it?
```
nickstenning, line 100: ?

python/cog/schema.py
```diff
@@ -43,7 +44,14 @@ class PredictionBaseModel(pydantic.BaseModel, extra=pydantic.Extra.allow):
...
 
 class PredictionRequest(PredictionBaseModel):
-    id: t.Optional[str]
+    # there's a problem here where the idempotent endpoint is supposed to
+    # let you pass id in the route and omit it from the input
+    # however this fills in the default
+    # maybe it should be allowed to be optional without the factory initially
+    # and be filled in later
+    #
+    # actually, this changes the public api so we should really do this differently
```
nickstenning, line 53: If this does in fact change the public API of Cog, then this comment needs addressing before this can merge.

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def common_headers() -> "dict[str, str]":
+    headers = {"user-agent": _user_agent}
+    return headers
+
+
+def webhook_headers() -> "dict[str, str]":
+    headers = common_headers()
+    auth_token = os.environ.get("WEBHOOK_AUTH_TOKEN")
```
nickstenning, line 65: I don't think this is needed any more.

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+                follow_redirects=False,
+            )
+            if resp1.status_code == 307 and resp1.headers["Location"]:
+                log.info("got file upload redirect from api")
+                url = resp1.headers["Location"]
+
+        log.info("doing real upload to %s", url)
+        # set connect timeout to slightly more than a multiple of 3 to avoid
+        # aligning perfectly with TCP retransmission timer
+        timeout = httpx.Timeout(10.0, read=15.0)
```
nickstenning, line 242: Isn't this already configured on the `file_client`?

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+        # otherwise fallback to the URL of the final request.
+        final_url = str(resp.url)
+        if "location" in resp.headers:
+            final_url = resp.headers.get("location")
+
+        # strip any signing gubbins from the URL
+        return urlparse(final_url)._replace(query="").geturl()
+
+    # this previously lived in json.upload_files, but it's clearer here
+    # this is a great pattern that should be adopted for input files
```
nickstenning, line 262: What pattern is being referred to here?

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+        self, obj: Any, *, url: Optional[str], prediction_id: Optional[str]
+    ) -> Any:
+        """
+        Iterates through an object from make_encodeable and uploads any files.
+        When a file is encountered, it will be passed to upload_file. Any paths will be opened and converted to files.
+        """
+        # # it would be kind of cleaner to make the default file_url
+        # # instead of skipping entirely, we need to convert to datauri
+        # if url is None:
+        #     return obj
```
nickstenning, line 273: What did this commented-out code do and why is it commented out?

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+            chunk = self.fh.read(1024 * 1024)
+            if isinstance(chunk, str):
+                chunk = chunk.encode("utf-8")
+            if not chunk:
+                log.info("finished reading file")
+                break
+            yield chunk
+
+
+# there's a case for splitting this apart or inlining parts of it
```
nickstenning, line 140: I don't think HTTP clients and data structure manipulations should be smooshed together like this. We should extract these back into separate business logic and client management pieces.

The current approach makes it hard to test, hard to iterate on (e.g. `TODO: upload concurrently` would be far easier to achieve if we separated the processes of working out what we have to upload from the process of actually uploading it), and hard to understand.

python/cog/server/eventtypes.py
```diff
@@ -9,6 +9,12 @@
 class PredictionInput:
     payload: Dict[str, Any]
 
+    # @classmethod
```
nickstenning, line 12: Why is this commented out? If we no longer use it, it should go.

python/cog/server/runner.py
```diff
@@ -110,168 +116,143 @@ def predict(
...
-                "worker": self._worker,
-                "request": prediction,
-                "event_handler": event_handler,
-                "should_cancel": self._should_cancel,
-            },
-            callback=cleanup,
-            error_callback=handle_error,
-        )
-
+                raise  # we don't actually want to raise anymore but w/e
```
nickstenning, line 164: What does this mean?

nickstenning (commented):
I don't think this is really reviewable. The tests aren't passing, and I don't think the changeset is any fit state for us to merge to main in good conscience.

I appreciate that this probably isn't the review you'd like. As a reviewer, anything more than a few hundred lines is quite hard to review effectively unless a substantial portion of the changes are trivial. Reviewing a diff five times that size that's full of half-finished code (stuff commented out, comments suggesting that the code isn't ready for primetime) is just not something anyone is likely to be able to do effectively.

Let's take a step back here and rather than trying to adapt what we have on the `async` branch by cutting stuff away from it, let's take what we have learned from that branch and make a series of smaller and more incremental changes -- possibly starting afresh from current main -- to get us to where we want to be.

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+        await self.retry_webhook_client.aclose()
+        await self.file_client.aclose()
+        await self.download_client.aclose()
+
+    # webhooks
+
+    async def send_webhook(
+        self, url: str, response: Dict[str, Any], event: WebhookEvent
+    ) -> None:
+        if Status.is_terminal(response["status"]):
```
mattt, line 165: `event` is unused. Should we be consulting `event` instead of `response.status`? Or should we drop it from the method signature entirely?

pyproject.toml
technillogue: I think the only real motivation for this when it was written is using http2 for file uploads which did not go through localhost at the time. h2 is not a big library so I don't think this is terribly consequential either way

python/cog/logging.py
technillogue: looks like it https://www.python-httpx.org/logging/

python/cog/predictor.py
technillogue: https://www.notion.so/replicate/Cog-Setup-Path-Problem-2fc41d40bcaf47579ccd8b2f4c71ee24

technillogue: besides downloading the weights twice as mentioned in that notion doc, there's a lot of code that refers to a bug that I don't quite understand

https://github.com/replicate/cog-flan-models/blob/f35f162f03b2559a755c5ffbe47a5f17009ab867/predict.py#L168-L170

https://github.com/replicate/cog-mpt-7b-storywriter-65k/blob/main/predict.py#L41-L43

https://github.com/replicate/cog-llama/blob/main/predict.py#L18-L20

https://github.com/replicate/cog-sdxl/blob/main/predict.py#L184-L185

https://github.com/replicate/cog-llama-template/blob/main/predict.py#L44-L46

python/cog/server/clients.py
technillogue: it seems like it's still in main? https://github.com/replicate/cog/blob/main/python/cog/server/webhook.py#L73

python/cog/server/clients.py
technillogue: yeah, I'm not sure what @mattt's motivation was adding this in https://github.com/replicate/cog/commit/3ccc7644b52fa5331584a7be0cf2f9cf40419216#diff-a2784bee5e78b50b83f93560b26adb124c350aad2ddfe5a98699ec5f114c4f06R240-R242

mattt: Sorry, that was a bad rebase.

mattt, line 242:
```suggestion
```

python/cog/server/clients.py
```diff
@@ -0,0 +1,319 @@
...
+        log.info("doing real upload to %s", url)
+        # set connect timeout to slightly more than a multiple of 3 to avoid
+        # aligning perfectly with TCP retransmission timer
+        timeout = httpx.Timeout(10.0, read=15.0)
+        resp = await self.file_client.put(
+            url,
+            content=ChunkFileReader(fh),
+            headers=headers,
+            timeout=timeout,
+        )
```
mattt, line 248:
```suggestion
        resp = await self.file_client.put(
            url,
            content=ChunkFileReader(fh),
            headers=headers,
        )
```

python/cog/server/clients.py
technillogue: recursively iterating through the entire object rather than only the top level

python/cog/server/clients.py
technillogue: I'm not entirely convinced, this is identical to json.upload_files. I am sympathetic to extracting some helper functions for data structure manipulations. 

you could have a helper function yield each file that has to be uploaded, but then those have to go back into the data structure anyway, it doesn't seem like less complexity to me.

the previous setup had about five layers of indirection across four files

technillogue: that said if you have a narrow proposal I would be open to that

python/cog/server/runner.py
```diff
@@ -110,168 +116,143 @@ def predict(
...
+        self._upload_url = upload_url
+        self._output_type = None
 
         # HACK: don't send an initial webhook if we're trying to optimize for
         # latency (this guarantees that the first output webhook won't be
         # throttled.)
         if not SKIP_START_EVENT:
-            self._send_webhook(schema.WebhookEvent.START)
+            # idk
+            # this is pretty wrong
```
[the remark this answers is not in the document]
technillogue: very reasonable, in retrospect this was in fact not wrong

python/cog/schema.py
technillogue: turns out this isn't needed yet, it only becomes necessary once we start tracking multiple predictions

python/cog/server/clients.py
technillogue: ah, right, this was moved up to make_webhook_sender

pyproject.toml
```diff
@@ -10,11 +10,13 @@ authors = [{ name = "Replicate", email = "team@replicate.com" }]
 license.file = "LICENSE"
 urls."Source" = "https://github.com/replicate/cog"
 
-requires-python = ">=3.7"
+requires-python = ">=3.8"
```
yorickvP, line 13: does this officially drop 3.7?

technillogue: yeah

aron:
@technillogue I think this is stale now right? Shall we close it and re-open when we pick it up again?

PR #1824, opened 2024-07-25
Fix Pyright `reportUnnecessaryComparison ` warning

mattt:
https://github.com/replicate/cog/blob/f7759a08ef89a39d06c287b0a279a2a6bb7e0ce8/python/cog/predictor.py#L395

```
python/cog/predictor.py:395:8 - warning: Condition will always evaluate to False since the types "type[BaseModel]" and "None" have no overlap (reportUnnecessaryComparison)
```

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -492,8 +492,7 @@ def get_training_output_type(predictor: BasePredictor) -> Type[BaseModel]:
     train = get_train(predictor)
 
     input_types = get_type_hints(train)
-    TrainingOutputType = input_types.pop("return", None)
-    if TrainingOutputType is None:
+    if "return" not in input_types:
         raise TypeError(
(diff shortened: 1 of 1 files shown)


python/cog/predictor.py
```diff
@@ -392,7 +392,7 @@ def get_output_type(predictor: BasePredictor) -> Type[BaseModel]:
     input_types = get_type_hints(predict)
 
     OutputType = input_types.pop("return", None)
-    if OutputType is None:
+    if not OutputType:
```
nickstenning, line 395:
```suggestion
    if "return" not in input_types:
```

nickstenning (commented):
This seems fine, but rather than doing `.pop(..., None)`, it might be clearer to avoid the problem entirely and do a `"return" not in input_types` and then `OutputType = input_types["return"]` if we don't hit that branch?

PR #1825, opened 2024-07-25
Resolve or ignore Ruff and Pylint warnings

mattt:
Ignoring warnings related to `TODO` and `FIXME` comments:

```
python -m pylint python/ --disable W0511
```

This is a lot, but I think it would help us to ratchet up code quality ahead of merging async.

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -525,14 +525,14 @@ For example:
 
     if name == "Output":  # pylint: disable=no-else-return
 
-        class TrainingOutput(TrainingOutputType):
+        class TrainingOutput(TrainingOutputType):  # type: ignore
             pass
 
         return TrainingOutput
     else:
 
         class TrainingOutput(BaseModel):
-            __root__: TrainingOutputType
+            __root__: TrainingOutputType  # type: ignore
 
         return TrainingOutput
 
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -276,7 +276,7 @@ def validate_input_type(type: Type[Any], name: str) -> None:
         raise TypeError(
             f"No input type provided for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}, or a Union or List of those types."
         )
-    elif type not in ALLOWED_INPUT_TYPES:
+    if type not in ALLOWED_INPUT_TYPES:
         if get_origin(type) in (Union, List, list) or (
             hasattr(types, "UnionType") and get_origin(type) is types.UnionType
         ):  # noqa: E721
@@ -540,9 +540,11 @@ def human_readable_type_name(t: Type[Union[Any, None]]) -> str:
 
     if hasattr(t, "__module__"):
         module = t.__module__
+
         if module == "builtins":
             return t.__qualname__
-        elif module.split(".")[0] == "cog":
+
+        if module.split(".")[0] == "cog":
             module = "cog"
 
         try:
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -20,17 +20,15 @@ from typing import (
     cast,
     get_type_hints,
 )
-from unittest.mock import patch
-
-import structlog
-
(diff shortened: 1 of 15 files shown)


python/cog/predictor.py
```diff
@@ -391,8 +394,8 @@ def get_output_type(predictor: BasePredictor) -> Type[BaseModel]:
 
     input_types = get_type_hints(predict)
 
-    OutputType = input_types.pop("return", None)
-    if OutputType is None:
+    OutputType = input_types.pop("return", None)  # pylint: disable=invalid-name
+    if not OutputType:
```
mattt, line 398: Duplicate of #1824. Will resolve that there first and rebase.

python/cog/predictor.py
```diff
@@ -492,7 +495,7 @@ def get_training_output_type(predictor: BasePredictor) -> Type[BaseModel]:
     train = get_train(predictor)
 
     input_types = get_type_hints(train)
-    TrainingOutputType = input_types.pop("return", None)
+    TrainingOutputType = input_types.pop("return", None)  # pylint: disable=invalid-name
```
mattt, line 498: We should apply the same pattern as suggested in https://github.com/replicate/cog/pull/1824#discussion_r1691339203 here as well.

python/cog/server/probes.py
```diff
@@ -24,9 +24,10 @@ def __init__(self, root: PathLike = None) -> None:
             self._root.mkdir(exist_ok=True, parents=True)
         except OSError:
             log.error(
-                f"Failed to create cog runtime state directory ({self._root}). "
+                "Failed to create cog runtime state directory (%s). "
                 "Does it already exist and is a file? Does the user running cog "
-                "have permissions?"
+                "have permissions?",
+                self._root,
```
mattt, line 30: Flagging this as a nontrivial change.

python/cog/server/runner.py
```diff
@@ -11,7 +11,7 @@
 import structlog
 from attrs import define
 from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry  # type: ignore
+from urllib3.util.retry import Retry
```
mattt, line 14: TIL. This is wild.

https://github.com/psf/requests/blob/0e322af87745eff34caffe4df68456ebc20d9068/src/requests/packages.py#L6

python/cog/types.py
```diff
@@ -150,9 +149,11 @@ def __init__(self, *, source: str, filename: str, fileobj: io.IOBase) -> None:
...
     def convert(self) -> Path:
         if self._path is None:
-            dest = tempfile.NamedTemporaryFile(suffix=self.filename, delete=False)
-            shutil.copyfileobj(self.fileobj, dest)
-            self._path = Path(dest.name)
+            with tempfile.NamedTemporaryFile(
+                suffix=self.filename, delete=False
+            ) as dest:
+                shutil.copyfileobj(self.fileobj, dest)
+                self._path = Path(dest.name)
```
mattt, line 156: Double-check that use of context manager `NamedTemporaryFile` with `delete=False` does what we expect.

python/cog/types.py
```diff
@@ -263,6 +261,9 @@ def get_filename(url: str) -> str:
 
 
 class ConcatenateIterator(Iterator[Item]):
+    def __next__(self) -> Item:
+        raise NotImplementedError("This is a stub class for typing purposes only")
```
mattt, line 265: Is this right?

technillogue:
are we using both pylint and ruff now?

python/cog/types.py
mattt: On review, I don't think this is equivalent. Backing out of this change.

mattt:
> are we using both pylint and ruff now?

@technillogue Yes. And Pyright, too. Taking a belt and suspenders and chainmail approach, I suppose.

python/cog/types.py
mattt: This isn't obviously correct, and not worth the heartache. Backing out and ignoring instead.

Issue #1783, opened 2024-07-02
Allow install script to be run in Dockerfile


andreasjansson:
The following issues prevent the install one-liner, `sh <(curl -fsSL https://cog.run/install.sh)`, to be run in a RUN statement in a Dockerfile:
* `<` throws a syntax error, which is easily fixed by doing `sh -c "$(curl -fsSL https://cog.run/install.sh)"` instead
* The `read` for install directory is non-interactive and cannot be set with environment variables. This can be worked around by piping `/usr/local/bin` into the shell command, but it's brittle.
* Even when the docker builder is running as root, the install script checks for `sudo` (which doesn't exist in the build environment)

8W9aG:
There is also a check to see whether `docker` is running, I'm not sure its reasonable to expect docker to be running during a `RUN` in a `Dockerfile`. This feels like this check should be done at runtime not at install time.

PR #1830, opened 2024-07-25
Make install docker friendly

8W9aG:
* Fixes: https://github.com/replicate/cog/issues/1783
* Adds venv to gitignore (not strictly part of this issue but noticed it on my system)
* Add support for specifying the `INSTALL_DIR` via environment variable outside of the script to prevent a `read`
* Allow the user to specify `SUDO_CMD` via environment variable outside of the script incase they are operating on a docker that doesn't contain it
* Downgrades the docker run check to a warning, in docker this can't be running. I also feel this should be a runtime check rather than an install check.
* Updates the docs to add a docker install instruction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ pkg/dockerfile/embed/cog.whl
 .DS_Store
 docs/README.md
 docs/CONTRIBUTING.md
+venv
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ sudo make install
 Or if you are on docker:
 
 ```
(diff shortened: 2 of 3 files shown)


mattt (approved):
Thanks, @8W9aG! This is a nice improvement. Only change I had was renaming `SUDO_CMD` to `SUDO` to match the convention in overrides for `Makefile`. Lemme know if you think this would cause a problem for anyone.

8W9aG:
> Thanks, @8W9aG! This is a nice improvement. Only change I had was renaming `SUDO_CMD` to `SUDO` to match the convention in overrides for `Makefile`. Lemme know if you think this would cause a problem for anyone.

Not at all, great change! I think it's ready to be merged now so let me know if its looking good.

Issue #1822, opened 2024-07-24
Cog docker labels for base images have incorrect hashes


8W9aG:
* When using `--use-cog-base-images` to build a container cog outputs the follow label: `run.cog.cog-base-image-last-layer-sha`.
* This label is filled with a hash that doesn't exist in the docker container.
* In cog `0.9.12` this is correct but in `0.9.13` it appears to be incorrect.

mattt:
This sounds like a regression from https://github.com/replicate/cog/pull/1779.

8W9aG:
Test case:

[hello-world.tar.gz](https://github.com/user-attachments/files/16365538/hello-world.tar.gz)
```
cog build -t dbg .
docker inspect dbg
```

PR #1833, opened 2024-07-26
Fix cog base layer hash

8W9aG:
* Fixes https://github.com/replicate/cog/issues/1822
* Adds a test case and develops against it, this test case creates a cog build, gets the labels from the docker image, and compares them against the layer hashes.
* Issue was the previous code was fetching the manifest, then the layers from the manifest, and the digest from those layers, but this was the hash for the layer manifest rather than the layer itself.
* Change to fetch layers rather than the manifest.
* Use the `DiffID` of the layers which is the uncompressed hash, this corresponds to what is in the docker inspect for the layers.


mattt (approved):
Excellent! Thanks so much, @8W9aG.

Issue #1826, opened 2024-07-25
No concrete output on invalid cog yaml


8W9aG:
If I use a `cog.yaml` like so:

```yaml
build:
  gpu: true
  cuda: "11.8"
  python_version: "3.11"
  python_packages:
    - "torch==2.0.1"

predict: "predict.py:Predictor"
concurrency: 54
```

I receive the following message:

```
ⅹ There is a problem in your cog.yaml file. 
ⅹ must be a mapping.
ⅹ 
ⅹ To see what options you can use, take a look at the docs:
ⅹ https://github.com/replicate/cog/blob/main/docs/yaml.md
ⅹ 
ⅹ You might also need to upgrade Cog, if this option was added in a
ⅹ later version of Cog.
```

I believe that the output should say "concurrency must be a mapping" but instead this value is being erased from the errors output.

PR #1836, opened 2024-07-29
Print field in the validator error

8W9aG:
* Currently we only know what the expected type was from validation error messages.
* Make sure we print the field as well to let the user know exactly which field is having issues.
* Add a test to confirm this error message.
* Fixes: https://github.com/replicate/cog/issues/1826


mattt (approved):
Brilliant. Thank you, @8W9aG!

Issue #1827, opened 2024-07-25
Cog hangs if docker hangs


8W9aG:
Sometimes when executing a cog build the docker daemon might be hanging, this leads to an output that looks like so:

`Building Docker image from environment in cog.yaml as mybuild...`

This isn't cogs fault and actually the fault of the docker daemon, but cog should have a heartbeat that checks the docker processes output and tell the user that its not progressing due to docker and a potential remedy is to restart docker.

PR #1838, opened 2024-07-30
Ping the docker daemon to check if it is up

8W9aG:
* Send a ping to the docker daemon directly using the golang docker client.
* Give it 5 seconds to answer, if it doesn’t answer send a warning to the user.
* Solves: https://github.com/replicate/cog/issues/1827


pkg/image/build.go
```diff
@@ -31,6 +34,8 @@ const bundledSchemaPy = ".cog/schema.py"
 func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache, separateWeights bool, useCudaBaseImage string, progressOutput string, schemaFile string, dockerfileFile string, useCogBaseImage bool) error {
 	console.Infof("Building Docker image from environment in cog.yaml as %s...", imageName)
 
+	ping()
```
mattt, line 37: What do you think about putting this in `cli` instead of `pkg`? If this is intended for interactive consumption by a user, I think that could be a better fit.

pkg/image/build.go
```diff
@@ -360,3 +365,22 @@ func restoreDockerignore() error {
...
+		return
+	}
+
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+
+	_, err = cli.Ping(ctxTimeout)
+
+	if err != nil {
+		fmt.Println("Failed to ping docker, please try restarting the docker daemon.")
```
mattt, line 382: Minor nitpick, but I'd prefer to have this method return an error and have the call site responsible for communicating with the user.

mattt (approved):
Thanks, @8W9aG! I reworked some of the code, moving the `Ping` method to the `docker` package and making the CLI responsible for communicating with the end user. I also copied that check to the other CLI subcommands that invoked Docker.

If this looks good to you, feel free to merge 😄

PR #1839, opened 2024-07-31
Extract release process into separate workflow

mattt:
The release job of the CI run for the latest release [failed](https://github.com/replicate/cog/actions/runs/10180772831/attempts/1) when building and pushing the Python package to PyPI. For some reason [^1], the package got built as `0.9.15.dev0+g592f768.d20240731` instead of `0.9.14`, and PyPI rejected that value.

[^1]: Any theories?

Our current CI/CD makes it impossible to re-run the failed release step without deleting or invalidating the GitHub release artifacts built by goreleaser. For sake of integrity, re-tagging or modifying artifacts after the fact is not an option.

This PR extracts the release step into its own workflow. It takes advantage of the [`workflow_run`](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run) event to trigger after CI finishes. This has the same effect as a dependent job in a single workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -7,7 +7,6 @@ on:
       - completed
     branches:
       - main
-  workflow_dispatch:
 
 concurrency:
   group: release-${{ github.ref }}
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,71 @@
+name: Release
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+    branches:
+      - main
+  workflow_dispatch:
+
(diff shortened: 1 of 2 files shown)


8W9aG (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,71 @@
+name: Release
+
+on:
+  workflow_run:
+    workflows: ["CI"]
+    types:
+      - completed
+    branches:
+      - main
+  workflow_dispatch:
```
mattt, line 10: I'm not sure `workflow_dispatch` actually makes sense for this.

```suggestion
```

PR #1841, opened 2024-07-31
Upload aarch64 cog binary release

8W9aG:
* aarch64 is actually exactly the same as arm64
* Unfortunately because of uname sometimes the architecture is reported as aarch64
* This how most people download our binaries, so if we don’t provide this path we end up not
supporting this platform as well as we should.

As an aside, I'm a bit unsure how to test this because it lives in the release deployment.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -118,6 +118,13 @@ jobs:
           args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Copy arm64 to aarch64
+        run: mkdir -p dist/cog_linux_aarch64 && cp dist/cog_linux_arm64/cog dist/cog_linux_aarch64/cog
+      - name: Upload aarch64 cog binary to artifacts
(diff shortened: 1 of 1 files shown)


mattt:
@8W9aG Nice! We're several versions behind on goreleaser, using v4 instead of v6 (#1766) and using v0 configuration instead of v2. Is this something we could fix instead by upgrading to newer version of goreleaser software or configuration?

8W9aG:
> @8W9aG Nice! We're several versions behind on goreleaser, using v4 instead of v6 (#1766) and using v0 configuration instead of v2. Is this something we could fix instead by upgrading to newer version of goreleaser software or configuration?

Certainly I can give that a go, would you happen to have any ideas on how to test it though? Perhaps its just a more involved release next time where we watch it in realtime to make sure it executes as expected.

mattt:
@8W9aG goreleaser should be runnable locally, so you could tweak the settings until it gives you what you need. Once you have a configuration that works for you, we can test out the integration with a new 0.9.x patch release (we need to do this anyway to get a new Python library package published)

PR #1842, opened 2024-07-31
Use uvloop for async event loop

mattt:
From https://github.com/MagicStack/uvloop:

> uvloop is a fast, drop-in replacement of the built-in asyncio event loop. uvloop is implemented in Cython and uses libuv under the hood.

Hard to argue with 2x speedup for a drop-in replacement. Curious to see whether that holds up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
   'typing-compat; python_version < "3.8"',
   "typing_extensions>=4.4.0",
   "uvicorn[standard]>=0.12,<1",
+  "uvloop",
 ]
 
 optional-dependencies = { "dev" = [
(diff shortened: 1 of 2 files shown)


mattt:
There's a good chance we're already using it implicitly. I just wanted to test this by making it explicit.

yorickvP:
How well does this combine with customers of our library who use asyncio? Should there maybe be an escape hatch?

mattt:
@technillogue pointed out that uvloop was already added to the async branch via httpx. Uvicorn advertises that the default `"auto"` loop setting selects uvloop when available, and I don't see any evidence to the contrary.

@yorickvP That's a good callout, but I don't anticipate a problem. The vast majority of consumers will run `cog.http` as a self-contained program, and anyone consuming as a library could pretty easily swap out their preferred event loop as desired.

yorickvP:
The reason I ask is that, with async cog, we saw people trying to combine it with their own event loops in their own threads, and it took a bit of hacking to make that work.

PR #1843, opened 2024-08-01
call terminate after the last prediction or after a timeout if shutting down

technillogue:
this addresses the problem described in https://www.notion.so/replicate/Short-Term-Async-Fixes-9d11cfc15bb84878aedd9d8ddf31840d?pvs=4#8a86d46ada954ff5b509231224c24f2f

prior to https://github.com/replicate/cog/commit/974675f9#diff-136cc6506e8aaea24081862f89e1cb778251b7726180fe237c8aaf560ae7ec69R315-R338, runner.shutdown() would call worker.terminate(). as far as I understand, neither in that branch nor in main is worker.shutdown() ever called. in that commit, I split runner.shutdown into runner.shutdown and runner.terminate, but never made anything call terminate. because I also removed the `should_exit = True` line, the uvicorn server would never exit, so the shutdown handler would eventually notice this and kill the whole process, so we never noticed that terminate was never called.

runner.terminate does:
* cancel any predict tasks
* mark us as defunct
* if the child is alive, terminate it and then waitpid
* close the events pipe
* cancel the read events task

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -369,13 +369,6 @@ class PredictionRunner:
                 # since this is just before this coroutine exits
                 self._predictions.pop(request.id)
 
-                # all predictions in flight have been processed, we can exit
-                if self._shutting_down and not self._predictions:
-                    # note: this will do a bunch of work
-                    # and potentially slow down the coroutine completing
-                    # however, this should be okay
-                    self.terminate()
-
         # this is still a little silly
         result = asyncio.create_task(async_predict_handling_errors())
         # result.add_done_callback(self.make_error_handler("prediction"))
@@ -384,7 +377,8 @@ class PredictionRunner:
 
         return (response, result)
 
-    def shutdown(self) -> None:
+    async def shutdown(self) -> None:
+        # this is called by the app's shutdown handler. server won't exit until this is done
         self.log.info("runner.shutdown called")
         if self._state == WorkerState.DEFUNCT:
             return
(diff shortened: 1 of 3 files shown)


mattt (approved):

python/cog/server/runner.py
```diff
@@ -198,8 +198,12 @@ async def inner() -> SetupResult:
             except Exception:
                 logs.append(traceback.format_exc())
                 status = schema.Status.FAILED
+            except asyncio.CancelledError:
+                self.log.info("caught CancelledError during setup")
+                logs.append(traceback.format_exc())
+                status = schema.Status.FAILED
             except BaseException:
```
nickstenning, line 205: Um, why are we catching `BaseException` and not re-raising it? That's generally a _very_ bad idea.

python/cog/server/runner.py
```diff
@@ -198,8 +198,12 @@ async def inner() -> SetupResult:
             except Exception:
                 logs.append(traceback.format_exc())
                 status = schema.Status.FAILED
+            except asyncio.CancelledError:
```
nickstenning, line 201: Is this just for the log line?

python/cog/server/runner.py
```diff
@@ -237,7 +241,9 @@ def handle_error(task: RunnerTask) -> None:
                 if self._shutdown_event is not None:
                     self._shutdown_event.set()
             except BaseException:
```
nickstenning, line 243: Again, I don't understand why we're catching `BaseException` and not re-raising it here. This will lead to some very confusing bugs.

python/cog/server/runner.py
```diff
@@ -363,6 +369,13 @@ async def async_predict_handling_errors() -> schema.PredictionResponse:
                 # since this is just before this coroutine exits
                 self._predictions.pop(request.id)
 
+                # all predictions in flight have been processed, we can exit
+                if self._shutting_down and not self._predictions:
+                    # note: this will do a bunch of work
+                    # and potentially slow down the coroutine completing
+                    # however, this should be okay
+                    self.terminate()
```
nickstenning, line 377: I don't think I understand this. In `main` the workflow is:

- any thread can request a shutdown by setting the `shutdown_event`
- when that happens, `Server.stop` is called, which starts a shutdown of fastapi
- when fastapi shuts down it triggers a `shutdown` event, which calls `runner.shutdown()`
- the runner shutdown terminates the worker and stops the runner thread
- if this sequence fails to complete within 10 seconds, server shoots itself in the head by doing `os.kill(os.getpid(), signal.SIGKILL)`

If I understand this correctly, this is happening in a per-prediction coroutine. It doesn't make any sense to me at all to call `self.terminate` when `self._shutting_down` is already set.

I think instead what's missing here is a way for the `shutdown` method itself to wait for all its child coroutines to complete.

python/cog/server/runner.py
```diff
@@ -381,16 +394,25 @@ def shutdown(self) -> None:
         if self._child.is_alive():
             self.log.info("child is alive during shutdown, sending Shutdown event")
             self._events.send(Shutdown())
+            asyncio.create_task(self.terminate_later())
```
nickstenning, line 397: This doesn't make sense to me. Generally in this codebase `shutdown` has meant "graceful shutdown" and `terminate` has meant "stop what you're doing right now". It doesn't really make sense for the former to call the latter.

It's also more than a little hairy for this kind of stuff to be coordinated in `Worker` rather than at the top level of the program as it is in `main` (see my comment above).

python/cog/server/runner.py
```diff
@@ -381,16 +394,25 @@ def shutdown(self) -> None:
...
+
+    async def terminate_later(self) -> None:
+        self.log.info("scheduling termination in 10s")
+        await asyncio.sleep(10)
+        self.log.info("10s has passed, terminating")
+        self.terminate()
 
     def terminate(self) -> None:
```
nickstenning, line 405: Does runner even need a `terminate` method? What is the scenario in which we need it? As far as I can tell we need it to have a single `shutdown` method which waits until things have been correctly shut down. If it fails to do that (within a timeout that should be coordinated at the top of the program), we SIGKILL ourselves?

python/cog/server/runner.py
technillogue: it had appeared that shutdown during setup was failing, in part because this would get cancelled and there would be no setup result. I've removed this

python/cog/server/runner.py
technillogue: re-raised

python/cog/server/runner.py
technillogue: that makes sense

PR #1844, opened 2024-08-02
Explain `NAME_UNKNOWN` error on failed cog push

mattt:
Sometimes, the `cog push` command returns the following error:

```
unknown: {"errors":[{"code":"NAME_UNKNOWN","message":"The model https://replicate.com/{owner}/{name} does not exist"}]}
ⅹ exit status 1
```

Often, this is because the user ran either `cog login` or `cog push` as `sudo`, which causes Docker to use the wrong credentials.

This PR updates the `cog push` command to explain this error.

### DIFF
--- a/pkg/cli/push.go
+++ b/pkg/cli/push.go
@@ -73,13 +73,27 @@ func push(cmd *cobra.Command, args []string) error {
 
 	console.Infof("\nPushing image '%s'...", imageName)
 
-	exitStatus := docker.Push(imageName)
-	if exitStatus == nil {
-		console.Infof("Image '%s' pushed", imageName)
-		if strings.HasPrefix(imageName, replicatePrefix) {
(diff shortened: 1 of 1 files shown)


yorickvP:
Could this also happen if the user never did `cog login` in the first place?

mattt:
@yorickvP No, that would result in an `UNAUTHORIZED` error (which would also benefit from a similar treatment, come to think of it)

yorickvP:
Wait, I don't understand. So running `cog login` as the wrong account results in a configuration that gives `NAME_UNKNOWN`, but not running `cog login` at all results in a configuration that gives `UNAUTHORIZED`.  What's the difference in configuration?

nickstenning:
FWIW it occurs to me that we can probably do better than this and know whether or not you have any valid credentials in the first place, before we start a push.

I wonder if Docker exposes an interface that would allow us to check the credentials before trying to use them? Even to check they exist at all?

aron (approved):

PR #1854, opened 2024-08-06
Pydantic v2 via pydantic.v1 compat module

yorickvP:
Tests pass locally, let's see what the CI says.


yorickvP:
We'd have to patch fastapi._compat to make this work.

PR #1687, opened 2024-05-23
Support Pydantic 2

mattt:
Alternative to #1625

PR #1858, opened 2024-08-06
Add support for Pydantic 2

mattt:
Rebase of #1687, applied to `main` instead of `async`.


python/cog/command/ast_openapi_schema.py
```diff
@@ -334,17 +334,24 @@ def to_serializable(val: "AstVal") -> "JSONObject":
...
 def get_value(node: ast.AST) -> "AstVal":
     """Return the value of constant or list of constants"""
-    if isinstance(node, ast.Constant):
-        return node.value
-    # for python3.7, were deprecated for Constant
-    if isinstance(node, (ast.Str, ast.Bytes)):
-        return node.s
-    if isinstance(node, ast.Num):
-        return node.n
+    # Deprecated since version 3.8:
```
8W9aG, line 337: In this case, do we need to check for constants regardless of python version and then drop down into the version check and check the str/bytes/num?

python/cog/command/ast_openapi_schema.py
```diff
@@ -519,7 +526,20 @@ def extract_info(code: str) -> "JSONDict":  # pylint: disable=too-many-branches,
...
                 kws[kw.arg] = to_serializable(get_value(kw.value))
-        elif isinstance(default, (ast.Constant, ast.List, ast.Tuple, ast.Str, ast.Num)):
+        elif isinstance(
+            default,
+            # Deprecated since version 3.8:
+            # Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis
+            # are still available, but they will be removed in Python 3.14.
+            ast.Constant
+            if sys.version_info > (3, 7)
+            else (
```
8W9aG, line 536: Same thing here

python/cog/predictor.py
```diff
@@ -99,18 +101,34 @@ def run_setup(predictor: BasePredictor) -> None:
...
-            weights = weights_url
+        if PYDANTIC_V2:
+            from pydantic import TypeAdapter
+
+            for t in [CogFile, CogPath]:
+                try:
+                    weights = TypeAdapter(t).validate_python(weights_url)
+                    break
+                except Exception:  # pylint: disable=broad-except # noqa: S110
+                    pass
```
8W9aG, line 112: If none of these work do we need to raise an exception?

python/cog/server/http.py
```diff
@@ -113,6 +112,98 @@ def create_app(  # pylint: disable=too-many-arguments,too-many-locals,too-many-s
...
+        def flatten_selected_allof_refs(
+            openapi_schema: Dict[str, Any],
+        ) -> None:
+            try:
+                response = openapi_schema["components"]["schemas"]["PredictionResponse"]
+                response["properties"]["output"] = {
+                    "$ref": "#/components/schemas/Output"
+                }
+            except KeyError:
+                pass
```
8W9aG, line 151: Would this be expected to happen if their was no prediction response? And if so is that valid according to our schema?

python/cog/server/http.py
```diff
@@ -113,6 +112,98 @@ def create_app(  # pylint: disable=too-many-arguments,too-many-locals,too-many-s
...
+                    del value["allOf"]
+
+            try:
+                path = openapi_schema["paths"]["/predictions"]["post"]
+                body = path["requestBody"]
+                body["content"]["application/json"]["schema"] = {
+                    "$ref": "#/components/schemas/PredictionRequest"
+                }
+            except KeyError:
+                pass
```
8W9aG, line 171: Same here

python/cog/types.py
```diff
@@ -6,12 +6,30 @@
...
 import requests
-from pydantic import Field, SecretStr
 from typing_extensions import NotRequired, TypedDict
 
+if pydantic.__version__.startswith("1."):
+    PYDANTIC_V2 = False
+else:
+    PYDANTIC_V2 = True
+    if TYPE_CHECKING:
+        pass  # type: ignore
```
8W9aG, line 30: What does this do?

8W9aG (approved):
Just some minor comments for my own knowledge but otherwise looks good!

python/cog/types.py
mattt: `if TYPE_CHECKING`: pass` is a fingerprint left by Ruff's formatter, which automatically removes unused imports. That used to have some import, but it got removed. But instead of removing it, it got replaced with a `pass`. But we should kill these on sight.

python/cog/predictor.py
mattt: That should be handled by the `else` clause below

python/cog/command/ast_openapi_schema.py
mattt: Sorry, this commit shouldn't have been applied

python/cog/server/http.py
mattt: It shouldn't and wouldn't, no. In practice, I don't think this exception can really happen. But if it _were_ missing, raising a stink here seems unhelpful. The alternative would be to do safe dict navigation with conditional `.get` chaining, but that felt cumbersome.

python/cog/json.py
```diff
@@ -26,7 +26,10 @@ def make_encodeable(obj: Any) -> Any:  # pylint: disable=too-many-return-stateme
     """
 
     if isinstance(obj, BaseModel):
-        return make_encodeable(obj.dict(exclude_unset=True))
+        if PYDANTIC_V2:
+            return make_encodeable(obj.model_dump(exclude_unset=True))
```
nickstenning, line 30: What's the deal with this rather than `obj.dict` or `obj.json`?

yorickvP: pydantic2 deprecates `obj.dict`

mattt:
Before merging, we should update CI to test against both Pydantic v1 and v2

python/cog/predictor.py
```diff
@@ -283,12 +308,20 @@ def validate_input_type(
             f"No input type provided for parameter `{name}`. Supported input types are: {readable_types_list(ALLOWED_INPUT_TYPES)}, or a Union or List of those types."
         )
     if type not in ALLOWED_INPUT_TYPES:
-        if get_origin(type) in (Union, List, list) or (
+        if get_origin(type) is Literal:
```
yorickvP, line 311: we should check if the AST parser can deal with Literals

mattt: Added by 095fd997cfdeb2ccf94bd63213b254dfad3cee9d

yorickvP:
Submitted a PR to pydantic-core to expose `.iterator` on SerializationIterator: https://github.com/pydantic/pydantic-core/pull/1399

python/cog/predictor.py
```diff
@@ -256,16 +276,21 @@ def load_predictor_from_ref(ref: str) -> BasePredictor:
...
+            # But, after validation, we want to pass the actual value to predict(), not the enum object
+            use_enum_values = True
 
     def cleanup(self) -> None:
         """
         Cleanup any temporary files created by the input.
         """
-        for _, value in self:
+
+        for _, value in self.__dict__.items():
```
nickstenning, line 293:
```suggestion
        for _, value in dict(self).items():
```
Perhaps makes it clearer that we want to iterate over the public fields of the model?

python/cog/server/http.py
```diff
@@ -383,12 +504,18 @@ def _predict(
 
         # ...and return the result.
         try:
-            response = PredictionResponse(**predict_task.result.dict())
+            response_object = (
+                unwrap_pydantic_serialization_iterators(
```
nickstenning, line 508: Sorry if I'm being slow here, but have you written anywhere about why this unwrapping is needed? I would kind of expect that `model_dump` should do the right thing?

mattt: Alright, just pushed a commit documenting that method and updated the PR with details.

nickstenning (commented):
Looking good to me!

I would suggest that given the size and scope of the change, we test it with some things in Replicate production _before_ merging the PR. I'll drop a note in our internal Slack about how to do that.

yorickvP:
Would it be worth it to additionally typecheck with pydantic v1?

mattt:
@yorickvP Yes, indeed! If you have time before next week, it'd be great to add that to the tox matrix.

yorickvP:
@mattt I might have time later, but don't wait for me! We can merge that later.

pyproject.toml
```diff
@@ -76,6 +71,9 @@ reportUnneesssaryContains = "warning"
 reportMissingTypeArgument = "error"
 reportUnusedExpression = "warning"
 
+[tool.pyright.defineConstant]
+PYDANTIC_V2 = true
```
yorickvP, line 75: is this unset somewhere for pydantic-1?

mattt: Gosh, that's news to me. Thanks for pointing that one out.

yorickvP:
Did some very basic testing.

- Observation 1: cog-base is still pydantic1, so it will use pydantic1 by default. `cog build --use-cog-base-image=false` uses pydantic2 by default.
- Observation 1.5: `cog push` doesn't allow you to configure `--use-cog-base-image`. Adding `pydantic>2` to the `python_packages` still installs pydantic2.

- Observation 2: fastapi deprecation errors
<details>

```
/usr/local/lib/python3.11/site-packages/cog/server/http.py:258: DeprecationWarning:
        on_event is deprecated, use lifespan event handlers instead.

        Read more about it in the
        [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).

  @app.on_event("startup")
/usr/local/lib/python3.11/site-packages/cog/server/http.py:272: DeprecationWarning:
        on_event is deprecated, use lifespan event handlers instead.

        Read more about it in the
        [FastAPI docs for Lifespan Events](https://fastapi.tiangolo.com/advanced/events/).

  @app.on_event("shutdown")

```

</details>

- Observation 3: it works!
  - Test with pydantic1: https://replicate.com/p/dffewqcem1rga0cj1d4ayehe2r
  - Test with pydantic2: https://replicate.com/p/nfd6tc1qk5rga0cj1d6vgafxp8

yorickvP:
my test with `python_dependencies: - pydantic>2` produced a pip install with `fastapi==0.98` and `pydantic==2.9.2`, which wouldn't work. Not sure what to do about that, except re-specifying fastapi constraints.

yorickvP:
We will need to rethink how `cog build` installs dependencies.

- Currently, there is a `pip install cog` step. This installs cog in the image, and picks pydantic2 and `fastapi==0.110.3`.
- (base image is built here)
- Then, user dependencies are installed. If this includes `pydantic < 2`, `pip` _won't actually downgrade pydantic_. It would have to run `pip uninstall pydantic fastapi` first.


## Possible solutions:

1. Depend on `pydantic1` in the `pip install cog` step.
   - This upgrades to pydantic2 as needed during user package installation.
   - Pros: Can work for users using pydantic 1 and 2.
   - Con: This defaults to pydantic 1, while we want to switch to 2.
   - Con: This won't upgrade fastapi when pydantic2 is upgraded by pip. We'd need to add another build step to upgrade fastapi when the pydantic version changes.
 
2. Make the `cog` wheel participate in the user's dependency resolution.
   - This means we add /tmp/cog-something.whl to the requirements.txt and run the package installation as a single step.
   - Pro: dependency sets are always valid. Users get clear errors when there's a conflicting dependency.
   - Con: Stops us from including cog in the base images
   - Neutral: Stops users from building packages where dependencies conflict with cog. Not sure if that's desired.
 
3. Add a `pydantic:` specifier to `cog.yaml`
    - Pro: makes it easy to have multiple base images
    - Con: not obvious for users when this is needed
    - Con: not backwards compatible
 
4. Infer `pydantic` version from dependencies.
    - We already do this with `torch`.
    - Pro: makes it easy to have multiple base images
    - Con: not backwards compatible
    - Con: pydantic dependencies are often transitive, so this will require extra work for users
 
5. Only allow building new images with pydantic2
    - Keep pydantic1 compat for existing images, but remove the functionality from `cog` go.
    - Pro: easiest to do (current behavior in this PR)
    - Con: Confusing to users (pydantic1 dependencies will break at runtime)
    - Con: not backwards compatible

6. Switch to something smarter than `pip`
   - Out of scope for this PR

yorickvP:
I went with option 1, which is the easiest for now. We can defer making pydantic2 a default to a later PR.

technillogue:
IMO that's the correct choice at least for the moment: keep things as they are now, but make it so that vllm doesn't break

PR #1859, opened 2024-08-06
Check release tag

8W9aG:
Check that the release tag is a version tag (v[0-9].[0-9]+.[0-9]+) and that the workflow is triggered by a new tag push.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,8 +5,7 @@ on:
     workflows: ["CI"]
     types:
       - completed
-    branches:
-      - main
+    branches: ["v*.*.*"]
 
 concurrency:
   group: release-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


mattt:
@8W9aG I think we need to update this line:

https://github.com/replicate/cog/blob/4efdcf386ec3f83c073d43ff9fcf316172b33773/.github/workflows/release.yaml#L9

If you look at past runs, the Release workflow only ran for merge commits to `main`. The tags triggered CI as expected, but not Release.

8W9aG:
> @8W9aG I think we need to update this line:
> 
> https://github.com/replicate/cog/blob/4efdcf386ec3f83c073d43ff9fcf316172b33773/.github/workflows/release.yaml#L9
> 
> If you look at past runs, the Release workflow only ran for merge commits to `main`. The tags triggered CI as expected, but not Release.

Done, let me know how its looking!

.github/workflows/release.yaml
```diff
@@ -23,8 +22,7 @@ jobs:
       - name: Check if commit has a version tag
         id: check-tag
         run: |
-          LATEST_TAG=$(git tag | sort --version-sort | tail -n1)
-          if [[ ${{ github.ref_name == $LATEST_TAG }} -a $LATEST_TAG =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+          if [[ ${{ github.event.workflow_run.head_branch }} =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
mattt: `github.event.workflow_run.head_branch` 🤨 

Weird, seems to be right.

mattt (approved):
Thanks for chasing this down, @8W9aG 🙏

PR #1865, opened 2024-08-12
Simplify `PredictionRunner`

nickstenning:
This simplifies `PredictionRunner`, removing its worker thread and making it a wrapper around the worker which also manages the lifecycle of `PredictTask` and `SetupTask` objects.

`PredictTask` is essentially `PredictionEventHandler` but with the additional responsibilities of a) tracking the future representing the predict task running in the worker, and b) handling the individual event objects returned by worker.

`SetupTask` is the equivalent but for a setup.

We expect that this code will change significantly in the near future to allow the runner to keep track of multiple predictions at once.

This also changes the behaviour of cog in the event of a fatal prediction exception: if `await_explicit_shutdown` is set, then rather than triggering instant shutdown, we will poison the healthcheck with a new DEFUNCT health status.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -13,7 +13,6 @@ from datetime import datetime, timezone
 from enum import Enum, auto, unique
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Optional
 
-import attrs
 import structlog
 import uvicorn
 from fastapi import Body, FastAPI, Header, HTTPException, Path, Response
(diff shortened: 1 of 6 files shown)


erbridge (approved):

python/cog/server/http.py
```diff
@@ -278,11 +278,6 @@ async def predict(
         """
         Run a single prediction on the model
         """
-        if runner.is_busy():
-            return JSONResponse(
-                {"detail": "Already running a prediction"}, status_code=409
-            )
```
nickstenning, line 284: This has been removed because we think it's not necessary -- we already catch the `RunnerBusyError` when attempting to call `predict` inside the `_predict` function.

PR #1866, opened 2024-08-12
Migrate to FastAPI Lifespan Events

mattt:
From https://fastapi.tiangolo.com/advanced/events/:

> You can define logic (code) that should be executed before the application starts up. This means that this code will be executed once, before the application starts receiving requests.

> [...]

>Let's start with an example use case and then see how to solve it with this.
>
> Let's imagine that you have some machine learning models that you want to use to handle requests. 🤖
>
> The same models are shared among requests, so, it's not one model per request, or one per user or something similar.
>
> Let's imagine that loading the model can take quite some time, because it has to read a lot of data from disk. So you don't want to do it for every request.

🎯

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -9,6 +9,7 @@ import sys
 import textwrap
 import threading
 import traceback
+from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from enum import Enum, auto, unique
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Optional
@@ -66,20 +67,8 @@ class Health(Enum):
     DEFUNCT = auto()
 
 
-class MyState:
-    health: Health
-    setup_result: Optional[SetupResult]
-
-
-class MyFastAPI(FastAPI):
-    # TODO: not, strictly speaking, legal
-    # https://github.com/microsoft/pyright/issues/5933
-    # but it'd need a FastAPI patch to fix
-    state: MyState  # type: ignore
-
-
 def add_setup_failed_routes(
-    app: MyFastAPI,  # pylint: disable=redefined-outer-name
+    app: FastAPI,  # pylint: disable=redefined-outer-name
     started_at: datetime,
     msg: str,
 ) -> None:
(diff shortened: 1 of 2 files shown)


8W9aG (approved):

python/cog/server/http.py
```diff
@@ -66,20 +67,8 @@ class Health(Enum):
     DEFUNCT = auto()
 
 
-class MyState:
-    health: Health
-    setup_result: Optional[SetupResult]
```
erbridge, line 71: Is this not still needed?

python/cog/server/http.py
```diff
@@ -104,20 +93,37 @@ async def healthcheck_startup_failed() -> Any:
...
     config: CogConfig,  # pylint: disable=redefined-outer-name
     shutdown_event: Optional[threading.Event],  # pylint: disable=redefined-outer-name
     threads: int = 1,  # pylint: disable=redefined-outer-name
     upload_url: Optional[str] = None,
     mode: str = "predict",
     is_build: bool = False,
     await_explicit_shutdown: bool = False,  # pylint: disable=redefined-outer-name
-) -> MyFastAPI:
-    app = MyFastAPI(  # pylint: disable=redefined-outer-name
-        title="Cog",  # TODO: mention model name?
```
nickstenning, line 117: Where was this title used? The generated OpenAPI?

mattt: Yes, I think so. It's now down here: https://github.com/replicate/cog/pull/1866/files#diff-424c9aeb9bb84b2114a86a5fdf84f4e2cbecf9797b03e6866caa03bd07e53b7cR124

python/cog/server/http.py
mattt: `FastAPI` instances have a `state: State` property. I haven't found a way to make that typed, so it's instead an opaque object with getters and setters. To the point of the comment, the way we had it before broke type invariance and shouldn't have worked (by some definition of should).

I'm relieved that your refactoring of runner / worker reduced how much state we were tracking. For just a pair of properties, it's not too bad.

erbridge: My editor typechecker was understanding these types fine, though? Are gaining something by removing it other than a bit of neatness?

Issue #1862, opened 2024-08-10
Unsupported base image configuration with torch==1.13.0


sonicbaume:
I'm struggling to get cog to work with the versions I'm interested in:

**cog.yaml:**
```yaml
build:
  gpu: true
  python_version: "3.9"
  python_packages:
    - "torch==1.13.0"
predict: "predict.py:Predictor"
```

```
cog build   
Building Docker image from environment in cog.yaml as cog-enclap-replicate...
⚠ Stripping patch version from Torch version 3.9 to 3.9
ⅹ Failed to get cog base image name: unsupported base image configuration: CUDA: 11.7 / Python: 3.9 / Torch: 1.13
```

This seems to suggest the versions are compatible - is that right?
https://github.com/replicate/cog/blob/d160cc29de9e6eea4932be3a8e3276819feff875/pkg/config/torch_compatibility_matrix.json#L497-L541

Thanks for your help 😊

8W9aG:
It looks like we have the following issues:
1. We do not support anything below Torch 1.13.1 by a hard limit defined in the code: https://github.com/replicate/cog/blob/2a5d7769a5f9e488c5c769c92312e3bee3db6d7b/pkg/dockerfile/base.go#L15
2. When this limit is removed and the compatibility matrix is tested against, it produces an image that does not exist: r8.im/cog-base:cuda11.7-python3.9-torch1.13
3. If this failure happens it cancels the whole build. What we should do here is fallback if the base image option is not explicitly sent.

Will get working on this now.

8W9aG:
I'll close this, it should be fixed in the next release with the fix that was merged moments ago

PR #1872, opened 2024-08-12
Fallback to regular if cog base image can't be determined and flag is not explicitly set

8W9aG:
* Add base-image to gitignore
* Use a boolean pointer for cog base image flag for a tri-state variable
* Handle not explicitly setting the cog base image and a failure to resolve the base image as a warning
* Add integration test for torch 1.13.0

* Solves problem 3 in: https://github.com/replicate/cog/issues/1862

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -21,7 +21,9 @@ var buildProgressOutput string
 var buildSchemaFile string
 var buildUseCudaBaseImage string
 var buildDockerfileFile string
-var buildUseCogBaseImage bool
+var buildUseCogBaseImage *bool = nil
+
+const useCogBaseImageFlagKey = "use-cog-base-image"
 
 func newBuildCommand() *cobra.Command {
 	cmd := &cobra.Command{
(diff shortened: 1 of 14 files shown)


pkg/cli/build.go
```diff
@@ -116,7 +112,11 @@ func addDockerfileFlag(cmd *cobra.Command) {
 }
 
 func addUseCogBaseImageFlag(cmd *cobra.Command) {
-	cmd.Flags().BoolVar(&buildUseCogBaseImage, "use-cog-base-image", true, "Use pre-built Cog base image for faster cold boots")
+	buildUseCogBaseImage = new(bool)
+	cmd.Flags().BoolVar(buildUseCogBaseImage, useCogBaseImageFlagKey, true, "Use pre-built Cog base image for faster cold boots")
+	if !cmd.Flags().Changed(useCogBaseImageFlagKey) {
+		buildUseCogBaseImage = nil
+	}
```
mattt, line 119: Wait, does this work? My mental model of `cmd.Flags().Changed()` was that it evaluated in `RunE` when args were passed in, like a variable initialized and then set later.

mattt (approved):

pkg/cli/build.go
8W9aG: You are completely correct, let me find a different way to express this.

8W9aG: I've got a fix and added some more tests, let me know if this is sufficient.

test-integration/test_integration/test_build.py
```diff
@@ -240,3 +240,49 @@ def test_build_base_image_sha(docker_image):
...
+            "-t",
+            docker_image,
+            "--openapi-schema",
+            "openapi.json",
+            "--use-cog-base-image=false",
+        ],
+        cwd=project_dir,
+        capture_output=True,
+    )
+    assert build_process.returncode == 0
```
mattt, line 288: Lovely to have all of these. Thank you.

mattt (approved):

PR #1873, opened 2024-08-13
Better build and release

nickstenning:
This PR contains several changes which attempt to improve how we build and release Cog.

1. When building `cog`'s Go binary, we embed the wheel with the full path generated by [`build`](https://build.pypa.io/en/stable/). This makes it clearer what version of cog's Python package is being injected by any given `cog` binary at runtime.
2. Omit test files from the built wheel.
3. Run tests against the `cog` package installed from the built wheel, not against an editable installation from the local directory.
4. Use `goreleaser` to build `cog` so we have only one way of building the `cog` binary, not two.
5. Make Python and Go "dev" versions aligned -- if the most recent tag is `v1.2.3`, a build on an untagged commit should now be `1.2.4-dev+g1234abcd` by default. The Python version will differ slightly as it is a [PEP440](https://peps.python.org/pep-0440/) version, not a semver version, but the generated versions should now be largely comparable.


nickstenning:
There's more I want to do here, specifically to replace the `release.yaml` workflow, but this is more than enough for one PR so I'm stopping here.

.github/workflows/ci.yaml
```diff
@@ -27,17 +41,20 @@ jobs:
...
         with:
           go-version-file: go.mod
       - uses: actions/setup-python@v5
         with:
           python-version: 3.12
       - name: Install Python dependencies
-        run: |
-          python -m pip install --no-cache uv
-          python -m uv pip install '.'
+        run: python -m pip install --no-cache uv setuptools_scm
```
technillogue, line 57:
```suggestion
        run: python -m pip install --no-cache uv && python -m uv pip install setuptools_scm
```

maybe like this? not a huge difference though

Makefile
```diff
@@ -16,56 +15,57 @@ PYTEST := $(PYTHON) -m pytest
 PYRIGHT := $(PYTHON) -m pyright
 RUFF := $(PYTHON) -m ruff
 
+COG_PYTHON_SOURCE := $(shell find python/cog -type f -name '*.py')
+COG_VERSION := $(shell $(PYTHON) -m setuptools_scm 2>/dev/null)
+
+ifndef COG_VERSION
+$(error Could not determine a version for cog! Did you `pip install -e '.[dev]'` first?)
+endif
```
technillogue, line 23: does this break running `make cog` locally?

technillogue: this seems to work fine on my machine, so I wonder if the hint is accurate

Makefile
```diff
@@ -16,56 +15,57 @@ PYTEST := $(PYTHON) -m pytest
...
 default: all
 
 .PHONY: all
 all: cog
 
-pkg/dockerfile/embed/cog.whl: python/* python/cog/* python/cog/server/* python/cog/command/*
-	@echo "Building Python library"
-	rm -rf dist
-	$(PYTHON) -m pip install build && $(PYTHON) -m build --wheel
+.PHONY: wheel
```
technillogue, line 30: :pray: :pray: :pray:, I've run `make pkg/dockerfile/embed/cog.whl` way too many times

pkg/dockerfile/generator.go
```diff
@@ -391,9 +390,19 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked apt-get update -qq &
...
+	}
+	if len(files) != 1 {
+		return "", fmt.Errorf("should only have one cog wheel embedded")
+	}
+	filename := files[0].Name()
+	data, err := cogEmbed.ReadFile("embed/" + filename)
+	if err != nil {
+		return "", err
+	}
+	lines, containerPath, err := g.writeTemp(filename, data)
```
technillogue, line 405: what's the rationale for reading and then writing it like this instead of the previous approach? just having the wheel name?

technillogue (approved):

.github/workflows/ci.yaml
nickstenning: I've changed this so we no longer need to install any Python deps unless we want to run Python tests.

Makefile
nickstenning: This shouldn't have ever broken `make cog` locally, because you're expected to have done `pip install -e '.[dev]'` before you do that.

pkg/dockerfile/generator.go
nickstenning: You can't embed a wildcard pattern `embed/*.whl` except into an `embed.FS`, which then requires that we extract it like this. And yes, the advantage is that we can use the built wheel output of `python -m build` without modification or renaming.

Makefile
technillogue: this hasn't been necessary though...? I've literally never done that in the past year and everything has worked fine...

PR #1874, opened 2024-08-13
Only send the "start" webhook when we start tracking a prediction

erbridge:
We create the task before we call `Worker.predict`, and sometimes, in case of errors, we never actually start the prediction. We should only send the webhook when we actually start.


technillogue (approved):

nickstenning (approved):
LGTM

PR #1875, opened 2024-08-13
Move the minimum torch version to 1.13.0

8W9aG:
* Allow the generate-matrix command to generate base images supporting torch 1.13.0

The intention here is to generate the necessary base image matrix to support running some torch `1.13.0` configurations. The new base images should be built after this is merged.

Solves the 1st problem here: https://github.com/replicate/cog/issues/1862


erbridge:
Do we know why we had that more specific limit before?

Issue #1867, opened 2024-08-12
Unsupported base image configuration for CUDA + Torch 2.0.1+cu118


8W9aG:
I am seeing the following error:

```
Cog doesn't know what CUDA version is compatible with torch==2.0.1+cu118. You might need to upgrade Cog: https://github.com/replicate/cog#upgrade
```

I believe the issue is [torch have stopped tagging their releases](https://pytorch.org/get-started/previous-versions/) explicitly as `cu118` after version 2. We do have a base image that should conform to this setup [here](https://github.com/replicate/cog/blob/main/pkg/config/torch_compatibility_matrix.json#L378-L391). I believe a correct approach here might be to wipe the `+cu118` modifier off the version if the cu118 index URL is being used.

PR #1876, opened 2024-08-13
Handle CUDA version modifiers on torch

8W9aG:
* If we get given a torch with a cuda version modifier such as 2.0.3+cu118 we fail if we do not
explicitly support this version.
* Instead extract the CUDA version from the version tag (11.8) and perform our cuda
compatibility lookup with those 2 versions.

Fixes: https://github.com/replicate/cog/issues/1867


pkg/config/compatibility_test.go
```diff
@@ -61,6 +61,12 @@ func TestGenerateTorchMinorVersionCompatibilityMatrix(t *testing.T) {
 	require.Equal(t, expected, actual)
 }
 
+func TestCudasFromTorchWithCUVersionModifier(t *testing.T) {
+	cudas, err := cudasFromTorch("2.0.1+cu118")
+	require.GreaterOrEqual(t, len(cudas), 1)
```
mattt, line 66:
```suggestion
	require.GreaterOrEqual(t, len(cudas), 1)
        require.Equal(t, cudas[0], "118")
```

pkg/config/compatibility.go
```diff
@@ -166,8 +166,38 @@ func generateTorchMinorVersionCompatibilityMatrix(matrix []TorchCompatibility) [
...
+		return "", ver
+	}
+	versionPlus := versionPluses[len(versionPluses)-1]
+	if !strings.HasPrefix(versionPlus, cudaVersionPrefix) {
+		return "", ver
+	}
+	cleanVersion := strings.TrimPrefix(versionPlus, cudaVersionPrefix)
+	cleanVersion = cleanVersion[:len(cleanVersion)-1] + "." + cleanVersion[len(cleanVersion)-1:]
+	return cleanVersion, versionPluses[0]
+}
```
mattt, line 182: I love Go, but string operations always feel so tedious and messy. Here's an (untested) pass at making this more legible.

```suggestion
func cudaVersionFromTorchPlusVersion(ver string) (string, string) {
	const cudaVersionPrefix = "cu"

	// Split the version string by the '+' character.
	versionParts := strings.Split(ver, "+")

	// If there is no '+' in the version string, return the original string with an empty CUDA version.
	if len(versionParts) <= 1 {
		return "", ver
	}

	// Extract the part after the last '+'.
	cudaVersionPart := versionParts[len(versionParts)-1]

	// Check if the extracted part has the CUDA version prefix.
	if !strings.HasPrefix(cudaVersionPart, cudaVersionPrefix) {
		return "", ver
	}

	// Trim the CUDA version prefix and reformat the version string.
	cleanVersion := strings.TrimPrefix(cudaVersionPart, cudaVersionPrefix)
	if len(cleanVersion) < 2 {
		return "", ver // Handle case where cleanVersion is too short to reformat.
	}

	// Insert a dot before the last character to format it as expected.
	cleanVersion = cleanVersion[:len(cleanVersion)-1] + "." + cleanVersion[len(cleanVersion)-1:]

	// Return the reformatted CUDA version and the main version.
	return cleanVersion, versionParts[0]
}
```

PR #1877, opened 2024-08-14
Download Path inputs in background

nickstenning:
PR #1865 introduced a regression: URLs for Path inputs were downloaded inside the request handler for `POST/PUT /predictions`.

This creates at least two problems:

1. It blocks the uvicorn thread, stopping us from serving healthchecks.
2. It makes the prediction create request take too long, even when `Prefer: respond-async` is specified.

Both of these issues cause unexpected failures in production.

This commit moves responsibility for fetching these files into Worker, and specifically into its background thread. This is probably not a long-term fix, as we want to make Worker (and ChildWorker) async eventually.

This also adds a test which fails on main and is fixed by this change.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -2,6 +2,7 @@ import multiprocessing
 import os
 import signal
 import sys
+import threading
 import traceback
 import types
 from concurrent.futures import Future, ThreadPoolExecutor
@@ -11,6 +12,7 @@ from typing import Any, Callable, Dict, Optional, TextIO, Union
 
 import structlog
 
+from ..types import URLPath
 from ..json import make_encodeable
 from ..predictor import BasePredictor, get_predict, load_predictor_from_ref, run_setup
 from .eventtypes import (
@@ -52,6 +54,9 @@ class Worker:
         self._result: Optional["Future[Done]"] = None
         self._subscribers: Dict[int, Optional[Callable[[_PublicEventType], None]]] = {}
 
+        self._predict_payload: Optional[Dict[str, Any]] = None
+        self._predict_start = threading.Event()  # set when a prediction is started
+
         self._pool = ThreadPoolExecutor(max_workers=1)
         self._event_consumer = None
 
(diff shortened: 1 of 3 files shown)


nickstenning:
I think this removes the need for #1874, @erbridge?

yorickvP:
Does something similar happen on uploads?

nickstenning:
> Does something similar happen on uploads?

Hmm, I don't think so? Uploads are handled synchronously when `set_output`/`append_output` is called on `PredictTask`, which I think is how it's always been. What are you thinking could happen?

yorickvP:
> > Does something similar happen on uploads?
> 
> Hmm, I don't think so? Uploads are handled synchronously when `set_output`/`append_output` is called on `PredictTask`, which I think is how it's always been. What are you thinking could happen?

files are uploaded synchronously in [_predict](https://github.com/replicate/cog/blob/main/python/cog/server/http.py#L392), called by the `/predictions` post handler. Would that not similarly block the uvicorn thread?

nickstenning:
> files are uploaded synchronously in [_predict](https://github.com/replicate/cog/blob/main/python/cog/server/http.py#L392), called by the `/predictions` post handler. Would that not similarly block the uvicorn thread?

Yes, but that's for a synchronous prediction request, which we don't use in production, and is not a change in behaviour.

python/cog/server/worker.py
```diff
@@ -215,10 +219,14 @@ def _consume_events_inner(self) -> None:
 
             self._predict_payload = None
             self._predict_start.clear()
-            self._result.set_result(done)
+
+            # We capture the predict future and then set state to READY before
+            # completing it, so that we can immediately accept work.
+            result = self._result
```
erbridge, line 225: I know it's the same in practice, but it feels safer to couple clearing `self._predict_start` with the rest of resetting the state of the worker:

```suggestion
            # We capture the predict future and then set state to READY before
            # completing it, so that we can immediately accept work.
            result = self._result
            self._predict_payload = None
            self._predict_start.clear()
```

erbridge (approved):
I think this makes sense to me.

I think #1874 still makes sense to do, because it feels weird that initializing a task is considered when the task has "started" but that could be much later. Once `track` is called, though, it's definitely started, because it has a future to wait for.

PR #1881, opened 2024-08-15
Publish on tag push rather than release

nickstenning:
Unfortunately the `release` event associated with the release created by `goreleaser` in the CI workflow does not trigger workflow runs. This is because the release is created using the `$GITHUB_TOKEN`, which doesn't trigger workflows to eliminate the possibility of recursive workflow triggers. See:

  https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

This changes the publishing workflow (which is separate so that PyPI trusted publishing can reference specifically this workflow, and not allow publishing from any CI workflow run) to trigger on tag push, just like the main CI workflow.

The drawback here is that this is not conditional on the tests running and passing. For now, this is probably ok, as we should not be pushing tags until we see that the relevant commit has passed CI.

### DIFF
--- a/.github/workflows/pypi-package.yaml
+++ b/.github/workflows/pypi-package.yaml
@@ -4,9 +4,7 @@ name: PyPI package
 on:
   push:
     branches: [main]
-    tags: ["*"]
-  release:
-    types: [published]
+    tags: ["v[0-9]+.[0-9]+.[0-9]+*"]
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


mattt:
> The drawback here is that this is not conditional on the tests running and passing. For now, this is probably ok, as we should not be pushing tags until we see that the relevant commit has passed CI.

Fortunately, it looks like we can create a [ruleset](https://docs.github.com/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/about-rulesets) to enforce that tags are created only if status checks pass.

mattt (approved):

PR #1888, opened 2024-08-19
Simplify TestWorkerState

nickstenning:
Switch to using bundles so all the state is visible to Hypothesis, and carry the payload/canceled status of each prediction with that state rather than tracking it as instance variables.


nickstenning:
Closing this in favour of a different approach.

PR #1890, opened 2024-08-20
Support new base images

8W9aG:
* Support patch versions of torch by not stripping the torch version to major/minor
* Add new compatibility matrix for torch 2.4.0 and CUDA 12.4.x and cuDNN
* Exclude torchaudio from pip installs like torchvision since the new base images should include the torchaudio package

This will introduce a race condition, since building the base images relies on this matrix, and until that is run these base images will be invalid, so it should be done in close connection to the generation of those base images.

### DIFF
--- a/pkg/config/compatibility.go
+++ b/pkg/config/compatibility.go
@@ -97,11 +97,6 @@ var TFCompatibilityMatrix []TFCompatibility
 var torchCompatibilityMatrixData []byte
 var TorchCompatibilityMatrix []TorchCompatibility
 
-// For minor Torch versions we use the latest patch version.
-// This is semantically different to pip, which uses the .0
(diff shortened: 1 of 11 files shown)


8W9aG:
Will separate https://github.com/replicate/cog/pull/1890/commits/43f0d662e9f17201fcd275e64aa80130e570f5c9 into a separate PR

8W9aG:
Making https://github.com/replicate/cog/pull/1890/commits/46375e3f40f343891f8a338ea1a58f33daed0c30 into a separate PR.

https://github.com/replicate/cog/pull/1893

8W9aG:
Isolating the torch patch version support in https://github.com/replicate/cog/pull/1894

PR #1893, opened 2024-08-21
Add new base images for torch 2.4.0

8W9aG:
* Adds torch 2.4.0 compatibility in the matrix
* Add CuDNN to CUDA 12.4.x compatibility

After this new version of torch is added we can build base images with it.

### DIFF
--- a/pkg/config/torch_compatibility_matrix.json
+++ b/pkg/config/torch_compatibility_matrix.json
@@ -1,5 +1,65 @@
 [
   {
+    "Torch": "2.4.0",
+    "Torchvision": "0.19.0",
+    "Torchaudio": "2.4.0",
+    "FindLinks": "",
+    "ExtraIndexURL": "https://download.pytorch.org/whl/cu124",
+    "CUDA": "12.4",
+    "Pythons": [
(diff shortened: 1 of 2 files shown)


nevillelyh (approved):

pkg/config/torch_compatibility_matrix.json
```diff
@@ -1,4 +1,64 @@
 [
+  {
+    "Torch": "2.4.0",
+    "Torchvision": "0.19.0",
+    "Torchaudio": "2.4.0",
+    "FindLinks": "",
+    "ExtraIndexURL": "https://download.pytorch.org/whl/cu124",
+    "CUDA": "12.4",
```
nevillelyh, line 8: So CUDA minor versions are not compatible? And there's no 12.2 or 12.3?

8W9aG: Going from this compatibility matrix here: https://pytorch.org/

Also this is somewhat cargo-culted from the rest of the compatibility matrix.

Screenshot 2024-08-21 at 4 34 19 PM

PR #1894, opened 2024-08-21
Support torch patch base images

8W9aG:
* Do not strip patch version from torch
* If the user specifies a torch version without a patch, resolve the highest patch version from our compatibility matrix
* Change the version struct to use a pointer for the patch version in order to handle cases where the pointer has not been supplied (and how that manifests when comparing versions)
* Standardise on StripModifier for the version rendering
* No longer generate a torch minor compatibility matrix, unnecessary now we explicitly define patch version compatibilities

### DIFF
--- a/pkg/dockerfile/base.go
+++ b/pkg/dockerfile/base.go
@@ -120,7 +120,7 @@ func BaseImageConfigurations() []BaseImageConfiguration {
 			}
 
 			cuda := *compat.CUDA
-			torch := version.StripPatch(compat.Torch)
+			torch := compat.Torch
 			conf := BaseImageConfiguration{
 				CUDAVersion:   cuda,
 				PythonVersion: python,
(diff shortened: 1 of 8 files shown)


pkg/dockerfile/base.go
```diff
@@ -110,17 +108,19 @@ func BaseImageConfigurations() []BaseImageConfiguration {
...
 
 	// Torch configs
-	for _, compat := range config.TorchMinorCompatibilityMatrix {
+	for _, compat := range config.TorchCompatibilityMatrix {
 		for _, python := range compat.Pythons {
 
-			// Only support fast cold boots for Torch with CUDA.
-			// Torch without CUDA is a rarely used edge case.
 			if compat.CUDA == nil {
+				configs = append(configs, BaseImageConfiguration{
```
andreasjansson, line 115: Is there a reason to build base images for torch without gpu? Feels like a very edgy edge case.

pkg/util/version/version.go
```diff
@@ -32,7 +32,9 @@ func NewVersion(s string) (version *Version, err error) {
 		}
 	}
 	if len(parts) >= 3 {
-		version.Patch, err = strconv.Atoi(parts[2])
+		patch, err := strconv.Atoi(parts[2])
+		version.Patch = new(int)
```
andreasjansson, line 36: Why is this necessary? Warrants a comment. And should this happen after the error check?

andreasjansson (commented):
We should also stop stripping out torch and torchvision from the generated requirements.txt in the build process, but that's probably a different PR?

8W9aG:
> We should also stop stripping out torch and torchvision from the generated requirements.txt in the build process, but that's probably a different PR?

Yeh the next PR after this one handles torchvision and torchaudio

pkg/util/version/version.go
8W9aG: The error check should be before the Patch pointer allocation you are right, I'll add a comment.

pkg/dockerfile/base.go
andreasjansson: I see, yes that makes sense.

PR #1896, opened 2024-08-22
Support async predictors

nickstenning:
This extracts some of the next steps from @technillogue's `async` branch, with a view to making `async def predict` signatures possible.

Things that don't work yet:

- actually concurrent predictions (I don't want to do this in this PR)
- ~cancelation~
- ~log redirection~

_This is now ready for review. I think this is just about the smallest meaningful change we can make that gets us closer to supporting real concurrent `async def predict` in Cog._

### DIFF
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -8,9 +8,48 @@ import sys
 import threading
 import uuid
 from types import TracebackType
-from typing import Callable, Sequence, TextIO
+from typing import BinaryIO, Callable, Sequence, TextIO
 
 from typing_extensions import Self
+
+
+class _SimpleStreamWrapper(io.TextIOWrapper):
+    """
+    _SimpleStreamWrapper wraps a binary I/O buffer and provides a TextIOWrapper
+    interface (primarily write and flush methods) which call a provided
+    callback function instead of (or, if `tee` is True, in addition to) writing
+    to the underlying buffer.
+    """
+
+    def __init__(
+        self,
+        buffer: BinaryIO,
+        callback: Callable[[str, str], None],
+        tee: bool = False,
+    ) -> None:
+        super().__init__(buffer, line_buffering=True)
+
+        self._callback = callback
+        self._tee = tee
+        self._buffer = []
+
+    def write(self, s: str) -> int:
+        length = len(s)
+        self._buffer.append(s)
+        if self._tee:
+            super().write(s)
+        else:
+            # If we're not teeing, we have to handle automatic flush on
+            # newline. When `tee` is true, this is handled by the write method.
+            if "\n" in s or "\r" in s:
+                self.flush()
+        return length
+
+    def flush(self) -> None:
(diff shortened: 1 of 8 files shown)


python/cog/server/helpers.py
```diff
@@ -6,6 +6,32 @@
...
+        hook: Callable[[str], None],
+    ) -> None:
+        self._hook = hook
+        self._buffer = []
+        self._tee = tee
+        super().__init__(buffer, line_buffering=True)
+
+    def write(self, data: str) -> int:
+        self._buffer.append(data)
+        if self._tee:
```
nickstenning, line 23: Ah, I've just realised that supporting `self._tee == False` requires more work as this currently is relying on the fact that `super().write(data)` will call `flush` if the data contains line terminators. Maybe we just don't support turning teeing off? I don't know we have a valid use case for it.

python/cog/server/worker.py
```diff
@@ -328,28 +342,57 @@ def _setup(self, redirector: StreamRedirector) -> None:
...
                     print(f"Got unexpected event: {ev}", file=sys.stderr)
 
+    async def _aloop(
+        self,
+        predict: Callable[..., Any],
+        redirector: AsyncStreamRedirector,
+    ) -> None:
+        # Unwrap and replace the events connection with an async one.
+        assert isinstance(self._events, LockedConnection)
+        self._events = AsyncConnection(self._events.connection)
```
technillogue, line 369: it may be cleaner to have this be a separate variable name instead of making _events.recv either sync or async

python/cog/server/worker.py
```diff
@@ -360,8 +403,46 @@ def _predict(
...
         except CancelationException:
             done.canceled = True
+        # async cancelation
+        except asyncio.CancelledError:
+            done.canceled = True
+            # We've handled the requested cancelation, so we uncancel the task.
+            # This ensures that any cleanup work we do won't be interrupted.
+            task = asyncio.current_task()
+            assert task
+            task.uncancel()
```
technillogue, line 445: whoa

I _think_ cleanup work isn't supposed to be interrupted, at least not in a way that's prevented by uncancelling, but it would be fantastic to test this

technillogue (approved):

python/cog/server/worker.py
nickstenning: I think you might be right that this isn't necessary. I found [this](https://github.com/python/cpython/issues/102780) which suggested to me that it was, but I think that was just a bug in the `asyncio.timeout` context manager.

technillogue: yeah, that doesn't seem like it's related to our use of TimeoutError, but that is interesting

meatballhat:
For anyone checking in on this one, I'm attempting to rebase and resolve conflicts _without_ overwriting the history in https://github.com/replicate/cog/pull/2009

nickstenning:
Closing in favour of the various rebases of this branch.

PR #1897, opened 2024-08-22
Add Torchvision and Torchaudio to base images

8W9aG:
* Exclude torchaudio from pip installs (like torchvision)
* Install torchvision and torchaudio in the base images

### DIFF
--- a/pkg/config/compatibility.go
+++ b/pkg/config/compatibility.go
@@ -301,7 +301,7 @@ func torchGPUPackage(ver string, cuda string) (name, cpuVersion, findLinks, extr
 	var latest *TorchCompatibility
 	for _, compat := range TorchCompatibilityMatrix {
 		compat := compat
-		if compat.TorchVersion() != ver || compat.CUDA == nil {
+		if !version.Matches(compat.TorchVersion(), ver) || compat.CUDA == nil {
 			continue
 		}
 		greater, err := versionGreater(*compat.CUDA, cuda)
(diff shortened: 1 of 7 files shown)


andreasjansson:
I don't think we should `excludePackages` at all. When we do that, pip doesn't know that a version is already installed. Instead I think we should _add_ torchvision and torchaudio to the requirements if they're not there already.

We probably also need some system packages for torchaudio to work. I'm guessing ffmpeg and possible some others.

8W9aG:
> I don't think we should `excludePackages` at all. When we do that, pip doesn't know that a version is already installed. Instead I think we should _add_ torchvision and torchaudio to the requirements if they're not there already.
> 
> We probably also need some system packages for torchaudio to work. I'm guessing ffmpeg and possible some others.

I'll add these packages to the base image: https://pytorch.org/audio/main/installation.html

Instead of excluding packages I will add them if they don't exist and if they exist I will overwrite them to the version we support, does that sound like a plan?

8W9aG:
> > I don't think we should `excludePackages` at all. When we do that, pip doesn't know that a version is already installed. Instead I think we should _add_ torchvision and torchaudio to the requirements if they're not there already.
> > We probably also need some system packages for torchaudio to work. I'm guessing ffmpeg and possible some others.
> 
> I'll add these packages to the base image: https://pytorch.org/audio/main/installation.html
> 
> Instead of excluding packages I will add them if they don't exist and if they exist I will overwrite them to the version we support, does that sound like a plan?

Should also add that it seems these base packages are already installed: https://github.com/replicate/cog/blob/main/pkg/dockerfile/base.go#L18

andreasjansson:
Another thing that occurred to me: Are we including the right index url in the generated requirements.txt for the user image when we include torch?
image

8W9aG:
> Another thing that occurred to me: Are we including the right index url in the generated requirements.txt for the user image when we include torch? image
> 
> Might warrant a test if no test is currently exercising that.

We weren't but I've got tests now that correct for that

andreasjansson (approved):
👍

PR #1900, opened 2024-08-23
Generate optimized python binaries with pyenv

8W9aG:
* Follow the pyenv guide for producing optimised python binaries for better performance.

https://github.com/pyenv/pyenv/blob/master/plugins/python-build/README.md#building-for-maximum-performance

If the user is not using a base image this adds some latency to the build, I am guessing this might be fine since not supported base images are rare


nickstenning (commented):
Generally this seems fine. I wonder how we'd test this?

More broadly, I do slightly wonder whether we should be switching away from `pyenv` altogether. `uv` now supports downloading and installing Python versions from 3.8 upwards. (`uv python install 3.8`) Should we evaluate that?

8W9aG:
> Generally this seems fine. I wonder how we'd test this?
> 
> More broadly, I do slightly wonder whether we should be switching away from `pyenv` altogether. `uv` now supports downloading and installing Python versions from 3.8 upwards. (`uv python install 3.8`) Should we evaluate that?

Beyond the integration tests we have there probably isn't much we can do testing wise, other than trust that compilers are generally dependable. For what its worth some of the other fixes I've got coming down the pipeline deal with pyenv and I'm not convinced it should be used in a high performance environment, evaluating `uv` can be done as a separate stream.

nickstenning:
LGTM, in that case.

PR #1902, opened 2024-08-23
Add strip to cog builds

8W9aG:
* Strip out all unnecessary debug symbols shipped by prebuilt shared objects.
* Careful not to touch the cpython binaries they don’t like strip, but they can be handled
separately with flags.
* This speeds up load times of the shared objects, especially when it comes to PyTorch.


nevillelyh:
`RUN` creates a new layer with the same files striped, so it might significantly increase image size. Is it possible to do them in the same `RUN pip install` step?

8W9aG:
> `RUN` creates a new layer with the same files striped, so it might significantly increase image size. Is it possible to do them in the same `RUN pip install` step?

Yeh great idea I'll do it there

technillogue (approved):
this looks reasonable, but it would be best to do this as a separate release from other optimizations so it can be rolled back if anything breaks.

it might also be nice to have a way to disable this, since it naturally makes segfaults a lot harder to debug

PR #1903, opened 2024-08-23
Rebuild ld cache in build

8W9aG:
* Find all the python related SO’s, and their corresponding folders and put them in a cog.conf
file in the /etc/ld.so.conf.d folder
* Run ldconfig to rebuild the cache


technillogue (approved):
this looks good, a related optimization I've used in the past is putting all the dependencies in `.` to reduce the number of ENOENT when importing. normally, python will go over everything in sys.path (usually starting with `.`) and try to open each library at each path. when I first dad I had some trouble with pyenv and installed binaries, but maybe something similar could work.

8W9aG:
> this looks good, a related optimization I've used in the past is putting all the dependencies in `.` to reduce the number of ENOENT when importing. normally, python will go over everything in sys.path (usually starting with `.`) and try to open each library at each path. when I first dad I had some trouble with pyenv and installed binaries, but maybe something similar could work.

I like that idea, I did see a tonne of `ENOENT` when I performed `strace` over the python invocation.

technillogue:
fwiw I believe nix using absolute paths for everything also helps

this is the current sys.path on a random container: `['', '/root/.pyenv/versions/3.10.14/lib/python310.zip', '/root/.pyenv/versions/3.10.14/lib/python3.10', '/root/.pyenv/versions/3.10.14/lib/python3.10/lib-dynload', '/root/.pyenv/versions/3.10.14/lib/python3.10/site-packages']`

we might be able to improve on this in a simpler way by just shuffling it so that site-packages are first, or just symlinking all of site-packages into `.`

PR #1904, opened 2024-08-23
Add Cython CFLAGS for package optimizations

8W9aG:
* These flags should make the code faster and smaller
* This will affect cpython generated shared libraries


pkg/dockerfile/generator.go
```diff
@@ -42,6 +42,7 @@ coverage.xml
 .pytest_cache
 .hypothesis
 `
+const CythonCFlags = "CFLAGS=\"-O3 -march=native -ffast-math -funroll-loops -fno-strict-aliasing -flto -mtune=native -S\""
```
nickstenning, line 45: Given this is an environment variable, would perhaps be easier/cleaner to set it with `ENV` and (if necessary) unset it at the end, rather than interpolating it in multiple places?

8W9aG: I'll add this to the preamble `ENV` commands instead, this will ensure these settings are applied to any compilation process, I can't think of a reason why that wouldn't be ultimately desirable.

8W9aG: I opt for wrapping pip installs in the set/unset of this ENV var to keep the blast area smaller

nickstenning (approved):

technillogue:
this is good but we really, really can't use ffast math, it breaks many ML things: https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html

8W9aG:
> this is good but we really, really can't use ffast math, it breaks many ML things: https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html

Thanks for that I had no idea, will remove `-ffast-math` flag.

technillogue (approved):

PR #1909, opened 2024-08-26
Add findutils to base image

8W9aG:
* Allow base images to use the find unix command
* This is required to use the python compileall command in conjunction with find

### DIFF
--- a/pkg/dockerfile/base.go
+++ b/pkg/dockerfile/base.go
@@ -20,6 +20,7 @@ var (
 		"cmake",
 		"curl",
 		"ffmpeg",
+		"findutils",
 		"g++",
 		"gcc",
 		"git",


nevillelyh:
Is `compileall` used for building image only? In that case you can `apt install findutils && python ... && apt remove ... && rm -rf /var/apt/cache ...`? But it'll be a lot of steps in 1 `RUN` maybe better off adding a few KBs to the base image :shrug:

8W9aG:
> Is `compileall` used for building image only? In that case you can `apt install findutils && python ... && apt remove ... && rm -rf /var/apt/cache ...`? But it'll be a lot of steps in 1 `RUN` maybe better off adding a few KBs to the base image 🤷

I think a bit of an issue here is this will be used in different places so it might be better just to bake it in if its only a few KBs

PR #1912, opened 2024-08-28
Clarify default install directory behaviour

aknoerig:
The install script seems to suggest `/usr/local/bin` as the default install directory. However, pressing `RETURN` will actually set the install dir to an empty string. The error message is not helpful as it then reads `The directory  does not exist.` (note the additional space).

This fix makes the script behave as expected: pressing `RETURN` accepts the default suggestion.
(Also factors out the directory selection into a function.)

### DIFF
--- a/tools/install.sh
+++ b/tools/install.sh
@@ -42,6 +42,22 @@
 # SOFTWARE.
 set -e
 
+set_install_dir() {
+  # Set install directory
+  DEFAULT_INSTALL_DIR="/usr/local/bin"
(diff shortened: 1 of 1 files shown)


aknoerig:
Addresses issue #1914

yorickvP (approved):

cpietsch:
ran into the same issue - was also expecting the default/placeholder to be on `RETURN`

mattt (approved):
Thanks so much, @aknoerig!

PR #1922, opened 2024-09-03
Fix base image build job

8W9aG:
* Set no-cache to true to avoid an issue with buildx where it doesn't recognise the cached sha digests
* Set max-parallel to 10 to process these builds faster, this is necessary due to the arm builds taking a very long time to compile python
* Allow continuation on error in case a build later down the pipe causes a small issue
* Remove torch version modifiers in the compatibility matrix, the extra URL does the job of routing them correctly and they make pip fail when using these extra URLs


8W9aG:
This is superceded by https://github.com/replicate/cog/pull/1923

PR #1926, opened 2024-09-04
Fix test build integration test

8W9aG:
* This string has escape sequences in it.

On a broader note I'm unsure why this is suddenly the case, it seems to have correlated with the release of new base images, but its unclear to me why it would add escape sequences as part of its stderr output.


8W9aG:
Some more observations here:

* Turning off cog base images does nothing
* This console output comes from go running the docker images, and is present in the go console, making me think this may not actually be related to the base images

timfallmk (approved):
Looks straightforward.

PR #1937, opened 2024-09-10
Gate base images behind a version flag

8W9aG:
* Give cog the ability to declare compatibility with base image versions.
* Add a hidden flag `--base-image-version` which applies a base image version to the base image tag
* The default of this flag will be the base image version corresponding to the current cog release


nevillelyh:
So it'll add `baseX` to docker tag?
* Who sets `X`? Should it be from the git commit hash of this repo?
* Should we set both `baseB-cudaX-pythonY-torchZ` and `cudaX-pythonY-torchZ` tags, so that the latter is always the "latest" for `cudaX`/`pythonY`/`torchZ` combo and the former is immutable?

8W9aG:
I think I might close this as we are thinking more and more about floating base images.

PR #1939, opened 2024-09-10
Remove torch vision GPU compatibility check

8W9aG:
* In the later versions of torch this is released 1:1 with torch versions


8W9aG:
Closing this, will go back to treating torchvision as a fully fledged dependency similar to pytorch.

PR #1943, opened 2024-09-11
Use COPY —-from when importing common images

8W9aG:
* This squashes the images allowing us faster resolution time in the snapshotter.

WIP: I want to figure out the impact this may have.


8W9aG:
Going with a different approach

PR #1944, opened 2024-09-11
Add mount caches to pip installs

8W9aG:
* Allow cached directories, but on every pip install use a volume mount to make the cache
invisible to the running container


technillogue (approved):
perfect, thank you

PR #1947, opened 2024-09-12
Fix openapi schema

philandstuff:
Generating the OpenAPI schema was broken under the circumstances:

- train.py#train had an enum-type input
- cog was running with --x-mode=train

I've verified that this change fixes it.  The current code is obviously wrong in any case.


aron (approved):
Good catch!

evilstreak (approved):
🙏

PR #1950, opened 2024-09-12
Bump go to 1.23.1

8W9aG:
* Fixes tests that require `1.23.1`

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-golang 1.22.6
+golang 1.23.1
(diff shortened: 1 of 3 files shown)


philandstuff:
Assuming it gets merged, we should revert https://github.com/replicate/cog/pull/1953 as part of this PR.

philandstuff:
I have no idea why golangci-lint is exploding under go 1.23 and not 1.22. For the moment #1953 has got `main` releasable again, but we should understand and fix this at some point.

Issue #1951, opened 2024-09-12
Resolve Torch to maximum compatible version


8W9aG:
If the user doesn't define which torch version to use, or a non-patch version, resolve this to the highest version we currently support manually.

PR #1952, opened 2024-09-12
Add Torch Dependency Resolution

8W9aG:
* Use our compatibility matrix to automatically resolve the torch version if the torch version is
blank but adjacent dependencies are defined.
* Make torchaudio capable of determining CUDA version compatibility like torchvision
* Rewrite dependency versions to align with resolved versions
* Remove version modifiers, this is now satisfied by adding in extra index URLs to the requirements

* Fixes: https://github.com/replicate/cog/issues/1951

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -118,7 +118,7 @@ jobs:
     name: "Test integration"
     needs: build-python
     runs-on: ubuntu-latest-16-cores
-    timeout-minutes: 10
+    timeout-minutes: 20
     steps:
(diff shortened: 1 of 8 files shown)


philandstuff:
Now that the tests run properly, that `test_torch_version_resolution_with_torchvision` test failure looks suspicious 👀

8W9aG:
> Now that the tests run properly, that `test_torch_version_resolution_with_torchvision` test failure looks suspicious 👀

Looks like we are hitting the timeout on integration tests, I will bump to 20

Issue #1949, opened 2024-09-12
Install packages before pip install


8W9aG:
Currently we run `pip install` as a build step before `apt-get install`. Some of those python packages depend on those packages but never the other way around, so run this build step first. Add good example is `dlib` which uses `cmake`, we do install `cmake` just after the python dependencies, so it errors because it can't find `cmake`.

AlexanderKozhevin:
has it been released?
I having issue with dlib which depends on cmake

PR #1954, opened 2024-09-13
Install apt packages before pip packages

8W9aG:
* Moves apt installs to just below the initial FROM line
* This prevents pip packages not installing due to depending on these apt packages

Fixes: https://github.com/replicate/cog/issues/1949

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -157,7 +157,7 @@ func (g *Generator) generateInitialSteps() (string, error) {
 		}), nil
 	}
 
-	pipInstallStage, err := g.pipInstallStage()
+	pipInstallStage, err := g.pipInstallStage(aptInstalls)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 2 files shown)


mattt (approved):
Thanks, @8W9aG.

I love this improvement. I don't love what's needed to get there. (And I'm sure you're also not super jazzed about `pipInstallStage` take `aptInstalls` as an argument).

How we generate Dockerfiles is long overdue for a rethink. I don't believe we have the appetite to pick that up anytime soon, so I'm fine to hold my nose on this one.

8W9aG:
> Thanks, @8W9aG.
> 
> I love this improvement. I don't love what's needed to get there. (And I'm sure you're also not super jazzed about `pipInstallStage` take `aptInstalls` as an argument).
> 
> How we generate Dockerfiles is long overdue for a rethink. I don't believe we have the appetite to pick that up anytime soon, so I'm fine to hold my nose on this one.

In this case if we removed the separate weights stuff I think we could come up with a very decent dockerfile generator which was more declarative than procedural.

PR #1960, opened 2024-09-19
Fix memory leak

nickstenning:
This naive implementation of the observer pattern both leaks memory and causes a very slow performance degradation over time as the `_subscribers` dictionary gets larger and larger, mostly full of `None` values.

This replaces the implementation with one that ensures that repeated subscribe/unsubscribe calls don't affect performance or memory consumption.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -5,6 +5,7 @@ import sys
 import threading
 import traceback
 import types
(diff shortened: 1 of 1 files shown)


philandstuff (approved):
this is the most hilariously minor memory leak I've seen, good work everyone

Issue #1961, opened 2024-09-19
Add .cog to a .dockerignore produces hard to debug states


8W9aG:
If you add `.cog` to a `.dockerignore` it produces issues like the following:

```
------
 > [stage-0 3/7] COPY .cog/tmp/build20240919160513.8228582876843814/requirements.txt /tmp/requirements.txt:
------
Dockerfile:4
--------------------
   2 |     FROM r8.im/cog-base:cuda11.8-python3.11
   3 |     RUN --mount=type=cache,target=/var/cache/apt,sharing=locked apt-get update -qq && apt-get install -qqy  && rm -rf /var/lib/apt/lists/*
   4 | >>> COPY .cog/tmp/build20240919160513.8228582876843814/requirements.txt /tmp/requirements.txt
   5 |     ENV CFLAGS="-O3 -funroll-loops -fno-strict-aliasing -flto -S"
   6 |     RUN pip install --no-cache-dir -r /tmp/requirements.txt
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 50e2db01-7591-4a90-a5ab-e68b1ae19a0e::qc781cqwiy4yk3g558ya3mubm: "/.cog/tmp/build20240919160513.8228582876843814/requirements.txt": not found
```

This is quite difficult to debug, we should either warn the user that adding `.cog` produces undesireable effects or filter this out of the `.dockerignore` file produced by the user.

@zsxkib

anish-sahoo:
Closing because .cog/ is now filtered at the BuildKit session level regardless of the user's .dockerignore.

PR #1962, opened 2024-09-20
Fix .dockerignore ignoring the .cog directory we produce

8W9aG:
* Before the build process sanitise the `.dockerignore` file to make sure it doesn't contain the `.cog` folder.
* Add an integration test that tests for this case.

Fixes: https://github.com/replicate/cog/issues/1961


technillogue (commented):
what we really want is for the dockerignore to look like this:

```
.cog
!.cog/tmp/<current build id>
```
the one annoying thing about that is that it means the dockerignore will keep changing, which would add noise to diffs if committed to git. it would best if there was an --extra-dockerignore flag to docker build (or similar), but I don't think that exists

PR #1972, opened 2024-09-25
use threading.Lock instead of mp.Lock

technillogue:
#1802 mentions
> although the lock isn't actually shared between processes and the more lightweight threading.Lock could be used, that means it has to be created in run() rather than __init__

The use of multiprocessing.Lock was introduced in #1773. We suspected that this might have been interfering with other libraries that might use multiprocessing. The specific comfy error turned out to be caused by something else, but it's still unnecessary to have a multiprocess lock.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -247,7 +247,7 @@ class Worker:
 class LockedConn:
     def __init__(self, conn: Connection) -> None:
         self.conn = conn
-        self._lock = _spawn.Lock()
+        self._lock = threading.Lock()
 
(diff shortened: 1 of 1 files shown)


philandstuff:
can you give a description of what problem you are trying to solve and why this PR solves it?

technillogue:
I've added a description of what problem you are trying to solve and why this PR solves it

yorickvP (approved):

philandstuff:
Just noting I think #2020 will make this unnecessary, by removing LockedConn entirely

PR #1974, opened 2024-10-03
Look up git metadata via GitHub env vars when available

meatballhat:
and fall back to `git` operations otherwise. This also changes the way that a `git` repo is detected to use `git` itself instead of checking for `.git` directory existence.

(linked to internal issue PLAT-228)


timfallmk (approved):
Looks good!

PR #1980, opened 2024-10-08
Prefix server errors with `Cog:`

meatballhat:
to allow for separation of error tracking.

Connected to internal issue PLAT-265

### DIFF
--- a/python/cog/errors.py
+++ b/python/cog/errors.py
@@ -8,7 +8,3 @@ class ConfigDoesNotExist(CogError):
 
 class PredictorNotSet(CogError):
     """Exception raised when 'predict' is not set in cog.yaml when it needs to be."""
-
-
-"""Prefix for all Cog errors attributable to internals or infrastructure."""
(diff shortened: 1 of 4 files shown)


nickstenning (commented):
These error messages can and will be seen by users, so I think we should probably avoid the word "internal." Maybe just `Cog:`?
 
What do you think about using a custom user exception type rather than a constant? Something like:

```python
class CogException(Exception):
     def __str__(self):
         return f"Cog: {super().__str__()}"
```

If we did that I'd imagine stuff like `UnknownPredictionError`, `RunnerBusyError`, etc. all become subclasses of `CogError`.

The only slightly tricksy one is likely the `TimeoutError` for which we could probably get by with:

```python
class CogTimeoutError(CogError, TimeoutError):
    pass
```

meatballhat:
@nickstenning I would much prefer a custom exception class 🎉

nickstenning (approved):
LGTM!

PR #1981, opened 2024-10-09
Fail prediction on unhandled error

evilstreak:
If the runner fails to handle an error this gets bubbled up to the worker, which logs a warning and then merrily carries on. Errors bubbling up in this way include things like failing to upload a file, which should fail the prediction. Currently if a model fails to upload one or more files for the prediction, the prediction will still be marked as succeeded, but it won't have any files in the output.


erbridge (approved):

meatballhat (approved):
🥅

Issue #1963, opened 2024-09-20
`cog build --use-cog-base-image=false` fails on invalid wheel filename


josephhaaga:
Looks like `cog build --use-cog-base-image=false` fails with this particular combination of CUDA 12.3 and Python 3.11 due to:
- no r8.im base image to start from
- the `nvidia/cuda:12.3.2-cudnn9-devel-ubuntu22.04` base image being picky about wheel filenames

I suspect the generated wheel needs a filename that will pass [this regex](https://github.com/pypa/pip/blob/102d8187a1f5a4cd5de7a549fd8a9af34e89a54f/src/pip/_internal/models/wheel.py#L17-L22)

<details><summary>My machine</summary>
<p>

```zsh
$ cog --version
cog version 0.9.23 (built 2024-09-13T09:49:08Z)

$ neofetch
                    'c.          josephhaaga@computer
                 ,xNMM.          ---------------------
               .OMMMMo           OS: macOS 14.7 23H124 arm64
               OMMM0,            Host: MacBookPro18,2
     .;loddo:' loolloddol;.      Kernel: 23.6.0
   cKMMMMMMMMMMNWMMMMMMMMMM0:    Uptime: 1 day, 1 hour, 23 mins
 .KMMMMMMMMMMMMMMMMMMMMMMMWd.    Packages: 183 (brew)
 XMMMMMMMMMMMMMMMMMMMMMMMX.      Shell: zsh 5.9
;MMMMMMMMMMMMMMMMMMMMMMMM:       Resolution: 1728x1117
:MMMMMMMMMMMMMMMMMMMMMMMM:       DE: Aqua
.MMMMMMMMMMMMMMMMMMMMMMMMX.      WM: yabai
 kMMMMMMMMMMMMMMMMMMMMMMMMWd.    Terminal: tmux
 .XMMMMMMMMMMMMMMMMMMMMMMMMMMk   CPU: Apple M1 Max
  .XMMMMMMMMMMMMMMMMMMMMMMMMK.   GPU: Apple M1 Max
    kMMMMMMMMMMMMMMMMMMMMMMd     Memory: 2481MiB / 32768MiB
     ;KMMMMMMMWXXWMMMMMMMk.
       .cooc,.    .,coo:.
```

`Use Rosetta for x86_64/amd64 emulation on Apple Silicon` is **disabled**

</p>
</details> 

## details

There doesn't seem to be a CUDA 12.3 + Python 3.11 base image available
[...]

PR #1982, opened 2024-10-09
Catch cog.whl invalid filenames

8W9aG:
* Check if the cog wheel is an invalid filename from the point of view of pip install
* Change the filename to conform to a valid pip filename.

Fixes: https://github.com/replicate/cog/issues/1963

I believe this was some kind of regression in https://github.com/replicate/cog/pull/1873 however I am not entirely sure what. It's important to note that this does not occur on any development branches, I am quite sure this only occurs in release, but the mechanism as to how this occurs I am not entirely sure of. This in turn makes it hard to create a test case against and makes the code here defensive.

### DIFF
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/replicate/cog/pkg/config"
+	"github.com/replicate/cog/pkg/global"
 	"github.com/replicate/cog/pkg/util/console"
 	"github.com/replicate/cog/pkg/util/slices"
 	"github.com/replicate/cog/pkg/util/version"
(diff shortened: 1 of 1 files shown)


nickstenning:
I'd honestly much rather we fix the place that's generating incorrect wheels. That shouldn't be the case either with a dev build or with released versions of cog, and we should fix it if it is.

nickstenning:
Ok, well that wasn't too hard. [This is the problem.](https://github.com/Homebrew/homebrew-core/blob/b1ecd61331fb78ea26ab34f9697185de85fe61a3/Formula/c/cog.rb#L29)

nickstenning:
I think we should probably aim to fix the Homebrew bottle build so it does the right thing rather than hack/patch it like this.

nickstenning:
https://github.com/Homebrew/homebrew-core/pull/193397/files

nickstenning:
This is now fixed:

```console
$ brew install cog
[...]
$ which cog
/opt/homebrew/bin/cog
$ strings $(which cog) | grep -Eo 'embed/.+.whl'
embed/cog-0.9.26-py3-none-any.whl
```

PR #1984, opened 2024-10-10
add a "quiet" flag to the HTTP API

danielkeyes-sony:
For some context, I was working on a use-case where the size of the **inputs** to my predictor were very large, but the size of the **outputs** are relatively small. I noticed that since both `PredictionResponse` and `PredictionRequest` are subclasses of `PredictionBaseModel` (which defines `input`), this meant that the inputs were transmitted in both the response and the request, which used up a lot of network bandwidth.

This PR adds an optional "quiet" flag to the http API, which just makes it so the `PredictionResponse` does not repeat the input.

Let me know if this looks good, or if you would like me to change anything.

### DIFF
--- a/python/cog/command/ast_openapi_schema.py
+++ b/python/cog/command/ast_openapi_schema.py
@@ -50,6 +50,11 @@ BASE_SCHEMA = """
             "default": ["start", "output", "logs", "completed"],
             "items": { "$ref": "#/components/schemas/WebhookEvent" },
             "type": "array"
+          },
+          "quiet": {
+            "title": "Quiet",
+            "type": "boolean",
+            "default": false
           }
         },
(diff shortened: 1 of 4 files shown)


tempusfrangit (changes requested):
I think this is reaosnable to pursue, however, we've moved away from the FASTAPI based http server. If you would like to respin this against the new Rust `coglet` implementation, I'd be happy to consider the change. 

Another approach to consider is adding an optional param to the prediction request itself to indicate how we send details back both via the http response and webhooks.

anish-sahoo:
Closing this PR as it is now out of scope. As @tempusfrangit mentioned, feel free to open a new PR with a rust/coglet implementation.

PR #1985, opened 2024-10-11
Support URLFile in the upload_file function

aron:
We would like `predict` functions to be able to return a remote URL rather than a local file on disk and have it behave like a file object. And when it is passed to the file uploader it will stream the file from the remote to the destination provided.

```py
    def predict(self, **kwargs) -> File:
        return URLFile("https://replicate.delivery/czjl/9MBNrffKcxoqY0iprW66NF8MZaNeH322a27yE0sjFGtKMXLnA/hello.webp")
```

This PR adds an additional check to the `upload_file()` handler to call `fh.seekable()` before attempting to seek. This allows instances of `io.IOBase` that do not support seek (like `URLFile`) to be uploaded.

We also add a `name` attribute to `URLFile`. This is used by the `upload_file` function to infer the file extension and mime type.

### DIFF
--- a/python/cog/types.py
+++ b/python/cog/types.py
@@ -104,6 +104,7 @@ class Secret(pydantic.SecretStr):
                 }
             )
             return json_schema
+
     else:
 
         @classmethod
@@ -223,6 +224,7 @@ class Path(pathlib.PosixPath):  # pylint: disable=abstract-method
             json_schema = handler(core_schema)
             json_schema.update(type="string", format="uri")
             return json_schema
+
     else:
 
         @classmethod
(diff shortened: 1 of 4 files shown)


python/cog/types.py
```diff
@@ -286,6 +288,7 @@ class URLFile(io.IOBase):
     __slots__ = ("__target__", "__url__")
 
     def __init__(self, url: str) -> None:
+        object.__setattr__(self, "name", os.path.basename(url))
```
nickstenning, line 291: Neat trick, but we should probably actually parse the URL here.

```python
In [9]: os.path.basename("http://google.com/florp.kpg?ga_id=abcd123#monkeys")
Out[9]: 'florp.kpg?ga_id=abcd123#monkeys'
```

nickstenning (commented):
Other than the filename extraction which doesn't really work for URLs in general, this LGTM.

nickstenning:
I think we need to limit this to `http/https/data` URLs for security reasons, as noted in #1987.

PR #1987, opened 2024-10-11
[async] Support URLFile in the upload_file function

aron:
We would like `predict` functions to be able to return a remote URL rather than a local file on disk and have it behave like a file object. And when it is passed to the file uploader it will stream the file from the remote to the destination provided.

```py
    def predict(self, **kwargs) -> File:
        return URLFile("https://replicate.delivery/czjl/9MBNrffKcxoqY0iprW66NF8MZaNeH322a27yE0sjFGtKMXLnA/hello.webp")
```

This PR modifies the URLFile class to use `urllib.request.openurl` instead of `requests` as the HTTP client for reading the file data.

The `urllib.response.addinfourl` interface conforms to the one used by `urllib3.response.HTTPResponse` so I don't think we have any gaps here, longer term we probably want to figure out a more reliable shim. But as this branch is mostly used for language models we're probably okay.

> [!NOTE]
> I've tested this locally by creating a function that returns a `FileURL()` pointing to a remote address and the file is uploaded successfully.

### DIFF
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -1,4 +1,3 @@
-from functools import partial
 import base64
 import io
 import mimetypes
@@ -22,7 +21,7 @@ from fastapi.encoders import jsonable_encoder
 
 from .. import types
 from ..schema import PredictionResponse, Status, WebhookEvent
-from ..types import Path, URLFile
+from ..types import Path
 from .eventtypes import PredictionInput
 from .response_throttler import ResponseThrottler
 from .retry_transport import RetryTransport
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -1,3 +1,4 @@
+from functools import partial
 import base64
 import io
 import mimetypes
@@ -21,7 +22,7 @@ from fastapi.encoders import jsonable_encoder
 
 from .. import types
 from ..schema import PredictionResponse, Status, WebhookEvent
-from ..types import Path
+from ..types import Path, URLFile
 from .eventtypes import PredictionInput
 from .response_throttler import ResponseThrottler
 from .retry_transport import RetryTransport
@@ -126,14 +127,10 @@ class ChunkFileReader:
         self.fh = fh
 
     async def __aiter__(self) -> AsyncIterator[bytes]:
-        self.fh.seek(0)
-        while True:
-            chunk = self.fh.read(1024 * 1024)
-            if isinstance(chunk, str):
-                chunk = chunk.encode("utf-8")
-            if not chunk:
-                log.info("finished reading file")
-                break
+        if self.fh.seekable():
+            self.fh.seek(0)
+
+        for chunk in iter(self.fh):
             yield chunk
 
 
@@ -288,7 +285,10 @@ class ClientManager:
             with obj.open("rb") as f:
                 return await self.upload_file(f, url=url, prediction_id=prediction_id)
         if isinstance(obj, io.IOBase):
-            return await self.upload_file(obj, url=url, prediction_id=prediction_id)
+            try:
+                return await self.upload_file(obj, url=url, prediction_id=prediction_id)
+            finally:
+                obj.close()
         return obj
 
     # inputs
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -293,10 +293,8 @@ class ClientManager:
             with obj.open("rb") as f:
                 return await self.upload_file(f, url=url, prediction_id=prediction_id)
         if isinstance(obj, io.IOBase):
-            try:
+            with obj:
                 return await self.upload_file(obj, url=url, prediction_id=prediction_id)
-            finally:
-                obj.close()
         return obj
 
     # inputs
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -131,9 +131,14 @@ class ChunkFileReader:
 
         while True:
             chunk = self.fh.read(1024 * 1024)
+
+            if isinstance(chunk, str):
+                chunk = chunk.encode("utf-8")
+
             if not chunk:
                 log.info("finished reading file")
                 break
+
             yield chunk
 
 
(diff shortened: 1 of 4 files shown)


python/cog/server/clients.py
```diff
@@ -126,11 +127,11 @@ def __init__(self, fh: io.IOBase) -> None:
...
 
     async def __aiter__(self) -> AsyncIterator[bytes]:
-        self.fh.seek(0)
+        if self.fh.seekable():
+            self.fh.seek(0)
+
         while True:
             chunk = self.fh.read(1024 * 1024)
-            if isinstance(chunk, str):
-                chunk = chunk.encode("utf-8")
```
aron, line 133: I'm unclear on why we do this...

python/cog/types.py
```diff
@@ -242,19 +236,25 @@ def __delattr__(self, name: str) -> None:
...
-            object.__setattr__(self, "__target__", resp.raw)
-            return resp.raw
+
+            # We create a streaming response here, much like the `requests`
+            # version in the main 0.9.x branch. The only concerning bit here
+            # is that the book keeping for closing the response needs to be
+            # handled elsewhere. There's probably a better design for this
+            # in the long term.
+            res = urllib.request.urlopen(url)
+            object.__setattr__(self, "__target__", res)
```
aron, line 255: This feels so much simpler than the requests version. What am I missing, why would we use `requests` here?

aron (commented):
Updated the code to use `urllib` instead of `requests` or `httpx`. The resulting `urllib.response.addinfourl` is a file-like object and works well with the design of URLFile.

python/cog/server/clients.py
aron: Tests fail, so apparently sometimes the `fh.read()` can return strings.

python/cog/server/clients.py
```diff
@@ -288,7 +285,10 @@ async def upload_files(
             with obj.open("rb") as f:
                 return await self.upload_file(f, url=url, prediction_id=prediction_id)
         if isinstance(obj, io.IOBase):
-            return await self.upload_file(obj, url=url, prediction_id=prediction_id)
+            try:
+                return await self.upload_file(obj, url=url, prediction_id=prediction_id)
+            finally:
+                obj.close()
```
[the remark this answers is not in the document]
nickstenning: Is there some reason this doesn't use `with` like the clause above it?

python/cog/types.py
```diff
@@ -195,22 +194,17 @@ def unlink(self, missing_ok: bool = False) -> None:
...
     object that is created lazily. It's a file-like object constructed from a
     URL that can survive pickling/unpickling.
-
-    This is the only place Cog uses requests
     """
 
     __slots__ = ("__target__", "__url__")
 
     def __init__(self, url: str) -> None:
+        object.__setattr__(self, "name", os.path.basename(url))
```
nickstenning, line 207: Ditto my comment on the other PR. This isn't quite enough work to extract a sensible name from a URL.

python/cog/types.py
```diff
@@ -242,19 +237,25 @@ def __delattr__(self, name: str) -> None:
...
         except AttributeError:
             url = object.__getattribute__(self, "__url__")
-            resp = requests.get(url, stream=True)
-            resp.raise_for_status()
-            resp.raw.decode_content = True
-            object.__setattr__(self, "__target__", resp.raw)
-            return resp.raw
+
+            # We create a streaming response here, much like the `requests`
+            # version in the main 0.9.x branch. The only concerning bit here
```
nickstenning, line 251: I think that's probably fine, especially if we guarantee that we read it in a `with fp:` elsewhere in the code. This is a fundamental limitation of `urlopen`, I think.

python/cog/server/clients.py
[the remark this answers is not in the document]
aron: Oooh is that what `__wrapped__` does? So that would look like this?

```py
with obj:
  return await self.upload_file(obj, url=url, prediction_id=prediction_id)
```

Nice!

python/cog/types.py
```diff
@@ -242,19 +237,25 @@ def __delattr__(self, name: str) -> None:
...
+            # is that the book keeping for closing the response needs to be
+            # handled elsewhere. There's probably a better design for this
+            # in the long term.
+            res = urllib.request.urlopen(url)  # noqa: S310
```
nickstenning, line 255: I think we need to pre-validate that this is an `http` or `https` URL. (Do we also support `data` URLs?)

Either way, we shouldn't allow this:

```
In [1]: len(urllib.request.urlopen("file:///etc/shadow").read())
Out[1]: 9050
```

python/tests/server/test_clients.py
```diff
@@ -1,12 +1,17 @@
...
 import tempfile
+from urllib.response import addinfourl
+from unittest import mock
 
 import cog
+import httpx
 import pytest
 from cog.server.clients import ClientManager
 
+pytest.mark.asyncio
```
nickstenning, line 13: Guessing this is a typo?

nickstenning (commented):
This mostly looks good, but I do think we're potentially giving ourselves a security headache if we don't constrain URL schemes.

python/cog/types.py
nickstenning: Hmmm, actually, thinking about this some more -- does this actually work? This doesn't spin off a background process to download the file, it blocks until the file is downloaded and stores the whole thing in memory.

aron: Can we chat about this sync? Is requests threaded by default?

aron: We discussed this in-person and this interface will do what we want. It looks like we can actually get something similar using httpx as well:

```py
with httpx.stream("GET", "https://example.com/") as response:
    raw_stream = response.extensions["network_stream"]
    print(raw_stream.read(2))
# b'\x1f\x8b'
```

https://github.com/encode/httpx/discussions/2296#discussioncomment-6827850

But for the moment we'll go with this simpler implementation and re-factor later.

PR #1990, opened 2024-10-14
Choices for Str not just list

tempusfrangit:
In the case of the Str (which translates to a StrEnum for choices) the explicit list isinstance check is incorrect. There are many cases where it would be valid to send an `Iterable` such as with `dict.keys()` returning a `dict_keys`. As we do not want to explictly cast to `list` type this commit results in a check of iterable type instead of list.


technillogue (approved):
LGTM

PR #1992, opened 2024-10-15
Increase test timeouts

8W9aG:
* Increase test timeouts to 20 minutes to allow for new tests focused around timeliness concepts
such as waiting for events.

This is a subset of https://github.com/replicate/cog/pull/1957/


meatballhat (approved):
🆙🕑

PR #1993, opened 2024-10-15
Separate BasePredictor and BaseInput from predictor

8W9aG:
* Currently these two classes live in predictor
* Separate them out into their own classes to allow for easier reasoning about the predictor code.

This is a subset of https://github.com/replicate/cog/pull/1957/.


technillogue (approved):
this seems like a pretty simple change that should be safe

PR #1994, opened 2024-10-15
Deduplicate choices

tempusfrangit:
Ensure choices are not duplicated, as that makes StrEnum impossible to resolve. Each element should only exist once.


meatballhat (approved):

tempusfrangit:
Not exactly sure why python 3.12 fails with pydantic 1. 

Clearly something is super wonky since other python versions succeeded.

PR #1996, opened 2024-10-16
Support more cases in code_xforms

8W9aG:
* Support multiple class names and functions to handle extracting both prediction and training
functions.
* Handle globals referenced in function inputs.
* Handle subclasses referenced in the major class.

This is necessary so we can put the stripped code into environment variables and load the types necessary for the OpenAPI schema before the file system is created around cog.
This is a subset of https://github.com/replicate/cog/pull/1957


python/cog/code_xforms.py
meatballhat, line 1: nit: using variable name other than `globals` so as not to shadow the `globals()` builtin

meatballhat (approved):
🙌🏼

python/cog/code_xforms.py
8W9aG: Resolved in [b0ad548](https://github.com/replicate/cog/pull/1996/commits/b0ad548c8637c8a488d1d8558ef75abe3b91995f)

PR #1999, opened 2024-10-17
Use environment for PyPI releases

nickstenning:
This allows us to control which branches are allowed to release to PyPI, and we can configure PyPI trusted publishing to expect releases to come from a specific environment.


nickstenning:
Current configuration for `test-pypi` environment:
image


And for the `pypi` environment:
image

aron (approved):
TIL about test.pypi.org

PR #2000, opened 2024-10-17
[async] Propagate `URLFile.name` attribute across serialization

aron:
The commit 3805e2e introduced the `filename` keyword argument to the `URLFile` constructor. However we do not correctly propagate that value when the instance is pickled resulting in the `URLFile` that is passed to the upload handler missing that attribute.

This PR updates the code to stash the `name` when pickling and extract it again when unpickling. The `__getattr__` function then supports returning the underlying `name` value rather than proxying to the underlying request object.

I also ran into a small bug whereby the `__del__` method was triggering a network request because of some private attributes being accessed during teardown would trigger the `__wrapper__` code. I've overridden the super class to disable this. Though I'm unclear if this is just the test suite doing this cleanup.

### DIFF
--- a/python/cog/types.py
+++ b/python/cog/types.py
@@ -201,7 +201,7 @@ class URLFile(io.IOBase):
     URL that can survive pickling/unpickling.
     """
 
-    __slots__ = ("__target__", "__url__")
+    __slots__ = ("__target__", "__url__", "name")
 
     def __init__(self, url: str, filename: Optional[str] = None) -> None:
         parsed = urllib.parse.urlparse(url)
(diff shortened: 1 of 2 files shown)


python/cog/types.py
```diff
@@ -218,13 +218,27 @@ def __init__(self, url: str, filename: Optional[str] = None) -> None:
...
 
+    def __del__(self) -> None:
+        try:
+            object.__getattribute__(self, "__target__")
+        except AttributeError:
+            # Do nothing when tearing down the object if the response object
+            # hasn't been created yet.
+            return
+
+        super().__del__()
```
aron, line 229: This showed up when I removed the requests mocking from the tests. It turns out that by default instances of `io.BaseIO` do some teardown (I assume when under test) that then results in trying to access various attributes which in turn triggers a request.

This change prevents that teardown unless we've actually made a request.

nickstenning (approved):
LGTM. Please do follow up and get support for `name` (and this fix) into `main` 🙇‍♂️

PR #2001, opened 2024-10-17
Fix cog not installing itself over present installation

8W9aG:
* Create an integration test to work against that performs a `pip freeze` on the resulting image to verify the install contents.
* Remove the `-t /dep` arguments when performing a `pip install` of the latest cog.

This is odd because `cog debug` presents the install step from docker very clearly:
```
COPY .cog/tmp/build20241017221314.7093841448192741/cog-0.11.1-py3-none-any.whl /tmp/cog-0.11.1-py3-none-any.whl
ENV CFLAGS="-O3 -funroll-loops -fno-strict-aliasing -flto -S"
RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir -t /dep /tmp/cog-0.11.1-py3-none-any.whl 'pydantic<2'
```

Indeed when running pip confirms the installation to the latest version:
```
#9 9.294 Successfully installed PyYAML-6.0.2 anyio-4.6.2.post1 attrs-23.2.0 certifi-2024.8.30 charset-normalizer-3.4.0 click-8.1.7 cog-0.11.1 fastapi-0.110.3 h11-0.14.0 httptools-0.6.4 idna-3.10 pydantic-1.10.18 python-dotenv-1.0.1 requests-2.32.3 sniffio-1.3.1 starlette-0.37.2 structlog-24.4.0 typing-extensions-4.12.2 urllib3-2.2.3 uvicorn-0.32.0 uvloop-0.21.0 watchfiles-0.24.0 websockets-13.1
```

However when `pip freeze` is run it references the old cog.

This is due to the install cog installing the wheel directly into `/dep`, in normal installs this is the then copied over into the `site-packages` of the pyenv shim at the very end, however in a base-image build this step is never performed, and since the pyenv aparatus is already set up a regular pip install will suffice without this tag, so detect if the build is a base image build and add the `-t /dep` flags as appropriate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,8 +65,8 @@ test-go: $(COG_EMBEDDED_WHEEL)
 	$(GO) run gotest.tools/gotestsum -- -timeout 1200s -parallel 5 ./... $(ARGS)
 
 .PHONY: test-integration
-test-integration: $(COG_BINARIES)
-	PATH="$(PWD):$(PATH)" $(TOX) -e integration
+test-integration: $(COG_BINARIES) $(COG_WHEEL)
+	PATH="$(PWD):$(PATH)" $(TOX) run --installpkg $(COG_WHEEL) -e integration
 
 .PHONY: test-python
 test-python: $(COG_WHEEL)
--- a/Makefile
+++ b/Makefile
@@ -65,8 +65,8 @@ test-go: $(COG_EMBEDDED_WHEEL)
 	$(GO) run gotest.tools/gotestsum -- -timeout 1200s -parallel 5 ./... $(ARGS)
 
 .PHONY: test-integration
-test-integration: $(COG_BINARIES) $(COG_WHEEL)
-	PATH="$(PWD):$(PATH)" $(TOX) run --installpkg $(COG_WHEEL) -e integration
+test-integration: $(COG_BINARIES)
+	PATH="$(PWD):$(PATH)" $(TOX) -e integration
 
 .PHONY: test-python
 test-python: $(COG_WHEEL)
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -426,9 +426,14 @@ func (g *Generator) installCog() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	pipInstallLine := "RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir "
+	if !g.IsUsingCogBaseImage() {
+		pipInstallLine += "-t /dep "
+	}
+	pipInstallLine += containerPath
 	// Install pydantic<2 for now, installing pydantic>2 wouldn't allow a downgrade later,
 	// but upgrading works fine
-	pipInstallLine := fmt.Sprintf("RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir -t /dep %s 'pydantic<2'", containerPath)
+	pipInstallLine += " 'pydantic<2'"
 	if g.strip {
 		pipInstallLine += " && " + StripDebugSymbolsCommand
 	}
--- a/pkg/dockerfile/generator.go
+++ b/pkg/dockerfile/generator.go
@@ -426,11 +426,11 @@ func (g *Generator) installCog() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	pipInstallLine := "RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir "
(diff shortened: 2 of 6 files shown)


test-integration/test_integration/test_build.py
```diff
@@ -315,3 +316,31 @@ def test_precompile(docker_image):
...
+    )
+    assert build_process.returncode == 0
+    pip_freeze_process = subprocess.run(
+        ["docker", "run", "-t", docker_image, "sh", "-c", "pip freeze"],
+        cwd=project_dir,
+        capture_output=True,
+    )
+    assert pip_freeze_process.returncode == 0
+    pip_freeze_stderr = pip_freeze_process.stdout.decode()
+    assert "cog @ file:///tmp/cog-0.1.dev" not in pip_freeze_stderr
```
nickstenning, line 346: Can we have a positive assertion here that the version reported is the version we're currently running, rather than a negative assertion specific to this particular variant of the bug?

pkg/dockerfile/generator.go
```diff
@@ -426,9 +426,14 @@ func (g *Generator) installCog() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	pipInstallLine := "RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir "
+	if !g.IsUsingCogBaseImage() {
+		pipInstallLine += "-t /dep "
+	}
+	pipInstallLine += containerPath
```
erbridge, line 433: For consistency, it would be nicer if we added a prefixing space instead of a suffix. That makes it easier to rearrange lines in the future.

```suggestion
	pipInstallLine := "RUN --mount=type=cache,target=/root/.cache/pip pip install --no-cache-dir"
	if !g.IsUsingCogBaseImage() {
		pipInstallLine += " -t /dep"
	}
	pipInstallLine += " " + containerPath
```

Makefile
```diff
@@ -65,8 +65,8 @@ test-go: $(COG_EMBEDDED_WHEEL)
 	$(GO) run gotest.tools/gotestsum -- -timeout 1200s -parallel 5 ./... $(ARGS)
 
 .PHONY: test-integration
-test-integration: $(COG_BINARIES)
-	PATH="$(PWD):$(PATH)" $(TOX) -e integration
+test-integration: $(COG_BINARIES) $(COG_WHEEL)
+	PATH="$(PWD):$(PATH)" $(TOX) run --installpkg $(COG_WHEEL) -e integration
```
nickstenning, line 69: Hmmm, I don't think we should be installing the built cog into the tox environment here. The integration tests are supposed to be testing `cog` the binary, not the python package.

nickstenning: I know _why_ you've done this... I just don't think it's a particularly pleasant solution to the problem.

A couple of alternatives to consider:

- compare the output of `cog --version` and `python -c "import cog; print(cog.__version__)"` from inside the built image -- they will not be identical except for release builds, so some string futzing (specifically, removing the local scheme parts of the PEP440 version) will be needed
- allow `cog --version` to report both the cog version and the version of the embedded wheel, which is contained in the file name and can be reliably parsed out

8W9aG: What would you suggest to check against the cog version?

pkg/dockerfile/generator_test.go
```diff
@@ -37,16 +37,20 @@ func getWheelName() string {
...
 }
 
-func testInstallCog(relativeTmpDir string, stripped bool) string {
+func testInstallCog(relativeTmpDir string, stripped bool, baseImage bool) string {
 	wheel := getWheelName()
 	strippedCall := ""
 	if stripped {
 		strippedCall += " && find / -type f -name \"*python*.so\" -not -name \"*cpython*.so\" -exec strip -S {} \\;"
 	}
+	baseImageArgs := " -t /dep"
```
nickstenning, line 46: What's the scenario in which `-t /dep` is needed?

nickstenning (approved):
Thank you! I've verified locally that this seems to fix the issue. I think we can probably do some smarter things with the version parsing in the integration test but this will do for now!

pkg/dockerfile/generator_test.go
8W9aG: There is a function called [copyPipPackagesFromInstallStage](https://github.com/replicate/cog/blob/main/pkg/dockerfile/generator.go#L535) which is executed at the end of an install that seems to copy a bunch of packages from `/dep` directly into the pyenv site-packages. It's unclear to me why this is done like this, I suspect it might be due to wanting to put these deps into both the system python and the pyenv python.

nickstenning: Hmm. That sounds like something we should probably understand and, if possible, simplify.

PR #2009, opened 2024-10-20
Support async predictors rebased

meatballhat:
⚠️ This is intentionally a draft rebase of #1896 for review and fixing tests prior to replacing the history of #1896.


meatballhat:
Not bothering with rebasing and instead moving forward in #2010

PR #2011, opened 2024-10-21
Provide User-Agent and Accept headers in URLFile requests

aron:
It turns out some providers require these headers to be present
otherwise they'll give an error response. This issue is not present
in the requests version of the codebase because the requests library
provides default headers on our behalf.


aron:
Gonna merge this as it's a minor change. I'm happy to do cleanup as a follow-up after audit.

PR #2012, opened 2024-10-21
Fix no COG_WHEEL in Make

8W9aG:
* make test currently fails if COG_WHEEL is not previously defined. This is the case when performing `make test` on a freshly cloned instance.
* Define COG_WHEEL if it doesn’t exist by finding the wheel file in pkg/dockerfile/embed


nickstenning:
Ooops! Sorry for missing this and thank you for fixing!

PR #2013, opened 2024-10-22
Make cog train call trainings endpoint

8W9aG:
* Fixes `/trainings` to use the proper `TrainingResponse` type as an input to perform proper deserialisation.
* Makes the `cog train` CLI call execute the `/trainings` endpoint.

This is a subset of https://github.com/replicate/cog/pull/1957

This is mostly tested by [test_train_takes_input_and_produces_weights](https://github.com/replicate/cog/blob/main/test-integration/test_integration/test_train.py#L7), however what this PR will do is allow this testing to occur over the `/trainings` endpoint to give it more coverage.


pkg/predict/predictor.go
```diff
@@ -197,30 +198,49 @@ func (p *Predictor) GetSchema() (*openapi3.T, error) {
...
 	}
 
+	command := "predict"
+	if p.isTrain {
+		command = "train"
+	}
+
 	return fmt.Errorf(
-		`The inputs you passed to cog predict could not be validated:
+		`The inputs you passed to cog %s could not be validated:
```
meatballhat, line 230: nit: using the `%[1]s` [explicit argument index syntax](https://pkg.go.dev/fmt#hdr-Explicit_argument_indexes) would cut down on repeated positional args and the likelihood of confusion that comes with them

8W9aG: Resolved in [0050e8e](https://github.com/replicate/cog/pull/2013/commits/0050e8e6ab2b926ef0a8860e71e675a9d4de4ef6)

Issue #1341, opened 2023-10-18
Introducing more specific input and output types for images, audio, video, etc


zeke:
The [`cog.Path`](https://github.com/replicate/cog/blob/main/docs/python.md#path) object is used to get files in and out of models. It represents a path to a file on disk. `Path` is used for all files, regardless of whether they're text files, zip files, videos, images, audio files, etc.

When looking at the schema for a model, it's not easy to tell what type of file is expected:

```
$ curl -s -H "Authorization: Token $REPLICATE_API_TOKEN" \
  https://api.replicate.com/v1/models/stability-ai/sdxl | jq ".latest_version.openapi_schema.components.schemas.Input.properties.mask"
```

SDXL's `mask` input expects an image file, but that's not clear from the schema. Unless the model author writes a description that says what kind of file is expected, users of the model can't reliably know what's expected:

```json
{
  "type": "string",
  "title": "Mask",
  "format": "uri",
  "x-order": 3,
  "description": "Input mask for inpaint mode. Black areas will be preserved, white areas will be inpainted."
}
```

## Being explicit about file types

What if, [instead of defining the mask in the predictor as a `Path`](https://github.com/replicate/cog-sdxl/blob/f64d7cefc9f00b64b83df306db62c6a48a1a3678/predict.py#L269-L272), it could be an `ImagePath`, which would really just be a `Path` under the hood with some extra constraints?

```python
from cog import BasePredictor, Input, ImagePath

class Predictor(BasePredictor):
    def predict(
        self,
        mask: ImagePath = Input(
            description="Input mask for inpaint mode. Black areas will be preserved, white areas will be inpainted.",
            default=None,
        )
    )
```

Alternatively, may `Input` could be extended to support some kind of property like `supportedFormats` which would allow model authors to specify the file types or MIME types supported by their inputs.

These are two naive suggestions about how to approach this, but I'm very open to other ideas that address the issue.

zeke:
Maybe it could be a property of the existing `Path`, like `mimetypes` or something.

zeke:
Related: https://github.com/replicate/cog/pull/2014

markphelps:
Closing as stale since there has been no activity for over 18 months. If this schema improvement is still needed, please open a new issue referencing this discussion and the related PR #2014.

PR #2014, opened 2024-10-22
Add an image mask file type

bfirsh:
Just wanted to demonstrate what it might be like to define richer file types in Cog that would affect the user interface.


zeke:
Related: https://github.com/replicate/cog/issues/1341

PR #2020, opened 2024-10-24
Tag worker events with prediction

meatballhat:


python/cog/server/worker.py
```diff
@@ -248,25 +281,24 @@ def _consume_events_inner(self) -> None:
...
+                        "publish failed",
+                        subscriber=subscriber,
+                        tag=e.tag,
+                        event=e.event,
+                        exc_info=True,
+                    )
+            else:
+                log.warn(
+                    f"skipping publish for {e.event}, because {requested_tag} is None or {e.tag} != {requested_tag}"
+                )
```
evilstreak, line 301:
```suggestion
            else:
                log.warn(
                    f"skipping publish for {e.event}, because {requested_tag} is None or {e.tag} != {requested_tag}"
                )
```

This looks like leftover debugging we probably don't want to ship.

evilstreak (approved):
One chunk of leftover debug code, but otherwise looks good ✨

PR #2021, opened 2024-10-24
add an OutputURL type with a streaming download/upload

technillogue:
https://github.com/replicate/cog/pull/1987 and followups aim to improve returning remote URLs. however, they use sync downloads, which can cause coroutines to queue up in the event loop under higher levels of concurrency. it also doesn't use a connection pool and can't overlap download+upload. these problems can be avoided by adding a new output type that's handled differently by ClientManager.upload_file. conveniently, ClientManager already holds a file_client with a connection pool suitable for downloading those images well, and httpx provides good ergonomics for streaming downloads/uploads.

### DIFF
--- a/python/cog/server/clients.py
+++ b/python/cog/server/clients.py
@@ -11,6 +11,7 @@ from typing import (
     Dict,
     Mapping,
     Optional,
+    Union,
     cast,
 )
 from urllib.parse import urlparse
@@ -21,7 +22,7 @@ from fastapi.encoders import jsonable_encoder
 
 from .. import types
 from ..schema import PredictionResponse, Status, WebhookEvent
-from ..types import Path
+from ..types import OutputURL, Path
 from .eventtypes import PredictionInput
 from .response_throttler import ResponseThrottler
 from .retry_transport import RetryTransport
@@ -199,7 +200,11 @@ class ClientManager:
     # files
 
     async def upload_file(
-        self, fh: io.IOBase, *, url: Optional[str], prediction_id: Optional[str]
+        self,
+        fh: Union[io.IOBase, OutputURL],
+        *,
+        url: Optional[str],
+        prediction_id: Optional[str],
     ) -> str:
         """put file to signed endpoint"""
         log.debug("upload_file")
@@ -213,7 +218,7 @@ class ClientManager:
 
         # this code path happens when running outside replicate without upload-url
         # in that case we need to return data uris
-        if url is None:
+        if url is None and not isinstance(fh, OutputURL):
             return file_to_data_uri(fh, content_type)
         assert url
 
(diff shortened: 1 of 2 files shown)


python/cog/server/clients.py
```diff
@@ -243,11 +248,14 @@ async def upload_file(
...
-            headers=headers,
-        )
+        if isinstance(fh, OutputURL):
+            async with self.file_client.stream("GET", fh.url) as resp:
+                content = resp.aiter_bytes()
+                resp = await self.file_client.put(url, content=content, headers=headers)
+        else:
+            resp = await self.file_client.put(
+                url, content=ChunkFileReader(fh), headers=headers
+            )
```
aron, line 258: Thanks for pulling this together. I'm trying to understand this, I'm not very familiar with how `async/await` works in Python. I understand by going with the `httpx` async client + iterator we get a non-blocking implementation but how does the existing `ChunkFileReader` accomplish the same thing?

If it doesn't and we still need to solve that issue we should probably figure out an interface that works for both.

python/cog/server/clients.py
```diff
@@ -213,7 +218,7 @@ async def upload_file(
 
         # this code path happens when running outside replicate without upload-url
         # in that case we need to return data uris
-        if url is None:
+        if url is None and not isinstance(fh, OutputURL):
             return file_to_data_uri(fh, content_type)
```
aron, line 222: I think this currently breaks the local development flow, it'll try and upload the `OutputURL` without a `url`... We need to return the URL as-is I think.

```suggestion
        if url is None:
            if isinstance(fh, OutputURL):
              return fh.url
            return file_to_data_uri(fh, content_type)
```

aron (commented):
I think conceptually this is a clean solution.

Though it does expand the surface area of Cog (This is a concern I had about repurposing `URLFile` too. But it already existed in a sort of clunky way).

We tried to extend the `io.IOBase` approach to keep the interface consistent, but if I understand correctly we need to rework that for async anyway? So perhaps we need an entirely new `File` and `Path` implementation?

I think it'd be worth getting a second review from someone else familar with cog.

python/cog/server/clients.py
technillogue: The PUT method takes bytes or an async iterator that returns bytes. aiter_bytes counts, as does ChunkFileReader, which is implemented earlier in the same file. ChunkFileReader does do blocking disk reads, but doing so 1MB at a time is likely short enough that we can do all the other networking we need in between.

if you wanted to be fancy you could have a single FileIterator that could take a local or remote URI, but that's kind of annoying to do while holding the context manager for the download request, and this approach is kind of simpler

aron:
Closing this while we figure out what the underlying need is. In the mean time model authors that require this functionality are recommended to use the `Path` type.

PR #2022, opened 2024-10-24
Revert "Support async predictors (#2010)"

meatballhat:
This reverts commit a86adcda6d511723bbe55f06555ffce8095fd6ac.

This is being done primarily to tag a ~`v0.11.6`~ ~`v0.12.0`~ `v0.11.6` release that does not contain the predict-time timeout issue seen with some models running `v0.11.5`.

After this is merged and ~`v0.11.6`~ ~`v0.12.0`~ `v0.11.6` is released, PLAT-410 may be considered done.


nickstenning:
If you'd like, I think we could probably pair on this and get a fixed (rather than reverted) version out.

meatballhat:
@nickstenning I'd much prefer that, yes!

meatballhat:
Update: we _do_ have a pretty good idea of how to not have to revert this, but reverting and tagging the next release first is the right call.

8W9aG (approved):

PR #2025, opened 2024-10-25
Support async predictors (✨ again ✨)

meatballhat:
Reverts #2022 to restore #2010 

plus:
- use async stream redirector in setup, which means writes from C will **not** be captured
- switch to sync stream redirector when not running in async mode
- adjusting accordingly for tests

Closes PLAT-373

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -292,14 +292,13 @@ class _ChildWorker(_spawn.Process):  # type: ignore
         # Initially, we ignore SIGUSR1.
         signal.signal(signal.SIGUSR1, signal.SIG_IGN)
 
-        redirector = StreamRedirector(
+        async_redirector = AsyncStreamRedirector(
             callback=self._stream_write_hook,
             tee=self._tee_output,
         )
 
-        # TODO: support async setup? see where `redirector` is redefined below if the predict is async
-        with redirector:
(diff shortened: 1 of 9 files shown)


python/cog/server/worker.py
```diff
@@ -292,41 +292,38 @@ def run(self) -> None:
...
-        redirector = StreamRedirector(
+        async_redirector = AsyncStreamRedirector(
             callback=self._stream_write_hook,
             tee=self._tee_output,
         )
 
-        # TODO: support async setup? see where `redirector` is redefined below if the predict is async
-        with redirector:
-            self._setup(redirector)
+        with async_redirector:
```
erbridge, line 300: Do we not need this to be conditional on the same conditions as predict?

meatballhat: It _was_, but this is intentionally different now since the async redirector will work for sync and async predict and _does not_ attempt to do low-level things with file descriptors. This means that setup will not capture writes from outside Python, which is judged (by me and @nickstenning) to be an acceptable trade-off.

erbridge: Ah. I see the problem. We might not have a predictor until after setup, so we're not able to work out if we need async or not before we run it.

meatballhat: Yep that!

PR #2027, opened 2024-10-28
Add integration tests around stream redirection

meatballhat:
to ensure we don't have regressions like that seen in `v0.11.4`.

The tests here are somewhat contrived, but based on how some known models work that failed when stream redirection didn't work as expected, and in fact they all fail when run against `v0.11.4`.

Connected to PLAT-410

### DIFF
--- a/test-integration/test_integration/fixtures/setup-subprocess-double-fork-http-project/cog.yaml
+++ b/test-integration/test_integration/fixtures/setup-subprocess-double-fork-http-project/cog.yaml
@@ -1,0 +1,5 @@
+build:
+  python_version: "3.8"
+  python_packages:
+  - requests
+predict: "predict.py:Predictor"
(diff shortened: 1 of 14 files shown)


8W9aG (approved):
Really amazing work dan, very impressive, only suggestion I might have would be to add a few comments about the issues the test itself is trying to fix, in the sense that looking at the test its not clear to me why it should ever fail, but thats just a nice to have

meatballhat:
> Really amazing work dan, very impressive, only suggestion I might have would be to add a few comments about the issues the test itself is trying to fix, in the sense that looking at the test its not clear to me why it should ever fail, but thats just a nice to have

Thank you, @8W9aG! I'll get some comments in place 👍🏼

8W9aG (approved):

PR #2028, opened 2024-10-28
Drop support for Python 3.7

meatballhat:
because python 3.7 has been EOL for ~1 year and we cannot continue to support it within Cog.

Closes PLAT-429

### DIFF
--- a/tools/compatgen/internal/torch.go
+++ b/tools/compatgen/internal/torch.go
@@ -209,7 +209,7 @@ func parseTorchInstallString(s string, defaultVersions map[string]string, cuda *
 	torchaudio := libVersions["torchaudio"]
 
 	// TODO: this could be determined from https://download.pytorch.org/whl/torch/
-	pythons := []string{"3.7", "3.8", "3.9", "3.10", "3.11"}
+	pythons := []string{"3.8", "3.9", "3.10", "3.11", "3.12"}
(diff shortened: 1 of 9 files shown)


tools/compatgen/internal/torch.go
```diff
@@ -209,7 +209,7 @@ func parseTorchInstallString(s string, defaultVersions map[string]string, cuda *
 	torchaudio := libVersions["torchaudio"]
 
 	// TODO: this could be determined from https://download.pytorch.org/whl/torch/
-	pythons := []string{"3.7", "3.8", "3.9", "3.10", "3.11"}
+	pythons := []string{"3.8", "3.9", "3.10", "3.11", "3.12"}
```
erbridge, line 212:
```suggestion
	pythons := []string{"3.8", "3.9", "3.10", "3.11", "3.12", "3.13"}
```

No?

meatballhat: I was going to save it for a follow up. WDYT?

erbridge: Sure. We added `3.12` here, though, so maybe either do both or neither?

meatballhat: Ah yes sorry! I should have called that out.

PR #2030, opened 2024-10-29
Exclude invalid output paths from infrastructure errors

meatballhat:
Closes PLAT-380

### DIFF
--- a/python/cog/server/runner.py
+++ b/python/cog/server/runner.py
@@ -408,10 +408,17 @@ class PredictTask(Task[schema.PredictionResponse]):
         try:
             # TODO: clean up output files
             return self._file_uploader(output)
+        except (FileNotFoundError, NotADirectoryError):
+            # These error case indicates that an output path returned by a prediction does
+            # not actually exist, so there is no way for us to even attempt to upload it.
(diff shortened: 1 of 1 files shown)


python/cog/server/runner.py
```diff
@@ -408,6 +408,10 @@ def _upload_files(self, output: Any) -> Any:
         try:
             # TODO: clean up output files
             return self._file_uploader(output)
+        except (FileNotFoundError, NotADirectoryError):
+            # Errors that indicate an output path does not exist are not considered
+            # infrastructure errors.
```
erbridge, line 413: Can we clarify what an "infrastructure error" is and why doing this makes it not one, please?

python/cog/server/runner.py
```diff
@@ -408,10 +408,17 @@ def _upload_files(self, output: Any) -> Any:
         try:
             # TODO: clean up output files
             return self._file_uploader(output)
+        except (FileNotFoundError, NotADirectoryError):
+            # These error case indicates that an output path returned by a prediction does
```
erbridge, line 412:
```suggestion
            # These error cases indicate that an output path returned by a prediction does
```

erbridge (approved):

PR #2034, opened 2024-10-29
Add python3.13 to package classifiers and test matrix

meatballhat:
including manual patches to CUDA base images given that the CuDNN version is no longer present in tags.


nickstenning (approved):

erbridge:
This was done by #2103

PR #2040, opened 2024-10-31
Add serve command

meatballhat:
to run an HTTP server command very similar to what is used in Replicate production. By making this command more within reach, my hope is that not only is it easier and more clear to run Cog in its HTTP server mode, but also that more Cog users will build familiarity with and opinions about the HTTP server interface.

Connected to PLAT-259

### DIFF
--- a/pkg/cli/root.go
+++ b/pkg/cli/root.go
@@ -45,8 +45,8 @@ https://github.com/replicate/cog`,
 		newPredictCommand(),
 		newPushCommand(),
 		newRunCommand(),
-		newTrainCommand(),
 		newServeCommand(),
+		newTrainCommand(),
 	)
 
 	return &rootCmd, nil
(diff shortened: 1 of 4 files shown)


pkg/cli/root.go
```diff
@@ -46,6 +46,7 @@ https://github.com/replicate/cog`,
 		newPushCommand(),
 		newRunCommand(),
 		newTrainCommand(),
+		newServeCommand(),
```
erbridge, line 49:
```suggestion
		newServeCommand(),
		newTrainCommand(),
```

We must serve the Great Alphabet 🙇

erbridge (approved):
This looks like a good addition to me.

nickstenning (approved):
I like it!

Let's run this past @bfirsh and/or @superhighfives as a quick reasonableness check? Otherwise LGTM.

superhighfives:
Agreed that this makes sense—I think my only request would be to add the details to the `README`.

bfirsh:
Love it!

PR #2042, opened 2024-11-01
Add Config class

8W9aG:
* Makes config a first class citizen
* All config variables are accessed through the class properties and resolved at runtime
* Cog Modes are now an enum
* Input/Output Types are managed by the config class

This PR is a subset of https://github.com/replicate/cog/pull/1957/


meatballhat (approved):
👏🏼 Y E S 🎉

PR #2046, opened 2024-11-06
Add integration test for multiprocessing usage

meatballhat:
in which the setup spawns a child process that accepts a multiprocessing Connection and Lock which are used in each prediction.

The intent of this test is to ensure Cog plays nicely with any predictor that uses these features of multiprocessing, especially in combination with a Path output, since these have been common elements of known failures.


test-integration/test_integration/fixtures/setup-subprocess-multiprocessing-project/bg.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def ponger(
+    conn: multiprocessing.connection.Connection, lock: multiprocessing.synchronize.Lock
+):
+    for i in range(100):
+        print(f"Getting ready for some serious ponginggg ({i+1}%)")
+        time.sleep(0.001 + (0.001 * (i + 1)))
+
+    print("ITS PONGIN TIME")
```
nickstenning, line 14: 😁

nickstenning (approved):

PR #2047, opened 2024-11-06
Inject an internal `_emit_metric` method into predictor classes

erbridge:
We use metrics internally to transport metadata about how the model ran out of the model. It's private because it's entirely invisible within our system and we make no claims of stability for how it works.


python/cog/base_predictor.py
```diff
@@ -10,6 +10,14 @@
 
 
 class BasePredictor(ABC):
+    def __init__(
+        self,
+        *,
+        _emit_metric: Callable[[str, Union[float, int]], None],
+    ) -> None:
+        # For internal use only.
+        self._emit_metric = _emit_metric
```
nickstenning, line 19: Should there be a no-op default?

nickstenning: Oh I see, that's handled later on.

python/cog/predictor.py
```diff
@@ -163,19 +163,38 @@ def load_slim_predictor_from_file(
...
     if inspect.isclass(predictor):
-        return predictor()
+        if emit_metric is None:
+
+            def noop_emit_metric(name: str, value: Union[float, int]) -> None:  # pylint: disable=unused-argument
+                pass
+
+            emit_metric = noop_emit_metric
+
+        try:
```
nickstenning, line 182: This shenanigans feels a little risky, given that it's interfering directly with the user's code via `BasePredictor`.

I know generally thread-local state bad, etc., but in this case I think I actually marginally prefer the approach taken by the `async` branch in which `emit_metric` is just a function you import from somewhere else in Cog, and the details of how predictions are matched up to the call are up to us.

nickstenning: Maybe it's helpful to be a bit concrete about how that would work. I would imagine that `emit_metric` is a module function which uses a `worker` ContextVar which holds a [`weakref`](https://docs.python.org/3/library/weakref.html) (probably a `weakref.proxy`) to the current child worker. 

Then `emit_metric` would call `worker_proxy.emit_metric(m)` and handle `ReferenceError` -- probably logging a warning if it happens.

I _think_ that this would generalise to the future with concurrent/async predictions, because `worker.emit_metric` could use its own `ContextVar` to track which prediction called it.

nickstenning (commented):
I think given all the trouble we have when touching user's code, I'd be more comfortable with an approach that doesn't touch `BasePredictor` (see my comments)

As to whether it's `emit_metric` or `_emit_metric` for now, I see @aron's point, but I think it's fine for us to unblock ourselves by making it private/experimental for now. We can always promote it to public API in future, and I think it's safer for us to get some data from production about whether this approach works well before we make it part of the public API of Cog.

python/cog/server/scope.py
```diff
@@ -0,0 +1,31 @@
...
+class Scope:
+    def __init__(
+        self,
+        *,
+        record_metric: Callable[[str, Union[float, int]], None],
+    ) -> None:
+        self.record_metric = record_metric
+
+
+SCOPE: ContextVar[Optional[Scope]] = ContextVar("scope", default=None)
```
nickstenning, line 15: Very much a nit, but I'd call this `_current_scope` or similar. It's not a "constant".

nickstenning (approved):
I like it! What do you make of it?

python/cog/server/worker.py
```diff
@@ -448,6 +453,9 @@ async def _apredict(
                     self._events.send(PredictionOutputType(multi=False))
                     self._events.send(PredictionOutput(payload=make_encodeable(output)))
 
+    def _loop_scope(self) -> Scope:
```
aron, line 456: Not a big deal but why is this called loop_scope? Isn't it a prediction scope?

aron (approved):
Nice, I like it.

python/cog/server/worker.py
erbridge: It's a scope to be used inside the loop. At least that was my reasoning :)

PR #2060, opened 2024-11-17
Pass environment variables from cog.yaml to build stage

okapusta:
Allow to pass environment variables to docker file template during build


mcam10:
This would be good to have... IMO..

bfirsh:
Thank you for the contribution! This was added in https://github.com/replicate/cog/pull/2274 at the top level, then https://github.com/replicate/cog/pull/2807 moves it into the `build:` section.

PR #2064, opened 2024-11-20
Bug Fix: Set timeout on requests.get

tempusfrangit:
If requests.get has no timeout it is possible to hang indefinitely on a ill behaved webserver. This timeout is explicitly a connect *and* read timeout, meaning if no bytes received.


philandstuff (approved):
I amended the commit to have timeouts I thought appropriate, with reasons in the commit message

Issue #2026, opened 2024-10-26
Cog MusicGen Fine Tune “Melody” Model Train Upload Error


dorukhanyaldiz:
When I train the model using Replicate through the URL https://replicate.com/sakemin/musicgen-fine-tuner/train, the training process completes without issues when selecting either the “melody” or “stereo melody” versions. However, I encounter the following error:

```
Training failed.
Cog: Got error trying to upload output files
```

I can’t figure out the reason for this error. Since the “melody” and “stereo melody” models have been trained before, I believe I should be able to train them as well.

Could you please assist with identifying the root cause of this issue?

aron:
Thanks for reporting your issue. This isn't a bug with cog but an issue with the server configuration on Replicate's size when trying to upload the very large weights produced by this model. We're working on a fix and I'll update you when it has landed.

dorukhanyaldiz:
Hi Aron,

Thank you for your response. Is there any possibility to upload the model
somewhere else?

Otherwise, Replicate might unfortunately no longer be a suitable platform
for my needs.

Best regards,
Dorukhan

Aron Carroll ***@***.***>, 28 Eki 2024 Pzt, 15:17 tarihinde
şunu yazdı:

PR #2068, opened 2024-11-27
Add deprecated `emit_metric()` helper to cog interface

aron:
This helps with the transition between the main cog branch and the experimental `async` branch. Models built for the `async` branch can be run on either without code changes.

New models should still be using the internal `current_scope().record_metric()` interface.

### DIFF
--- a/python/cog/__init__.py
+++ b/python/cog/__init__.py
@@ -4,7 +4,7 @@ from pydantic import BaseModel
 
 from .base_predictor import BasePredictor
 from .mimetypes_ext import install_mime_extensions
-from .server.scope import current_scope
+from .server.scope import current_scope, emit_metric
 from .types import (
     ConcatenateIterator,
     ExperimentalFeatureWarning,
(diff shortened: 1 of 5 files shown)


philandstuff:
This looks good. However it will conflict with #2066 because I've updated tests that work with async models to specify a minimum python version (see for ex https://github.com/replicate/cog/pull/2066/files#diff-922f17529df722a8c4e93788b47410def10f269ce3cd5cff2f768feb2f066336R80). I think the easiest thing would be to change the base of this PR to `support-concurrent-predictions-in-child`, then get the tests passing (probably just by specifying minimum python version). Then we can either merge this into the support-concurrent-predictions-in-child branch before merging #2066, or merge #2066 first and then this.

aron:
Perfect, I'll cherry-pick and fix the tests in #2066

aron:
@philandstuff I'm going to merge this into main, then clean-up #2026 ready to be merged.

PR #2069, opened 2024-11-28
Use a single stream redirector for both setup and predict

aron:
Currently for models that provide a non-async predict function we have two stream redirectors wrapping stdout/stderr.

We think this entering/exiting the stream redirector context multiple times may have been causing unusual bugs, particularly with models that use libraries like `tqdm` to render progress bars and other manipulation.

This PR changes this behaviour to use either the `StreamRedirector` or (newly renamed) `SimpleStreamRedirector` based on whether the predict is defined with an async function or not. We now use the same redirector for module loading, setup and prediction.

The interrogation of the predict function now happens higher up the stack in http.py when we read the input/output types using `config.get_predictor_types`. This now returns an additional value `is_async` which is `True` when an async function is defined.

This is then passed down the stack into the worker which uses it to select an appropriate stream redirector instance. In future when we come to support async setup functions we may need to revisit whether `is_async` is the correct term for the worker argument, `use_simple_stream` or similar might be more appropriate.

This has been tested with both sync and async models including flux-dev and seems to be working correctly.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -50,7 +50,7 @@ from .exceptions import (
     FatalWorkerException,
     InvalidStateException,
 )
-from .helpers import AsyncStreamRedirector, StreamRedirector
+from .helpers import SimpleStreamRedirector, StreamRedirector
 from .scope import Scope, scope
 
 if PYDANTIC_V2:
@@ -377,7 +377,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
         signal.signal(signal.SIGUSR1, signal.SIG_IGN)
 
         if self._is_async:
-            redirector = AsyncStreamRedirector(
+            redirector = SimpleStreamRedirector(
                 callback=self._stream_write_hook,
                 tee=self._tee_output,
             )
@@ -397,7 +397,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
 
             predict = get_predict(self._predictor)
             if self._is_async:
-                assert isinstance(redirector, AsyncStreamRedirector)
+                assert isinstance(redirector, SimpleStreamRedirector)
                 self._setup(redirector)
                 asyncio.run(self._aloop(predict, redirector))
             else:
@@ -450,7 +450,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
         return None
 
     def _setup(
-        self, redirector: Union[StreamRedirector, AsyncStreamRedirector]
+        self, redirector: Union[StreamRedirector, SimpleStreamRedirector]
     ) -> None:
         done = Done()
         try:
@@ -504,7 +504,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
     async def _aloop(
         self,
         predict: Callable[..., Any],
-        redirector: AsyncStreamRedirector,
+        redirector: SimpleStreamRedirector,
     ) -> None:
         # Unwrap and replace the events connection with an async one.
         assert isinstance(self._events, LockedConnection)
@@ -583,7 +583,7 @@ class _ChildWorker(_spawn.Process):  # type: ignore
         tag: Optional[str],
         payload: Dict[str, Any],
         predict: Callable[..., Any],
-        redirector: AsyncStreamRedirector,
+        redirector: SimpleStreamRedirector,
     ) -> None:
         _tag_var.set(tag)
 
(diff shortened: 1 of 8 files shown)


python/cog/server/worker.py
```diff
@@ -417,11 +420,32 @@ def _current_tag(self) -> Optional[str]:
...
+            raise
+        finally:
+            self._events.send(Envelope(event=done))
+
+    def _setup(
+        self, redirector: Union[StreamRedirector, AsyncStreamRedirector]
+    ) -> None:
+        done = Done()
+        try:
+            assert self._predictor
```
aron, line 447: I'll refactor this into a context helper before merging.

python/cog/server/worker.py
```diff
@@ -373,32 +373,35 @@ def run(self) -> None:
...
 
         predict = get_predict(self._predictor)
         if inspect.iscoroutinefunction(predict) or inspect.isasyncgenfunction(predict):
+            async_redirector = AsyncStreamRedirector(
+                callback=self._stream_write_hook,
+                tee=self._tee_output,
+            )
+            with async_redirector:
+                self._setup(async_redirector)
             asyncio.run(self._aloop(predict, async_redirector))
```
philandstuff, line 390: I think we still want `_setup` and `_aloop` to be inside the same `with` statement, just like the sync case below, so we're not exiting and re-entering the context manager

aron: Good point. I've updated the code to remove the additional `with redirector:` in the `_loop` and `_aloop`.

python/cog/server/worker.py
```diff
@@ -456,7 +480,7 @@ def _loop(
         predict: Callable[..., Any],
         redirector: StreamRedirector,
     ) -> None:
-        with scope(self._loop_scope()), redirector:
+        with scope(self._loop_scope()):
```
philandstuff, line 483: I think we can lift `_loop_scope` up to where we now set up the redirector. It's not strictly identical - we now set loop scope in setup, which we didn't before - but I don't think that's a real problem.

python/cog/server/worker.py
```diff
@@ -417,11 +420,32 @@ def _current_tag(self) -> Optional[str]:
...
+            done.error_detail = str(e)
+        except BaseException as e:
+            # For SystemExit and friends we attempt to add some useful context
+            # to the logs, but reraise to ensure the process dies.
+            traceback.print_exc()
+            done.error = True
+            done.error_detail = str(e)
+            raise
+        finally:
+            self._events.send(Envelope(event=done))
```
philandstuff, line 440: this is a bug. we send a `done` when setup is complete, but we don't send one when we finish importing the predictor

python/cog/server/worker.py
```diff
@@ -373,33 +376,40 @@ def run(self) -> None:
...
-                ),
+            redirector = StreamRedirector(
+                callback=self._stream_write_hook,
+                tee=self._tee_output,
             )
 
+        with scope(Scope(record_metric=self.record_metric)), redirector:
+            self._predictor = self._load_predictor()
+
+            # If setup didn't set the predictor, we're done here.
```
erbridge, line 393: Is this comment right? We're not calling `setup` until after this are we?

aron: Nope, it's out of date, I'll fix it now.

python/tests/server/test_worker.py
```diff
@@ -110,11 +110,15 @@
 SETUP_LOGS_FIXTURES = [
     (
         WorkerConfig("logging", setup=False),
-        ("writing to stdout at import time\n" "setting up predictor\n"),
+        (
+            "writing some stuff from C at import time\n"
```
aron, line 114: This changed because we now use the `StreamRedirector` for setup of normal predictors which captures libc output again. The async version uses `SimpleStreamRedirector` which does not.

philandstuff (approved):

python/cog/server/worker.py
```diff
@@ -373,33 +376,41 @@ def run(self) -> None:
...
+                tee=self._tee_output,
             )
 
+        with scope(Scope(record_metric=self.record_metric)), redirector:
+            self._predictor = self._load_predictor()
+
+            # If _load_predictor hasn't returned a predictor instance then
+            # it has sent a error Done event and we're done here.
+            if not self._predictor:
+                return
```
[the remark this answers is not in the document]
aron: Chatted in person. I think it does clean up the flow of the run function yes.

PR #2074, opened 2024-12-03
Change torch vision to 0.20.0 for torch 2.5.0 cpu

8W9aG:
* Torchvision 0.20.1 doesn’t work on 2.5.0 cpu torch

### DIFF
--- a/pkg/config/torch_compatibility_matrix.json
+++ b/pkg/config/torch_compatibility_matrix.json
@@ -57,7 +57,7 @@
   },
   {
     "Torch": "2.5.0+cpu",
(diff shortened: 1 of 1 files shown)


meatballhat (approved):
I thought that this file was generated 🤔 So does this mean the generation code also needs to be updated?

8W9aG:
> I thought that this file was generated 🤔 So does this mean the generation code also needs to be updated?

This is a good point, i forgot that this was the case, I imagine that this might be the reason it's a bit borked lately. Come next torch update we should attempt with the script and verify heavily.

Issue #2054, opened 2024-11-13
Failing on copying .git directories on image build


alexeevit:
Hi, I've got an ComfyUI project (based on [fofr/cog-comfyui](https://github.com/fofr/cog-comfyui)) with custom_nodes being git submodules.
When I'm trying to build the image and push it with separate-weights, it fails.

Here's the full log:

```
$ cog push --separate-weights r8.im/username/model
Building Docker image from environment in cog.yaml as r8.im/username/model...
⚠ Stripping patch version from Python version 3.10.6 to 3.10
⚠ Stripping patch version from Python version 3.10.6 to 3.10
[+] Building 1.3s (9/9) FINISHED                                                                                                             docker:orbstack
 => [internal] load build definition from Dockerfile                                                                                                    0.0s
 => => transferring dockerfile: 378B                                                                                                                    0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1.4                                                                             0.8s
 => [auth] docker/dockerfile:pull token for registry-1.docker.io                                                                                        0.0s
 => CACHED docker-image://docker.io/docker/dockerfile:1.4@sha256:9ba7531bd80fb0a858632727cf7a112fbfd19b17e94c4e84ced81e24ef1a0dbc                       0.0s
 => [internal] load .dockerignore                                                                                                                       0.0s
 => => transferring context: 1.32kB                                                                                                                     0.0s
 => [internal] load build context                                                                                                                       0.0s
 => => transferring context: 8.94kB                                                                                                                     0.0s
 => CACHED [1/3] COPY checkpoints /src/checkpoints                                                                                                      0.0s
 => ERROR [2/3] COPY ComfyUI/custom_nodes/ComfyUI-Impact-Pack/.git/objects/pack /src/ComfyUI/custom_nodes/ComfyUI-Impact-Pack/.git/objects/pack         0.0s
 => ERROR [3/3] COPY ComfyUI/custom_nodes/comfyui_controlnet_aux/.git/objects/pack /src/ComfyUI/custom_nodes/comfyui_controlnet_aux/.git/objects/pack   0.0s
------
 > [2/3] COPY ComfyUI/custom_nodes/ComfyUI-Impact-Pack/.git/objects/pack /src/ComfyUI/custom_nodes/ComfyUI-Impact-Pack/.git/objects/pack:
------
------
[...]

PR #2087, opened 2024-12-09
Ignore files within a .git directory

8W9aG:
* Do not consider files within a .git directory when we are using separate-weights

Fixes https://github.com/replicate/cog/issues/2054

### DIFF
--- a/pkg/weights/weights.go
+++ b/pkg/weights/weights.go
@@ -101,7 +101,14 @@ func isCodeFile(path string) bool {
 }
 
 func isGitFile(path string) bool {
-	return strings.Contains(path, ".git")
+	dir, _ := filepath.Split(path)
+	folders := strings.Split(filepath.Clean(dir), string(filepath.Separator))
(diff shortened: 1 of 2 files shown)


pkg/weights/weights.go
```diff
@@ -97,6 +100,10 @@ func isCodeFile(path string) bool {
 	return ext == ".py" || ext == ".ipynb"
 }
 
+func isGitFile(path string) bool {
+	return strings.Contains(path, ".git")
+}
```
meatballhat, line 105: Any concern about this being too broad? Would it be reasonable to check for `"/.git/"` instead, for example?

8W9aG: Resolved in [34000e9](https://github.com/replicate/cog/pull/2087/commits/34000e9a34bdacb799cda1c82fdfffded364772b)

meatballhat (approved):
Nice!

PR #2088, opened 2024-12-11
Make dockerfile generators abstract

8W9aG:
* This allows us to add a fast build dockerfile generator that we can sub out based on the fast
build flag.


pkg/dockerfile/generator.go
```diff
@@ -1,689 +1,19 @@
...
-	}
-
-	strippedVersion := fmt.Sprintf("%d.%d", v.Major, v.Minor)
-	changed := strippedVersion != versionString
-
-	return strippedVersion, changed, nil
+import "github.com/replicate/cog/pkg/weights"
+
+type Generator interface {
+	generateInitialSteps() (string, error)
```
meatballhat, line 6: I think you can safely leave this method out of the interface 👍🏼 
```suggestion
```

meatballhat (approved):

pkg/dockerfile/generator.go
8W9aG: This is used by the base image generator here: https://github.com/replicate/cog/blob/main/pkg/dockerfile/base.go#L188

meatballhat: Ah I totally missed that! In that case I think it should be made a public method `GenerateInitialSteps`.

8W9aG: Resolved in [83769a2](https://github.com/replicate/cog/pull/2088/commits/83769a254c101e5017c2b7a9c6e75bf5c182d99a)

PR #2089, opened 2024-12-12
Async setup

aron:
This commit introduces the ability to define an async `setup` function on your predictor. For simplicity an async `setup()` function is only supported alongside an async `predict()` function. An error will be raised during setup if this is not the case.

Various pieces of the code have been extracted into smaller methods in order to achieve this. A new `_handle_setup_error` context manager has been created to handle setup errors and send appropriate `Done` event over the worker channel.

The `_setup()` method has been split into two phases, first we perform validation on the requirements for async/concurrency support. Then we attempt to run the `setup()` method either as a direct call for the non-async path or as part of the event loop in the async path.

### TODO

 - [x] Add tests for async setup
 - [ ] Perform async validation at build time, we should be able to perform the concurrency checks as part of the app start-up which would mean errors are caught during build/publish.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -31,7 +31,12 @@ from attrs import define
 
 from ..base_predictor import BasePredictor
 from ..json import make_encodeable
-from ..predictor import get_predict, load_predictor_from_ref, run_setup
+from ..predictor import (
+    get_predict,
+    load_predictor_from_ref,
+    has_setup_weights,
+    extract_setup_weights,
+)
 from ..types import PYDANTIC_V2, URLPath
 from ..wait import wait_for_env
 from .connection import AsyncConnection, LockedConnection
@@ -462,7 +467,12 @@ class _ChildWorker(_spawn.Process):  # type: ignore
 
             # Could be a function or a class
             if hasattr(self._predictor, "setup"):
-                run_setup(self._predictor)
+                if not has_setup_weights(predictor):
+                    predictor.setup()
+                    return
+
+                weights = extract_setup_weights(predictor)
+                predictor.setup(weights=weights)  # type: ignore
 
             predict = get_predict(self._predictor)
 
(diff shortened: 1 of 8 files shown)


python/cog/server/worker.py
```diff
@@ -458,22 +472,13 @@ def _load_predictor(self) -> Optional[BasePredictor]:
...
-                if not has_setup_weights(predictor):
-                    predictor.setup()
-                    return
-
-                weights = extract_setup_weights(predictor)
-                predictor.setup(weights=weights)  # type: ignore
-
             predict = get_predict(self._predictor)
 
             is_async_predictor = inspect.iscoroutinefunction(
```
philandstuff, line 484: this exactly duplicates `self._is_async`. I think we should get rid of this local variable and rename `self._is_async` to `self._is_async_predictor`. (I know your code didn't introduce this but it confused me while reviewing this PR)

python/tests/server/test_worker.py
```diff
@@ -351,6 +351,57 @@ def test_async_predictor_on_python_3_10_or_older_raises_error(worker):
...
+    predict_result = Result()
+    predict_sid = worker.subscribe(predict_result.handle_event, tag="p1")
+    worker.predict({}, tag="p1").result()
+
+    assert predict_result.done
+    assert predict_result.output == "output"
+    assert predict_result.stdout_lines == ["running prediction\n"]
+
+    worker.unsubscribe(predict_sid)
+
```
philandstuff, line 381: can we also have a test that uses the `"async_setup_uses_same_loop_as_predict"` fixture? (It already exists in the code tree but nothing currently references it, and some models rely on setup and predict using the same event loop.

philandstuff (commented):
Looks great! I especially like the refactoring of ChildWorker into smaller methods.  A couple of small comments.

philandstuff (approved):

PR #2090, opened 2024-12-13
Concurrency cancellation

aron:
Currently a cancellation event sent to a worker with `concurrency` > 1 will always cancel the most recent task.

For the async `_aloop` we now use a weak map to track all in flight prediction tasks. On receipt of a `Cancel` event, we retrieve the task from the map and cancel it.

For the standard `_loop` method we continue to ignore the `Cancel` event. Relying on calling the `send_cancel_signal()` to terminate the worker method before sending the `Cancel` event. I'm not 100% clear on why the parent needs to do this but calling `send_cancel_signal` or raising `CancelationException` directly in the `_loop` method causes the tests to fail so I've left it as-is.

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -12,6 +12,7 @@ import traceback
 import types
 import uuid
 import warnings
+import weakref
 from concurrent.futures import Future, ThreadPoolExecutor
 from enum import Enum, auto, unique
 from multiprocessing.connection import Connection
@@ -95,6 +96,10 @@ class PredictionState:
 
 
 class Worker:
+    @property
+    def uses_concurrency(self) -> bool:
+        return self._max_concurrency > 1
+
     def __init__(
         self, child: "_ChildWorker", events: Connection, max_concurrency: int = 1
     ) -> None:
@@ -289,7 +294,7 @@ class Worker:
                     with self._predictions_lock:
                         predict_state = self._predictions_in_flight.get(ev.tag)
                         if predict_state and not predict_state.cancel_sent:
-                            self._child.send_cancel()
+                            self._child.send_cancel_signal()
                             self._events.send(Envelope(event=Cancel(), tag=ev.tag))
                             predict_state.cancel_sent = True
                 else:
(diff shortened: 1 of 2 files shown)


python/cog/server/worker.py
```diff
@@ -516,7 +521,8 @@ def _loop(
         while True:
             e = cast(Envelope, self._events.recv())
             if isinstance(e.event, Cancel):
-                continue  # Ignored in sync predictors.
+                # this is handled via SIGUSR1 signals from the parent via send_cancel_signal
```
philandstuff, line 524:
```suggestion
                # for sync predictors, this is handled via SIGUSR1 signals from the parent via send_cancel_signal
```

python/cog/server/worker.py
```diff
@@ -533,17 +539,27 @@ async def _aloop(
...
                 e = cast(Envelope, await self._events.recv())
-                if isinstance(e.event, Cancel) and task and self._cancelable:
+                if isinstance(e.event, Cancel):
+                    if not self._cancelable:
+                        continue
+
+                    task = tasks.get(e.tag)
+                    if not task:
+                        print(
+                            f"Got cancel event for unrecognized prediction",
```
philandstuff, line 553:
```suggestion
                            "Got cancel event for unrecognized prediction",
```

python/cog/server/worker.py
```diff
@@ -533,17 +539,27 @@ async def _aloop(
...
 
-        task = None
-
         async with asyncio.TaskGroup() as tg:
+            tasks = weakref.WeakValueDictionary[str | None, asyncio.Task[Any]]()
             while True:
                 e = cast(Envelope, await self._events.recv())
-                if isinstance(e.event, Cancel) and task and self._cancelable:
+                if isinstance(e.event, Cancel):
+                    if not self._cancelable:
```
philandstuff, line 547: this isn't a useful flag for concurrent predictions (it's set and cleared in _handle_predict_error; but in the concurrent context it's set whenever *any* prediction starts and cleared whenever *any* prediction finishes); but I think we can rely entirely on whether the prediction is in the weak value dictionary.

aron:
Hmm, I've broken the sync tests somewhere, putting this back into draft while I figure it out.

aron:
I had an accidental `setup=False` in a WorkerConfig of the failing test... fixed now.

PR #2092, opened 2024-12-20
Download inputs off the event consumer loop

philandstuff:
Currently, we download inputs (via URLPath's convert() method) in
_prepare_payload, in the main Worker event consumer loop.  This is bad, because
downloading inputs is a blocking operation and we cannot process other events
while downloads are in progress.

This is a substantial restructuring:

- the event consumer loop now only processes events from the child; prediction
  requests and cancel requests from the http server are now processed directly.
- there are new concurrent.future.ThreadPoolExecutor instances for preparing
  predictions and downloading inputs.
  - as part of this, this commit also supports downloading inputs concurrently
    rather than serially.  I have hardcoded a maximum of 8 concurrent downloads
    based on the ThreadPoolExecutor size.  I picked this number arbitrarily.
- when the web server receives a request to start a prediction, we submit a task
  on the prediction start executor, which in turn submits downloads on the input
  download executor. Once all inputs are downloaded, we send the prediction
  request to the child.
- cancel requests are handled similarly, but as there is no input download this
  is much simpler.
- all calls to self._events.send() now happen within the http server async event
  loop, not the Worker ThreadPoolExecutor.  (This is maybe not strictly correct,
  as we're calling blocking i/o from an async event loop, but it's a local
  Connection object and it's a tiny amount of data)

### DIFF
--- a/python/cog/server/worker.py
+++ b/python/cog/server/worker.py
@@ -4,7 +4,6 @@ import contextvars
 import inspect
 import multiprocessing
 import os
-import select
 import signal
 import sys
 import threading
@@ -13,6 +12,7 @@ import types
 import uuid
 import warnings
 import weakref
+from concurrent import futures
 from concurrent.futures import Future, ThreadPoolExecutor
 from enum import Enum, auto, unique
 from multiprocessing.connection import Connection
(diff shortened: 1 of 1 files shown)


python/cog/server/worker.py
```diff
@@ -163,9 +161,57 @@ def predict(
...
+                        to_await.append(futs[k])
+                    # Check if v is a list of URLPath instances
+                    elif isinstance(v, list) and all(
+                        isinstance(item, URLPath) for item in v
+                    ):
+                        futs[k] = [
+                            self._input_download_pool.submit(item.convert) for item in v
+                        ]
+                        to_await += futs[k]
+                futures.wait(to_await, return_when=futures.FIRST_EXCEPTION)
```
erbridge, line 189: What happens to the rest of the work if we exit on first failure? Do we need to clean them up?

philandstuff: i suppose we could go cancel the unfinished futures. I'll do that.

PR #2094, opened 2024-12-23
Do not run a separate python install stage

8W9aG:
* Perform all the necessary installations on the pyenv python.


meatballhat (approved):
Yay!

PR #2105, opened 2025-01-08
Support run tokens

andreasjansson:
"Run tokens" are used to run downstream Replicate models from within a model.

They are not yet exposed in an API, but can be found inside a Cog predictor with some ugly frame inspection:

```python
def _find_api_token() -> str:
    frame = inspect.currentframe()
    while frame:
        if "self" in frame.f_locals:
            self = frame.f_locals["self"]
            if hasattr(self, "_current_run_token"):
                token = self._current_run_token
                return token
        frame = frame.f_back
    raise ValueError("No run token found in call stack")
```


python/cog/include.py
```diff
@@ -0,0 +1,37 @@
...
+            if hasattr(self, "_current_run_token"):
+                token = self._current_run_token()
+                return token
+        frame = frame.f_back
+    raise ValueError("No run token found in call stack")
+
+
+def include(model_path: str) -> Callable[..., Any]:
+    def run(**inputs: dict[str, Any]) -> Any:
+        client = replicate.Client(api_token=_find_api_token())
```
philandstuff, line 20: This can use `current_scope().run_token`, which is the affordance that exposes the run token to the model. In future, if the prototype is successful, we should make replicate-python automatically detect this and use it if no token was explicitly provided.

Further, I think the whole `include` function here probably belongs on replicate-python, not in cog?

tempusfrangit:
Closing this out since we're no longer pursuing pipelines via cog.

PR #2108, opened 2025-01-10
Add fast generator for cog build

8W9aG:
* When the `x-fast` flag is enabled, switch the generator to use the `FastGenerator` instead of the `StandardGenerator`
* This loads from monobase, then performs the following steps:
1. Loads the monobase generation, it uses a userspace cache for the monobase to share generation hydrations across projects.
2. Copies the weights (if found) via hardlinking (so no copying big files is performed). This performs a digest on the weights with SHA256 but then records the modified time and size of the file, if the file has the same modified time and size in a subsequent build it does not recalculate the SHA256.
3. Installs the apt-get packages in a layer.
4. Installs the python packages in a layer through monobase (to de-duplicate packages from the base).
5. Copies the source files without the weights (again using hardlinking).
6. Symlinks the weights to their position in the r8 srv directory.


8W9aG:
TODOs:
1. Add a user cache into the monobase build step
2. Do not copy over weights in the source copy, instead make sure they are symlinked to the weights in the copy weights section
3. Add unit tests
4. Add integration tests

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,261 @@
...
+	for sha256, file := range weights {
+		rel_path, err := filepath.Rel(g.Dir, file)
+		if err != nil {
+			return nil, err
+		}
+		commands = append(commands, "cp /src/"+rel_path+" "+filepath.Join(FUSE_RPC_WEIGHTS_PATH, sha256))
+	}
+
+	return append(lines, []string{
+		"RUN --mount=type=bind,ro,source=.,target=/src mkdir -p " + FUSE_RPC_WEIGHTS_PATH + " && " + strings.Join(commands, " && "),
```
nevillelyh, line 196: Might be easier and more reliable to weird file names if you do `rsync -a --include-from=weights-list.txt`. We can also bake the list into container FS & as a label for easy retrieval.

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	// Install python packages
+	packages, err := g.pythonPackages()
+	if err != nil {
+		return nil, err
+	}
+	if len(packages) > 0 {
+		mounts = append(mounts, "--mount=type=cache,target=/root/.cache,id=pip-cache")
+		pipCommand := "uv pip install "
```
nevillelyh, line 226: This could install package already in monobase right? Like if user include `python_packages: torch==...` instead of having it under `torch: ...`.

https://github.com/replicate/monobase/blob/main/mini.Dockerfile#L24
Monobase build script can take a user `--requirements` arg which exclude pkgs already in monobase's corresponding `requirements.txt`. This ensures user venv is mutually exclusive from monobase venv.

BTW where/how is user venv created?

pkg/dockerfile/fast_generator.go
nevillelyh: I guess we need to bake a file with SHA <-> path mapping, so that we can recreate symlinks later in prod. OTOH `rsync`'s include file is just the paths.

8W9aG: Ended up going with `COPY --link` on separate layers, makes builds superfast since it doesn't copy the weights, added some more logic to create a manifest then check the files mod time and size against that manifest (to prevent sha256 digests on the big weights files): https://github.com/replicate/cog/pull/2108/commits/23a109a85767d6577246094d301a139e23644d89

8W9aG: Ended up going with `COPY --link` on separate layers, makes builds superfast since it doesn't copy the weights, added some more logic to create a manifest then check the files mod time and size against that manifest (to prevent sha256 digests on the big weights files): https://github.com/replicate/cog/pull/2108/commits/23a109a85767d6577246094d301a139e23644d89

pkg/dockerfile/fast_generator.go
8W9aG: Interestingly there isn't currently one, but I wonder whether thats ok? I think the layer should reflect the changes to the main venv which might be enough for the fuse config? lmk if thats a good assumption

pkg/dockerfile/fast_generator.go
nevillelyh: Didn't know `--link` nice!

pkg/dockerfile/fast_generator.go
nevillelyh: We need to create a new user venv. The monobase venv is in `/srv/r8/monobase/venv/...` where `/srv/r8/monobase` is a read-only volume mount. So we can't untar user files on top.

The dedup part (removing pkgs already in monobase) is so that when we can layer `PYTHONPATH=<monobase>:<user>:<cog>` because:
* To ensure user always depend on pkg version in monobase, not random ones pulled in by pip
* We can bump pkgs in monobase for security fixes, and be sure that it's picked up

8W9aG:
I intend to write more unit tests before I declare this complete, but the basic structure is ready for review.

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,273 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if requirementsFile != "" {
+		lines = append(lines, "RUN "+strings.Join([]string{
+			buildTmpMount,
+			g.monobaseUsercacheMount(),
+			APT_CACHE_MOUNT,
+			PIP_CACHE_MOUNT,
+		}, " ")+" /opt/r8/monobase/run.sh monobase.build --requirements=/buildtmp/requirements.txt --mini --skip-cuda --cache="+MONOBASE_CACHE_PATH)
```
nevillelyh, line 229: We don't need `APT_CACHE_MOUNT` here right?

I decoupled this from mini-mono build to ensure it doesn't touch the existing layer.
So the command should be: `/opt/r8/monobase/run.sh monobase.user --requirements /src/requirements.txt`

We might want to mount `/srv/r8/monobase/uv/cache` with as a cache bind mount to work around this:
https://github.com/replicate/monobase/pull/63#issuecomment-2591478537

We might also need to exclude anything outside of `/root/.venv` to make sure the resulting tarball doesn't fail to write into read-only `/srv/r8/monobase` in K8S.

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,273 @@
...
+	"strings"
+
+	"github.com/replicate/cog/pkg/config"
+	"github.com/replicate/cog/pkg/weights"
+)
+
+const FUSE_RPC_WEIGHTS_PATH = "/srv/r8/fuse-rpc/weights"
+const MONOBASE_CACHE_PATH = "/var/cache/monobase"
+const APT_CACHE_MOUNT = "--mount=type=cache,target=/var/cache/apt,id=apt-cache"
+const PIP_CACHE_MOUNT = "--mount=type=cache,target=/root/.cache,id=pip-cache"
```
nevillelyh, line 17: This should be `/srv/r8/monobase/uv/cache` as defined here:
https://github.com/replicate/monobase/blob/main/Dockerfile#L28

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,273 @@
...
+	buildTmpMount, err := g.buildTmpMount(tmpDir)
+	if err != nil {
+		return nil, err
+	}
+
+	return append(lines, []string{
+		"RUN " + strings.Join([]string{
+			buildTmpMount,
+			g.monobaseUsercacheMount(),
+			APT_CACHE_MOUNT,
```
nevillelyh, line 189: Don't need `APT_CACHE_MOUNT` here either now that `apt-get install` is in a separate layer right?

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,273 @@
...
+	cogPath, err := g.copyCog(tmpDir)
+	if err != nil {
+		return nil, err
+	}
+
+	lines = append(lines, []string{
+		"ENV R8_COG_VERSION=\"file:///buildtmp/" + filepath.Base(cogPath) + "\"",
+	}...)
+
+	skipCudaArg := "--skip-cuda"
```
nevillelyh, line 156: We should get rid of `--skip-cuda` everywhere. It's to speed up CI only and I renamed to `CI_SKIP_CUDA` to reduce confusion.

nevillelyh (commented):
Comments to sync up with https://github.com/replicate/monobase/pull/63

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [0e019cb](https://github.com/replicate/cog/pull/2108/commits/0e019cb77300143a81a59328184501725c7eabe5)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [0e019cb](https://github.com/replicate/cog/pull/2108/commits/0e019cb77300143a81a59328184501725c7eabe5)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [0e019cb](https://github.com/replicate/cog/pull/2108/commits/0e019cb77300143a81a59328184501725c7eabe5)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [9dbbe5a](https://github.com/replicate/cog/pull/2108/commits/9dbbe5a3c184061c20d71397e2bf740aa76b7e9d)

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,271 @@
...
+	buildTmpMount, err := g.buildTmpMount(tmpDir)
+	if err != nil {
+		return nil, err
+	}
+	if requirementsFile != "" {
+		lines = append(lines, "RUN "+strings.Join([]string{
+			buildTmpMount,
+			g.monobaseUsercacheMount(),
+			UV_CACHE_MOUNT,
+		}, " ")+" UV_CACHE_DIR=\""+UV_CACHE_DIR+"\" /opt/r8/monobase/run.sh monobase.build --requirements=/buildtmp/requirements.txt --cache="+MONOBASE_CACHE_PATH)
```
nevillelyh, line 227: Should be `monobase.user` once https://github.com/replicate/monobase/pull/63 is merged.

Should also add `UV_COMPILE_BYTECODE=0` for this one step, otherwise UV compiles byte code of the managed Python interpreter in `/srv/r8/monobase/uv/python`, polluting that directory.

nevillelyh (commented):
A few more minor comments.
Once all converged, the monobase vs user venv layers should be cleanly separate.

pkg/dockerfile/fast_generator.go
```diff
@@ -0,0 +1,271 @@
...
+		lines = append(lines, []string{
+			"ENV R8_CUDA_VERSION=" + cudaVersion,
+			"ENV R8_CUDNN_VERSION=" + cudnnVersion,
+			"ENV R8_CUDA_PREFIX=https://monobase.replicate.delivery/cuda",
+			"ENV R8_CUDNN_PREFIX=https://monobase.replicate.delivery/cudnn",
+		}...)
+	}
+
+	lines = append(lines, []string{
+		"ENV R8_PYTHON_VERSION=" + g.Config.Build.PythonVersion,
```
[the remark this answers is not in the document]
8W9aG: Resolved in [f72607a](https://github.com/replicate/cog/pull/2108/commits/f72607ae8db899623288dc557cc12a345ca539ac)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [339abfd](https://github.com/replicate/cog/pull/2108/commits/339abfd2ace253151e2726b6fef0b47264d7e469)

pkg/config/requirements.go
```diff
@@ -0,0 +1,130 @@
...
+	err = os.WriteFile(requirementsFile, []byte(requirementsContent), 0o644)
+	if err != nil {
+		return "", err
+	}
+	return requirementsFile, nil
+}
+
+// SplitPinnedPythonRequirement returns the name, version, findLinks, and extraIndexURLs from a requirements.txt line
+// in the form name==version [--find-links=<findLink>] [-f <findLink>] [--extra-index-url=<extraIndexURL>]
+func SplitPinnedPythonRequirement(requirement string) (name string, version string, findLinks []string, extraIndexURLs []string, err error) {
```
nevillelyh, line 85: Just realized this. Do we still need this, now that `monobase.user` dedups the requirements and handles Torch index anyway?

8W9aG: Yeh we need to perform a dedup ourselves on the `cog.yaml`, it offers both `python_packages` and `python_requirements` as an option.

nevillelyh (approved):

Issue #2112, opened 2025-01-17
Using `pydantic>2` requires specifying `fastapi` version, too


nickstenning:
I encountered today the surprising fact that the following `cog.yaml` results in a broken image:

```yaml
build:
  gpu: false
  python_version: "3.12"
  python_packages:
    - "pydantic>2"
predict: "predict.py:Predictor"
```

The build fails with the following error:

```console
[...]
#14 DONE 0.7s
Validating model schema...

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/local/lib/python3.12/site-packages/cog/command/openapi_schema.py", line 13, in <module>
    from ..server.http import create_app
  File "/usr/local/lib/python3.12/site-packages/cog/server/http.py", line 18, in <module>
    from fastapi import Body, FastAPI, Header, Path, Response
  File "/usr/local/lib/python3.12/site-packages/fastapi/__init__.py", line 7, in <module>
    from .applications import FastAPI as FastAPI
  File "/usr/local/lib/python3.12/site-packages/fastapi/applications.py", line 16, in <module>
    from fastapi import routing
  File "/usr/local/lib/python3.12/site-packages/fastapi/routing.py", line 22, in <module>
    from fastapi import params
  File "/usr/local/lib/python3.12/site-packages/fastapi/params.py", line 4, in <module>
    from pydantic.fields import FieldInfo, Undefined
ImportError: cannot import name 'Undefined' from 'pydantic.fields' (/usr/local/lib/python3.12/site-packages/pydantic/fields.py)

ⅹ Failed to get type signature: exit status 1
```

As far as I can tell the problem here is that `fastapi==0.98` is installed by default in the `cog-base` images, but this is actually not compatible with Pydantic v2.

It's possible to make a working image by specifying a tighter range of acceptable versions for `fastapi`, but I don't think this is ideal for users:

```yaml
build:
  gpu: false
  python_version: "3.12"
  python_packages:
    - "fastapi>0.100.0,<0.111.0"
    - "pydantic>2"
predict: "predict.py:Predictor"
```

8W9aG:
https://github.com/replicate/cog/pull/2115

PR #2113, opened 2025-01-17
Workaround for fastapi in base image not working with pydantic 2

nevillelyh:
https://github.com/replicate/cog/issues/2112


nevillelyh:
Discussed IRL should be bumped in Cog instead.

PR #2115, opened 2025-01-17
Fix pydantic2 cog builds

8W9aG:
* Add an integration test to validate the problem
* Upgrade fastapi to >= 0.100 to support pydantic 2


nickstenning (approved):
LGTM!

PR #2122, opened 2025-01-23
Add test for ffmpeg in base images

8W9aG:
* Add an integration test to check for `ffmpeg` in the base image.


meatballhat (approved):
Yay! 🤘🏼

PR #2128, opened 2025-01-27
Add upload and verification to the fast pusher

8W9aG:
* Creates an S3 client for uploading based on the response from the upload service
* Begins a verification step after the upload is completed
* Check the verification state in a loop


pkg/docker/fast_push.go
```diff
@@ -128,12 +145,30 @@ func startUploadURL(objectType string, digest string) url.URL {
...
+	return checkResp.StatusCode == http.StatusOK, nil
+}
+
 func uploadFile(objectType string, digest string, path string, token string) error {
 	console.Debug("uploading file: " + path)
 
 	uploadUrl := startUploadURL(objectType, digest)
 	client := &http.Client{}
-	req, _ := http.NewRequest("POST", uploadUrl.String(), nil)
+	req, err := http.NewRequest("POST", uploadUrl.String(), nil)
```
meatballhat, line 168: I don't see a `context.Context` available, but getting one passed down so that cancellation and cleanup are handled correctly will be real nice 😁 and so that you can use `http.NewRequestWithContext`.

pkg/docker/fast_push.go
```diff
@@ -148,7 +183,76 @@ func uploadFile(objectType string, digest string, path string, token string) err
...
+	s3Client := s3.NewFromConfig(*cfg)
+	uploadParams := &s3.PutObjectInput{
+		Bucket: aws.String(data.Bucket),
+		Key:    aws.String(data.Key),
+		Body:   file,
+	}
+	_, err = s3Client.PutObject(context.Background(), uploadParams)
+	if err != nil {
+		return err
+	}
```
meatballhat, line 222: This will need to be changed to use the [managed uploader](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#NewUploader) since:
- files will often be over 5GB
- the uploader takes care of retry / is much more resilient

meatballhat (approved):
With the understanding that `uploadFile` will need to grow a `ctx context.Context` first arg in the signature and the use of the [managed uploader](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager#NewUploader) is almost certainly going to be needed.

🎉

pkg/docker/fast_push.go
8W9aG: Resolved in [8809dea](https://github.com/replicate/cog/pull/2128/commits/8809deac27ece2069bad51328b1f4fefab75ff6c)

pkg/docker/fast_push.go
```diff
@@ -26,6 +33,17 @@ const requirementsTarFile = "requirements.tar.zst"
...
+	Bucket          string `json:"bucket"`
+	Endpoint        string `json:"endpoint"`
+	AccessKeyId     string `json:"access_key_id"`
+	SecretAccessKey string `json:"secret_access_key"`
+	SessionToken    string `json:"session_token"`
+	Expires         int64  `json:"expires"`
+	Uuid            string `json:"uuid"`
+}
+
 func FastPush(image string, projectDir string, command Command, ctx context.Context) error {
```
meatballhat, line 47: This is a nit for later, but `ctx` really should be the first argument in the signature 👍🏼

pkg/docker/fast_push.go
```diff
@@ -128,12 +146,30 @@ func startUploadURL(objectType string, digest string) url.URL {
...
+	return verificationUrl
+}
+
+func checkVerificationStatus(req *http.Request, client *http.Client) (bool, error) {
+	checkResp, err := client.Do(req)
+	if err != nil {
+		return false, err
+	}
+	defer checkResp.Body.Close()
+	return checkResp.StatusCode == http.StatusOK, nil
```
meatballhat, line 161: I think this will need to look at the body, too, since completed != verified.

Here's where the e2e test script does it, fwiw: https://github.com/replicate/monobeam/blob/1fb65095db29e1b07368e0e09cd544735b033805/script/e2e-test#L92-L108

pkg/docker/fast_push.go
```diff
@@ -128,12 +146,30 @@ func startUploadURL(objectType string, digest string) url.URL {
...
+}
+
+func uploadFile(ctx context.Context, objectType string, digest string, path string, token string) error {
 	console.Debug("uploading file: " + path)
 
 	uploadUrl := startUploadURL(objectType, digest)
 	client := &http.Client{}
-	req, _ := http.NewRequest("POST", uploadUrl.String(), nil)
+	req, err := http.NewRequestWithContext(ctx, "POST", uploadUrl.String(), nil)
```
meatballhat, line 169: nit: `http.MethodPost` instead of `"POST"`

meatballhat (approved):
🙌🏼

pkg/docker/fast_push.go
8W9aG: Resolved in [9f53230](https://github.com/replicate/cog/pull/2128/commits/9f53230ed8d7eac9d1102b7e605ec6e587b8e9f1)

pkg/docker/fast_push.go
8W9aG: Resolved in [e7ff752](https://github.com/replicate/cog/pull/2128/commits/e7ff752774997f3282db15b69eee7f5267f96ac8)

pkg/docker/fast_push.go
8W9aG: Resolved in [3975cbb](https://github.com/replicate/cog/pull/2128/commits/3975cbb80308e1dcd83f064040d47afad8111cc4)

meatballhat (approved):

PR #2131, opened 2025-01-28
Change ffmpeg int test to 3.12

8W9aG:
* Python 3.12 doesn’t work for ffmpeg


nevillelyh:
What does "doesn't work" mean? There's a Python ffmpeg binding?

8W9aG:
This is a hangover from when we werent installing apt packages correctly in base images, i think I forgot to execute the python 3.12 base image builds

nevillelyh (approved):

PR #2132, opened 2025-01-31
Fix file too long on tmp file create

8W9aG:
* Splits the extension on a ? query parameter
* Ensure that the filename in URLPath is truncated

As an addendum I don't actually know how this comes to be, in the sense that this was originally created by an input that is a `Path` type which gets converted to a `URLPath` and during that conversion the truncation of the filename happens. The truncation is based of parsing the URL and using its `path` attribute which should not include queries. However they somehow sneak past this check and show up in the tempfile `NamedTemporaryFile` method. In light of this I've opted to check the length on creation of the `URLPath` construction and provided some tests to verify that.


daanelson:
lgtm! odd that we don't know exactly how these filenames sneak in there (you'd think they'd all be shortened w/`get_filename`); regardless looks like this fixes it.

daanelson (approved):

PR #2134, opened 2025-02-04
Add multi progress bar to uploads

8W9aG:
* Shows the progress of each upload as a distinct item.

### DIFF
--- a/pkg/docker/fast_push.go
+++ b/pkg/docker/fast_push.go
@@ -58,7 +58,7 @@ func userAgent() string {
 func FastPush(ctx context.Context, image string, projectDir string, command Command) error {
 	g, _ := errgroup.WithContext(ctx)
 	p := mpb.New(
-		mpb.WithRefreshRate(180*time.Millisecond),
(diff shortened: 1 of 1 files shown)


pkg/docker/fast_push.go
```diff
@@ -55,6 +57,10 @@ func userAgent() string {
 
 func FastPush(ctx context.Context, image string, projectDir string, command Command) error {
 	g, _ := errgroup.WithContext(ctx)
+	p := mpb.New(
+		mpb.WithWidth(60),
```
philandstuff, line 61: Why did we decide 60 columns? Are the defaults bad?

In general I'd expect the library to do the right thing if I resize my terminal (if, say, I go below 60 columns wide), so I'm a little suspicious of customizing to a hardcoded value.

(This isn't a blocking issue)

philandstuff (approved):
looks good! one non-blocking comment

pkg/docker/fast_push.go
8W9aG: This is a cargo cult, I'll remove the width requirement.

PR #2135, opened 2025-02-07
Create multipart uploader for fast push

NikhilSinha1:
### Summary

We are having loads of problems with the AWS managed uploader so let's replace it with a different managed uploader that we have more control over.

### Test Plan

Try to upload a large file using `cog push --debug --x-fast`


pkg/docker/docker_command.go
```diff
@@ -97,23 +104,26 @@ func (c *DockerCommand) Inspect(image string) (*command.Manifest, error) {
...
-	var manifest command.Manifest
-	err = decoder.Decode(&manifest)
+	var manifests []command.Manifest
+	err = decoder.Decode(&manifests)
 	if err != nil {
 		return nil, err
 	}
 
-	return &manifest, nil
+	return &manifests[0], nil // Docker inspect returns us a list of manifests
```
meatballhat, line 119: probably worth checking the `len(manifests)` and returning a non-nil `err` if it is empty rather than potentially panicking here

pkg/web/client.go
```diff
@@ -102,19 +108,19 @@ func (c *Client) PostNewVersion(ctx context.Context, image string, weights []Fil
 func (c *Client) versionFromManifest(image string, weights []File, files []File) (*Version, error) {
 	manifest, err := c.dockerCommand.Inspect(image)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to inspect docker image: %v", err)
```
meatballhat, line 111: make sure to use `fmt.Errorf` with `%w` instead of `%v` so that the error is wrapped instead of stringified

tools/uploader/s3.go
```diff
@@ -0,0 +1,318 @@
...
+	return strconv.ParseInt(s, 0, 64)
+}
+
+func NewS3Uploader(client *s3.Client) Uploader {
+	uploader := &s3Uploader{
+		S3Client: client,
+	}
+
+	// 64 GB default limit
+	uploader.MaxPartUploads = util.GetEnvOrDefault(UPLOADER_MAX_PARTS_UPLOAD_KEY, 1024*4, parseInt32)
```
meatballhat, line 66: What is the 64 GB referencing here in the comment?

tools/uploader/s3.go
```diff
@@ -0,0 +1,318 @@
...
+	// 30 retries per part default
+	uploader.MaxPartUploads = util.GetEnvOrDefault(UPLOADER_MAX_PART_UPLOAD_RETRIES_KEY, 30, parseInt32)
+
+	return uploader
+}
+
+func (s *s3Uploader) Initialize() {
+	if !s.initialized {
+		s.mpuBufferPool = sync.Pool{
+			New: func() interface{} {
```
meatballhat, line 80: prefer `any` over `interface{}` 🎉

tools/uploader/s3.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return nil
+}
+
+func (s *s3Uploader) uploadMultipartObjectToS3(ctx context.Context, fInfo uploadFileDetails) error {
+
+	// Initialize if we haven't already
+	if !s.initialized {
+		s.Initialize()
+	}
```
meatballhat, line 161: IIUC this should be safe, right?
```suggestion
	s.Initialize()
```

meatballhat (approved):
Nothing blocking pending the test fixes 👍🏼 

Do I understand correctly that most of the uploader code was adapted from another codebase?

PR #2139, opened 2025-02-07
Call the web when layers have been pushed

8W9aG:
* Creates an http client that decorates requests with user agent and authentication.
* Adds a monobeam client
* Adds a web client
* Adds a manifest function to the command
* Looks up the manifest from the image and fills in the necessary variables to send to web on
version push

TODO: Will be adding more tests to make sure this is covered

### DIFF
--- a/pkg/web/client.go
+++ b/pkg/web/client.go
@@ -1,0 +1,240 @@
+package web
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/replicate/cog/pkg/config"
+	"github.com/replicate/cog/pkg/docker/command"
+	"github.com/replicate/cog/pkg/env"
+	"github.com/replicate/cog/pkg/global"
+)
+
+type Client struct {
+	command command.Command
+	client  *http.Client
+}
+
+type File struct {
+	Path   string `json:"path"`
+	Digest string `json:"digest"`
+	Size   int64  `json:"size"`
+}
+
+type Env struct {
+	CogGpu              string `json:"COG_GPU"`
+	CogPredictTypeStub  string `json:"COG_PREDICT_TYPE_STUB"`
+	CogTrainTypeStub    string `json:"COG_TRAIN_TYPE_STUB"`
+	CogPredictCodeStrip string `json:"COG_PREDICT_CODE_STRIP"`
+	CogTrainCodeStrip   string `json:"COG_TRAIN_CODE_STRIP"`
+	CogPyEnvPath        string `json:"COG_PYENV_PATH"`
+	CogEagerImports     string `json:"COG_EAGER_IMPORTS"`
+	R8CudaVersion       string `json:"R8_CUDA_VERSION"`
+	R8CudnnVersion      string `json:"R8_CUDNN_VERSION"`
+	R8PythonVersion     string `json:"R8_PYTHON_VERSION"`
+	R8TorchVersion      string `json:"R8_TORCH_VERSION"`
+}
+
+type RuntimeConfig struct {
+	Weights []File `json:"weights"`
+	Files   []File `json:"files"`
+	Env     Env    `json:"env"`
+}
+
+type Version struct {
+	Annotations   map[string]string      `json:"annotations"`
+	CogConfig     config.Config          `json:"cog_config"`
+	CogVersion    string                 `json:"cog_version"`
+	OpenAPIScheme map[string]interface{} `json:"openapi_schema"`
+	RuntimeConfig RuntimeConfig          `json:"runtime_config"`
+	Virtual       bool                   `json:"virtual"`
+}
+
+func NewClient(command command.Command, client *http.Client) *Client {
+	return &Client{
+		command: command,
+		client:  client,
+	}
+}
+
+func (c *Client) PostNewVersion(ctx context.Context, image string, weights []File, files []File) error {
+	userInfo, err := c.command.LoadUserInformation(global.ReplicateRegistryHost)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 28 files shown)


pkg/web/client.go
```diff
@@ -0,0 +1,239 @@
...
+	Weights []File `json:"weights"`
+	Files   []File `json:"files"`
+	Env     Env    `json:"env"`
+}
+
+type Version struct {
+	Annotations   map[string]string      `json:"annotations"`
+	CogConfig     config.Config          `json:"cog_config"`
+	CogVersion    string                 `json:"cog_version"`
+	OpenAPIScheme map[string]interface{} `json:"openapi_schema"`
```
meatballhat, line 55: nit: we can use `any` instead of `interface {}` :tada:

pkg/web/client.go
```diff
@@ -0,0 +1,239 @@
...
+	RuntimeConfig RuntimeConfig          `json:"runtime_config"`
+	Virtual       bool                   `json:"virtual"`
+}
+
+func NewClient(command command.Command, client *http.Client) *Client {
```
meatballhat, line 60: nit: I'd change the local `command` to something else so that it's not shadowing the imported package name

pkg/web/client.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	var cogConfig config.Config
+	err = json.Unmarshal([]byte(manifest.Config.Labels[command.CogConfigLabelKey]), &cogConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	var openAPIScheme map[string]interface{}
+	err = json.Unmarshal([]byte(manifest.Config.Labels[command.CogOpenAPISchemaLabelKey]), &cogConfig)
```
meatballhat, line 115: I'm pretty sure this is supposed to be passing a pointer to the openapi schema struct

pkg/web/client.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	predictCode, err := stripCodeFromStub(cogConfig, true)
+	if err != nil {
+		return nil, err
+	}
+	trainCode, err := stripCodeFromStub(cogConfig, false)
```
meatballhat, line 124: are predictCode and trainCode supposed to be the same value?

pkg/web/client.go
```diff
@@ -0,0 +1,239 @@
...
+		cogGPU = 1
+	}
+
+	torchVersion := ""
+	cudaVersion := ""
+	cudnnVersion := ""
+	pythonVersion := ""
+	pythonPath := ""
+	for _, env := range manifest.Config.Env {
+		envComponents := strings.Split(env, "=")
```
meatballhat, line 140: ooh! strings.Cut is nice for this 

https://pkg.go.dev/strings#Cut

meatballhat (approved):
This is quite a big lift! :tada: Very nicely done!

pkg/docker/env.go
```diff
@@ -0,0 +1,13 @@
...
+import "os"
+
+const DockerCommandEnvVarName = "R8_DOCKER_COMMAND"
+
+func DockerCommandFromEnvironment() string {
+	scheme := os.Getenv(DockerCommandEnvVarName)
+	if scheme == "" {
+		scheme = "docker"
+	}
+	return scheme
```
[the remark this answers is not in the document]
8W9aG: Resolved in [9234c1f](https://github.com/replicate/cog/pull/2139/commits/9234c1f80ea3860076e0054cd29331aedba22f8d)

pkg/dockerfile/monobase_matrix.go
```diff
@@ -0,0 +1,57 @@
...
+	Id             int            `json:"id"`
+	CudaVersions   []string       `json:"cuda_versions"`
+	CudnnVersions  []string       `json:"cudnn_versions"`
+	PythonVersions []string       `json:"python_versions"`
+	TorchVersions  []string       `json:"torch_versions"`
+	Venvs          []MonobaseVenv `json:"venvs"`
+}
+
+func NewMonobaseMatrix(client *http.Client) (*MonobaseMatrix, error) {
+	resp, err := client.Get("https://raw.githubusercontent.com/replicate/monobase/refs/heads/main/matrix.json")
```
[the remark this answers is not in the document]
8W9aG: In the future we should, this was a separate of the monobase matrix concern from the fast generator concern so not a new detail from this PR specifically

pkg/web/client.go
8W9aG: Resolved in [0f13d1a](https://github.com/replicate/cog/pull/2139/commits/0f13d1a7bb1fe109a46ee9ab26695ba7a0c585fc)

pkg/web/client.go
8W9aG: Resolved in [5ee1230](https://github.com/replicate/cog/pull/2139/commits/5ee1230d0507082d8be36be2a2973c06bd46680e)

pkg/web/client.go
8W9aG: This is a great catch, I'm going to see if I can make some tests to verify this is happening.

8W9aG: Resolved in [4e6fb46](https://github.com/replicate/cog/pull/2139/commits/4e6fb462d3eda39dca1e0ce70a6b6c7e4d80b994)

pkg/web/client.go
8W9aG: Resolved in [0517e96](https://github.com/replicate/cog/pull/2139/commits/0517e96f1a23dd4bdb1cf946dce80d81702bc663)

pkg/web/client.go
meatballhat: oohhhhh I see why I was confused 😅 I was only looking at the function name `stripCodeFromStub` and didn't read through the implementation with the second positional boolean arg

PR #2146, opened 2025-02-11
Enforce usage of r8.im image name

8W9aG:
* Do not use the username from the authentication credentials, the user could always push to an org
that they have access to.
* Make sure the URL is a r8.im URL before breaking it up.

### DIFF
--- a/pkg/web/client.go
+++ b/pkg/web/client.go
@@ -194,8 +194,11 @@ func (c *Client) versionFromManifest(image string, weights []File, files []File)
 func newVersionURL(image string) (url.URL, error) {
 	imageComponents := strings.Split(image, "/")
 	newVersionUrl := webBaseURL()
+	if len(imageComponents) != 3 {
+		return newVersionUrl, errors.New("The image URL must have 3 components in the format of " + global.ReplicateRegistryHost + "/username/modelname")
+	}
 	if imageComponents[0] != global.ReplicateRegistryHost {
-		return newVersionUrl, errors.New("The image name must have the r8.im prefix in fast push.")
+		return newVersionUrl, errors.New("The image name must have the " + global.ReplicateRegistryHost + " prefix in fast push.")
 	}
 	newVersionUrl.Path = strings.Join([]string{"", imageComponents[1], imageComponents[2], "versions"}, "/")
 	return newVersionUrl, nil
--- a/pkg/web/client.go
+++ b/pkg/web/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/replicate/cog/pkg/config"
 	"github.com/replicate/cog/pkg/docker/command"
 	"github.com/replicate/cog/pkg/env"
+	"github.com/replicate/cog/pkg/global"
 )
 
 type Client struct {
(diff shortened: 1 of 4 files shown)


pkg/web/client_test.go
```diff
@@ -74,3 +74,8 @@ func TestVersionFromManifest(t *testing.T) {
 	require.Equal(t, openAPISchema, version.OpenAPISchema)
 	require.Equal(t, cogConfig, version.CogConfig)
 }
+
+func TestVersionURLErrorWithoutR8IMPrefix(t *testing.T) {
+	_, err := newVersionURL("docker.com/thing/thing")
+	require.Error(t, err)
+}
```
erbridge, line 81: We should have a test for `thing/thing`, too, shouldn't we?

pkg/web/client.go
```diff
@@ -192,10 +191,14 @@ func (c *Client) versionFromManifest(image string, weights []File, files []File)
 	return &version, nil
 }
 
-func newVersionURL(username string, image string) url.URL {
+func newVersionURL(image string) (url.URL, error) {
+	imageComponents := strings.Split(image, "/")
 	newVersionUrl := webBaseURL()
-	newVersionUrl.Path = strings.Join([]string{"", username, image, "versions"}, "/")
-	return newVersionUrl
+	if imageComponents[0] != global.ReplicateRegistryHost {
```
NikhilSinha1, line 197: Should we also check if we have the 3 parts? Maybe we just return like a "The image name must be r8.im/username<or org name>/model name"

NikhilSinha1 (approved):

pkg/web/client.go
```diff
@@ -192,10 +191,14 @@ func (c *Client) versionFromManifest(image string, weights []File, files []File)
 	return &version, nil
 }
 
-func newVersionURL(username string, image string) url.URL {
+func newVersionURL(image string) (url.URL, error) {
+	imageComponents := strings.Split(image, "/")
```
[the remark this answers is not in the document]
8W9aG: Resolved in [d74b018](https://github.com/replicate/cog/pull/2146/commits/d74b018d5783ccaafcfd3a38acd247b944720aca)

pkg/web/client_test.go
8W9aG: This is covered by the test for component lengths: [d74b018](https://github.com/replicate/cog/pull/2146/commits/d74b018d5783ccaafcfd3a38acd247b944720aca)

pkg/web/client.go
```diff
@@ -192,10 +191,17 @@ func (c *Client) versionFromManifest(image string, weights []File, files []File)
...
 	newVersionUrl := webBaseURL()
-	newVersionUrl.Path = strings.Join([]string{"", username, image, "versions"}, "/")
-	return newVersionUrl
+	if len(imageComponents) != 3 {
+		return newVersionUrl, errors.New("The image URL must have 3 components in the format of " + global.ReplicateRegistryHost + "/username/modelname")
+	}
+	if imageComponents[0] != global.ReplicateRegistryHost {
+		return newVersionUrl, errors.New("The image name must have the " + global.ReplicateRegistryHost + " prefix in fast push.")
```
erbridge, line 201: Do users know what "fast push" is? Should this say something like 
```suggestion
		return newVersionUrl, errors.New("The image name must have the " + global.ReplicateRegistryHost + " prefix when using --x-fast.")
```

pkg/web/client_test.go
```diff
@@ -74,3 +74,13 @@ func TestVersionFromManifest(t *testing.T) {
...
+
+func TestVersionURLErrorWithout3Components(t *testing.T) {
+	_, err := newVersionURL("r8.im/username")
```
erbridge, line 84: I'd expect a more likely real situation would be 
```suggestion
	_, err := newVersionURL("username/test")
```

PR #2147, opened 2025-02-11
Validate config by checking the run commands

8W9aG:
* Ensure we send an error if the user tries to use fast push with run commands, they are not
supported.


meatballhat (approved):
🚫🏃🏼

PR #2148, opened 2025-02-11
Fix monobeam client not authenticating properly

8W9aG:
* Monobeam client was recreating a standard HTTP client rather than using the one injected.


meatballhat (approved):
🔏

PR #2149, opened 2025-02-12
Require a test to be added for fast push

8W9aG:
* Ensure that the user adds a test in the form of a cog predict command before performing a fast
build of their model.


pkg/config/config_test.go
```diff
@@ -693,3 +693,16 @@ torch==2.4.0
...
 	require.Equal(t, expected, requirements)
 }
+
+func TestParseTests(t *testing.T) {
+	yamlString := `
+build:
+  run:
+  - command: "echo 'Hello, World!'"
+tests:
+  - command: "cog predict -i s=world"
```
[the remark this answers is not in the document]
8W9aG: Is your intention to have something that looks like:
```yaml
tests:
  - s: world
```

In this case?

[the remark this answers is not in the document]
erbridge: Something like that, yeah?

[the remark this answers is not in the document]
8W9aG: Resolved in [6ce8145](https://github.com/replicate/cog/pull/2149/commits/6ce8145eae612393b5f42fd01a45d1fd59c71d55)

pkg/config/data/config_schema_v1.0.json
```diff
@@ -161,6 +184,19 @@
           "description": "The default target for number of concurrent predictions. This setting can be used by an autoscaler to determine when to scale a deployment of a model up or down."
         }
       }
+    },
+    "tests": {
+      "$id": "#/properties/build/properties/run",
```
meatballhat, line 189: This `$id` doesn't make sense to me 🤔  Shouldn't it be `#/properties/tests`?

pkg/dockerfile/fast_generator.go
```diff
@@ -296,5 +296,8 @@ func (g *FastGenerator) validateConfig() error {
 	if len(g.Config.Build.Run) > 0 {
 		return errors.New("cog builds with --x-fast do not support build run commands.")
 	}
+	if len(g.Config.Tests) < 1 {
+		return errors.New("cog builds with --x-fast require 1 test.")
```
erbridge, line 300:
```suggestion
		return errors.New("cog builds with --x-fast require at least 1 test.")
```

pkg/config/data/config_schema_v1.0.json
```diff
@@ -161,6 +184,19 @@
           "description": "The default target for number of concurrent predictions. This setting can be used by an autoscaler to determine when to scale a deployment of a model up or down."
         }
       }
+    },
+    "tests": {
+      "$id": "#/properties/build/properties/tests",
```
meatballhat, line 189: This `$id` still doesn't make sense to me 😅 In the test yaml above, `tests` is at the same level as `build`, but this `$id` indicates `tests` is a sub-key of `build`. I expected the value to be `#/properties/tests`.

8W9aG: Resolved in [aa83e8b](https://github.com/replicate/cog/pull/2149/commits/aa83e8bbb0d70c2dc2bd8184e8982a13e827189b)

meatballhat (approved):

PR #2152, opened 2025-02-13
Send push timings to the server

NikhilSinha1:
### Summary

We want to know about how long it takes to build and push images. If we are introducing something called "fast push", we should probably know if it is actually faster or not than the same thing through the standard push flow.

### Test Plan

Do a push with the fast push flow and the standard push flow and see if they still work


pkg/docker/push.go
```diff
@@ -2,22 +2,58 @@ package docker
...
-			return err
+			console.Warnf("Failed to inspect image: %v", err)
 		}
-		webClient := web.NewClient(command, client)
+		parts := strings.Split(imageMeta.ID, ":")
+		if len(parts) != 2 {
+			console.Warn("Image ID was not of the form sha:hash")
+		} else {
+			imageID = parts[1]
+		}
```
meatballhat, line 46: nit: Ooh! Use [strings.Cut](https://pkg.go.dev/strings#Cut) instead!

pkg/docker/push.go
```diff
@@ -2,22 +2,58 @@ package docker
...
+	}
+
+	err = webClient.PostBuildStart(ctx, imageID, buildTime)
+	if err != nil {
```
meatballhat, line 50: nit: I don't see `err` being used later, so

```suggestion
	if err := webClient.PostBuildStart(ctx, imageID, buildTime); err != nil {
```

pkg/web/client.go
```diff
@@ -65,12 +69,47 @@ func NewClient(dockerCommand command.Command, client *http.Client) *Client {
...
+		"push_start_time": time.Now().UTC(),
+	}
+
+	jsonData, err := json.Marshal(jsonBody)
+	if err != nil {
+		return util.WrapError(err, "failed to marshal JSON for build start")
+	}
+
+	url := webBaseURL()
+	url.Path = strings.Join([]string{"", "api", "models", "build-start"}, "/")
```
meatballhat, line 85: nit: Given there's no variable data in this string, I think it should be a `const` instead.

pkg/web/client.go
```diff
@@ -65,12 +69,47 @@ func NewClient(dockerCommand command.Command, client *http.Client) *Client {
...
+	}
+
+	resp, err := c.client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusCreated {
+		return errors.New("Bad response from new version endpoint: " + strconv.Itoa(resp.StatusCode))
```
meatballhat, line 99: nit: it's not a huge deal in this context since the process is not long-lived, but `errors.New` should generally not be used like this. Instead, define a module-level `var` of the base error and then use error wrapping to include the variable bit.

meatballhat (approved):
⏱️👀

meatballhat (approved):
Nice 🙌🏼 !

PR #2153, opened 2025-02-14
Include `sha256:` prefix in generated runtime config files

meatballhat:
Since all previous runtime configs include the `sha256:` prefix expected by other components in the whole thing.

Connected to PLAT-848


NikhilSinha1 (approved):
I think overriding the json.Marshal is a potential future optimization but eh don't care rn

PR #2159, opened 2025-02-18
fix(python): invalid URL creation when output_file_prefix didn't end w…

KunalSin9h:
…ith `/`

In file upload docs (https://cog.run/http/#file-uploads): 

```json
{
    "output_file_prefix": "https://example.com/upload"
}
```

But in code, we are appending the `filename` directly into it without `/` in between. Then this url is used to send `PUT` request. 

https://github.com/replicate/cog/blob/e12f2374827d64a3e3b90c96436bddf8963eda58/python/cog/files.py#L17

What it caused is 404 not found, since url become like `/uploadimage.png`.

consiquences

To fix this, user `output_file_prefix` need `/` at the end of url all the time. Hence this will give error when user forgot this. 

```diff
{
-    "output_file_prefix": "https://example.com/upload",
+   "output_file_prefix": "https://example.com/upload/"
}
```


anish-sahoo:
Closing as the code path this PR touched (python/cog/files.py) no longer exists. File uploads moved to Rust Coglet, which now ensures a trailing slash before constructing the upload URL.

PR #2162, opened 2025-02-19
Fail python_packages on --x-fast

nevillelyh:
So that we don't attempt to parse requirements specs and instead pass
them as is to monobase, i.e. `uv pip compile` for validation


NikhilSinha1 (approved):
👍🏾

PR #2185, opened 2025-03-06
Convert files to data URLs on webhook post

8W9aG:
* When posting a webhook we can end up with files referencing /tmp folders that can’t be accessed by the consumer of the webhook
* Convert these files to data URLs so they can be consumed.

### DIFF
--- a/python/cog/server/webhook.py
+++ b/python/cog/server/webhook.py
@@ -7,6 +7,8 @@ from fastapi.encoders import jsonable_encoder
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
+from ..files import upload_file
+from ..json import upload_files
 from ..schema import PredictionResponse, Status, WebhookEvent
 from ..types import PYDANTIC_V2
 from .response_throttler import ResponseThrottler
@@ -52,6 +54,13 @@ def webhook_caller(webhook: str) -> Callable[[Any], None]:
                 )
             else:
                 dict_response = jsonable_encoder(response.dict(exclude_unset=True))
+
+            if "output" in dict_response:
+                dict_response["output"] = upload_files(
+                    dict_response["output"],
+                    upload_file=upload_file,  # type: ignore
+                )
+
             if Status.is_terminal(response.status):
                 # For terminal updates, retry persistently
                 retry_session.post(webhook, json=dict_response)


python/cog/server/webhook.py
```diff
@@ -47,11 +49,18 @@ def webhook_caller(webhook: str) -> Callable[[Any], None]:
...
+                response_object = response.model_dump(exclude_unset=True)
             else:
-                dict_response = jsonable_encoder(response.dict(exclude_unset=True))
+                response_object = response.dict(exclude_unset=True)
+
+            if "output" in response_object:
+                response_object["output"] = upload_files(
+                    response_object["output"],
+                    upload_file=upload_file,  # type: ignore
+                )
```
aron, line 60: What was the thinking of doing this here in the webhook handler rather than in the worker where we currently do the file uploads?

https://github.com/replicate/cog/blob/018aca50f9d9d43ff784f1ea239c7cede72dc7c0/python/cog/server/runner.py#L452-L453

python/tests/server/test_webhook.py
```diff
@@ -153,3 +156,28 @@ def test_webhook_caller_connection_errors():
...
+
+        c = webhook_caller("https://example.com/webhook/123")
+
+        payload = {
+            "status": Status.SUCCEEDED,
+            "output": Path(tmpfile.name),
+            "input": {},
+        }
+        response = PredictionResponse(**payload)
+        payload["output"] = "data:None;base64,aGVsbG8gd29ybGQ="
```
aron, line 175: Is `None` supposed to be here?
```suggestion
        payload["output"] = "data:text/plain;base64,aGVsbG8gd29ybGQ="
```

aron: Actually what is this test doing?

8W9aG: It tests that the webhook is sent data URLs rather than tmp file directories. I can change the mime-type however it seems odd to me that the `upload_file` function uses `None` as a default if that default is considered itself invalid. Can you comment on that?

python/cog/server/webhook.py
8W9aG: Had a chat with @aron and we will move this to the runner and see in the CI what breaks in terms of tests, updated in: [8707aba](https://github.com/replicate/cog/pull/2185/commits/8707abac640c5bff663512c0ad7134408488e4ee)

python/tests/server/test_webhook.py
```diff
@@ -153,3 +156,28 @@ def test_webhook_caller_connection_errors():
...
+            handle.write("hello world")
+
+        c = webhook_caller("https://example.com/webhook/123")
+
+        payload = {
+            "status": Status.SUCCEEDED,
+            "output": Path(tmpfile.name),
+            "input": {},
+        }
+        response = PredictionResponse(**payload)
```
NikhilSinha1, line 174: Ok just for my understanding, without the change we would've passed back the path to the `/tmp` file which then the caller could not access, so we're uploading the file (somewhere) and passing a URL back? And then the line below has a b64 encoded URL

Issue #2180, opened 2025-03-04
Support OpenAI-compatible ChatMessage type


andreasjansson:
For language models, it'd be good to support an input type that's consistent with the OpenAI API.

I imagine something like this:

```python
from cog import Input
from cog.types import ChatMessage

def run(
    messages: list[ChatMessage] = Input(),
) -> str:
    print(messages)
    return f"HELLO {messages[0]['role']}"
```

zeke:
Progress: https://github.com/replicate/cog/pull/2261

zeke:
Implementation in cog-runtime: https://github.com/replicate/cog-runtime/pull/71

anish-sahoo:
This is now supported in a similar way using `TypedDict/list[TypedDict]` or `list[dict]`, including JSON input through cog run/cog predict. Third-party chat message types can use `Annotated[..., Opaque]`. If a first-class `cog.types.ChatMessage` compatible with OpenAI’s API is still something you want to see in Cog, please reopen this issue or file a new one specifically for that type. Closing this issue due to inactivity.

PR #2186, opened 2025-03-06
Add ChatMessage as a first class cog type

8W9aG:
* Solves https://github.com/replicate/cog/issues/2180


python/cog/__init__.py
```diff
@@ -36,4 +37,5 @@
     "Input",
     "Path",
     "Secret",
+    "CommonChatSchemaChatMessage",
```
erbridge, line 40: Please use alphanumeric ordering :)

python/cog/predictor.py
```diff
@@ -48,14 +49,15 @@
 
 log = structlog.get_logger("cog.server.predictor")
 
-ALLOWED_INPUT_TYPES: List[Type[Any]] = [
+ALLOWED_INPUT_TYPES: List[Union[Type[Any], Type[CommonChatSchemaChatMessage]]] = [
```
erbridge, line 52: I don't think this adds anything does it? `Any` covers `CommonChatSchemaChatMessage`, doesn't it?

python/cog/__init__.py
8W9aG: Resolved in [ec7ac31](https://github.com/replicate/cog/pull/2186/commits/ec7ac313c0dbe7466c3973a584f8d4c9f067b83f)

python/cog/predictor.py
8W9aG: This solves the linting issue as seen here: https://github.com/replicate/cog/actions/runs/13705936291/job/38331101920

`UnionType` is apparently not covered by `Type[Any]`

erbridge: Wild. Ok!

andreasjansson:
Niceee!

Is there a reason to use `CommonChatSchemaChatMessage` instead of just calling it `ChatMessage`?

How do you test this with `cog predict`? Does `cog predict -i messages='[{"role": "user", "content": "hello"}]'` work?

8W9aG:
> Niceee!
> 
> Is there a reason to use `CommonChatSchemaChatMessage` instead of just calling it `ChatMessage`?
> 
> How do you test this with `cog predict`? Does `cog predict -i messages='[{"role": "user", "content": "hello"}]'` work?

Reason for the namespace `CommonChatSchema` was to namespace all the sub objects such as `Audio` and `Function`, however I can change the root to `ChatMessage` if it suits since its a union anyway.

I couldn't get `cog predict` to work on this but I can get it to recognise an API call, would you like me to extend the support to `cog predict`?

andreasjansson:
Yes I think `ChatMessage` is cleaner. And it would be great if we could get it working with cog predict so these functions can be tested locally.

8W9aG:
> Yes I think `ChatMessage` is cleaner. And it would be great if we could get it working with cog predict so these functions can be tested locally.

I believe this now complete

zeke:
This is sweet! @8W9aG can you take a pass at adding some docs for it?

The bottom of `docs/python.md` would probably be the best spot for it, alongside the other types like `Path` and `Secret`.

bfirsh:
See also https://replicatehq.slack.com/archives/C05600FDYTE/p1741707361371319

PR #2188, opened 2025-03-07
Tweak fast build cache and file exposure

nevillelyh:
- **Clean up build cache**
- **Only overwrite env.txt if different**
- **Bind mount . => /src during monobase.user build**


pkg/dockerfile/fast_generator.go
```diff
@@ -227,12 +228,12 @@ func (g *FastGenerator) generateMonobase(lines []string, tmpDir string) ([]strin
 
 	// The only input to monobase.build are these ENV vars
 	// Write them in tmp mount for layer caching
-	err := os.WriteFile(path.Join(tmpDir, "env.txt"), []byte(strings.Join(envs, "\n")), 0o644)
+	err := files.WriteIfDifferent(path.Join(tmpDir, "env.txt"), strings.Join(envs, "\n"))
```
nevillelyh, line 231: Overwriting files with same content changes its timestamp and invalidates layer cache. So only write if content changed.

pkg/dockerfile/fast_generator.go
```diff
@@ -16,10 +17,10 @@ import (
 )
 
 const FUSE_RPC_WEIGHTS_PATH = "/srv/r8/fuse-rpc/weights/sha256"
-const MONOBASE_CACHE_PATH = "/var/cache/monobase"
-const APT_CACHE_MOUNT = "--mount=type=cache,target=/var/cache/apt,id=apt-cache,sharing=locked"
```
nevillelyh, line 20: We're not caching `/var/cache/apt` for the APT tarball building stage. But those are usually small and rarely change. We can add cache later if it becomes problematic.

pkg/dockerfile/fast_generator.go
```diff
@@ -288,10 +288,12 @@ func (g *FastGenerator) installPython(lines []string, tmpDir string) ([]string,
 		return nil, err
 	}
 	if requirementsFile != "" {
+		srcMount := "--mount=type=bind,ro,source=.,target=/src"
```
nevillelyh, line 291: Python wheel/zip files in `requirements.txt` cannot be in `.dockerignore` or they'll not be exposed via this bind mount. OTOH they'll be baked into the `/src` layer unless we parse requirements, check existence and add them to `--exclude` in the `COPY` step, overkill for now assuming most private packages are small.

8W9aG (approved):

PR #2192, opened 2025-03-12
update token config in `cog init` actions workflow template

zeke:
You used to be able to use a Replicate API token as a CLI auth token, but that's no longer supported.

This PR updates the `push.yaml` GitHub Actions workflow that's generated by `cog init`, so users know to use a "CLI auth token" instead of an API token.

See:

- https://github.com/replicate/setup-cog/pull/45
- https://github.com/replicate/frontend/pull/677


superhighfives (approved):
LGTM 👍

PR #2194, opened 2025-03-12
Fix fast tarball tmp dir

nevillelyh:
Follow up to #2191
Make sure that separate temp dirs in `fast_push.go` are consistent with `fast_generator.go`.
Add a new tmp dir for tarballs extracted from images, so that they don't cause cache invalidation.


NikhilSinha1 (approved):
Confirmed works - actually used this to push a fast push model

PR #2200, opened 2025-03-14
Add documentation for concurrency support

philandstuff:
This adds documentation for the new concurrency support in cog 0.14.0.

I moved the "streaming output" section to be under the "output" heading; it fits better here, and it means I can introduce async predictors before talking about AsyncIterator and AsyncConcatenateIterator.  As a result this is probably better viewed as a split diff instead of a unified diff.


erbridge (approved):

zeke:
I missed the boat on reviewing this. Followup PR here: https://github.com/replicate/cog/pull/2205

PR #2203, opened 2025-03-14
Warn the user when they use deprecated fields

8W9aG:
* These fields have been deprecated for a while, make it clear to the user that they will be removed in future versions.
* Add the necessary documentation.

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -7,8 +7,7 @@ It has three keys: [`build`](#build), [`image`](#image), and [`predict`](#predic
 ```yaml
 build:
   python_version: "3.11"
-  python_packages:
-    - pytorch==2.0.1
+  python_requirements: requirements.txt
   system_packages:
     - "ffmpeg"
     - "git"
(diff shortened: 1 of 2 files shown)


zeke (approved):
This looks good.

There is some still-relevant leftover content in the `python_packages` section of these docs about installing git-hosted packages and pinning package versions. I think it would be good to yank that out of the now-deprecated docs for `python_packages` and give them a new home under the `python_requirements` section, so they're not overlooked or assumed to be outdated.

8W9aG:
> This looks good.
> 
> There is some still-relevant leftover content in the `python_packages` section of these docs about installing git-hosted packages and pinning package versions. I think it would be good to yank that out of the now-deprecated docs for `python_packages` and give them a new home under the `python_requirements` section, so they're not overlooked or assumed to be outdated.

Done! Let me know if it looks good.

zeke (approved):
Excellent 👍🏼

NikhilSinha1 (approved):
LGTM

PR #2205, opened 2025-03-14
Tweak new concurrency docs

zeke:
This a little followup to https://github.com/replicate/cog/pull/2200 with some minor changes to the concurrency docs.

Commits are significant.


docs/yaml.md
```diff
@@ -154,6 +154,13 @@ This stanza describes the concurrency capabilities of the model. It has one opti
 
 The maximum number of concurrent predictions the model can process.  If this is set, the model must specify an [async `predict()` method](python.md#async-predictors-and-concurrency).
 
+For example:
```
zeke, line 157: How do model authors determine a good value to use for this?

Is there a default?

Can we give some guidance on reasonable values here?

PR #2210, opened 2025-03-18
Check that dockerignore is compatibile with cog

8W9aG:
* We can’t ignore the .cog tmp directory
* Parse the ignore file and check if it ignores the .cog directory
* If it does, fail and alert the user


go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/moby/term v0.5.0
 	github.com/replicate/go v0.0.0-20250205165008-b772d7cd506b
+	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
```
meatballhat, line 23: Any alternatives out there that haven't been abandoned? 😬

pkg/image/build.go
```diff
@@ -44,6 +49,11 @@ func Build(cfg *config.Config, dir, imageName string, secrets []string, noCache,
 	_ = os.Remove(bundledSchemaFile)
 	_ = os.Remove(bundledSchemaPy)
 
+	err := checkCompatibleDockerIgnore(dir)
+	if err != nil {
```
meatballhat, line 53: looks like this and below could use an `iferr` fix to prevent variable redeclaration

pkg/image/build.go
```diff
@@ -403,3 +413,41 @@ func restoreDockerignore() error {
...
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		patterns = append(patterns, line)
+	}
+	return patterns, scanner.Err()
+}
+
+func checkCompatibleDockerIgnore(dir string) error {
+	dockerIgnorePath := path.Join(dir, ".dockerignore")
```
meatballhat, line 434: This should use `path/filepath` instead of `path` 👍🏼

go.mod
8W9aG: I can't find any, but let me know if you can suggest an alternative

pkg/image/build.go
8W9aG: Resolved in [e474b84](https://github.com/replicate/cog/pull/2210/commits/e474b8492886f6521760f8277c1f1c44f4eb916f)

go.mod
meatballhat: 🤷🏼

meatballhat (approved):
🙌🏼

PR #2160, opened 2025-02-18
Support Python requirements package[extra,...]==version

nevillelyh:
Apparently this works:
`uv pip install 'imageio[ffmpeg]==2.37.0'`

And pulls in `imageio-ffmpeg==0.6.0`

PR #2211, opened 2025-03-18
Revert "Support Python requirements package[extra,...]==version (#2160)"

nevillelyh:
This reverts commit bc4b95e070b271e06d796d335986791908271fd0.


nevillelyh:
Raised by #2211

No longer needed as we're no longer parsing requirements in `python_packages`.

PR #2214, opened 2025-03-19
Allow cancelation on async models

### LINKED ISSUE #2212: Async cancelation returns 404 when concurrency.max is set.
`cog==0.14.1`

`cog.yaml`:
```
build:
  python_version: "3.8"
concurrency:
  max: 2
```

`predict.py`:
```
import asyncio

from cog import BasePredictor


    async def predict(self, i: int) -> str:
        try:
            print('starting async prediction')
            for x in range(i):
                print(f'prediction in progress {x + 1}/{i}')
                await asyncio.sleep(1)
            print('completed async prediction')
            return f'*{i}*'
        except asyncio.CancelledError as e:
            print('prediction canceled')
            raise e
```

Steps to repro:
```
# start cog
python3 -m cog.server.http

# prediction request
curl -v -X POST -H 'Prefer: respond-async' \
  -H 'Content-Type: application/json' \
  --data '{"id":"p01", "input":{"i":600},"webhook":"http://localhost:5150/webhook"}' \
  http://localhost:5000/predictions

# cancellation request
curl -v -X POST -H 'Content-Type: application/json' http://localhost:5000/predictions/p01/cancel
```

And got:
```
*   Trying 127.0.0.1:5000...
* Connected to localhost (127.0.0.1) port 5000 (#0)
> POST /predictions/p01/cancel HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.81.0
> Accept: */*
> Content-Type: application/json
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< date: Tue, 18 Mar 2025 19:05:16 GMT
< server: uvicorn
< content-length: 2
< content-type: application/json
<
* Connection #0 to host localhost left intact
{}%
```

Cancellation works correctly and responds with 200 if I comment out `concurrency.max` from `cog.yaml`.

philandstuff:
Fixes #2212.

Previously, we had `runner.is_busy()`, which for sync models returned True if a prediction was currently being processed and False otherwise.  But for concurrent models, we can be partially busy.

The cancel method had a fail-fast check to return if we is_busy() returned False.  But with concurrent models, is_busy() has been generalized to mean "is completely full; cannot accept more work".  This meant we could only cancel predictions if we were currently processing concurrency.max predictions.

As it turns out, this fail-fast check is unnecessary.  runner.cancel() checks that the targeted prediction exists and isn't done already.  So we can just drop this check.

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -486,8 +486,6 @@ def create_app(  # pylint: disable=too-many-arguments,too-many-locals,too-many-s
         """
         Cancel a running prediction
         """
-        if not runner.is_busy():
-            return JSONResponse({}, status_code=404)
         try:
             runner.cancel(prediction_id)
         except UnknownPredictionError:


python/cog/server/http.py
```diff
@@ -486,8 +486,6 @@ async def cancel(prediction_id: str = Path(..., title="Prediction ID")) -> Any:
         """
         Cancel a running prediction
         """
-        if not runner.is_busy():
-            return JSONResponse({}, status_code=404)
         try:
```
aron, line 489: 404 was a weird response code to pick…

aron (approved):
Thanks @philandstuff!

python/cog/server/http.py
philandstuff: the URL identifies a specific prediction, so if we're not running that prediction then 404 is appropriate.

There's maybe an argument that if we _did_ run that prediction but now it's completed, we could return 409.

erbridge (approved):
Great!

PR #2216, opened 2025-03-19
Support Optional in Input

8W9aG:
* Add support for `Optional` arguments into a `predict` or `train` method


NikhilSinha1 (approved):
Discussed async, LGTM

Issue #2206, opened 2025-03-17
Pydantic >2 and optional image Inputs


jschoormans:
Hi, 
I'm developing a cog model that requires a dependeny, which needs pydantic >2. 
Now, I'm noticing an issue with optional image inputs. The model returns an error in that case. I'll provide a minimal working example: 

```
from cog import BasePredictor, Path, Input

class Predictor(BasePredictor):
    def predict(
        self,
        test_image: Path = Input(description="Test image", default=None),
    ) -> Path:
        """Run a single prediction on the model"""

        return Path("./hello.webp")
```

```
build:
  python_version: "3.11"

  python_packages:
    - "pydantic==2.1.0"


predict: "predict.py:Predictor"

```

Runs perfectly fine with cog predict -i test_image=@...; however with no input it gives this error:

```

[...]

PR #2223, opened 2025-03-24
add Nonetype

jschoormans:
See this issue: https://github.com/replicate/cog/issues/2206#issuecomment-2749298089 
Optional inputs give an error because Nonetype is not in the allowed types list.


anish-sahoo:
Closing this as this has been solved as of Issue [#2206](https://github.com/replicate/cog/issues/2206).

PR #2228, opened 2025-03-26
Add more data to error strings

8W9aG:
* Allow us to debug fast push better by including URLs and UUIDs when appropriate

### DIFF
--- a/pkg/monobeam/client.go
+++ b/pkg/monobeam/client.go
@@ -85,7 +85,7 @@ func (c *Client) UploadFile(ctx context.Context, objectType string, digest strin
 	if resp.StatusCode == http.StatusConflict {
 		return nil
 	} else if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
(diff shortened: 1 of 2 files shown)


pkg/monobeam/client.go
```diff
@@ -85,7 +85,7 @@ func (c *Client) UploadFile(ctx context.Context, objectType string, digest strin
 	if resp.StatusCode == http.StatusConflict {
 		return nil
 	} else if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusCreated {
-		return errors.New("Bad response from monobeam: " + strconv.Itoa(resp.StatusCode))
+		return errors.New("Bad response from monobeam for URL " + uploadUrl.String() + ": " + strconv.Itoa(resp.StatusCode))
```
NikhilSinha1, line 88: nit:

```suggestion
		return errors.New("Bad response from monobeam for upload " + uploadUrl.String() + ": " + strconv.Itoa(resp.StatusCode))
```

NikhilSinha1 (approved):
Yup, prints more information for errors so we have better understanding of what went wrong

PR #2230, opened 2025-03-27
Create scanner for removing comments and continuations

8W9aG:
* When reading the requirements txt, handle line continuations and remove comments

### DIFF
--- a/pkg/requirements/requirements.go
+++ b/pkg/requirements/requirements.go
@@ -49,7 +49,11 @@ func ReadRequirements(path string) ([]string, error) {
 	scanner.Split(scanLinesWithContinuations)
 	requirements := []string{}
 	for scanner.Scan() {
-		requirements = append(requirements, scanner.Text())
+		requirementsText := strings.TrimSpace(scanner.Text())
+		if len(requirementsText) == 0 {
+			continue
+		}
+		requirements = append(requirements, requirementsText)
 	}
 	return requirements, nil
 }
@@ -93,10 +97,6 @@ func scanLinesWithContinuations(data []byte, atEOF bool) (advance int, token []b
 		if advance == len(data) {
 			break
 		}
-	}
-	strTok := strings.TrimSpace(string(token))
-	if len(strTok) == 0 {
-		return advance, nil, err
 	}
 	return advance, token, err
 }
--- a/pkg/requirements/requirements.go
+++ b/pkg/requirements/requirements.go
@@ -1,29 +1,19 @@
 package requirements
 
 import (
+	"bufio"
 	"errors"
-	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/replicate/cog/pkg/util/files"
-
-	"github.com/replicate/cog/pkg/config"
 )
 
 const REQUIREMENTS_FILE = "requirements.txt"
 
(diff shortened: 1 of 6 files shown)


pkg/requirements/requirements_test.go
```diff
@@ -7,36 +7,49 @@ import (
...
+	reqFile := path.Join(srcDir, "requirements.txt")
+	err := os.WriteFile(reqFile, []byte("torch==\\\n2.5.1"), 0o644)
+	require.NoError(t, err)
+
+	requirements, err := ReadRequirements(reqFile)
+	require.NoError(t, err)
+	require.Equal(t, []string{"torch==2.5.1"}, requirements)
+}
+
+func TestReadRequirementsStripHashes(t *testing.T) {
```
aron, line 46:
```suggestion
func TestReadRequirementsStripComments(t *testing.T) {
```

Also I think this needs to test line comments not just trailing comments.

pkg/requirements/requirements.go
```diff
@@ -48,3 +38,59 @@ func CurrentRequirements(tmpDir string) (string, error) {
...
+		if atEOF || len(data) == 0 {
+			break
+		}
+		if token == nil {
+			token = []byte{}
+		}
+		if data[advance] == '#' {
+			inHash = true
+		}
+		if data[advance] == '\n' {
```
aron, line 76: I think based on the comment on line 47, this should handle CRLF characters too.

pkg/requirements/requirements.go
```diff
@@ -48,3 +38,59 @@ func CurrentRequirements(tmpDir string) (string, error) {
...
+	}
+	// Use scanner to handle CRLF endings
+	scanner := bufio.NewScanner(fh)
+	scanner.Split(scanLinesWithContinuations)
+	requirements := []string{}
+	for scanner.Scan() {
+		requirementsText := strings.TrimSpace(scanner.Text())
+		if len(requirementsText) == 0 {
+			continue
+		}
```
aron, line 55: Not a big deal but as we're implementing a custom splitter we could have `scanLinesWithContinuations` return `nil` for any whitespace only and comment lines and then they'll be skipped here entirely.

aron (changes requested):
Thanks for fixing this! In general the approach looks great. Before we merge I'd like to see tests for CRLF line endings and full line comments just to complete the coverage.

pkg/requirements/requirements.go
8W9aG: Resolved in [16a4ca8](https://github.com/replicate/cog/pull/2230/commits/16a4ca850e8dd3b46527aadeec01c6b3498cbee6)

pkg/requirements/requirements.go
8W9aG: Resolved in [d4397ba](https://github.com/replicate/cog/pull/2230/commits/d4397bafd8e53a37f13ea408fc66d359991a4766)

8W9aG: I revert this in [36131fa](https://github.com/replicate/cog/pull/2230/commits/36131fa1b5839b8c73029621f566d76a96623bc9). It seems that if you return nil as a token the scanner stops scanning in the loop.

aron (approved):
Nice, thanks for fixing this!

PR #2232, opened 2025-03-28
Remove checks for static schema

8W9aG:
* Static schema checks are now made redundant by Pydantic 2.11 adding in additional metadata
* Use the tests to check for targeted dictionary equality instead.


michaeldwan (approved):
lgtm, though I don't trust my judgement yet. 

Also, the `web.TestDoFileChallenge` test in go failed on ubuntu, but succeeded after I retried it. I don't think it's related though.

PR #2233, opened 2025-03-28
Configure `uv` + misc fixes while setting up a new dev environment

### LINKED ISSUE #2167: Use uv instead of pip/pyenv
This is where the world is going.

michaeldwan:
I ran into some issues while setting up a new dev environment (macOS+vscode, but the improvements are universal) so I aggregated the fixes into this PR. 

I broke them up into individual commits for easy  review or cherry picking:
1. I had troubles with python from asdf, but uv was nearly a drop in replacement. 
2. added scripts/setup for bootstrapping the environment
3. remove deprecated ruff lint configs that were causing warnings
4. go.mod and go.sum were getting rewritten anytime I ran `go`, these were probably accidentally left out of another commit, so I'm just adding them here to stop seeing them change

Closes #2167


meatballhat (approved):
👏🏼 🎉  Thrilled to see `uv` here!

zeke:
Nice!

PR #2235, opened 2025-03-31
Allow `pathlib.Path` and descendants as uploadables

meatballhat:
so that more types than `cog.types.Path` will be treated as path-like uploadables.

Connected to PLAT-1107

### DIFF
--- a/python/cog/json.py
+++ b/python/cog/json.py
@@ -1,4 +1,5 @@
 import io
+import pathlib
 from datetime import datetime
 from enum import Enum
 from types import GeneratorType
@@ -61,9 +62,9 @@ def upload_files(obj: Any, upload_file: Callable[[io.IOBase], str]) -> Any:
         return {key: upload_files(value, upload_file) for key, value in obj.items()}
     if isinstance(obj, list):
         return [upload_files(value, upload_file) for value in obj]
-    if isinstance(obj, io.IOBase):
-        return upload_file(obj)
-    if callable(getattr(obj, "open", None)):
+    if isinstance(obj, pathlib.Path):
         with obj.open("rb") as f:
             return upload_file(f)
+    if isinstance(obj, io.IOBase):
+        return upload_file(obj)
     return obj
--- a/python/cog/json.py
+++ b/python/cog/json.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     np = None
 
-from .types import PYDANTIC_V2, Path
+from .types import PYDANTIC_V2
 
 
 def make_encodeable(obj: Any) -> Any:  # pylint: disable=too-many-return-statements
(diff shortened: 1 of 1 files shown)


python/cog/json.py
```diff
@@ -61,9 +61,9 @@ def upload_files(obj: Any, upload_file: Callable[[io.IOBase], str]) -> Any:
...
     if isinstance(obj, list):
         return [upload_files(value, upload_file) for value in obj]
-    if isinstance(obj, Path):
-        with obj.open("rb") as f:
-            return upload_file(f)
     if isinstance(obj, io.IOBase):
         return upload_file(obj)
+    if callable(getattr(obj, "open", None)):
+        with obj.open("rb") as f:
+            return upload_file(f)
```
aron, line 68: This makes me a bit nervous. The contract isn't purely that the object conforms to `open()` right? It also needs `seek`, `read` etc...

meatballhat: @aron it's definitely a bit more loose than I'd like 😅 I'm happy to tighten it up.

aron: Hmm, I think it's okay, perhaps we should assert that `f` is an instance of `io.IOBase`…

meatballhat: mmm yeah I like that

nevillelyh: Asking coz cog-runtime doesn't handle upload/download in Python, but rather in the Go server. So the `Path` seen by Python predictor code is always a local file. The Python runner serializes it as `file:///<path>` in the JSON output file & passes it to Go server, which then picks up the file for upload.
Just for context so we keep it in mind for converging cog runtimes later.

aron: @meatballhat I think we're looking at this wrong. `cog.Path` really only exists to provide a Pydantic variant of `pathlib.Path` for schema generation/validation (If we ignore the `URLPath` shenanigans).

Anywhere we are dealing with outputs I think we only care about `pathlib.Path`, so that's probably the simplest change here.

meatballhat: @aron I'm good with assuming `pathlib.Path`, but in order to handle both that and `cog.types.Path` we'll need to do something like this change to move away from `isinstance`, or use `isinstance` with a typing protocol that both `pathlib.Path` and `cog.types.Path` fulfill.

meatballhat: Partial retraction:

```
[ins] In [6]: cog_path = cog.types.Path("./ok-bud")

[ins] In [7]: pathlib_path = pathlib.Path("./ok-bud")

[ins] In [8]: isinstance(cog_path, pathlib.Path)
Out[8]: True

[ins] In [9]: isinstance(pathlib_path, cog.types.Path)
Out[9]: False 

```

aron: Yeah, this is what I meant. I think as long as we don't extend `cog.Path` with behaviour beyond the Pydantic types we're fine in this specific case and I think pretty much in the general case too.

meatballhat: I think my extreme dislike of `isinstance` is showing 😅

aron: You nerd-sniped me. It looks like Python 3.6 introduced `os.PathLike` so you can detect a path-like object (that can be passed to `open()`) by implementing `__fspath__()`. Though I still think that probably still requires an `isinstance` check...

https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol

PR #2236, opened 2025-03-31
Add local image flag for fast context loading

8W9aG:
* Add a new experimental flag —x-localimage for use with —x-fastpush
* When activated, the flag uses the .cog tmp directory as an rsync for the source code without
weights
* It separates the different build steps into different build contexts
* On run/predict/train/serve it loads the weights in as volume mounts
* This allows it to not reload the context each time a small file is changed for weights


pkg/dockerfile/fast_generator.go
```diff
@@ -362,3 +442,145 @@ func (g *FastGenerator) validateConfig() error {
...
+	return nil
+}
+
+func copyFileWithTimes(destination string, src string) error {
+	console.Debug("Copying " + destination + " to " + src)
+	info, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+	err = files.CopyFile(src, destination)
```
nevillelyh, line 535: Do we want to hard link instead? That might save you from setting TS, ownership, etc.

pkg/dockerfile/fast_generator.go
```diff
@@ -362,3 +442,145 @@ func (g *FastGenerator) validateConfig() error {
...
+	info, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+	mode := info.Mode()
+
+	if !exists {
+		switch {
+		case mode.IsDir():
+			err := os.Mkdir(destination, 0o777)
```
nevillelyh, line 561: Is it always 777 though? Unlikely but could use rely on perm? Or what if perm changes?

pkg/dockerfile/fast_generator.go
```diff
@@ -362,3 +442,145 @@ func (g *FastGenerator) validateConfig() error {
...
+			if err != nil {
+				return err
+			}
+		case mode.IsRegular():
+			err := copyFileWithTimes(destination, src)
+			if err != nil {
+				return err
+			}
+		}
+	} else if mode.IsRegular() {
```
nevillelyh, line 571: Hard links work on Mac OS, so maybe it's easier than copying? IIRC they're guaranteed to have the same time, ownership etc. We just need to check src & dst have the same inode.

You might still wanna deal with symlinks though?

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [4fa0ed5](https://github.com/replicate/cog/pull/2236/commits/4fa0ed545dc5606a00dde29247dea5f5f5e6c941)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [817ec3d](https://github.com/replicate/cog/pull/2236/commits/817ec3d92af297853f08f2148b048d7724f4feb3)

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [5d27e08](https://github.com/replicate/cog/pull/2236/commits/5d27e0886afc2fceb29be1f2d5963905a1138e91)

nevillelyh (approved):

pkg/cli/predict.go
```diff
@@ -76,22 +77,27 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
 		}
+
 		if cfg.Build.Fast {
 			buildFast = cfg.Build.Fast
 		}
 
-		if imageName, err = image.BuildBase(cfg, projectDir, buildUseCudaBaseImage, DetermineUseCogBaseImage(cmd), buildProgressOutput); err != nil {
-			return err
-		}
+		if buildFast {
```
NikhilSinha1, line 85: nit: can you just throw this into the block above
```
if cfg.Build.Fast {
    buildFast = true
    imageName = config.DockerImageName(projectDir)
} else {
    ...
}

File: pkg/dockerfile/fast_generator.go [hunk 5]
```diff
@@ -362,3 +442,150 @@ func (g *FastGenerator) validateConfig() error {
...
+	info, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+	mode := info.Mode()
+
+	if !exists {
+		switch {
+		case mode.IsDir():
+			err := os.Mkdir(destination, mode.Perm())
```
NikhilSinha1, line 566: May need mkdirall?

pkg/dockerfile/fast_generator.go
```diff
@@ -362,3 +442,150 @@ func (g *FastGenerator) validateConfig() error {
...
+			return filepath.SkipDir
+		}
+
+		relPath, err := filepath.Rel(g.Dir, path)
+		if err != nil {
+			return err
+		}
+
+		// Skip weights, we handle them separately
+		for _, weight := range weights {
```
NikhilSinha1, line 482: Hm for every file in the filepath.walk we go through every weight? Maybe better to just make a set (go doesn't have sets I think so a `map[string]bool`) before the `filepath.walk` and then check if the `relPath` is in the set

pkg/dockerfile/fast_generator.go
```diff
@@ -315,16 +376,43 @@ func (g *FastGenerator) installPython(lines []string, tmpDir string) ([]string,
...
 	// Install /src
 
+	srcDir, err := BuildCogTempDir(g.Dir, srcBuildDir)
+	if err != nil {
+		return nil, err
+	}
+
+	// Rsync local src with our srcdir
+	if g.localImage {
+		err := g.rsyncSrc(srcDir, weights)
```
NikhilSinha1, line 386: Could just be me not understanding correctly but what we're doing here
- Symlinking everything in g.Dir to a subdirectory
- Hard link to all the symlinks?

I guess I'm a bit confused why we're doing the `rsync` first?

pkg/cli/predict.go
8W9aG: Unfortunately I don't think we can, config fast build overwrites any CLI flag which is the first block, but in the 2nd block we need to know whether the fast build flag is in general on and overwrite the `imageName`

pkg/dockerfile/fast_generator.go
8W9aG: Resolved in [1d038ef](https://github.com/replicate/cog/pull/2236/commits/1d038ef58ef27d3ea15549e1aebfd77da5db472f)

pkg/dockerfile/fast_generator.go
8W9aG: Our process is:
1. If a local image flag is used, created a src directory in `.cog`
2. Create the necessary hard links and directories to mirror whats in the users current directory (sans weights and dockerignore directives), if it is symlink use the target path
3. Tag that as a unique build context so it only changes the docker build when the user changes their code

8W9aG:
Going to merge this to build a new release

PR #2242, opened 2025-04-02
Update `cog init` command to use `python_requirements` and `requirements.txt`

zeke:
The `python_packages` key in cog.yaml is [deprecated](https://github.com/replicate/cog/blob/4ce6fb90b3812d46b753413e5dc4b3b689ba5785/docs/yaml.md#python_packages) and may eventually stop working entirely.

This PR updates the `cog init` command to use `python_requirements` instead, and adds a starter `requirements.txt` file to the mix with some helpful comments.


aron (approved):
Nice!

PR #2244, opened 2025-04-02
Allow fast as a build property in cog.yaml

8W9aG:
* Allows fast to be defined as a build property in cog.yaml to prevent the constant use of —x-fast flags on the CLI.


nevillelyh (approved):
How is this related to #2222?

8W9aG:
> How is this related to #2222?

There is an additional check to validate the schema beyond go JSON marshalling, this adds it to that schema

PR #2264, opened 2025-04-17
Add python_overrides

8W9aG:
* Allows users to specify pip package versions in direct conflict with pips dependency management in case they need to force through a specific version


test-integration/test_integration/fixtures/overrides-project/overrides.txt
```diff
@@ -0,0 +1 @@
+numpy==1.26.4
```
nevillelyh, line 1: We’re only testing whether build passes but not which version was installed? Maybe better if we add a `numpy<=1.26.3` here, basically 2 conflicting req

nevillelyh (approved):

test-integration/test_integration/fixtures/overrides-project/overrides.txt
8W9aG: I can add an int test for checking the numpy version in the predict method

8W9aG: Resolved in [1a0625c](https://github.com/replicate/cog/pull/2264/commits/1a0625c545c577ff43e664df67966c8ffaf996dc)

Issue #2253, opened 2025-04-11
Don't force pydantic to downgrade to <2


bjhargrave:
https://github.com/replicate/cog/blob/beba0424aba813033c1d03eeb3bae9f3d17a5be7/pkg/dockerfile/standard_generator.go#L435-L437

This code overrides the version of pydantic which may have been installed via 

```yaml
python_requirements: requirements.txt
```

To fix this undesired downgrade, I need to add

```yaml
  run:
    # Restore pydantic version which cog forced to <2
    - --mount=type=cache,target=/root/.cache/pip pip install "$(grep 'pydantic==' /tmp/requirements.txt)"
```

But I don't think cog should be forcing a downgrade of pydantic.

Clement-Lelievre:
This might be the reason why I [get a freeze](https://github.com/replicate/replicate-python/issues/424) with no logs whatsoever? My model outputs an instance of pydantic's `BaseModel` and I pinned a pydantic version >2 in my requirements

bjhargrave:
> can you let me know more about your setup so I can attempt to replicate this and test against it

See https://github.com/bjhargrave/cog-models/tree/main/ibm-granite/granite-3.3-8b-instruct for my setup.

8W9aG:
> > can you let me know more about your setup so I can attempt to replicate this and test against it
> 
> See https://github.com/bjhargrave/cog-models/tree/main/ibm-granite/granite-3.3-8b-instruct for my setup.

I can confirm the clobbering in this case, I will investigate more and hopefully have a fix soon!

PR #2273, opened 2025-04-22
Prevent pydantic 2 clobbering

8W9aG:
* Fixes: https://github.com/replicate/cog/issues/2253


michaeldwan (approved):
Take this approval with a grain of salt... Change seems sensible, but I don't know enough about pedantic to understand the reason. :)

PR #2274, opened 2025-04-23
Environment variables in cog.yaml

### LINKED ISSUE #291: Environment variables in cog.yaml

michaeldwan:
This adds support for declaring environment variables in the `cog.yaml` file. Closes #291 

For example, this `cog.yaml` file:
```yaml
environment:
  - COLOR=orange
```

would result in the `COLOR` env variable being saved in the model image. 

Note: this is only the cog portion, we have a separate PR that enables env vars on replicate.


nevillelyh:
While impossible to be exhaustive, should we at least add some heuristics to disallow any env we might over ride in prod, to prevent surprises?

* Basic paths: `PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH`, `VIRTUAL_ENV`, `PYTHONUNBUFFERED`
* Replicate: `R8_*`, `REPLICATE_*`
* Nvidia: `LIBRARY_PATH`, `CUDA_*`, `NVIDIA_*`, `NV_*`
* PGet: `PGET_*`, `HF_ENDPOINT`, `HF_HUB_ENABLE_HF_TRANSFER`
* K8s: `KUBERNETES_*`

Sources:
* https://github.com/replicate/monobase/blob/main/src/monobase/activate.sh
* https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/internal/config/image/envvars.go
* https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime#environment-variables-oci-spec
* Spot checking some prod model containers

michaeldwan:
@nevillelyh awesome thanks. I'm all for a denylist. I'll update

michaeldwan:
@meatballhat would love to hear your thoughts on variable name restrictions!

michaeldwan:
Current thinking based on @nevillelyh's excellent list...

I'm going to add a `build.environment` entry that'll land in the image at _**build time only**_ as `ARG`s. Since these don't impact runtime I don't think there's as many foot guns with overriding things like `PATH`, `LD_LIBRARY_PATH`, and python stuff. 

For `environment` we restrict the following:
- Replicate vars `R8_*` and `REPLICATE_*`, though I'm wondering if there's a legit use case for user code to namespace variables with `REPLICATE_`, like `REPLICATE_TOKEN_`, or settings that apply to code running on us. Might be fine though, we always remove this restriction later!
- Nvidia `CUDA_*`, `NVIDIA_*`, `NV_*`
- pget: `PGET_*`, `HF_ENDPOINT`, `HF_HUB_ENABLE_HF_TRANSFER`
- K8s: `KUBERNETES_*`

I'm still waffling on `PATH` and `LD_LIBRARY_PATH`. There's valid reasons to set these, and they're already a list of paths, so maybe we allow them?

nevillelyh:
If the user wants to pass `REPLICATE_TOKEN_`, it'll likely be to code they control, so they can prefix it with `<MYORG>_REPLICATE_TOKEN` to de-conflict?

We actually manipulate `PATH`, `PYTHONPATH`, `LD_LIBRARY_PATH` at build time, more specifically:
* Install CUDA, CuDNN, and base venv in `/srv/r8/monobase` as "base layer", i.e. layers we skip from pushing
* Install "user layer" with `source activate.sh && monobase.user --requirements user-requirements.txt` which:
    * Adds `python`, `python3`, etc. from base venv to `PATH`
    * Adds base venv `site-packages` to `PYTHONPATH`
    * Adds CUDA `bin`, `lib`, etc. to `PATH` & `LD_LIBRARY_PATH`

These are needed for Python packages that compile native code during install and/or require dependencies in the base layer, e.g. Torch, Cython, etc. And since we don't allow `run` commands and therefore custom installs (like clone repo, install additional headers, compilers, etc) in fast build, I can't see many causes where a package needs to override these to install, and in those cases we usually just add the extra deps in monobase, e.g. [Cython](https://github.com/replicate/monobase/pull/140) most recently.

pkg/config/data/config_schema_v1.0.json
```diff
@@ -207,6 +207,20 @@
...
         "additionalItems": true
       }
+    },
+    "environment": {
+      "$id": "#/properties/properties/environment",
+      "type": [
+        "array",
+        "null"
+      ],
+      "description": "A list of environment variables to make available at build time, in the format `NAME=value`",
```
aron, line 217: Is this correct? I think this is at runtime no?

test-integration/test_integration/test_predict.py
```diff
@@ -555,3 +555,24 @@ def test_predict_fast_build(docker_image):
...
+def test_predict_env_vars(docker_image, cog_binary):
+    project_dir = Path(__file__).parent / "fixtures/env-project"
+    build_process = subprocess.run(
+        [cog_binary, "build", "-t", docker_image],
+        cwd=project_dir,
+        capture_output=True,
+    )
+    assert build_process.returncode == 0
+    result = subprocess.run(
+        [cog_binary, "predict", "--debug", docker_image, "-i", "name=TEST_VAR"],
```
aron, line 569: Can we update the other call sites to use `cog_binary` for consistency.

aron (approved):
This looks good to me! Great to see this in here. A couple of questions:

1. Why is the `environment` field typed as a list of strings rather than a map of key/value pairs?
2. Have we thought about being able to provide environment variables dynamically the calling environment much like docker compose e.g. `- "XYZ_API_TOKEN=${XYZ_API_TOKEN:-default-value}"`? This would allow a user to change configuration without editing the file. Doesn't need to be in this PR but curious on your thoughts.

test-integration/test_integration/test_predict.py
michaeldwan: Yep! I'll submit another PR to keep this tidy.

michaeldwan:
Thanks @aron, and good questions!

> 1. Why is the `environment` field typed as a list of strings rather than a map of key/value pairs?
> 2. Have we thought about being able to provide environment variables dynamically the calling environment much like docker compose e.g. `- "XYZ_API_TOKEN=${XYZ_API_TOKEN:-default-value}"`? This would allow a user to change configuration without editing the file. Doesn't need to be in this PR but curious on your thoughts.

I made this a string array instead of `key: value` to match docker compose's rules, and to avoid a host of yaml deserialization quirks around quotes, multiline values, and boolean-ish values (like `true`) that we want to pass through unaltered. 

Along with `KEY=VALUE`, compose supports `KEY` without a value to set it from the host environment. I backed that out until I understood how it would behave in prod. But I think that feature would cover half the GH issues asking for env vars at build time. I'll make an issue to track that and we can revisit shortly.

Clement-Lelievre:
hey,
I can see this was released in the v0.14.9, current latest is 0.15.2 but I still do not see this reflected in the `[latest docs](https://github.com/replicate/cog/blob/main/docs/yaml.md)` nor does it work locally, am I missing smthg?

Issue #2280, opened 2025-04-24
Cog malformed requirements.txt


dsinha244:
Following is my requirements.txt file

[requirements.txt](https://github.com/user-attachments/files/19897527/requirements.txt)

This is my cog file

```
build:
  system_packages:
    - "libgl1-mesa-glx"
    - "libglib2.0-0"
    - wget
  python_version: "3.10"
  python_requirements: "lib/requirements.txt"
  run:
    - wget -q https://storage.googleapis.com/labelbox-assets/clip_market1501.pt
    - python -c 'from boxmot import DeepOCSORT;from pathlib import Path;DeepOCSORT(model_weights=Path("clip_market1501.pt"), device="cpu", fp16=False)'
predict: "aws_rekognition.py:AWSRekognition"

```
When i run cog build it is erroring out with the following message
```

9.21 ERROR: Could not find a version that satisfies the requirement ython_version<"3.13" (from versions: none)
19.21 ERROR: No matching distribution found for ython_version<"3.13"
19.23 
19.23 [notice] A new release of pip is available: 23.0.1 -> 25.0.1
19.23 [notice] To update, run: pip install --upgrade pip
```
Upon further inspection when i check the requirements.txt file in .cog/tmp/build.****. I see a malformed requirements.txt as follows

```
--extra-index-url https://download.pytorch.org/whl/cpu
annotated-types==0.7.0
anyio==4.8.0
asttokens==2.4.1
attrs==22.2.0
backports-tarfile==1.2.0
beautifulsoup4==4.13.3
build==1.2.2.post1
cachecontrol==0.14.2
cachetools==5.5.2
certifi==2025.1.31
cffi==1.17.1
charset-normalizer==3.4.1
cleo==2.1.0
click==8.1.8
cog==0.14.7
colorama==0.4.6
crashtest==0.4.1
cryptography==43.0.3
decorator==5.2.1
deepmerge==1.1.1
devtools==0.12.2
distlib==0.3.9
distro==1.9.0
dulwich==0.22.8
exceptiongroup==1.2.2
executing==2.2.0
fastapi==0.115.11
fastjsonschema==2.21.1
filelock==3.18.0
findpython==0.6.3
google-api-core==1.34.1
google-auth==2.38.0
google-cloud-secret-manager==2.16.4
google-crc32c==1.6.0
googleapis-common-protos==1.69.0
grpc-google-iam-v1==0.14.1
grpcio-status==1.48.2
grpcio==1.70.0
[...]

PR #2281, opened 2025-04-28
fix: Requirements format parser

danfairs:
This change modifies the requirements file parser to improve very long line handling. Previously, lines (including continuations) that were longer than the data chunk that `scanner.Text()` passed in (4096 bytes) would be truncated at the boundary. This caused invalid requirements files to be generated, as was observed in #2280.

I have added a test which failed with the old code, but passes with the new code. My own problematic image also proceeds past the requirements generation step.

It would be good if a few other people could confirm this works on their images. I only have one - the first one I built with `cog` ran into this problem!


8W9aG:
Thanks for this @danfairs ! The implementation and test look great, I'll see if we can get this in

8W9aG:
Moved the PR to https://github.com/replicate/cog/pull/2292 and merged it, will close now

PR #2288, opened 2025-04-29
Replace `docker.Xxx` helpers with functions on `command.Command`

michaeldwan:
The TLDR is:
1. Consolidate package functions that shell out to the docker CLI into functions on the `docker.DockerClient` struct
2. Add them to the `command.Command` interface (which'll get renamed to something like `docker.Client` shortly) so I can add the direct engine client without changing build logic
3. Upgrade `github.com/docker/docker` and `github.com/docker/cli` packages. Unfortunately there was an irreconcilable conflict with a far flung goreleaser dependency, so I removed it from `go tools`.
4. add integration tests for the docker client that directly hit a docker daemon and registry.

Since this one is big (though mostly benign) I split each logical chunk into a commit for easier review.

This PR moves the following operations to `command.Command`:
- Inspect
- ContainerInspect
- ContainerStop
- ContainerLogs
- Pull

I'll knock the chonky ones out in subsequent PRs, like Build and Run.


pkg/cli/train.go
```diff
@@ -88,17 +91,17 @@ func cmdTrain(cmd *cobra.Command, args []string) error {
 		// Use existing image
 		imageName = args[0]
 
-		exists, err := docker.ImageExists(ctx, imageName)
+		manifest, err := dockerCommand.Inspect(ctx, imageName)
```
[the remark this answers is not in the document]
michaeldwan: wow copilot found an actual bug. fixing!

pkg/cli/predict.go
```diff
@@ -112,17 +115,18 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("Invalid image name '%s'. Did you forget `-i`?", imageName)
 		}
 
-		exists, err := docker.ImageExists(ctx, imageName)
+		manifest, err := dockerCommand.Inspect(ctx, imageName)
```
8W9aG, line 118: Is there a quicker way to get an existance check than an inspect call here?

8W9aG (approved):

pkg/cli/predict.go
michaeldwan: just realized you reviewed before I pushed up the changes to Pull that I mentioned. Here's the commit if you're curious: https://github.com/replicate/cog/pull/2288/commits/9e0b967cc12f1b839c7bd69ce500b341ab343d84

PR #2300, opened 2025-05-06
Build the image on predict in fast boots

8W9aG:
* Build the image on `cog predict` if we are running a fast boot image


michaeldwan (approved):
lgtm!

PR #2301, opened 2025-05-06
Add test for cog migration with GPU

8W9aG:


pkg/migrate/migrator_v1_v1fast.go
```diff
@@ -175,10 +175,7 @@ func (g *MigratorV1ToV1Fast) flushConfig(cfg *config.Config, dir string, configF
 		cfg.Build = config.DefaultConfig().Build
 	}
 	cfg.Build.Fast = true
-	err := cfg.ValidateAndComplete("")
-	if err != nil {
-		return err
-	}
+	console.Infof("cfg = %v", cfg)
```
michaeldwan, line 178: does this still print when `--debug` is disabled?

michaeldwan (approved):
lgtm, one question about what looks like a debug print to info

pkg/migrate/migrator_v1_v1fast.go
8W9aG: Resolved in [b973f50](https://github.com/replicate/cog/pull/2301/commits/b973f5087a6c3039581f12570d8e13cc882a0832)

PR #2304, opened 2025-05-07
Update all integration tests to use the `cog_binary` fixture

michaeldwan:
This PR updates all the integration tests to use [the `cog_binary` fixture](https://github.com/replicate/cog/pull/2274/commits/ae288968b3bed5102996913592b24086d61dcdd2). This allows the test runner to specify a cog binary to test. When not provided it'll fallback to the existing behavior, which is a PATH lookup for `"cog"`.


aron (approved):
Thanks for updating this!

PR #2307, opened 2025-05-08
Move container ops to `command.Command`

michaeldwan:
This change is part of the ongoing work to abstract all docker cli calls behind the `command.Command` interface. Once complete (almost there!) we can replace the cli client with one that connects directly to the daemon and exposes a swath of low level api calls we need.

This PR:
- moves run (execute and wait for completion) and start (execute and detach) to the interface
- uses the client's `ContainerInspect` function to get a container port rather than the `docker port` command
- adds `mockery` to codegen interface mocks for testing (just giving it a try, works okay so far, will rip out if it's a pita) 
- adds tests for handling ports in container inspect responses

The only intentional change to behavior is that the entire host environment is no longer passed to containers by default. The motivation for this change is to reduce hard-to-debug foot guns caused by host env invisibly altering the container. Going forward the env passed to containers will be explicit.


8W9aG (approved):
🙌  nice job!

PR #2315, opened 2025-05-09
Add environment variable for defining coglet version

8W9aG:
* Allow environment variables to define what version we use for coglet
* This can be used for forcing a coglet version
* Primarily we use this in the CI to prevent github rate limits


michaeldwan (approved):
awesome thanks!

PR #2316, opened 2025-05-09
Annotate nullable values as such in the OpenAPI schema

erbridge:
Prior to this change, we allowed the return of `None` values from models, serializing them as `null`, but the schema didn't specify them as nullable, which would cause model output to fail validation.

This only supports Pydantic V2, because, from what I can tell, Pydantic V1 doesn't give us any clues that a field allows `None` values.

Note: this is probably a breaking change.


8W9aG (approved):

aron (approved):
I think this is good to merge (bar the `input_seed.py` file).

Regarding whether this is a breaking change. From what I understand, optional types will now include the `nullable` field on the OpenAPI schema for both inputs and outputs. In practice this allows a client to pass `field: null` in addition to omitting it from the request body, it also correctly types the output to indicate that output can be `null`, this has always been the case when a prediction is in a non-terminal state.

The only two failure modes I can think of are:

 1. Where someone is using the generated OpenAPI schema for something else. The new presence of the `nullable` field could change the behavior of whatever system is using the schema.
 2. When the output schema is used to determine the type of the output. However this has always been a nullable field on the PredictionResponse so I would expect most consumers to handle this already despite the incorrect schema.

I _think_ based on this we can class it as a bug fix without a major version bump.

python/cog/server/helpers.py
```diff
@@ -361,12 +378,268 @@ def unwrap_pydantic_serialization_iterators(obj: Any) -> Any:
...
+                openapi_schema.get("components", {}).get("schemas", {}).items()
+            ):
+                # Look for anyOf with more than one entry
+                if (
+                    "anyOf" in schema
+                    and isinstance(schema["anyOf"], list)
+                    and len(schema["anyOf"]) > 1
+                ):
+                    # If it's missing nullable, and it's meant to represent an Optional/Union output
+                    if "nullable" not in schema:
```
erbridge, line 405: Shouldn't this be checking for `null` in the types or something? How do we know that just because it's a union of 2 or more types that it's nullable?

8W9aG:
> I think this is good to merge (bar the `input_seed.py` file).
> 
> Regarding whether this is a breaking change. From what I understand, optional types will now include the `nullable` field on the OpenAPI schema for both inputs and outputs. In practice this allows a client to pass `field: null` in addition to omitting it from the request body, it also correctly types the output to indicate that output can be `null`, this has always been the case when a prediction is in a non-terminal state.
> 
> The only two failure modes I can think of are:
> 
> 1. Where someone is using the generated OpenAPI schema for something else. The new presence of the `nullable` field could change the behavior of whatever system is using the schema.
> 2. When the output schema is used to determine the type of the output. However this has always been a nullable field on the PredictionResponse so I would expect most consumers to handle this already despite the incorrect schema.
> 
> I _think_ based on this we can class it as a bug fix without a major version bump.

Luckily we have another breaking change going in right now, the next bump will be to `0.15.0` so we can make this part of that bump.

python/cog/server/helpers.py
8W9aG: I've reworked this so its simpler, now each nullable is the result of a type check: [9e17599](https://github.com/replicate/cog/pull/2316/commits/9e17599d5591ef1932515c8a1b359301b8873dae)

python/cog/server/helpers.py
```diff
@@ -361,12 +378,268 @@ def unwrap_pydantic_serialization_iterators(obj: Any) -> Any:
...
+    def update_nullable_optional(openapi_schema: Dict[str, Any], app: FastAPI) -> None:
+        def patch_nullable_parameters(openapi_schema: Dict[str, Any]) -> None:
+            for _, methods in openapi_schema.get("paths", {}).items():
+                for _, operation in methods.items():
+                    for param in operation.get("parameters", []):
+                        # If the parameter is optional (required: false), make it nullable
+                        if not param.get("required", True):
+                            schema = param.get("schema", {})
+                            if "nullable" not in schema:
+                                schema["nullable"] = True
```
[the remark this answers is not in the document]
8W9aG: Removed header nullability in [9e17599](https://github.com/replicate/cog/pull/2316/commits/9e17599d5591ef1932515c8a1b359301b8873dae)

8W9aG:
Going to merge this since we need to do a `0.15.0` release and I believe all the comments have been taken care of

PR #2318, opened 2025-05-12
on_event of FastAPI is deprecated and replaced with lifespan

pranitbauva1997:
Removing the two `startup` and `shutdown` methods and merging them together into the lifespan method.

Here is the docs for the FastAPI lifespan method:
https://fastapi.tiangolo.com/advanced/events/#lifespan


michaeldwan:
@8W9aG you're more familiar with the python than I am. mind taking a look at this?

michaeldwan (changes requested):
This looks good, thank you! All that's left is the [linter issue](https://github.com/replicate/cog/actions/runs/15311398942/job/43076703666?pr=2318#step:9:119). Would you like me to clean that up?

pranitbauva1997:
@michaeldwan Let me fix that.

PR #2320, opened 2025-05-12
Support Torch 2.7.0 and 2.6.0

andreasjansson:
Also required updating compatgen with `manylinux_2`


8W9aG (approved):

michaeldwan:
@andreasjansson Doh! I opened #2400 before seeing this one, but it also updates the compatibility matrix files. I worked around the missing CuDNN version in the image tags (which is causing the test failures here) by fetching it from the image config in the registry. 

Would you be cool if I closed this PR in favor of the other one? I cherry-picked your `manylinux_2` change, so it should be the same.

PR #2322, opened 2025-05-12
Add callout to move users towards fast-boots

8W9aG:
* If the build is not a fast build, send the user a message about using fast boots


michaeldwan:
I wonder why the integration tests were still running after 20m. I've noticed them creeping up but never past 13m. If this becomes a normal issue we can split the integration tests up across multiple runners.

8W9aG:
> change looks good but the url printed 404s. is it incorrect, or just not deployed yet?

We will make the link live before the next release

gandalfhz (approved):

PR #2332, opened 2025-05-14
Fix keeping comments in yaml

8W9aG:
* Can't blindly reassign contents, need to consider when to add/remove on mapping nodes


nevillelyh (approved):

michaeldwan (approved):
Looks good. Thanks for adding tests!

PR #2334, opened 2025-05-14
Add call_graph tool

8W9aG:
* Determines the call graph from within a model
* Fails if the input is not a static string

### DIFF
--- a/python/tests/command/call_graph_test.py
+++ b/python/tests/command/call_graph_test.py
@@ -1,0 +1,51 @@
+import os
+import tempfile
+from pathlib import Path
+
+import pytest
+
+from cog.command.call_graph import analyze_python_file
+
+
+def test_call_graph():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        filepath = os.path.join(tmpdir, "predict.py")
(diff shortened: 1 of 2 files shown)


python/tests/command/call_graph_test.py
```diff
@@ -0,0 +1,51 @@
...
+    seed: int = Input(description="A seed", default=0)
+) -> Path:
+    output_url = flux_schnell(prompt=prompt, seed=seed)[0]
+    return output_url
+""")
+        with pytest.raises(ValueError) as excinfo:
+            analyze_python_file(Path(filepath))
+        assert str(excinfo.value).endswith(
+            "Unresolvable argument at line 5: Not a string literal"
+        )
```
aron, line 51: Nice, the other constraint currently is that we only support calling `include()` at the top level of a module. Dynamically calling it within a predictor for example is no good, so we should either, a) limit the parser to only walk the module scope or b) walk the entire file and error if a call is found outside of the module scope.

We could also do this as a follow up if needed.

aron (approved):
Nice! Looks great.

PR #2335, opened 2025-05-15
fix: Always pass `--load` during docker build

danfairs:
This change modifies Cog so that it always passes `--load` during docker build.

Previously, it would only pass this flag if running on Apple Silicon. I don't think that is quite the right check - it is actually down to the build driver in use. Cog seems to assume that the image will always be loaded into Docker, so unconditionally passing `--load` seems to be the right thing to do.

I had to include this change to allow builds to complete successfully on a self-hosted Linux GH CI runner.


michaeldwan:
@danfairs I'm happy to rebase and fix conflicts with main, but I'll need permission to your fork. I want to get this in, just let me know how to help 😄

PR #2342, opened 2025-05-16
Add integration test for setup run in trainer

8W9aG:
* Test that the `setup` function is run in a `train` predictor


.github/workflows/ci.yaml
```diff
@@ -121,7 +121,7 @@ jobs:
     name: "Test integration"
     needs: build-python
     runs-on: ubuntu-latest-16-cores
-    timeout-minutes: 20
+    timeout-minutes: 25
```
michaeldwan, line 124: thanks!

michaeldwan (approved):
👍

PR #2343, opened 2025-05-16
Call train function

8W9aG:
* Calls `train` when a class is defined in the `train` key in `cog.yaml`.

This is a breaking change, since currently it calls `predict` but uses the type definition of the `train` method.


daanelson (commented):
broadly looks good; my general worry here is that this might have unexpected consequences for our existing high volume trainers (e.g. a trainer that has `train` but no `setup`). 

and to be clear, this is a worry driven by paranoia as opposed to anything I see here. 😄  do we have a test somewhere that would catch an issue with that kind of trainer?

8W9aG:
> broadly looks good; my general worry here is that this might have unexpected consequences for our existing high volume trainers (e.g. a trainer that has `train` but no `setup`).
> 
> and to be clear, this is a worry driven by paranoia as opposed to anything I see here. 😄 do we have a test somewhere that would catch an issue with that kind of trainer?

Luckily setup is always optional so I bet that this won't matter so much

michaeldwan (approved):

PR #2345, opened 2025-05-19
Support kwargs predictors

philandstuff:
This change allows a predictor input to be defined exclusively as `**kwargs`. This is an experimental change that allows a model to be used as a proxy where any inputs will be passed through directly without modification.

```py
import json

def predict(**kwargs) -> str:
  return json.dumps(kwargs)
```

This change restricts the use of `**kwargs` to functions/methods that accept only that as an argument to avoid the complexity of having to support both dynamic and fixed inputs.

The OpenAPI schema for such a model will be:

```json
{
   "properties": {},
   "additionalProperties": true,
   "type": "object",
   "title": "Input"
}
```

### DIFF
--- a/python/cog/predictor.py
+++ b/python/cog/predictor.py
@@ -241,29 +241,29 @@ def get_input_create_model_kwargs(signature: inspect.Signature) -> Dict[str, Any
     for name, parameter in signature.parameters.items():
         InputType = parameter.annotation
 
-        validate_input_type(InputType, name)
-
         if parameter.kind == inspect.Parameter.VAR_POSITIONAL:
-            raise TypeError(f"Unsupported varargs parameter *{name}.")
+            raise TypeError(f"Unsupported positional parameter *{name}.")
+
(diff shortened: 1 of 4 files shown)


python/cog/predictor.py
```diff
@@ -231,13 +231,38 @@ def validate_input_type(
...
 
     for name, parameter in signature.parameters.items():
         InputType = parameter.annotation
 
+        if parameter.kind == inspect.Parameter.VAR_POSITIONAL:
+            raise TypeError(f"Unsupported positional parameter *{name}.")
+
+        if parameter.kind == inspect.Parameter.VAR_KEYWORD:
+            if order != 0:
+                raise TypeError(f"Unsupported keyword parameter **{name}")
```
philandstuff, line 249:
```suggestion
                raise TypeError(f"Unsupported variadic keyword parameter **{name}")
```

philandstuff (commented):
@aron thanks for the polish. LGTM

PR #2359, opened 2025-05-23
Add docs for training classes

8W9aG:


docs/training.md
```diff
@@ -37,6 +37,30 @@ $ cat weights
 hello train
 ```
 
+You can also use classes if you want to run many model trainings and save on setup time, this works the same way as the [Predictor](./python.md) class with the only difference being the `train` method.
```
superhighfives, line 40: I think this could this could two sentences. So, "....on setup time. This works the same way..."

superhighfives (approved):
Minor suggestions but looks good to me. Merge away!

docs/training.md
8W9aG: Resolved in [46b12fd](https://github.com/replicate/cog/pull/2359/commits/46b12fd30ffed36261d3e084ddbbdf005297885c)

PR #2366, opened 2025-05-27
Only overwrite header if it doesn’t exist

8W9aG:
* Do not overwrite the content-type header if it is explicitly set by the user

### DIFF
--- a/pkg/http/transport_test.go
+++ b/pkg/http/transport_test.go
@@ -28,3 +28,26 @@ func TestTransportAddsHeaders(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, resp.Request.Header.Get(testHeader), testValue)
 }
+
+func TestTransportOnlyAddsHeaderIfMissing(t *testing.T) {
+	// Setup mock http server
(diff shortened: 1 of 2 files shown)


pkg/http/transport_test.go
```diff
@@ -28,3 +28,26 @@ func TestTransportAddsHeaders(t *testing.T) {
...
 }
+
+func TestTransportOnlyAddsHeaderIfMissing(t *testing.T) {
+	// Setup mock http server
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	}))
+	defer server.Close()
+
+	const testHeader = "X-Test-Header"
```
michaeldwan, line 39: is `const` required for the test? it's fine, just uncommon to see in a function body

michaeldwan (approved):
lgtm, question about const

pkg/http/transport_test.go
8W9aG: I suppose I don't expect it to change jajaja

PR #2369, opened 2025-05-28
Add recent torch compatibility matrix

8W9aG:
* Adds torch 2.6.0 and torch 2.7.0 to the compatibility matrix
* This will temporarily break builds against these versions on main until we build the base images (which will come after this PR is merged)


8W9aG:
Merging due to expedited urgency

PR #2375, opened 2025-05-29
Add checking of pipelines runtime requirements before pushing

8W9aG:
* Make sure the user uses the correct pipelines runtime requirements before pushing


pkg/global/global.go
```diff
@@ -1,12 +1,13 @@
...
-	LabelNamespace        = "run.cog."
+	Version                 = "dev"
+	Commit                  = ""
+	BuildTime               = "none"
+	Debug                   = false
+	ProfilingEnabled        = false
+	ReplicateRegistryHost   = "r8.im"
+	ReplicateWebsiteHost    = "replicate.com"
+	LabelNamespace          = "run.cog."
+	CogBuildArtifactsFolder = ".cog"
```
aron, line 12: FWIW I noticed today that this needs to be added to the generated `.dockerignore`, unless it's intentional that this is bundled up as part of a deploy...

8W9aG: Unfortunately it can't be, because we rely on this store artifacts we use to build with docker, theres an error if you add it in.

aron: Do we strip it out of the final build? I guess it doesn't matter, it just surprised me today when I saw it.

PR #2376, opened 2025-05-30
Use Python 3.13 in generated cog.yaml

aron:
Python 3.11 is now two versions old and 3.13 has been out for over a
year at this point.

This commit also removes the `fast: true` flag which is unlikely
to be supported going forward.


michaeldwan:
We should get this shipped! Would you like me to fix the failing tests?

PR #2378, opened 2025-05-30
Refactor PostNewPipeline to use Replicate public API

aron:
This follows the procedure for shipping an experimental "pipeline" model via the API used by other areas of the Replicate product.

We use the new cog.replicate.com/api/token/{account} endpoint to exchange the cog token for a short lived Replicate API token that is restricted only to read/write actions on model resources.

Usage:

    cog push --x-pipeline r8.im/username/my-model

### DIFF
--- a/pkg/docker/pipeline_push_test.go
+++ b/pkg/docker/pipeline_push_test.go
@@ -18,18 +18,53 @@ import (
 )
 
 func TestPipelinePush(t *testing.T) {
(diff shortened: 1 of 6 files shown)


pkg/docker/pipeline_push_test.go
aron, line 1: This same test exists three times in the codebase 😓

PR #2388, opened 2025-06-03
Support `os.PathLike` output from models

aron:
This commit updates two pieces of our file handling code to work with `os.PathLike` objects.

 1. When transforming the output into a data-structure suitable for encoding in `make_encodable` we convert the `PathLike` into a `pathlib.Path` instance.
 2. When reading the file data from disk in `upload_files` we use `open()` instead of `Path.open()`.

This provides broader support for path-like objects, particularly the new replicate-python client which will return path-like objects from the new `use()` helper.

### DIFF
--- a/python/cog/json.py
+++ b/python/cog/json.py
@@ -1,4 +1,5 @@
 import io
+import os
 import pathlib
 from datetime import datetime
 from enum import Enum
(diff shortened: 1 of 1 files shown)


8W9aG (approved):
Great change! It seems that this was introduced in python `3.6` so should be well within our compatibility window.

PR #2400, opened 2025-06-09
Update compatibility matrix files

michaeldwan:
## Summary

Fix `compatgen` tool to correctly extract CuDNN versions from newer nvidia/cuda images, and regenerate all compatibility matrices.

### Problem

Newer nvidia/cuda image tags no longer include the CuDNN version number (e.g. `12.9.1-cudnn-devel-ubuntu24.04` instead of `12.6.3-cudnn9-devel-ubuntu22.04`). The tag parser in `compatgen` couldn't extract CuDNN from these tags, so new CUDA versions had to be manually patched into the JSON (see #2036).

### Fix

- **`tools/compatgen/internal/cuda.go`**: Fetch the Docker image config via `go-containerregistry` and read `NV_CUDNN_VERSION` from environment variables instead of parsing the tag string. Uses `authn.DefaultKeychain` for Docker Hub auth to avoid rate limits. Adds deterministic sorting of output.
- **`tools/compatgen/internal/torch.go`**: Filter out torch entries with no supported Python versions (all below 3.10 minimum) instead of emitting them with `"Pythons": []`.
- **`tools/compatgen/main.go`**: Pass `context.Context` to support the image fetching.

### Regenerated files

- **`cuda_compatibility.json`**: Adds CUDA 13.0.x and 13.1.x. CuDNN values now correctly extracted from image configs.
- **`torch_compatibility.json`**: Adds torch 2.10.0, 2.9.x. Drops 41 entries for torch versions with no supported Python versions.
- **`tf_compatibility.json`**: Adds TensorFlow 2.20.0.

### Test updates

Updated 4 tests that referenced dropped torch versions (1.7.x, 1.8.0) to use versions still in the matrix (1.11.0, 1.13.1, 2.0.1).

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -97,9 +97,9 @@ func TestPythonPackagesAndRequirementsCantBeUsedTogether(t *testing.T) {
 
 func TestPythonRequirementsResolvesPythonPackagesAndCudaVersions(t *testing.T) {
 	tmpDir := t.TempDir()
-	err := os.WriteFile(path.Join(tmpDir, "requirements.txt"), []byte(`torch==1.7.1
-torchvision==0.8.2
-torchaudio==0.7.2
+	err := os.WriteFile(path.Join(tmpDir, "requirements.txt"), []byte(`torch==1.13.1
+torchvision==0.14.1
+torchaudio==0.13.1
 foo==1.0.0`), 0o644)
 	require.NoError(t, err)
 
@@ -112,15 +112,15 @@ foo==1.0.0`), 0o644)
 	}
 	err = config.Complete(tmpDir)
 	require.NoError(t, err)
-	require.Equal(t, "11.0", config.Build.CUDA)
+	require.Equal(t, "11.7", config.Build.CUDA)
 	require.Equal(t, "8", config.Build.CuDNN)
 
 	requirements, err := config.PythonRequirementsForArch("", "", []string{})
 	require.NoError(t, err)
-	expected := `--find-links https://download.pytorch.org/whl/torch_stable.html
-torch==1.7.1
-torchvision==0.8.2
-torchaudio==0.7.2
+	expected := `--extra-index-url https://download.pytorch.org/whl/cu117
+torch==1.13.1
+torchvision==0.14.1
+torchaudio==0.13.1
 foo==1.0.0`
 	require.Equal(t, expected, requirements)
 }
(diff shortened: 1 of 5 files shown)


pkg/util/version/version.go
```diff
@@ -14,12 +14,12 @@ type Version struct {
 }
 
 func NewVersion(s string) (version *Version, err error) {
+	// TODO[md]: handle prerelease versions (0.1.2-rc1) so they aren't appended to the previous component
+	// todo[md]: tbh just switch to hashicorp/go-version or github.com/Masterminds/semver/v3
```
markphelps: 💯 ive used semver/v3 in the past, its pretty nice

tools/compatgen/internal/cuda.go
```diff
@@ -70,17 +81,56 @@ func fetchCUDABaseImageTags(url string) ([]string, error) {
 	return tags, nil
 }
 
-func parseCUDABaseImage(tag string) (*config.CUDABaseImage, error) {
-	parts := strings.Split(tag, "-")
-	if len(parts) != 4 {
-		return nil, fmt.Errorf("Tag must be in the format <cudaVersion>-cudnn<cudnnVersion>-{devel,runtime}-ubuntu<ubuntuVersion>. Invalid tag: %s", tag)
+func parseCUDABaseImage(ctx context.Context, tag string) (*config.CUDABaseImage, error) {
+	fmt.Println("parsing", tag)
```
markphelps, line 95: debug printlns ? / do we want to keep these ?

pkg/util/version/version.go
michaeldwan: I always go to that one too, but I found it the other day it doesn't support "invalid" semver input like ubuntu's "22.04" with leading zeros. I was hoping to use the same version code in cog and the new base image generator code 😒

tools/compatgen/internal/cuda.go
```diff
@@ -18,16 +23,22 @@ func FetchCUDABaseImages() ([]config.CUDABaseImage, error) {
...
-		image, err := parseCUDABaseImage(tag)
-		if err != nil {
-			return nil, err
-		}
-		images = append(images, *image)
-	}
+	images := make([]config.CUDABaseImage, len(tags))
+	eg, egctx := errgroup.WithContext(context.TODO())
+	// set a concurrency limit to avoid throttling by the docker hub api (since these are authenticated requests)
+	eg.SetLimit(1)
```
markphelps: why use error group at all then if we are running them serially?

markphelps (approved):
one question, but overall lgtm

tools/compatgen/internal/cuda.go
michaeldwan: natural evolution of intermittent issues and sloppy code. fixing :)

markphelps (dismissed):
lgtm!

PR #2402, opened 2025-06-10
Fix FastAPI deprecation warnings by migrating to lifespan handlers

hafeezhmha:
### Problem
Cog currently uses deprecated `@app.on_event()` handlers which cause deprecation warnings in FastAPI.

### Solution
- Migrated from `@app.on_event("startup")` and `@app.on_event("shutdown")` to FastAPI's lifespan handlers
- Updated imports to include `asynccontextmanager`
- No functional changes, just API modernization

### Testing
- [x] Tested locally with `cog predict`
- [x] No deprecation warnings appear
- [x] Server startup/shutdown works correctly

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -12,7 +12,16 @@ import traceback
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from enum import Enum, auto, unique
-from typing import TYPE_CHECKING, Any, AsyncGenerator, Awaitable, Callable, Dict, Optional, Type
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AsyncGenerator,
+    Awaitable,
+    Callable,
+    Dict,
+    Optional,
+    Type,
+)
 
 import structlog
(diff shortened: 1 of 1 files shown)


michaeldwan (changes requested):
Thanks @hafeezhmha! It looks like every file in the repo changed line endings. If you can get that sorted out we'll review!

hafeezhmha:
> Thanks @hafeezhmha! It looks like every file in the repo changed line endings. If you can get that sorted out we'll review!

@michaeldwan Hi, can you have a look, I have sorted them out.

hafeezhmha:
> Thanks @hafeezhmha! It looks like every file in the repo changed line endings. If you can get that sorted out we'll review!

@michaeldwan for some reason, the P.ge works , there are no deprecation warnings anymore. Can you please test it out. The file already has some todos and fixes which i have kept as is. I have only updated the fastAPI part.

michaeldwan (approved):
Looks great, thanks for the quick fixes!

PR #2404, opened 2025-06-11
Refactor `cog predict` to support `--json` flag

aron:
This PR introduces `cog predict --json` that represents both input and output as JSON.

The intention behind this PR is to provide a better interface for machines to interface with `cog predict`. Such as running automated tests. In these cases trying to parse the existing unstable output is non-trivial. Having access to the full JSON prediction makes things much easier.

In order to do this correctly the PR also fixes a number of bugs with the `--output` behavior.

 * We now support passing a directory as an output value e.g. `--output ./outputs`
 * We now honor the `--output` flag for lists of files. Previously if a model returned a list of files they would all be `output.<idx>.<ext>` regardless of the --output flag.
 * We now error with a non-zero exit code if the prediction is not successful.

There are several ways to use `cog predict` with the `--json` flag:

 1. Passing a JSON string directly.
    ```
    cog predict --json '{"prompt": "make me a sandwich"}'
    ```
 1. Passing a path to a JSON file.
    ```
    cog predict --json @inputs.json
    ```
 1. Passing JSON via stdin, using the special `@-` reference.
    ```
    cog predict --json @- < inputs.json
    ```

The output will be the entire prediction printed to stdout by default.

```
cog predict --json '{"prompt": "make me a sandwich"}'
{"status": "successful", "output": "blah blah", error: null}
```

The output can also be written to a file:

```
cog predict --json '{"prompt": "make me a sandwich"}' --output output.json
Output written to: output.json
```

### DIFF
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -364,6 +364,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 		return fmt.Errorf("Failed to predict: %w", err)
 	}
 
+	// FIXME: Return JSON if needsJSON is true.
 	if prediction.Output == nil {
 		console.Warn("No output generated")
 		return nil
@@ -388,7 +389,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 			fileOutputPath = trimExt(fileOutputPath)
 		}
 
-		path, err := writeDataURLOutput(outputStr, fileOutputPath)
+		path, err := writeDataURLToFile(outputStr, fileOutputPath)
 		if err != nil {
 			return fmt.Errorf("Failed to write output: %w", err)
 		}
@@ -403,7 +404,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 				return fmt.Errorf("Failed to encode prediction output as JSON: %w", err)
 			}
 			if writeOutputToDisk {
-				path, err := writeOutput(rawJSON, outputPath)
+				path, err := writeFile(rawJSON, outputPath)
 				if err != nil {
 					return fmt.Errorf("Failed to write output: %w", err)
 				}
@@ -433,7 +434,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 				return fmt.Errorf("Failed to convert prediction output to string")
 			}
 
-			path, err := writeDataURLOutput(outputStr, fileOutputPath)
+			path, err := writeDataURLToFile(outputStr, fileOutputPath)
 			if err != nil {
 				return fmt.Errorf("Failed to write output %d: %w", i, err)
 			}
@@ -451,7 +452,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 				return fmt.Errorf("Failed to encode prediction output as JSON: %w", err)
 			}
 			if writeOutputToDisk {
-				path, err := writeOutput(rawJSON, outputPath)
+				path, err := writeFile(rawJSON, outputPath)
 				if err != nil {
 					return fmt.Errorf("Failed to write output: %w", err)
 				}
@@ -479,7 +480,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 			}
 
 			if writeOutputToDisk {
-				path, err := writeOutput(rawJSON, outputPath)
+				path, err := writeFile(rawJSON, outputPath)
 				if err != nil {
 					return fmt.Errorf("Failed to write output: %w", err)
 				}
@@ -491,7 +492,7 @@ func runPrediction(predictor predict.Predictor, inputs predict.Inputs, outputPat
 		}
 
 		if writeOutputToDisk {
-			path, err := writeOutput([]byte(s), outputPath)
+			path, err := writeFile([]byte(s), outputPath)
 			if err != nil {
 				return fmt.Errorf("Failed to write output: %w", err)
 			}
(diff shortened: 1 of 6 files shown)


pkg/cli/predict.go
```diff
@@ -67,10 +72,79 @@ the prediction on that.`,
 	cmd.Flags().StringVarP(&outPath, "output", "o", "", "Output path")
 	cmd.Flags().StringArrayVarP(&envFlags, "env", "e", []string{}, "Environment variables, in the form name=value")
 	cmd.Flags().BoolVar(&useReplicateAPIToken, "use-replicate-token", false, "Pass REPLICATE_API_TOKEN from local environment into the model context")
+	cmd.Flags().StringVar(&inputJSON, "json", "", "Pass inputs as JSON object, read from file (@inputs.json) or via stdin (@-)")
```
michaeldwan, line 77: `-` (without the `@` prefix) is the usual arg for "read from stdin". It's fine to support both but `-` is the preferred one. If you haven't seen it [this](https://clig.dev/?trk=public_post-text#arguments-and-flags) is a lovely CLI ux guide written by Ben :)

pkg/cli/predict.go
```diff
@@ -67,10 +72,79 @@ the prediction on that.`,
...
 }
 
+func parseJSONInput(jsonInput string) (map[string]any, error) {
+	var jsonStr string
+
+	if strings.HasPrefix(jsonInput, "@") {
+		// Read from file or stdin
+		source := jsonInput[1:]
+
+		if source == "-" {
```
michaeldwan: Same as previous comment, this shouldn't need a `@` prefix, but fine if it's supported as well.

pkg/cli/predict.go
```diff
@@ -277,152 +395,264 @@ func predictIndividualInputs(predictor predict.Predictor, inputFlags []string, o
...
-		return unix.Access(outputPath, unix.W_OK)
-	} else if os.IsNotExist(err) {
-		// File doesn't exist, check if the directory is writable
-		dir := filepath.Dir(outputPath)
-		return unix.Access(dir, unix.W_OK)
+	// If a directory was provided, use that with the fallback filename
+	if stat.IsDir() {
+		// If the fallback path already exists as a directory error.
+		if usingFallback {
+			return "", fmt.Errorf("Default output name \"%s\" conflicts with directory, provide --output", outputPath)
```
michaeldwan: No need to change, just a tip. But I recently discovered the format verb `%q` for wrapping the input with double quotes.

michaeldwan (approved):
This is excellent, nice work! Just one suggestion about the stdin ux. And I'd like some tests around inputs and outputs someday, but this looks great for now. Nice job.

pkg/cli/predict.go
aron: Yeah, but that's usually added at the end of the command as a positional argument right? I can support it though.

The `@-` is lifted from cURL.

```
curl http://example.com --json @-
```

https://curl.se/docs/manpage.html#--json

pkg/cli/predict.go
```diff
@@ -277,152 +395,264 @@ func predictIndividualInputs(predictor predict.Predictor, inputFlags []string, o
...
 		}
 
-		return nil
-	case outputSchema.Type.Is("array") && isURI(outputSchema.Items.Value):
-		outputs, ok := (*prediction.Output).([]interface{})
-		if !ok {
-			return fmt.Errorf("Failed to decode output")
+		// Exit with non-zero code if the prediction has failed.
+		if prediction.Status != "succeeded" {
+			os.Exit(1)
```
8W9aG, line 457: I think we can just return an error here and let it propagate up the call stack

pkg/cli/predict.go
```diff
@@ -277,152 +395,264 @@ func predictIndividualInputs(predictor predict.Predictor, inputFlags []string, o
...
+		return nil
+	}
 
-			outputStr, ok := output.(string)
-			if !ok {
-				return fmt.Errorf("Failed to convert prediction output to string")
-			}
+	if prediction.Status != "succeeded" {
+		console.Errorf("Prediction failed with status \"%s\": %s", prediction.Status, prediction.Error)
+		os.Exit(1)
```
8W9aG: Same here

pkg/cli/predict.go
aron: Doesn't that then dump the error into the console? I can test this, but the goal is for it to not print anything, just dump the prediction to Stdout and exit.

pkg/cli/predict.go
aron: This one makes more sense, try returning the above string as an error.

aron:
Thanks both, I'll fix the failing integration test and add a couple more before merging.

pkg/cli/predict.go
8W9aG: Resolved in https://github.com/replicate/cog/pull/2404/commits/fc780cedc52bf0fb42dabb50b7a7fc577eb8f95e

pkg/cli/predict.go
8W9aG: Resolved in https://github.com/replicate/cog/pull/2404/commits/fc780cedc52bf0fb42dabb50b7a7fc577eb8f95e

pkg/cli/predict.go
8W9aG: Resolved in [b29b967](https://github.com/replicate/cog/pull/2404/commits/b29b967d610cec4d1ba1eb8d5df7226c245e332c)

pkg/cli/predict.go
8W9aG: I see what you mean, I think your way of doing this is the correct one.

pkg/cli/predict.go
8W9aG: Resolved in [6986609](https://github.com/replicate/cog/pull/2404/commits/698660996aa1d713ec8530c19419736797216f14)

PR #2406, opened 2025-06-12
Send explicit error if no token supplied

8W9aG:
* Send an explicit error that no token has been supplied for an authenticated endpoint

### DIFF
--- a/pkg/http/transport.go
+++ b/pkg/http/transport.go
@@ -26,7 +26,7 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 		authorisation, ok := t.authentication[req.URL.Host]
 		if ok {
 			if authorisation == BearerHeaderPrefix {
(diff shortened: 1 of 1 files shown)


pkg/http/transport.go
```diff
@@ -22,6 +25,9 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if req.Header.Get(AuthorizationHeader) == "" {
 		authorisation, ok := t.authentication[req.URL.Host]
 		if ok {
+			if authorisation == BearerHeaderPrefix {
+				return nil, errors.New("No token supplied for HTTP authorization.")
```
NikhilSinha1: nit:
```suggestion
				return nil, errors.New("No token supplied for HTTP authorization. Have you run 'cog login'?")
```

NikhilSinha1 (approved):

PR #2409, opened 2025-06-13
Handle badly formatted data URIs in cog predict

8W9aG:
* Currently cog returns data:None;base64
* The data URI parser does not recognise None, and requires data:;base64
* Handle receiving these URIs by removing the None format specified


nevillelyh (approved):

yorickvP:
I think cog python should not return `data:None;base64` in these cases, but `data:application/octet-stream;base64`

PR #2415, opened 2025-06-17
In event of error, checking equality

8W9aG:
* When checking if 2 version strings are greater or equal, if an error occurs parsing the version
fall back to string equality instead of panicking

### DIFF
--- a/pkg/util/version/version_test.go
+++ b/pkg/util/version/version_test.go
@@ -79,3 +79,9 @@ func TestVersionMatchesModifier(t *testing.T) {
 	matchVersion := "2.3.2+cu118"
 	require.True(t, Matches(version, matchVersion))
 }
+
+func TestGreaterThanOrEqualToWithInvalidPatch(t *testing.T) {
+	leftVersion := "1.1.0b2"
+	rightVersion := "1.1.0b2"
+	require.True(t, GreaterOrEqual(leftVersion, rightVersion))
+}
--- a/pkg/util/version/version.go
+++ b/pkg/util/version/version.go
@@ -112,7 +112,15 @@ func Greater(v1 string, v2 string) bool {
 }
 
 func GreaterOrEqual(v1 string, v2 string) bool {
-	return MustVersion(v1).GreaterOrEqual(MustVersion(v2))
+	leftVersion, err := NewVersion(v1)
+	if err != nil {
+		return v1 == v2
+	}
+	rightVersion, err := NewVersion(v2)
+	if err != nil {
+		return v1 == v2
+	}
+	return leftVersion.GreaterOrEqual(rightVersion)
 }
 
 func (v *Version) Matches(other *Version) bool {


pkg/util/version/version_test.go
```diff
@@ -79,3 +79,9 @@ func TestVersionMatchesModifier(t *testing.T) {
 	matchVersion := "2.3.2+cu118"
 	require.True(t, Matches(version, matchVersion))
 }
+
+func TestGreaterThanOrEqualToWithInvalidPatch(t *testing.T) {
+	leftVersion := "1.1.0b2"
+	rightVersion := "1.1.0b2"
+	require.True(t, GreaterOrEqual(leftVersion, rightVersion))
+}
```
aron, line 87: These aren't invalid patches... they need to be handled... We use [go-pep440-version](https://github.com/aquasecurity/go-pep440-version) elsewhere to do this. How about we use this here?

8W9aG: Resolved in [422a29d](https://github.com/replicate/cog/pull/2415/commits/422a29dc48b2edee5569d3408edb3ee8519d4da2)

go.mod
```diff
@@ -69,6 +69,7 @@ require (
 	github.com/alexkohler/prealloc v1.0.0 // indirect
 	github.com/alingse/asasalint v0.0.11 // indirect
 	github.com/alingse/nilnesserr v0.1.2 // indirect
+	github.com/aquasecurity/go-version v0.0.1 // indirect
```
aron, line 73: This isn't the right repository, we want go-pep440-version which is compatible with Python packaging. go-version is for SemVer and similar versioning schemes.

https://github.com/aquasecurity/go-pep440-version

pkg/docker/pipeline_push.go
```diff
@@ -210,7 +211,19 @@ func validateRequirements(projectDir string, client *http.Client, cfg *config.Co
...
+					if err != nil {
+						return err
+					}
+					rightVersion, err := version.Parse(pipelineVersion)
+					if err != nil {
+						return err
+					}
+					found = leftVersion.GreaterThanOrEqual(rightVersion)
+				}
 				break
```
aron: I think what we want to do here is parse the pipelinePackage and ensure that the range provided by the project requirements.txt is satisfied. From the docs this looks like:

```go
v, _ := version.Parse("2.1")
c, _ := version.NewSpecifiers(">= 1.0, < 1.4 || > 2.0")

if c.Check(v) {
	fmt.Printf("%s satisfies specifiers '%s'", v, c)
}
```

aron (changes requested):
Sorry, I'm still not sure this is doing what we need it to do. To be clear if a user project provides a version range then we need to ensure that that range is satisfied by the explicit version pinned in the requirements we get from the CDN.

Some examples of this would be:

```
replicate==1.0.0
replicate>=1.0.0b1
replicate>=1.0.0b1,<2.0.0a1
```

It's important to note that Python does not use SemVer, instead it uses it's own syntax defined in pep440 and implemented in the `packaging` module.

go.mod
8W9aG: Resolved in [89d5f26](https://github.com/replicate/cog/pull/2415/commits/89d5f263e10f79f8fc3e2ee1c78a8ee27f7fe4ab)

pkg/docker/pipeline_push.go
8W9aG: Resolved in [89d5f26](https://github.com/replicate/cog/pull/2415/commits/89d5f263e10f79f8fc3e2ee1c78a8ee27f7fe4ab)

aron (approved):
👍 Thanks for working through that, this works perfectly.

PR #2422, opened 2025-06-20
Add dotenv environment variable defaults

8W9aG:
* Allow a .env file to specify the default environment variables to be passed to predict

### DIFF
--- a/test-integration/test_integration/test_predict.py
+++ b/test-integration/test_integration/test_predict.py
@@ -803,11 +803,11 @@ def test_predict_future_annotations(cog_binary):
     assert result.returncode == 0
 
 
-def test_predict_dotenv(docker_image, cog_binary):
+def test_predict_dotenv(cog_binary):
     project_dir = Path(__file__).parent / "fixtures/env-cli-project"
 
     result = subprocess.run(
-        [cog_binary, "predict", "--debug", docker_image, "-i", "name=TEST_VAR"],
+        [cog_binary, "predict", "--debug", "-i", "name=TEST_VAR"],
         cwd=project_dir,
         capture_output=True,
         text=True,


8W9aG:
I agree I'm not sure it makes sense to scope the `.env` file to just cherrypick environment variables, this will be surprising to users. @michaeldwan is there a precedent for `.env` files being used in build rather than run workflows?

michaeldwan:
Sorry I missed your reply @aron! You're both right, env variables are the right thing to use. I'm only (gently) pushing back on scope and timing. Supporting .env files, and host-to-container env in general, comes with ux footguns we need to think through. I haven't thought through it more than the discussion above, so I'll think it through in words here and we can figure it out together 🙃 

The main question is if `.env` is to configure cog, or to configure the container. Loading `.env` to the cog process is fine, and loading it to the container is also fine. But using a single file to configure both is the footgun I'm stuck on.

---

The default/expected behavior of an `.env` file is to get loaded into the process environment, which makes sense. We want the token in `os.Environ()` so cog can access it, and we probably want it in the container as well. NBD. We're probably fine to pass all the other `COG_` or `R8_*` variables to the container as well. 

What we don't want is to load `.env` into the current process environment and _then_ pass the current env to the container. Most of the variables would be fine in the container, but obviously we don't want to overwrite things like `PATH`/`USER`/`SHELL`/`HOME` which would break the container, or config like `HTTP_PROXY` or `SSH_*` which would probably not break it but behavior would be unexpected. The way docker-like tooling contains the blast radius is requiring explicit ENV mapping (`-e KEY`, `-e KEY=$HOSTKEY`, or `--env-file .env`) when containers run. And this is a build-time concern as well since build is just a series of running containers. 

There's also a number of collisions I'd expect between cog and the container, like `LOG_LEVEL`. Or configuring python with `UV_*` that would impact the host, the build env, and the runtime env. 

We can go into more detail with all that, but that's the gist.

So what options do we have for a path forward? The quickest is to not address this yet and use `.env` as a convenience for the token until we can properly come back and think this through. But if we do want to move forward with making env variables available in the container, perhaps we should consider multiple files. For example, `.env` autoloads on the host (expected behavior) and `.env.cog` autoloads in the container, and we can even support `.env.cog.build` or similar to target variables for a specific specific stage.

Dunno... just wanted to dump some thoughts for discussion. What do y'all think?

PR #2423, opened 2025-06-20
Add replicate-username flag

8W9aG:
* Allow a user to specify their replicate username for the purposes of performing predictions
* This adds REPLICATE_API_TOKEN to the environment variables once it resolves the token


michaeldwan:
no longer needed, closing

PR #2430, opened 2025-06-30
feat: add ability to skip files that already exist, overwrite or quit

markphelps:
This updates the `cog init` command to allow for skipping files that already exists either one by one or all at once

It also adds two cli flags `--skip-existing` and `--overwrite-existing` which will either:
1. skip all existing flags without prompting
2. overwrite all existing flags without prompting

I chose to not just skip existing by default as to be more consistent with the existing behavior, but I could also see the argument that we just want to always skip all existing files.. so would like to get feedback here

## Example Run

```

File cog.yaml already exists. What would you like to do? (default: skip, options: skip, skip-all, quit): skip
Skipped existing cog.yaml
✅ Created /Users/markphelps/workspace/replicate/pipelines-test/predict.py
File .dockerignore already exists. What would you like to do? (default: skip, options: skip, skip-all, quit): skip-all
Skipped existing .dockerignore (and will skip all remaining)
Skipped existing .github/workflows/push.yaml
Skipped existing requirements.txt

```

### DIFF
--- a/pkg/cli/init.go
+++ b/pkg/cli/init.go
@@ -68,7 +68,7 @@ func initCommand(cmd *cobra.Command, args []string) error {
 		"requirements.txt":            requirementsTxtContent,
 	}
 
-	var globalChoice string
+	globalSkip := false
 
 	for filename, content := range fileContentMap {
 		filePath := path.Join(cwd, filename)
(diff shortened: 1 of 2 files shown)


zeke:
Whoop whoop! ✨

PR #2437, opened 2025-07-02
feat: support cog init with pipeline template

markphelps:
Fixes: DP-85

Adds the `--x-pipeline` flag to `cog init` to create a pipeline template for cog

```
/tmp/cog-pipeline-test » /usr/local/bin/cog init --x-pipeline

✅ Created /tmp/cog-pipeline-test/.dockerignore
✅ Created /tmp/cog-pipeline-test/cog.yaml
✅ Created /tmp/cog-pipeline-test/main.py
✅ Created /tmp/cog-pipeline-test/requirements.txt

/tmp/cog-pipeline-test » ls -la
total 32
drwxr-xr-x@ 6 markphelps  wheel  192 Jul  2 12:57 .
drwxrwxrwt  9 root        wheel  288 Jul  2 12:55 ..
-rw-r--r--@ 1 markphelps  wheel  366 Jul  2 12:57 .dockerignore
-rw-r--r--@ 1 markphelps  wheel  455 Jul  2 12:57 cog.yaml
-rw-r--r--@ 1 markphelps  wheel  679 Jul  2 12:57 main.py
-rw-r--r--@ 1 markphelps  wheel  652 Jul  2 12:57 requirements.txt
```

## Example

```
/tmp/cog-pipeline-test » /usr/local/bin/cog predict \
  -i prompt="A tiny tiger in the rain" \
  --use-replicate-token
```

output

## Notes/Questions

1. I didnt include a github actions template like the regular init as I wasnt sure what inputs it would take/how it should work
2. I also chose to keep `--x-pipeline` flag hidden (from cli usage) as this matches `cog push` where `-x-pipeline` is also a hidden flag
3. Should we output or link to run instructions? Ie that you have to use  the `--x-pipeline` `--use-replicate-token` flags?


pkg/cli/init-templates/pipeline/requirements.txt
```diff
@@ -0,0 +1,27 @@
+# This is a normal Python requirements.txt file.
```
8W9aG: Can we remove this file? We don't support additional python packages in pipelines so it might lead to user confusion

markphelps: in the web editor it has these contents:

```
# IMPORTANT: The requirements.txt is only for local development,
# it will be ignored in production.
replicate>=1.1.0b1,<2.0.0a1
```

What is considered 'local' ?

pkg/cli/init-templates/pipeline/cog.yaml
```diff
@@ -0,0 +1,19 @@
...
+# Reference: https://cog.run/yaml
+
+build:
+  # a list of ubuntu apt packages to install
+  # system_packages:
+  #   - "libgl1-mesa-glx"
+  #   - "libglib2.0-0"
+
+  # path to a Python requirements.txt file
+  python_requirements: requirements.txt
```
markphelps: ill remove this line as well

pkg/cli/init-templates/pipeline/requirements.txt
ethulia: Since we don't support requirements.txt (aka custom dependencies) for pipelines running on Replicate, and we expect pretty much everyone to be developing pipelines so that they can eventually run them on Replicate (vs wanting to run them purely locally), we should just build the init under that assumption.

8W9aG (approved):

PR #2455, opened 2025-07-08
Remove python 3.12 from older torch versions

8W9aG:
* Torch >=2.0.0<=2.1.2 doesn’t support python 12
* This causes the base image builds to error


nevillelyh (approved):

8W9aG:
@michaeldwan I can look into this tomorrow but right now I need this unblocked

PR #2459, opened 2025-07-09
Fix torch compatibility matrix generation script

8W9aG:
* Expand the pypi scraper to handle newer version formats
* Explicitly warn if a bad format is encountered


michaeldwan:
awesome thank you!!!

PR #2479, opened 2025-07-30
Bump OpenCV

8W9aG:
* Bump opencv-python to a version that supports python 3.13


8W9aG:
Will merge to test this out on the base image builds

PR #2500, opened 2025-08-20
Fix -f flag

nevillelyh:
- **Fix cog build -f/--file flag**
- **Fix non-default cog.yaml inside container**

### DIFF
--- a/test-integration/test_integration/util.py
+++ b/test-integration/test_integration/util.py
@@ -12,7 +12,7 @@ import httpx
 from packaging.version import VERSION_PATTERN
 
 # From the SemVer spec: https://semver.org/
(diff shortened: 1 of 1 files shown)


pkg/cli/build.go
```diff
@@ -207,8 +207,7 @@ func addLocalImage(cmd *cobra.Command) {
 }
 
 func addConfigFlag(cmd *cobra.Command) {
-	const configFlag = "f"
-	cmd.Flags().StringVar(&configFilename, configFlag, "cog.yaml", "The name of the config file.")
+	cmd.Flags().StringVarP(&configFilename, "file", "f", "cog.yaml", "The name of the config file.")
```
nevillelyh, line 210: `-f/--file` instead of `--f` to be more consistent with Docker. Breaking is fine since it didn't work before anyway.

pkg/dockerfile/fast_generator.go
```diff
@@ -428,6 +428,11 @@ func (g *FastGenerator) installSrc(lines []string, weights []weights.Weight) ([]
 		lines = append(lines, "RUN "+strings.Join(linkCommands, " && "))
 	}
 
+	if filepath.Base(g.Config.Filename()) != "cog.yaml" {
+		// Absolute filename doesn't work anyway, so it's always relative
```
nevillelyh, line 432: Cog mistakenly uses `/` as project root if `-f /abs/path/to/cog.yaml` and fails on creating `/.cog` :shrug:

8W9aG (approved):

PR #2505, opened 2025-08-25
Disable coglet install codepath

michaeldwan:
Considering the significant changes under way over in cog-runtime, I'm disabling the codepath that installs coglet during model builds to minimize blast radius. Once we've sorted out concurrency issues with long-running models we can reassess.


michaeldwan:
Replaced by #2522

PR #2514, opened 2025-09-02
Swap out deprecated `ast.Str` for `ast.Constant`

aron:
This API will be removed in Python 3.14 and is creating noisy deprecation warnings.


erbridge (approved):
Thank you

PR #2515, opened 2025-09-04
Keep requirements.txt in sync for pipeline models

michaeldwan:
The packages installed in the pipelines-runtime image are drifting from what's specified by cog, which means that models may work locally but fail when deployed.

The proper fix for this is to bundle requirements/venv with the procedure. But that's out of scope. Until then this PR, which I admit is gross, will ensure the local requirements.txt file is always updated from the requirements.txt file generated by the pipelines-runtime build.

### DIFF
--- a/pkg/cli/init.go
+++ b/pkg/cli/init.go
@@ -5,12 +5,14 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path"
 	"time"
 
 	"github.com/spf13/cobra"
 
+	"github.com/replicate/cog/pkg/env"
 	"github.com/replicate/cog/pkg/util/console"
 	"github.com/replicate/cog/pkg/util/files"
 )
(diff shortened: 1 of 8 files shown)


pkg/cli/init.go
```diff
@@ -133,7 +137,21 @@ func processTemplateFile(fs embed.FS, templateDir, filename, cwd string) error {
 		} else {
 			content = downloadedContent
 		}
-	} else {
+	case filename == "requirements.txt" && pipelineTemplate:
+		// Special handling for requirements.txt in pipeline templates - download from runtime
```
8W9aG, line 141: Why include this at all? Don't we end up downloading it anyway even if it doesn't exist?

pkg/docker/pipeline_push.go
```diff
@@ -37,6 +38,25 @@ func createTarball(folder string) (*bytes.Buffer, error) {
...
 
+	// Track if we need to add downloaded requirements to the tarball
+	var downloadedRequirementsPath string
+	var downloadedRequirementsContent []byte
+
+	// If config points to downloaded requirements (outside project directory),
+	// we need to include them in the tarball as requirements.txt
+	if cfg.Build.PythonRequirements != "" {
+		reqPath := cfg.RequirementsFile(folder)
+		if !strings.HasPrefix(reqPath, folder) || strings.Contains(reqPath, ".cog") {
```
8W9aG: Instead of using `.cog` you can make use of the constant `CogBuildArtifactsFolder`

pkg/cli/init.go
michaeldwan: Yeah we download it on demand, but if the download fails for whatever reason it uses the embedded version during init. That whole process can get tidied but I didn't want to change the way the AGENTS.md file was handled as part of this.

PR #2541, opened 2025-09-30
feat: add registry migration support with credential fallback

markphelps:
## Summary

Implements support for other registries besides hardcoded `r8.im`

## Changes

- **Add `COG_REGISTRY_HOST` environment variable** for persistent registry override
- **Add `--registry` global persistent flag** (hidden) for one-off overrides  
- **Implement credential fallback**: when alternate registry has no credentials, automatically reuse `r8.im` credentials
- **Update API client** to dynamically load registry authentication
- **Update login messages** to show actual registry host being authenticated

## Configuration Priority

```
--registry flag (highest - one-off overrides)
       ↓
COG_REGISTRY_HOST env var (session/testing)
       ↓
Default "r8.im" (lowest - production)
```

## Usage

Internal teams can now test alternate registries:

```bash
# Set for current shell session
export COG_REGISTRY_HOST=x.r8.im
cog login
cog push x.r8.im/user/model

# Or use flag for one-off override
cog --registry x.r8.im login
```

## Credential Fallback

Existing `r8.im` credentials are automatically reused for alternate registries, displaying:
```
Using existing r8.im credentials for x.r8.im
```

## Testing

- ✅ All Go tests pass
- ✅ Code formatted and linted
- ✅ No regressions - default `r8.im` behavior unchanged
- ✅ Generic fallback works for any alternate registry, not just `x.r8.im`

### DIFF
--- a/pkg/cli/login.go
+++ b/pkg/cli/login.go
@@ -25,7 +25,7 @@ type VerifyResponse struct {
 }
 
 func newLoginCommand() *cobra.Command {
-	var cmd = &cobra.Command{
+	cmd := &cobra.Command{
 		Use:        "login",
 		SuggestFor: []string{"auth", "authenticate", "authorize"},
 		Short:      "Log in to Replicate Docker registry",
@@ -75,7 +75,7 @@ func login(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	console.Infof("You've successfully authenticated as %s! You can now use the %s registry.", username, registryHost)
+	console.Infof("You've successfully authenticated as %s! You can now use the '%s' registry.", username, registryHost)
 
 	return nil
 }
@@ -93,7 +93,7 @@ func readTokenInteractively(registryHost string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	console.Infof("This command will authenticate Docker with Replicate's %s Docker registry. You will need a Replicate account.", registryHost)
+	console.Infof("This command will authenticate Docker with Replicate's '%s' Docker registry. You will need a Replicate account.", registryHost)
 	console.Info("")
 
 	// TODO(bfirsh): if you have defined a registry in cog.yaml that is not r8.im, suggest to use 'docker login'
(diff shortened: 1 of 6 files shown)


pkg/cli/login.go
```diff
@@ -79,7 +75,7 @@ func login(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	console.Infof("You've successfully authenticated as %s! You can now use the '%s' registry.", username, registryHost)
+	console.Infof("You've successfully authenticated as %s! You can now use the %s registry.", username, registryHost)
```
tempusfrangit: This seems like an unrelated change
```suggestion
	console.Infof("You've successfully authenticated as %s! You can now use the '%s' registry.", username, registryHost)
	```

File: pkg/cli/login.go [hunk 2]
```diff
@@ -97,7 +93,7 @@ func readTokenInteractively(registryHost string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	console.Infof("This command will authenticate Docker with Replicate's '%s' Docker registry. You will need a Replicate account.", registryHost)
+	console.Infof("This command will authenticate Docker with Replicate's %s Docker registry. You will need a Replicate account.", registryHost)
```
tempusfrangit: Same as above
```suggestion
	console.Infof("This command will authenticate Docker with Replicate's '%s' Docker registry. You will need a Replicate account.", registryHost)
```

pkg/cli/login.go
markphelps: yup unintentional

pkg/cli/login.go
markphelps: yup unintentional

pkg/cli/login.go
michaeldwan: just pointing out for posterity: the %q format verb wraps the string value in escaped double quotes so we don't need to do manual quoting that linters/AI enjoy fixing

markphelps: yeah i thought about that but wanted to keep the changes minimal

PR #2562, opened 2025-11-06
Fix x-order bug

8W9aG:
* Pydantic 2.12.0 introduced a behaviour change that meant it copied `json_schema_extra` into `_attributes_set` making changes to the dictionary of `json_schema_extra` after the object construction not take effect
* We rely on making these changes, so instead reference the dictionary carried by the `_attributes_set`
* Also ensure that no `nullable` properties are `required` in the openapi schema render

### DIFF
--- a/python/cog/server/helpers.py
+++ b/python/cog/server/helpers.py
@@ -502,7 +502,7 @@ def _update_nullable_anyof(
             if len(non_null_items) < len(value) and not in_header:
                 openapi_schema["nullable"] = True
 
-    elif isinstance(openapi_schema, list): # type: ignore
+    elif isinstance(openapi_schema, list):  # type: ignore
         for item in openapi_schema:
             _update_nullable_anyof(item, in_header=in_header)
 


python/cog/server/helpers.py
```diff
@@ -593,3 +594,22 @@ def _restore_allof_for_prediction_id_put(
             ref = value["$ref"]
             del value["$ref"]
             value["allOf"] = [{"$ref": ref}]
+
+
+def _ensure_nullable_properties_not_required(openapi_schema: Dict[str, Any]) -> None:
```
andreasjansson, line 599: I found this function quite hard to read. Could it be simplified to something like

```py
def _ensure_nullable_properties_not_required(openapi_schema: Dict[str, Any]) -> None:
    schemas = openapi_schema["components"]["schemas"]
    for schema in schemas.values():
        properties = schema.get("properties", {})
        nullable = {k for k, v in properties.items() if v.get("nullable", False)}
        
        if "required" in schema and nullable:
            schema["required"] = [k for k in schema["required"] if k not in nullable]
```

8W9aG: Resolved in [7cba18b](https://github.com/replicate/cog/pull/2562/commits/7cba18be7c73fab52769c4d0db20669d7bf37910)

python/cog/predictor.py
```diff
@@ -282,20 +282,30 @@ class Config:
...
+        extra: Dict[str, Any] = {}
         if PYDANTIC_V2:
             # https://github.com/pydantic/pydantic/blob/2.7/pydantic/json_schema.py#L1436-L1446
             # json_schema_extra can be a callable, but we don't set that and users shouldn't set that
             if not default.json_schema_extra:  # type: ignore
-                default.json_schema_extra = {}  # type: ignore
+                default.json_schema_extra = {"x-order": order}  # type: ignore
             assert isinstance(default.json_schema_extra, dict)  # type: ignore
-            extra = default.json_schema_extra  # type: ignore
+            # In Pydantic 2.12.0 the json_schema_extra field is copied into a variable called "_attributes_set"
```
markphelps, line 297: we should probably add a test in a followup PR so we dont break this again going forward

markphelps (approved):
one comment about adding a test later but lgtm!

python/cog/predictor.py
8W9aG: Interesting thing here is tests are already failing on main due to this pydantic upgrade, so I do believe the tests are covering this

PR #2564, opened 2025-11-20
Secrets v1

NikhilSinha1:
### Summary

Start building implementation for passing secrets through. Locally, the secrets are pulled through the `.env` file. In production, cog creates an RSA key pair when it starts up and makes a request to a passed in URL for the secret, which it then decodes and can use.

### DIFF
--- a/test-integration/test_integration/test_build.py
+++ b/test-integration/test_integration/test_build.py
@@ -373,7 +373,7 @@ def test_pip_freeze(docker_image, cog_binary):
     )
     assert (
         pip_freeze
-        == "anyio==4.4.0\nattrs==23.2.0\ncertifi==2024.8.30\ncharset-normalizer==3.3.2\nclick==8.1.7\nexceptiongroup==1.2.2\nh11==0.14.0\nhttptools==0.6.1\nidna==3.8\npydantic==1.10.18\npython-dotenv==1.0.1\nPyYAML==6.0.2\nrequests==2.32.3\nsniffio==1.3.1\nstructlog==24.4.0\ntyping_extensions==4.12.2\nurllib3==2.2.2\nuvicorn==0.30.6\nuvloop==0.20.0\nwatchfiles==0.24.0\nwebsockets==13.0.1\n"
+        == "anyio==4.4.0\nattrs==23.2.0\ncertifi==2024.8.30\ncffi==2.0.0\ncharset-normalizer==3.3.2\nclick==8.1.7\ncryptography==46.0.3\nexceptiongroup==1.2.2\nh11==0.14.0\nhttptools==0.6.1\nidna==3.8\npycparser==2.23\npydantic==1.10.18\npython-dotenv==1.0.1\nPyYAML==6.0.2\nrequests==2.32.3\nsniffio==1.3.1\nstructlog==24.4.0\ntyping_extensions==4.15.0\nurllib3==2.2.2\nuvicorn==0.30.6\nuvloop==0.20.0\nwatchfiles==0.24.0\nwebsockets==13.0.1\n"
     )
 
 


python/cog/server/runner.py
meatballhat, line 1: It's a bit weird to me that `self._secrets_url` will only be present if passed into the `__init__`. Could the design be to always set `self._secrets_url` and then handle a `None` value elsewhere?

```python
class Always:
    def __init__(self, bla: str | None):
        self.bla = bla


class Sometimes:
    def __init__(self, bla: str | None):
        if bla is not None:
            self.bla = bla

a0 = Always("ok")
a1 = Always(None)

s0 = Sometimes("ok")
s1 = Sometimes(None)

print("a0", a0.bla)
print("a1", a1.bla)

print("s0", s0.bla)
print("s1", s1.bla)
```

```sh-session
$ python ~/tmp/always_sometimes.py
a0 ok
a1 None
s0 ok
Traceback (most recent call last):
  File "/Users/me/tmp/always_sometimes.py", line 21, in <module>
    print("s1", s1.bla)
                ^^^^^^
AttributeError: 'Sometimes' object has no attribute 'bla'
```

python/cog/secret.py
```diff
@@ -0,0 +1,97 @@
...
+                    label=None,
+                ),
+            )
+
+            return plaintext_bytes.decode("utf-8")
+        except Exception:
+            return self.env.get(secret_name, "")
+
+
+secret_provider = SecretProvider()
```
meatballhat: nit: I would prefer `default_secret_provider` since that reads more to me as "this is an instance and it is the default and you can make another one if you want" like it's a convenience, but not a strict singleton or borg.

python/cog/secret.py
```diff
@@ -0,0 +1,97 @@
...
+        public_pem = self.key.public_key().public_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PublicFormat.SubjectPublicKeyInfo,
+        )
+        public_key_path_raw = os.getenv(COG_PUBLIC_KEY_LOCATION_ENV_VAR_KEY)
+        if not public_key_path_raw:
+            self.no_public_key = True
+            return
+        public_key_path = Path(public_key_path_raw)
+        os.makedirs(os.path.basename(public_key_path), exist_ok=True)
```
meatballhat: nit
```suggestion
        public_key_path = Path(public_key_path_raw)
        public_key_path.parent.mkdir(mode=0700, exist_ok=True)
```

python/cog/secret.py
```diff
@@ -0,0 +1,97 @@
...
+        if not os.path.isfile(COG_ENV_LOCATION):
+            return
+        with open(COG_ENV_LOCATION) as f:
+            for line in f:
+                line = line.strip()
+                kv = line.split("=", 1)
+                # Skip bad entries with no equals sign
+                if len(kv) != 2:
+                    continue
+                self.env[kv[0]] = kv[1]
```
meatballhat: It looks like `python-dotenv` is already pulled in via `uvicorn[standard]`, so I'd opt for that instead of this implementation.

python/cog/secret.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def set_secret_url(self, secret_url: Optional[str] = None) -> None:
+        if self.once:
+            return
```
meatballhat: Can this be treated as an error condition?

python/cog/secret.py
meatballhat, line 1: I would love to see `COG_ENV_LOCATION` and `COG_PUBLIC_KEY_LOCATION_ENV_VAR_KEY` injected as defaults rather than referencing the module-level variables directly in the `__init__`.

Is the intended flow for the `cog` process to generate the RSA key pair and the remote `secret_url` also has access to the public key? I guess I don't follow the order of operations here, since it seems like the values loaded from `.cog/.env` need to be encrypted with a key pair that isn't known until _just before reading_ the `.cog/.env` values 🤔

python/cog/__init__.py
meatballhat, line 1: The existence of both `secret` and `Secret` with very different implementations is a source of confusion I would really love to avoid 😅 WDYT about a verb prefix like `get_secret` or `load_secret` or rolling it together with non-secret config values like a `cog.getenv` that automatically handles encrypted values?

PR #2619, opened 2026-01-08
Avoid panic and return helpful error when torch version is not specified

SharkFishDeveloper:
Fixes #2559

Cog previously panicked when no torch version was specified and CUDA
validation was triggered.

This change adds a guard in cudasFromTorch to return a helpful
user-facing error instead of panicking.

PR #2620, opened 2026-01-08
Fix/configurable user agent

### LINKED ISSUE #2554: please make user agent string configurable
Hello Replicate team,

We're using your whisperx containers for transcription services. We pass in audio file urls that need to be fetched.
However, some services reject the request due to the user agent string for the request. 

Tracing the code, the request gets made on this line with no way to configure the headers or request before it's made: https://github.com/replicate/cog/blob/1087c982a8b9306cc791db6db9d4b0d5c03b9a1d/python/cog/types.py#L383

SharkFishDeveloper:
Fixes #2554

Some external services reject requests made by Cog when fetching input URLs
because the default `python-requests` User-Agent is blocked or disallowed.

Previously, Cog used `requests.get()` without any way for users to configure
request headers. This caused URL fetches to fail for certain providers.

This change allows users to configure the User-Agent via an environment
variable (`COG_USER_AGENT`). If the variable is set, it is included in the
request headers; otherwise, the existing behavior remains unchanged.

This provides a simple, backward-compatible way to avoid request rejections
without modifying Cog internals or re-hosting input files.


markphelps:
> Nice improvement! I'm good moving it forward once we add an integration test. @markphelps can you do that?

I can in a bit. @SharkFishDeveloper this looks like it also contains the fix in #2619 . Can we rebase this onto main instead without that change since its in another PR?

PR #2631, opened 2026-01-15
fix: make TestPredictionPathUploadIterator deterministic

markphelps:
## Summary

Fixes the flaky `TestPredictionPathUploadIterator` test that was failing in CI due to race conditions in webhook delivery ordering.

## Problem

The test assumed webhooks would arrive in strict chronological order with specific statuses at specific indices. However, webhooks are delivered asynchronously and can arrive out of order, especially in CI environments with varying performance.

Example failure:
```
path_test.go:296: 
    Error:      	Not equal: 
                	expected: "processing"
                	actual  : "succeeded"
```

This occurred when the prediction completed faster than expected and the "succeeded" webhook arrived before some "processing" webhooks were consumed from the channel.

## Solution

Refactored the test to:
1. **Collect all webhooks** until the final "succeeded" webhook arrives (non-blocking)
2. **Validate the collected set** rather than assuming specific order
3. **Check for presence** of expected states (initial processing, outputs with 1/2/3 items) using a set-based approach
4. **Make upload validation order-independent** by collecting all uploads first, then verifying their contents

## Testing

- ✅ Test passes 10 consecutive runs locally
- ✅ All path tests pass 5 consecutive runs
- ✅ Code formatted with `make fmt`

The test is now resilient to timing variations and webhook delivery order.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -200,7 +200,6 @@ jobs:
           COG_WHEEL: ${{ matrix.runtime }}
           COG_BINARY: ./cog
           TEST_PARALLEL: 4
-          BUILDKIT_PROGRESS: plain
         run: make test-integration-go
 
   # TODO[md]: This is a gross hack, remove once this is sorted out: https://github.com/replicate/cog/pull/2353
--- a/integration-tests/harness/harness.go
+++ b/integration-tests/harness/harness.go
@@ -187,7 +187,6 @@ func (h *Harness) cmdCog(ts *testscript.TestScript, neg bool, args []string) {
 	}
 
 	// Default: run cog command normally
-	// Note: BUILDKIT_PROGRESS=quiet is set in Setup() to suppress Docker build output
 	expandedArgs := make([]string, len(args))
 	for i, arg := range args {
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -199,7 +199,8 @@ jobs:
         env:
           COG_WHEEL: ${{ matrix.runtime }}
           COG_BINARY: ./cog
-          TEST_PARALLEL: 8
+          TEST_PARALLEL: 4
+          BUILDKIT_PROGRESS: plain
```
michaeldwan: I'm surprised we need this. It should default to plain unless there's a PTY attached. Build code is changing soon and this will get addressed, but until then pay attention to regressions on the other side where we're not rendering build output correctly in dev.

markphelps: yeah i changed it in a previous PR to default to quiet cause it was hard to debug failures

markphelps: i will undo that change and have it default back to plain in code

markphelps: oh nvm that change was just in the integration test harness, not actual cog code

PR #2635, opened 2026-01-16
Add design document

bfirsh:
Just a brain dump. Could do with some organization.


michaeldwan (approved):
This is great! Merging now and we'll iterate as a living doc.

PR #2641, opened 2026-01-16
Cog Server Rewrite

tempusfrangit:
This is a complete rewrite of the cog http server moving away from python and to a rust pyo3, ABI3 wheel. The aim is to provide a significant uplift in control of how we run predictors and isolate the predictor run from the core interface used to run the predictor. 

```
═══════════════════════════════════════════════════════════════════════════════════
                           COMPONENT OWNERSHIP
═══════════════════════════════════════════════════════════════════════════════════
  PredictionSupervisor (DashMap - lock-free concurrent access)
  ├── owns: prediction state (id, status, input, output, logs, error, timestamps)
  ├── owns: webhook sender (sends terminal webhook, then cleans up entry)
  ├── owns: cancel_token (propagates cancellation to worker)
  └── owns: completion notifier (for waiting on result)
  PredictionSlot (RAII container)
  ├── owns: Prediction (logs, outputs from worker event loop)
  ├── owns: Permit (concurrency token, returns to pool on drop)
  └── Drop: marks permit idle, releases back to pool
  PredictionHandle (returned to route handler)
  ├── has: reference to supervisor (for state queries)
  ├── has: cancel_token clone (for cancellation)
  ├── has: completion notifier clone (for waiting)
  └── method: sync_guard() → SyncPredictionGuard (cancels on drop)
═══════════════════════════════════════════════════════════════════════════════════
                         WORKER SUBPROCESS PROTOCOL
═══════════════════════════════════════════════════════════════════════════════════
  Control Channel (stdin/stdout - JSON framed)
  ┌─────────────────────────────────────────────────────────────────────────────┐
  │  Parent → Child                    Child → Parent                           │
  │  ──────────────                    ──────────────                           │
  │  Init { predictor_ref, slots }     Ready { schema }                         │
  │  Cancel                            Cancelled                                │
  │  Shutdown                          Failed { error }                         │
  │                                    Idle                                     │
  └─────────────────────────────────────────────────────────────────────────────┘
  Slot Channel (Unix socket per slot - JSON framed)
  ┌─────────────────────────────────────────────────────────────────────────────┐
  │  Parent → Child                    Child → Parent                           │
  │  ──────────────                    ──────────────                           │
  │  Predict { id, input }             Log { data }                             │
  │                                    Output { value }  (streaming)            │
  │                                    Done { output }                          │
  │                                    Failed { error }                         │
  │                                    Cancelled                                │
  └─────────────────────────────────────────────────────────────────────────────┘
[...]

PR #2643, opened 2026-01-20
Replace boolean flags with Rust enums

markphelps:
## Summary

Addresses review feedback from #2641 by replacing boolean "soup" with strongly-typed Rust enums across four key areas of the coglet codebase.

### 1. Predictor Method Characteristics (`predictor.rs`)
**Before**: 5 boolean fields
- `is_async`, `is_async_gen`, `has_train`, `is_train_async`, `is_standalone_function`

**After**: `PredictorKind` enum with nested structure
```rust
enum PredictorKind {
    Class { predict: PredictKind, train: TrainKind },
    StandaloneFunction(PredictKind),
}

enum PredictKind { Sync, Async, AsyncGen }
enum TrainKind { None, Sync, Async }
```

**Benefits**: Invalid combinations like `is_async=false` + `is_async_gen=true` are now impossible at compile time.

### 2. Prediction Results (`worker.rs`)
**Before**: `success: bool` + `error: Option<String>` with string matching
```rust
if result.error.as_deref() == Some("Cancelled") { ... }
```

**After**: `PredictionOutcome` enum
```rust
enum PredictionOutcome {
    Success { output: Value, predict_time: f64 },
    Failed { error: String, predict_time: f64 },
    Cancelled { predict_time: f64 },
}
```

**Benefits**: Type-safe pattern matching, no string comparisons for cancellation detection.

### 3. Worker Bridge Mode (`worker_bridge.rs`)
**Before**: `is_train: bool`

**After**: `HandlerMode` enum
```rust
enum HandlerMode { Predict, Train }
```

**Benefits**: Explicit intent, clearer than boolean flag.

### 4. Slot State (`worker_bridge.rs`)
**Before**: 4 fields tracking state
- `cancelled: bool`, `in_progress: bool`, `is_async: bool`, `async_future: Option<Py<PyAny>>`

**After**: `SlotState` enum with proper state machine
```rust
enum SlotState {
    Idle,
    SyncPrediction { cancelled: bool },
    AsyncPrediction { future: Py<PyAny>, cancelled: bool },
}
```

**Benefits**: State machine clarity, cancellation context preserved for choosing the right cancel mechanism (SIGUSR1 vs future.cancel()).

## Impact

- **Type Safety**: Invalid states are impossible at compile time
- **Readability**: Self-documenting code with clear semantics
- **Maintainability**: Easier to extend with new variants
- **No External Changes**: Internal Rust refactoring only, no API impact

## Testing

- ✅ Compiles cleanly with no warnings (`cargo check --all`)
- ✅ All existing functionality preserved
- ✅ No changes to bridge protocol or Python API

## Stats

- 3 files changed
- +385 insertions, -268 deletions

### DIFF
--- a/crates/coglet-python/src/worker_bridge.rs
+++ b/crates/coglet-python/src/worker_bridge.rs
@@ -52,11 +52,6 @@ impl SlotState {
             SlotState::Idle => { /* no-op */ }
         }
     }
-
-    #[allow(dead_code)]
-    pub fn is_in_progress(&self) -> bool {
-        !matches!(self, SlotState::Idle)
-    }
 }
 
 /// Wraps PythonPredictor to implement the PredictHandler trait.
@@ -304,12 +299,8 @@ impl PredictHandler for PythonPredictHandler {
         tracing::trace!(%slot, %id, is_async, "Got predictor");
 
         // Track that we're starting a prediction on this slot
-        if is_async {
-            // Note: we'll update with the actual future later
-            self.start_sync_prediction(slot); // Temporary - will be replaced when we get the future
-        } else {
-            self.start_sync_prediction(slot);
-        }
+        // Note: we'll update with the actual future later if async
+        self.start_sync_prediction(slot);
 
         // Check cancellation first (in case cancel was called before we started)
         if self.take_cancelled(slot) {


crates/coglet-python/src/worker_bridge.rs
```diff
@@ -250,23 +287,29 @@ impl PredictHandler for PythonPredictHandler {
...
             }
         };
+        let is_async = pred.is_async();
         tracing::trace!(%slot, %id, is_async, "Got predictor");
 
         // Track that we're starting a prediction on this slot
-        self.start_prediction(slot, is_async);
+        if is_async {
+            // Note: we'll update with the actual future later
+            self.start_sync_prediction(slot); // Temporary - will be replaced when we get the future
```
mfainberg-cf: I see the comment that we're replacing with the future, but is there a reason for the if/else if we call self.start_sync_prediction(slot) in either case?

PR #2649, opened 2026-01-23
Use recently built image instead of latest for src

NikhilSinha1:
### Summary

We have been investigating a caching issue that has lead to versions of cog including an old version of the `src` folder when running `cog build` or `cog push` on a model that already has a version. We are fairly sure that in certain circumstances, when the `FROM r8.im/<username>/<model-name>` command is run, it is pulling the latest from the upstream registry instead of using the recently built image from the local machine. To disambiguate, we should directly pass the SHA of the image on the local machine instead of letting docker just look for `latest`, as it would do by default.

NOTE: this problem seems to be happening on machines with Docker desktop. Machines with orbstack may not see this issue

### Test Plan

A script, `./script/test-e2e-rebuild`, was introduced, that simply makes a small change to the `predict.py` function and tries to rebuild the image. It then checks that the hash of the `predict.py` built into the docker image is the same as the one in the `predict.py` source directory for building the image.

To verify the correctness of this change, build `cog` from source on the `main` branch. Then checkout this branch and run the test. The exit code of the script should be 1. Rebuild `cog` from this branch, and re-run the test. The exit code of the script should be 0.


pkg/docker/docker.go
```diff
@@ -277,6 +278,34 @@ func (c *apiClient) Inspect(ctx context.Context, ref string) (*image.InspectResp
 	return &inspect, nil
 }
 
+func (c *apiClient) LocalImageID(ctx context.Context, ref string) (string, error) {
+	console.Debugf("=== APIClient.LocalImageTag %s", ref)
```
meatballhat-cf: I was expecting this debug line to say `APIClient.LocalImageID` 🤔

script/test-e2e-rebuild
meatballhat-cf, line 1: I haven't found a way to run this that produces the incorrect hash message 🤔

pkg/image/build.go
```diff
@@ -321,8 +321,16 @@ func Build(
 //
 // The new image is based on the provided image with the labels and schema file appended to it.
 func BuildAddLabelsAndSchemaToImage(ctx context.Context, dockerClient command.Command, image string, labels map[string]string, bundledSchemaFile string, progressOutput string) error {
-	dockerfile := "FROM " + image + "\n"
-	dockerfile += "COPY " + bundledSchemaFile + " .cog\n"
+	if strings.HasPrefix("image", "r8.im") {
```
meatballhat-cf: did you mean?

```suggestion
	if strings.HasPrefix(image , "r8.im") {
```

NikhilSinha1: I did

script/test-e2e-rebuild
meatballhat-cf: We talked in a separate channel and current working theory is that this bug manifests in Docker Desktop but *not* in Orbstack 🙃

PR #2652, opened 2026-01-24
Add user-defined healthcheck

NikhilSinha1:
### Summary

We want to add a user-defined healthcheck function to allow users to tell us if their container is healthy or not, given they may have some more information about what determines a "healthy" system than we do. This PR adds hooks for the user to pass a `healthcheck` function to us and we can run it on behalf of the user whenever we hit the `/health-check` endpoint

### Test Plan

Unit tests added to verify this works when a user's healthcheck succeeds, fails, times out and errors

### DIFF
--- a/integration-tests/tests/healthcheck_custom.txtar
+++ b/integration-tests/tests/healthcheck_custom.txtar
@@ -1,0 +1,44 @@
+# Skip for coglet_alpha which does not support this
+[coglet_alpha] skip
+# Skip for coglet_rust which does not support this
+[coglet_rust] skip
+# Skip for cog-dataclass which does not support this
+[cog_dataclass] skip
+
+# Test custom healthcheck functionality
+# This tests the user-defined healthcheck() method in predictors
+
+# Build the image
+cog build -t $TEST_IMAGE
+
+# Start the server
+cog serve
+
+# Test 1: Healthy healthcheck returns READY status
+curl GET /health-check
+stdout '"status":"READY"'
+! stdout 'user_healthcheck_error'
+
+# Test 2: Make a prediction to ensure predictor works
+curl POST /predictions '{"input":{"text":"world"}}'
+stdout '"output":"hello world"'
+
+# Test 3: Health check still works after prediction
+curl GET /health-check
+stdout '"status":"READY"'
+
+-- cog.yaml --
+build:
+  python_version: "3.12"
+predict: "predict.py:Predictor"
+
+-- predict.py --
+from cog import BasePredictor
+
+class Predictor(BasePredictor):
+    def predict(self, text: str) -> str:
+        return f"hello {text}"
+    
+    def healthcheck(self) -> bool:
+        """Custom healthcheck that always returns healthy."""
+        return True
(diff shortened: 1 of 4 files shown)


python/cog/server/worker.py
```diff
@@ -742,6 +779,75 @@ async def _apredict(
...
+                    )
+                else:
+                    # If sync, run in executor with timeout
+                    loop = asyncio.get_event_loop()
+                    result = await asyncio.wait_for(
+                        loop.run_in_executor(None, healthcheck_fn),
+                        timeout=HEALTHCHECK_TIMEOUT,
+                    )
+
+                if result is False or result is None:
```
meatballhat-cf: I find this a bit confusing that `None` counts as unhealthy since it's the default value if the function has a bare `return` or no `return` statement at all. WDYT about only treating `result is False` as failure?

python/cog/server/worker.py
```diff
@@ -742,6 +779,75 @@ async def _apredict(
...
+                    done.error = True
+                    done.error_detail = (
+                        "Healthcheck failed: user-defined healthcheck returned False"
+                    )
+        except asyncio.TimeoutError:
+            done.error = True
+            done.error_detail = f"Healthcheck failed: user-defined healthcheck timed out after {HEALTHCHECK_TIMEOUT} seconds"
+            print(f"Healthcheck timed out after {HEALTHCHECK_TIMEOUT} seconds")
```
meatballhat-cf: Where is this output intended to go?

python/cog/server/http.py
```diff
@@ -358,10 +358,31 @@ async def root() -> Any:
...
             health = Health.BUSY if runner.is_busy() else Health.READY
+
+            # Run custom healthcheck. If it doesn't exist, this will
+            # always return healthy (healthcheck_result.error = False)
+            healthcheck_result = await runner.healthcheck()
+            custom_health_ok = not healthcheck_result.error
+            custom_health_error = healthcheck_result.error_detail
+
+            if not custom_health_ok:
+                health = Health.SETUP_FAILED
```
meatballhat-cf: You're sure we shouldn't introduce a new value like `Health.UNHEALTHY`?

python/cog/server/worker.py
mfainberg-cf: Can we require a non-None response instead? [I know, python, so no not rtequire but set typdefs to indicate not none?] - I agree with dan here. I'd encourage the response to be True for healthy, and false-y or exception = unhealthy.

python/cog/server/worker.py
NikhilSinha1: I think to the user visible logs right, since it's the user defined one I think it's good for them to see this. Guess it should be a log.warn instead though

python/cog/server/worker.py
NikhilSinha1: Yeah I think false only simplifies it. But I think we should fail always if it's not a bool? Or should we succeed always

NikhilSinha1: We decided to go with whatever `bool(x)` returns

python/cog/server/http.py
NikhilSinha1: Oh, whoops, I didn't even notice I did this

PR #2656, opened 2026-01-27
chore: add useragent support to cog-dataclass

markphelps:
like #2620 but for new pydanticless cog

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -168,6 +168,8 @@ jobs:
         run: |
           uv venv
           uv pip install $(ls dist/cog_dataclass-*.whl)[dev]
+      - name: Remove src to ensure tests run against wheel
+        run: rm -rf cog-dataclass/python/cog
       - name: Test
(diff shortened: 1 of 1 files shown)


cog-dataclass/python/cog/_version.py
```diff
@@ -28,7 +28,7 @@
 commit_id: COMMIT_ID
 __commit_id__: COMMIT_ID
 
-__version__ = version = '0.1.dev2006+gbd4814d49.d20260123'
-__version_tuple__ = version_tuple = (0, 1, 'dev2006', 'gbd4814d49.d20260123')
+__version__ = version = "0.1.dev2006+gbd4814d49.d20260123"
+__version_tuple__ = version_tuple = (0, 1, "dev2006", "gbd4814d49.d20260123")
 
-__commit_id__ = commit_id = 'gbd4814d49'
+__commit_id__ = commit_id = "gbd4814d49"
```
mfainberg-cf: We need to unstage this. This is auto-generated at compile time.

PR #2664, opened 2026-01-29
Fix build with older installed golang versions

lkolbly:
I tried building cog on my Linux machine (with `make cog`), and it failed unless I did these things.

Because of https://stackoverflow.com/questions/78519711/toolchain-not-available-error-prevents-me-from-using-any-go-commands and also because it seems like part of the build step needs some wheels built?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,8 @@
 module github.com/replicate/cog
 
-go 1.25
+go 1.25.3
 
 require (
-	github.com/alecthomas/kong v1.12.1
 	github.com/anaskhan96/soup v1.2.5
 	github.com/aquasecurity/go-pep440-version v0.0.1
 	github.com/aws/aws-sdk-go-v2 v1.36.3
@@ -12,7 +11,6 @@ require (
 	github.com/docker/cli v28.3.0+incompatible
 	github.com/docker/docker v28.1.1+incompatible
 	github.com/docker/go-connections v0.5.0
-	github.com/gabriel-vasile/mimetype v1.4.10
 	github.com/getkin/kin-openapi v0.128.0
 	github.com/google/go-containerregistry v0.20.5
 	github.com/hashicorp/go-version v1.7.0
@@ -35,7 +33,6 @@ require (
 	github.com/vincent-petithory/dataurl v1.0.0
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/xeonx/timeago v1.0.0-rc5
-	go.uber.org/zap v1.27.0
 	golang.org/x/crypto v0.38.0
 	golang.org/x/exp v0.0.0-20250408133849-7e4ce0ab07d0
 	golang.org/x/sync v0.14.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,9 +1,8 @@
 module github.com/replicate/cog
 
-go 1.25
```
markphelps, line 3: undo this please

markphelps (changes requested):

go.mod
lkolbly: Ah, no can do mi amigo:
```
(base) lkolbly@B0K3C14:~/cog$ make cog
... lots of stuff ...
go: downloading go1.25 (linux/amd64)
go: download go1.25 for linux/amd64: toolchain not available
make: *** [Makefile:77: generate] Error 1
(base) lkolbly@B0K3C14:~/cog$ 
```

michaeldwan: @lkolbly how is go installed? what version? this should "just work" since we perform builds and tests in CI on linux

lkolbly: @michaeldwan I have 1.22.3:
```
(base) lkolbly@B0K3C14:~$ go version
go version go1.22.3 linux/amd64
```
Installed presumably by following the instructions at go.dev (which is to... manually unpack a tarball? Hmm someone should invent a tool to do that for us :D)

Okay so I'm reading https://go.dev/doc/toolchain#config, and it says that:

> If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line. ... For example, a go.mod that says go 1.21.0 with no toolchain line is interpreted as if it had a toolchain go1.21.0 line.

This is a `go.mod` with no toolchain line, so I guess presumably the toolchain would default to `go1.25`?

...but, `go1.25` is not a valid toolchain, which is the error I'm seeing:

> For example, go1.21rc1 and go1.21.0 are toolchain names; go1.21 and go1.22 are not (the initial releases are go1.21.0 and go1.22.0)

So my reading is that, if you don't have a toolchain file, it's invalid to not specify the full patch version for the language? It's weird that go wouldn't lint that though.

(of course, if you have a higher toolchain installed, then go will just use that - in CI presumably you have at least 1.25 installed, so it doesn't try to install 1.25)

But, I think also, this means that I should set the patch version to 1.25.0 (because otherwise it changes the toolchain (and language?) for people with versions 1.25.0 through 1.25.5), I've updated the PR with this change.

(specifying 1.25.0 does exclude all the RCs, I think, where 1.25 maybe does not. Don't quote me on that)

Makefile
```diff
@@ -24,7 +24,7 @@ default: all
 .PHONY: all
 all: cog
 
-$(COG_BINARIES): $(COG_GO_SOURCE) generate
+$(COG_BINARIES): wheel $(COG_GO_SOURCE) generate
```
lkolbly: This change seems to break a lot of things, I'll drop it and just live with having to build wheels 😔

anish-sahoo:
closing due to inactivity and `mise` support that installs the right toolchain.

PR #2665, opened 2026-01-29
Fix: use tmpImageId for schema validation when pushing to r8.im

lucataco:
When pushing to r8.im/*, commit 2ad2c7f6 introduced temporary image naming to fix OrbStack compatibility, but missed updating three function calls that run containers to extract metadata:

- GenerateOpenAPISchema
- GeneratePipFreeze
- GenerateModelDependencies

These functions were still using the original imageName (e.g., r8.im/replicate/model-test) instead of tmpImageId (cog-tmp:<hash>), causing 'No such image' errors during the 'Validating model schema' step.

This fix ensures all three functions use the temporary image ID that was actually built by Docker.

### DIFF
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -213,7 +213,7 @@ func Build(
 		schemaJSON = data
 	} else {
 		console.Info("Validating model schema...")
-		schema, err := GenerateOpenAPISchema(ctx, dockerCommand, imageName, cfg.Build.GPU)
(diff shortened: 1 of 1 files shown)


gandalfhz-cf (approved):
LGTM

pkg/image/build.go
bzikst, line 178: Do we need tmpImageId instead of imageName here as well?

PR #2666, opened 2026-01-30
chore: EOL python below python 3.10

mfainberg-cf:
- Raise minimum supported Python to 3.10 across config validation, dockerfile logic, CI/test matrices, and docs.
- Update integration fixtures to use Python 3.10, add explicit 3.8/3.9 deprecation tests, and drop the legacy optional_input test.
- Refresh AST/migration helpers for 3.10+ parsing and allow rust/dataclass runs in the CUDA baseimage IT.

### DIFF
--- a/tools/compatgen/internal/torch_test.go
+++ b/tools/compatgen/internal/torch_test.go
@@ -20,8 +20,8 @@ func TestFetchTorchPackages(t *testing.T) {
 		if err != nil {
 			log.Fatalf("Error reading file: %v", err)
 		}
-		w.Write(content)
 		w.WriteHeader(http.StatusOK)
+		w.Write(content)
 	}))
 	defer server.Close()
 	url, err := url.Parse(server.URL)
@@ -67,13 +67,6 @@ func TestFetchTorchPackages(t *testing.T) {
 			Variant:       "cu128",
 			CUDA:          &cuda128,
 			PythonVersion: "3.13",
-		},
-		{
-			Name:          "2.7.1+cu128",
-			Version:       "2.7.1",
-			Variant:       "cu128",
-			CUDA:          &cuda128,
-			PythonVersion: "3.10",
 		},
 	}, torch271Packages)
 }
--- a/tools/compatgen/internal/torch.go
+++ b/tools/compatgen/internal/torch.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"slices"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/anaskhan96/soup"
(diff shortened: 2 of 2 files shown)


integration-tests/tests/apt_packages.txtar
```diff
@@ -14,7 +11,7 @@ stdout 'hello world'
 -- cog.yaml --
 build:
   gpu: true
-  python_version: "3.9"
+  python_version: "3.10"
```
michaeldwan, line 14: Random thought, but I wonder if we should omit the python version from fixtures unless we're specifically testing version behavior. When missing it'll use the default, 3.13 atm.

integration-tests/tests/optional_input.txtar
```diff
@@ -1,32 +0,0 @@
-# Test optional input parameters with default None
-# Skip for cog-dataclass and coglet (Rust) which require Python 3.10+
```
michaeldwan, line 2: Why was this test removed?

pkg/config/config_test.go
```diff
@@ -154,7 +154,7 @@ func assertMinorVersion(t *testing.T, expected, actual string) {
 func TestPythonPackagesAndRequirementsCantBeUsedTogether(t *testing.T) {
 	config := &Config{
 		Build: &Build{
-			PythonVersion: "3.8",
+			PythonVersion: "3.10",
```
michaeldwan, line 157: omg maybe we should use a constant or something so we don't need this again when 3.10 is removed in 6mo

tools/compatgen/internal/torch_test.go
```diff
@@ -20,8 +20,8 @@ func TestFetchTorchPackages(t *testing.T) {
 		if err != nil {
 			log.Fatalf("Error reading file: %v", err)
 		}
-		w.Write(content)
 		w.WriteHeader(http.StatusOK)
+		w.Write(content)
```
michaeldwan, line 24: lol nice fix

michaeldwan (approved):
approved, but a few nits for those who follow

integration-tests/tests/apt_packages.txtar
tempusfrangit: We should. that seems like a useful enhancement.

integration-tests/tests/optional_input.txtar
tempusfrangit: This was testing a python 3.8 specific branch of code in the pydantic cog. In 3.10+ it would fail because it was looking for a particular non-type default mapping peculular to the 3.8 we had. Dataclass was agnostic to this.

pkg/config/config_test.go
tempusfrangit: or... i like your previouos take... stop specifying.

PR #2675, opened 2026-02-02
Remove deprecated fast/monobase build system

michaeldwan:
Remove the deprecated fast/monobase/monobeam build system and `cog_runtime` config option. This functionality was experimental, never shipped to users, and is being removed.

## Removed

- `pkg/monobeam/` package
- `pkg/docker/fast_push.go`, `monobase.go`
- `pkg/dockerfile/fast_generator.go`, `monobase_matrix.go`
- `pkg/weights/fast_weights.go`
- `--x-fast` CLI flag
- `build.fast`, `build.cog_runtime`, `build.python_overrides` config fields
- Related integration tests

## Result

62 files changed, ~2500 lines removed. The standard build path is now the only path.


integration-tests/tests/build_base_image_sha.txtar
```diff
@@ -1,4 +1,4 @@
-# Skip for cog-dataclass and coglet (Rust) which do not support monobase
```
markphelps, line 1: `%s/monobase/'cog base images'/g`

but srs should we be skipping these now?

michaeldwan: hah good catch. ngl I was waiting to see how tests do before adjusting anything, but GHA has been down all day and I forgot 😭. Will revise because that's a bogus gsub!

michaeldwan: This change is actually somewhat correct, just oddly phrased because of terminology overload. `--use-cog-base-image` skips the build by injecting user code atop a literal base image, the kind we'd use in `FROM <base>`, but it also skips the wheel installation in some paths. Accidentally similar to monobase. So any test that uses a non-standard wheel might not work. I need to verify and fix the naming, but we might need to continue skipping for now and rip out that footgun flag when we have fast builds by default.

michaeldwan: Okay this works correctly, removing the skips.

PR #2681, opened 2026-02-04
feat: Replace Pydantic with native Python dataclasses for cog.BaseModel

tempusfrangit:
This PR replaces Pydantic with native Python dataclasses as the foundation for `cog.BaseModel` and the input/output type system. This is a significant architectural change that simplifies the codebase, removes a major dependency, and provides a more predictable, lightweight runtime.

## Motivation
- **Simplification**: Pydantic added complexity for maintaining compatibility across v1 and v2 APIs
- **Performance**: Native dataclasses have lower overhead than Pydantic models
- **Predictability**: Removes magic behavior and implicit coercion from Pydantic
- **Maintenance**: Eliminates need to support both Pydantic 1.x and 2.x codepaths

### Python SDK (`python/cog/`)
- **New `cog.BaseModel`**: Now a native Python dataclass instead of Pydantic BaseModel
- **New ADT system** (`_adt.py`): Algebraic Data Type utilities for type-safe unions and enums
- **New type coders** (`coder/`): Modular encode/decode system for all supported types (primitives, files, paths, secrets, lists, etc.)
- **Simplified input handling** (`input.py`): Clean dataclass-based input validation
- **Updated `cog.Field()`**: Provides metadata (default, description, ge, le, choices) without Pydantic dependency
### Rust Coglet (`crates/`)
- **Removed Pydantic detection**: No longer checks for or handles Pydantic-specific code paths
- **Simplified input processing**: Direct dataclass field extraction
- **User-defined healthcheck support**: New `healthcheck()` method support with 5-second timeout
  - Sync and async healthcheck support
  - Proper error handling and timeout messages
  - `UNHEALTHY` status in health response
### Integration Tests
- Removed Pydantic-specific tests (`build_pydantic1_none`, `complex_types`, etc.)
- Added async healthcheck tests (4 new test files)
- Enabled `[coglet_rust]` for healthcheck tests (removed skips)
- Updated `complex_output` to use `cog.BaseModel` instead of `pydantic.BaseModel`

### Removed Features
- Pydantic 1.x/2.x compatibility layer
 
## Breaking Changes
1. **`cog.BaseModel` is now a dataclass** - Models must be defined as dataclasses, not Pydantic models
2. **No implicit type coercion** - Types must match exactly (no automatic string→int conversion)
3. **`cog.Field()` API changes** - Uses dataclass field metadata instead of Pydantic Field

### DIFF
--- a/python/cog/server/http.py
+++ b/python/cog/server/http.py
@@ -62,6 +62,35 @@ if TYPE_CHECKING:
 
 log = structlog.get_logger("cog.server.http")
 
+# Known safe validation messages - these exact messages (or prefixes) are safe to return
+# We return only the safe portion, never the full message which may contain sensitive data
+_SAFE_VALIDATION_MESSAGES = {
+    "Field required": "Field required",
+    "Invalid value": "Invalid value",
+    "fails constraint": "Value fails constraint",
+    "does not match regex": "Value does not match required pattern",
+    "does not match choices": "Value does not match allowed choices",
+}
+
+
+def _sanitize_validation_message(msg: str) -> str:
+    """
+    Sanitize validation error messages to prevent information leakage.
+
+    Returns only predefined safe messages, never the original message content
+    which could contain sensitive internal details.
+    """
+    if not msg:
+        return "Invalid value"
+
+    # Check if message contains any known pattern and return ONLY the safe version
+    for pattern, safe_msg in _SAFE_VALIDATION_MESSAGES.items():
+        if pattern in msg:
+            return safe_msg
+
+    # For unknown messages, return generic error
+    return "Invalid value"
+
 
 @unique
 class Health(Enum):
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -141,47 +131,13 @@ jobs:
...
-          uv pip install $(ls dist/cog_dataclass-*.whl)[dev]
-      - name: Remove src to ensure tests run against wheel
-        run: rm -rf cog-dataclass/python/cog
-      - name: Test
-        run: uv run pytest cog-dataclass/python/tests -v
+        run: uvx --with tox-uv tox run --installpkg dist/cog-*.whl -e ${{ env.TOX_PYTHON }}-tests
 
   test-coglet-python:
-    name: "Test Coglet Python bindings (${{ matrix.runtime }})"
+    name: "Test Coglet Python bindings"
```
michaeldwan, line 137: Probably out of scope for this PR, but I wonder if we should be testing this, and all the other python code, in a matrix across supported python versions.

michaeldwan (approved):
This beefy PR looks good. It's a lot, but no blocking issues... so @tempusfrangit merge it and we'll chat through a few gaps in test coverage that were accidentally covered in the deleted tests.

.github/workflows/ci.yaml
tempusfrangit: It should be something we consider but if we fail on any version of python it's a bug in Maturin/pyo3 since we're compiling to pure ABI3, 3.10+

PR #2688, opened 2026-02-05
feat: add real-time progress tracking for multipart weight uploads

markphelps:
## Summary

- Implements multipart uploads for large weight blobs using Content-Range headers
- Adds real-time progress tracking during chunk uploads (updates every ~32-64KB instead of every 25MB chunk)
- Fixes bug where commit used stale location URL instead of the final location with updated state hash

## Changes

- Modified `uploadChunk` to wrap the chunk reader with progress tracking, reporting progress as bytes are written to the network
- Updated `uploadBlobChunks` to return the final location URL for proper commit handling
- Unified channel handling to use non-blocking sends across both single-part and multipart uploads
- Added defensive progress capping at `totalSize` to prevent overshoot

### DIFF
--- a/pkg/registry/registry_client.go
+++ b/pkg/registry/registry_client.go
@@ -606,7 +606,7 @@ func (c *RegistryClient) initiateUpload(ctx context.Context, client *http.Client
 func (c *RegistryClient) uploadBlobChunks(ctx context.Context, client *http.Client, repo name.Repository, layer v1.Layer, location string, totalSize int64, progressCh chan<- v1.Update) (string, error) {
 	// Multipart upload settings:
 	// - Threshold: Use multipart only for blobs larger than 50MB (avoids MPU overhead for smaller files)
-	// - Chunk size: 25MB per chunk (good balance for R2 object store)
+	// - Chunk size: 25MB per chunk (good balance for object stores and typical network conditions)
 	const multipartThreshold = 50 * 1024 * 1024
 	const chunkSize = 25 * 1024 * 1024
 
@@ -722,11 +722,8 @@ func (c *RegistryClient) uploadBlobSingle(ctx context.Context, client *http.Clie
 		onRead: func(n int) {
 			uploaded += int64(n)
 			if progressCh != nil {
-				complete := uploaded
 				// Cap at totalSize defensively
-				if complete > totalSize {
-					complete = totalSize
-				}
+				complete := min(uploaded, totalSize)
 				select {
 				case progressCh <- v1.Update{Complete: complete, Total: totalSize}:
 				default:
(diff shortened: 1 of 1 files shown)


pkg/registry/registry_client.go
```diff
@@ -474,6 +469,385 @@ func (c *RegistryClient) WriteLayer(ctx context.Context, opts WriteLayerOptions)
...
+	blob, err := layer.Compressed()
+	if err != nil {
+		return "", fmt.Errorf("getting compressed blob: %w", err)
+	}
+	defer blob.Close()
+
+	// Try multipart upload first, fall back to single-part if not supported
+	// Chunk size: 25MB for multipart uploads
+	const chunkSize = 25 * 1024 * 1024
+	tryMultipart := totalSize > chunkSize // Only try multipart for large blobs
```
mfainberg-cf: When we move to kong we should consider making this a tunable. There are reasons for changing the MPU size depending on the objectstore backing for R2 this is a really good starting opklace

I hihgly recommend we make this two distinc axis:

Threshold for MPU and MPU chunk size. this means we can be a little fatter on the single upload size. I'd probably pick 50M and a 25M chunk otherwise.  Or at minimum set the value for MPU triggering to be significantly more than 26M it's a bit silly to MPU 25M and 1M.

pkg/registry/registry_client.go
```diff
@@ -474,6 +469,385 @@ func (c *RegistryClient) WriteLayer(ctx context.Context, opts WriteLayerOptions)
...
+			return "", err
+		}
+	}
+
+	// Single-part upload for small blobs or servers that don't support multipart
+	err = c.uploadBlobSingle(ctx, client, location, blob, totalSize, progressCh)
+	if err != nil {
+		return "", err
+	}
+	return location, nil
```
mfainberg-cf: Nit: I'd likely pull the logic for try MPU into a function so it's not a lot of location logic. IT's fine as is...

pkg/registry/registry_client.go
```diff
@@ -474,6 +469,385 @@ func (c *RegistryClient) WriteLayer(ctx context.Context, opts WriteLayerOptions)
...
+	req.Header.Set("Content-Type", "application/octet-stream")
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	return transport.CheckError(resp, http.StatusCreated)
+}
```
mfainberg-cf, line 859: If we're terminally failing here maybe we should cancel the MPU? Are we planning to have a resume MPU function in the case the CLI crashes or is SIGINT'd? These are questions not any changes needed.

markphelps: good questions. yeah we should do cleanup on early exit. will do in a followup. first let me check if the registry has gc too

PR #2701, opened 2026-02-06
fix: cross-compile all coglet wheels from ubuntu using zig

tempusfrangit:
Drop macOS runners and macOS x64 target. Build linux-x64, linux-arm64, and macos-arm64 wheels from ubuntu-latest using maturin --zig.

Fixes uv cache error on ARM runners and reduces runner costs.


tempusfrangit:
Bypassing as this change only impacts release workflow.

PR #2703, opened 2026-02-09
chore: modernize go

markphelps:
- sets go version in mise.toml instead of latest
- run `golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize` on go code


tempusfrangit:
can we also switch yaml from the unmaintained yaml lib to https://github.com/yaml/go-yaml ? This could be it's own PR or logically part of this one.

PR #2717, opened 2026-02-13
chore(coglet): Rust safety improvements and wheel platform fix

tempusfrangit:
## Summary 
- **Pass transport info from Init** — Remove `unsafe { std::env::set_var }` from coglet-python, pass `ChildTransportInfo` as a parameter instead
- **Add Fatal IPC message and panic hook** — Worker panics now send a Fatal message over IPC, orchestrator poisons slots and fails predictions cleanly
- **Move slot poisoning to pool level** — Per-slot `AtomicBool` poison flags in `PermitPool`, lazy filtering on acquire, removes poison tracking from predictions
- **Coalesce concurrent healthcheck requests** — All concurrent callers wait on the same in-flight healthcheck and get the real result instead of fake "healthy"
- **Fix coglet wheel platform selection** — `cog build` now selects the correct platform-specific coglet wheel based on Docker target arch; also fixes `GOOS`/`GOARCH` initialization bug in `StandardGenerator`

### DIFF
--- a/pkg/dockerfile/standard_generator.go
+++ b/pkg/dockerfile/standard_generator.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"runtime"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


tempusfrangit:
This should be reviewed commit-by-commit. Each entry is a commit. I bundled it since it was relatively contained to cleanup/safety. The wheel thing was discovered while trying to test locally.

PR #2719, opened 2026-02-16
Add mise to install instructions

bfirsh:


tempusfrangit:
@bfirsh there is a mise task to update LLM Docs. `mise run docs:llm` this should update the LLM Docs and we can get this added quickly

PR #2722, opened 2026-02-16
fix: suppress verbose coglet logs during cog predict

bfirsh:
## Summary
- Rename `COG_LOG` to `COG_LOG_LEVEL` in coglet Rust code to match the env var the Go CLI already sets
- Configure Python structlog from `COG_LOG_LEVEL` in the server entry point
- Demote "Starting Rust coglet server" to debug level
- Demote "Large frame being encoded" from warn to info

The Go CLI sets `COG_LOG_LEVEL=warning` by default and `COG_LOG_LEVEL=debug` with `--debug`, but coglet was reading `COG_LOG` instead, so the setting was ignored and all info-level JSON log lines were printed during `cog predict`.

## Test plan
- [ ] Run `cog predict` on a model — should no longer show coglet JSON log spam
- [ ] Run `cog predict --debug` — should show all log output including coglet info/debug lines
- [ ] Verify "Starting Rust coglet server" only appears with `--debug`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/coglet-python/README.md
+++ b/crates/coglet-python/README.md
@@ -247,7 +247,7 @@ worker_bridge.setup()
 
 **Worker Startup:**
 1. `set_active()` - Mark as worker subprocess
-2. `init_tracing()` - Configure logging (stderr, COG_LOG env)
+2. `init_tracing()` - Configure logging (stderr, COG_LOG_LEVEL env)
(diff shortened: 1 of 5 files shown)


tempusfrangit (approved):
Perfect! Good catch!

PR #1696, opened 2024-05-29
Generate CLI docs with Cobra [WIP]

zeke:
This is an experiment. Nothing to see here yet...

PR #2736, opened 2026-02-17
Generate CLI docs from Go code

bfirsh:
This implements CLI documentation generation using cobra/doc, building on the work started in #1696.

## Changes

- Add `tools/gendocs/` Go tool to generate CLI reference docs from cobra commands
- Generate `docs/cli.md` programmatically from Go source code
- Add mise tasks:
  - `mise run docs:cli` - generate CLI docs
  - `mise run docs:cli:check` - verify docs are up to date
- Run CLI docs check in CI alongside llms.txt check
- Update AGENTS.md with CLI docs maintenance instructions
- Update docs workflow to generate CLI docs before deploying

## Usage

```bash
# Generate CLI docs
mise run docs:cli

# Check CLI docs are up to date (runs in CI)
mise run docs:cli:check
```

## Notes

- CLI docs are auto-generated and should not be edited manually
- The check runs in CI when `cmd/` or `pkg/cli/` files change
- Docs include all commands, options, and examples from cobra definitions

Closes #1696

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,0 +1,178 @@
+# CLI reference
+
+<!-- This file is auto-generated. Do not edit manually. -->
+
+## `cog`
+
+Containers for machine learning.
+
+To get started, take a look at the documentation:
+https://github.com/replicate/cog
+
+**Examples**
+
+```
+   To run a command inside a Docker environment defined with Cog:
+      $ cog run echo hello world
+```
+
+**Options**
+
+```
+      --debug     Show debugging output
+  -h, --help      help for cog
+      --version   Show version of Cog
+```
+## `cog build`
+
+Build an image from cog.yaml
+
+```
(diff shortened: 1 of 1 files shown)


mise.toml
```diff
@@ -565,6 +565,28 @@ fi
 echo "docs/llms.txt is up to date"
 """
 
+[tasks."docs:cli"]
```
tempusfrangit, line 569: I would add the 

`depends = ["//:docs:cli"]` to the `docs:llm` task that way docs:llm will always generate the cli docs if needed.

tempusfrangit (changes requested):
This looks fine but we should wire up mise to do some lifting for us!

mise.toml
bfirsh: Ah good idea! @copilot

PR #2476, opened 2025-07-29
feat: use uv instead of pip in container builds

paulnegz:
This change replaces pip with uv for Python package installation in container builds. Key changes:
- Update StandardGenerator to use uv for package installation
- Add proper uv caching configuration
- Update tests to expect uv-based commands
- Update documentation to reflect uv usage

Fixes #2167

PR #2747, opened 2026-02-19
Use uv and uv run pip for Dockerfile generation

tempusfrangit:
- Replace pyenv/pip with uv for Python installation (uv python install)
- Use UV_SYSTEM_PYTHON=true so uv operates on the system Python
- For user requirements (pip install -r), use 'uv run pip' which invokes Python's own pip/OpenSSL stack, working around rustls's lack of P-521 ECDSA support needed for Cloudflare Warp MITM certificates
- Append CA cert directly to bundle file in addition to update-ca-certificates for robustness across base images
- Update all tests to match new Dockerfile output

closes: #2744 
closes: #2167 
closes: #2476


pkg/dockerfile/standard_generator.go
```diff
@@ -425,47 +435,26 @@ func (g *StandardGenerator) installPython() (string, error) {
 	return "", nil
 }
 
+func (g *StandardGenerator) installUV() string {
```
markphelps, line 438: nit, this could just be a regular function since it doesnt need `g`. or could just be a constant. super nit tho

markphelps (approved):
nit picks but lgtm

pkg/dockerfile/standard_generator.go
tempusfrangit: good point! happy to convert for cleanliness.

tempusfrangit: I'll spin up a new change to do some cleanup after deep in some refactors.

PR #2755, opened 2026-02-20
feat: add Homebrew cask to replicate/tap for prebuilt macOS binaries

### LINKED ISSUE #2738: Fix codesigning on release binaries
On macos: 

Image

Should be a quick fix.

markphelps:
## Summary

- Configures GoReleaser to auto-publish a Homebrew **cask** (prebuilt binary) to `replicate/homebrew-tap` on each stable release
- The cask includes a `post_install` hook that clears the macOS quarantine attribute, so users never hit the Gatekeeper "cannot be opened" warning
- Updates docs, README, and install script to recommend `brew install replicate/tap/cog` as the primary macOS install method
- Pre-releases are automatically skipped (`skip_upload: auto`)

## Why a cask instead of the existing Homebrew Core formula?

| | Homebrew Core (current) | Own Tap (this PR) |
|---|---|---|
| Install speed | ~60s (compiles from source) | ~5s (prebuilt download) |
| Version freshness | Lags (PR to homebrew-core) | Instant (auto-pushed on release) |
| Binary matches release | No (Homebrew rebuilds) | Yes (exact same binary) |
| Quarantine handling | N/A (source build) | Cleared via post-install hook |

We keep the Homebrew Core formula for discoverability — this tap gives users a faster, always-current alternative.

## Setup required

A `HOMEBREW_TAP_TOKEN` repository secret must be added — a GitHub PAT (classic) with `repo` scope that has write access to `replicate/homebrew-tap`.

## Related

- Follows up on #2754 (xattr workaround)
- Closes #2738


tempusfrangit:
I am working on this app option instead.

PR #2760, opened 2026-02-23
feat: chunked image push via OCI compatible push

markphelps:
## Summary

Replace Docker's monolithic `ImagePush` with chunked layer uploads, gated behind `COG_PUSH_OCI=1`. The image is exported from Docker via `ImageSave`, loaded lazily using go-containerregistry's `tarball.ImageFromPath` (layers are read on-demand from a temp file, not into memory), then each layer is pushed through the existing `RegistryClient.WriteLayer` chunked upload path — the same infrastructure weight artifacts already use. This bypasses request body limits that block Docker's native push for large layers.

Off by default. Set `COG_PUSH_OCI=1` to enable. Falls back to Docker push on any non-fatal error (except context cancellation/timeout and auth errors), so there's zero regression risk.

### Chunked image push
- `ImagePusher` tries OCI chunked push first when `COG_PUSH_OCI=1`, falls back to Docker push
- Exports image via `docker.ImageSave()` → tar → lazy v1.Image (via `tarball.ImageFromPath`) → concurrent layer upload
- `shouldFallbackToDocker()` blocks fallback on context cancellation/timeout and authentication errors (401/403) — auth errors won't be fixed by Docker fallback, so the original error is surfaced directly
- Added `ImageSave(ctx, imageRef)` to the Docker `Command` interface
- `ImagePusher.Push()` takes `*ImageArtifact` directly (not a raw string ref), preserving the resolved reference from the Model type

### Unified push infrastructure
- Single `PushProgress` type replaces separate image/weight progress types
- `writeLayerWithProgress()` helper deduplicates progress channel boilerplate
- `GetPushConcurrency()` (default 5, `COG_PUSH_CONCURRENCY`) shared by image layers, weight pushes, and CLI progress
- `BundlePusher.pushWeights()` uses `errgroup` with concurrency limit
- `NewBundlePusher` takes `docker.Command` + `registry.Client` and creates both sub-pushers (image + weight) internally — unified construction

### Progress rendering
- Replaced the `mpb` progress bar library with Docker's `jsonmessage.DisplayJSONMessagesStream` — the same rendering code used by `docker push`
- Progress output now matches `docker push` exactly: layer ID, status, and progress bar on each line
- Handles terminal resizing correctly: each line is erased and rewritten individually (`ESC[2K` + per-line cursor up/down), rather than relying on a bulk cursor-up count that desyncs when lines wrap after a resize
- Terminal width is re-queried via `ioctl(TIOCGWINSZ)` on every render, so progress bars adapt dynamically
- `ProgressWriter` adapter in `pkg/docker/progress.go` converts `PushProgress` callbacks into JSON-encoded `JSONMessage` structs fed to `DisplayJSONMessagesStream` via an `io.Pipe`
- Retry messages are only logged via `console.Warnf` in non-TTY mode — in TTY mode the progress writer handles display inline, avoiding duplicate output
- Removed `mpb` dependency entirely

### HTTP/1.1 transport for all registry operations
- All registry operations (push, inspect, HEAD, etc.) now use HTTP/1.1 instead of allowing HTTP/2 negotiation
- **Problem**: Go's default `http.Transport` negotiates HTTP/2 via TLS ALPN. High-throughput HTTP/2 suffers from head-of-line blocking when multiplexing large blob uploads on a single connection. Additionally, certain CDN/proxy edges can send `RST_STREAM INTERNAL_ERROR` on large requests — killing uploads before they reach the origin.
- **Fix**: `RegistryClient` stores an `http1OnlyTransport()` and passes it to all `remote.*` calls via a `remoteOptions()` helper. Multiple concurrent HTTP/1.1 connections outperform a single multiplexed HTTP/2 connection for large blob uploads.
- HTTP/2 stream errors (`RST_STREAM`) are also recognized by `isRetryableError()` as a belt-and-suspenders measure, so they trigger the existing retry-from-scratch logic in `WriteLayer`.

### Retry and resource management
- Retry jitter is now properly randomized (`rand.Float64()`) to avoid thundering herd when multiple clients retry simultaneously
- Chunk buffers (up to 96 MB each) are pooled via `sync.Pool` to reduce memory pressure when pushing multiple layers concurrently

### Server-driven chunk sizing (OCI-Chunk-Min/Max-Length)
- `initiateUpload()` now parses `OCI-Chunk-Min-Length` and `OCI-Chunk-Max-Length` headers from the registry's 202 response
- The server-advertised maximum always takes precedence over client defaults — chunk size is set to `max - 64KB` margin to stay safely under the limit
- The result is clamped to be at least the server-advertised minimum
- `COG_PUSH_DEFAULT_CHUNK_SIZE` is only used as a fallback when the registry doesn't advertise limits
- New `uploadSession` struct carries location + chunk constraints from initiation through to upload

### Registry config
- Extracted `getDefaultChunkSize()` and `getMultipartThreshold()` env var helpers into `pkg/registry/config.go`
- Default chunk size: 96 MB (multiple of 8), stays under common CDN/proxy request body limits (used only when registry doesn't advertise `OCI-Chunk-Max-Length`)
- Default multipart threshold: 128 MB (higher than chunk size so blobs that fit in a single chunk skip multipart entirely)

### Cleanup
- Deleted `pkg/oci/` package — OCI image loading inlined into `ImagePusher`
- Deleted `tools/uploader/` — unused S3 multipart uploader (363 lines, zero imports)
- Removed `Pusher` interface, `OCIImagePusher`, `pushImageWithFallback()` — consolidated into single `ImagePusher`
- Unexported `DefaultFactory` → `defaultFactory`, `NewImagePusher` → `newImagePusher`

## Environment variables

| Variable | Default | Description |
|---|---|---|
| `COG_PUSH_OCI` | unset (off) | Set to `1` to enable OCI chunked image push |
| `COG_PUSH_CONCURRENCY` | `5` | Max concurrent layer/weight uploads |
| `COG_PUSH_DEFAULT_CHUNK_SIZE` | `100663296` (96 MiB) | Fallback chunk size when registry doesn't advertise `OCI-Chunk-Max-Length` |
| `COG_PUSH_MULTIPART_THRESHOLD` | `134217728` (128 MiB) | Blobs above this size use chunked upload; below this they upload in a single request |

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -3,7 +3,7 @@ resolver = "2"
 members = ["coglet", "coglet-python"]
 
 [workspace.package]
-version = "0.17.0"
+version = "0.17.0-rc.1"
 edition = "2024"
 license = "Apache-2.0"
 repository = "https://github.com/replicate/cog"


pkg/model/push_helpers.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/replicate/cog/pkg/registry"
+)
+
+const (
+	// DefaultPushConcurrency is the default number of concurrent uploads
+	// for both image layers and weight artifacts.
+	DefaultPushConcurrency = 4
+
+	// envPushConcurrency is the environment variable that overrides DefaultPushConcurrency.
+	envPushConcurrency = "COG_PUSH_CONCURRENCY"
```
mfainberg-cf, line 20: Thanks for having this!

pkg/registry/config.go
```diff
@@ -0,0 +1,52 @@
...
+	// DefaultMultipartThreshold is the minimum blob size (in bytes) before using multipart upload.
+	// Blobs smaller than this are uploaded in a single request to avoid multipart overhead.
+	DefaultMultipartThreshold = 50 * 1024 * 1024 // 50 MB
+
+	// DefaultChunkSize is the size (in bytes) of each chunk in a multipart upload.
+	// This is used as a fallback when the registry does not advertise chunk size
+	// limits via OCI-Chunk-Min-Length / OCI-Chunk-Max-Length headers.
+	// 95 MB stays under common CDN/proxy request body limits while still being
+	// large enough to reduce HTTP round-trips for multi-GB files.
+	DefaultChunkSize = 95 * 1024 * 1024 // 95 MB
```
mfainberg-cf: Shouldn't these be inverted, typically we want the threshold to be higher than the chunksize by not a small margin, this prevents MPU for slightly larger than N. It's fine as is, but we're going to MPU for 1 chunk for the default.

pkg/registry/registry_client.go
```diff
@@ -348,6 +350,40 @@ func (c *RegistryClient) PushIndex(ctx context.Context, ref string, idx v1.Image
...
+// http1OnlyTransport returns an http.Transport that only speaks HTTP/1.1.
+// This avoids HTTP/2 stream errors (RST_STREAM INTERNAL_ERROR) that occur when
+// uploading large blobs through certain CDN/proxy edges.
+func http1OnlyTransport() *http.Transport {
+	t := http.DefaultTransport.(*http.Transport).Clone()
+	t.TLSClientConfig = tlsConfigHTTP1Only(t.TLSClientConfig)
+	// ForceAttemptHTTP2 is true by default on cloned transports; disable it.
+	t.ForceAttemptHTTP2 = false
+	return t
+}
```
mfainberg-cf, line 360: I hate how broken HTTP2 is. if we move to 1.26 we have some additional knobs with adjusting new connections per stream. Additionally, we should really consdier if we want HTTP2 at all vs multiple H1 connections only. High throughput http2 often has issues with the internal muxing of the streams on a single connection. This isn't tiny files a browser is downloading.

markphelps: right, head of line blocking and all that. good point. we can switch to http1 always

pkg/cli/progress.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/replicate/cog/pkg/util/console"
+)
+
+// progressWriter adapts push progress callbacks to Docker's jsonmessage rendering.
+//
+// This uses the same ANSI cursor movement and progress display as `docker push`,
+// which handles terminal resizing correctly: each line is erased and rewritten
+// individually (ESC[2K + cursor up/down per line), rather than relying on a
+// bulk cursor-up count that can desync when lines wrap after a terminal resize.
+type progressWriter struct {
```
michaeldwan: This is really a docker thing and I wonder if the type should live in the docker package. we'll soon need a standalone progress reporter for non-docker things, like fetching/processing weights, which feels more appropriate for a CLI UI concern.

pkg/cli/weights.go
```diff
@@ -214,31 +203,32 @@ func weightsPushCommand(cmd *cobra.Command, args []string) error {
 		err  error
 	}
 
-	const maxConcurrency = 4
-	sem := make(chan struct{}, maxConcurrency)
+	sem := make(chan struct{}, model.GetPushConcurrency())
```
michaeldwan: This should probably be an error group with a concurrent limit set

pkg/model/image_pusher.go
```diff
@@ -1,36 +1,322 @@
...
 		return fmt.Errorf("image has no reference")
 	}
-	return p.docker.Push(ctx, artifact.Reference)
+	return p.Push(ctx, artifact.Reference)
+}
+
+// canOCIPush returns true if OCI chunked push is enabled.
+// Requires COG_PUSH_OCI=1 and a registry client.
+func (p *ImagePusher) canOCIPush() bool {
+	return os.Getenv("COG_PUSH_OCI") == "1" && p.registry != nil
```
michaeldwan: What would cause the registry to be nil? curious if we could get into a place where a wonky code path was expecting to push a model bundle but failed because we didn't set a registry so a standard push happened instead, which is not a model bundle.

pkg/model/image_pusher.go
```diff
@@ -1,36 +1,322 @@
...
+}
+
+// ociPush exports the image from Docker daemon as a tar, then pushes all layers,
+// config, and manifest to the registry using chunked uploads.
+func (p *ImagePusher) ociPush(ctx context.Context, imageRef string, opt ImagePushOptions) error {
```
michaeldwan: this is a regression. the pusher originally took a fully resolved Model type, which already has a parsed ref and all that. going back to a string sidesteps that recent improvement and opens the door to issues like the ref we thought we were pushing resolved to something else immediately before hitting docker. basically the only place we're dealing with a string reference is in resolve, the returned Model is either fully resolved and correct, or not, and we use that everywhere.

pkg/model/pusher.go
```diff
@@ -49,9 +53,9 @@ type BundlePusher struct {
 }
 
 // NewBundlePusher creates a new BundlePusher.
-func NewBundlePusher(docker command.Command, reg registry.Client) *BundlePusher {
+func NewBundlePusher(imagePusher *ImagePusher, reg registry.Client) *BundlePusher {
 	return &BundlePusher{
-		imagePusher:  NewImagePusher(docker),
+		imagePusher:  imagePusher,
```
michaeldwan: This is exposing some of the leaky design recently replaced. Why are we creating a weight pusher in this function but the image pusher in another? Part of the recent refactor was to unify the way we create these things so there's only one way to do it. ie given docker and registry clients, we can create a BundlePusher or ImagePusher and both satisfy the interface the CLI/model package operates on. nothing knows or cares what we're dealing with, only the code that determined which pusher to return. for example, we're going to probably add the json schema file as another bundle artifact, which might force knowledge of that (creation and consumption) onto the CLI and all this other code.

pkg/model/pusher.go
```diff
@@ -121,50 +132,28 @@ func (p *BundlePusher) Push(ctx context.Context, m *Model, opts PushOptions) err
...
-		result *WeightPushResult
-		err    error
-	}
+func (p *BundlePusher) pushWeights(ctx context.Context, repo string, weights []*WeightArtifact) ([]*WeightPushResult, error) {
+	ordered := make([]*WeightPushResult, len(weights))
 
-	results := make(chan indexedResult, len(weights))
-	var wg sync.WaitGroup
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(GetPushConcurrency())
```
michaeldwan, line 146: nice

pkg/model/image_pusher.go
```diff
@@ -1,36 +1,322 @@
...
+		tag = ref.Context().Tag("latest")
+	}
+
+	img, err := tarball.ImageFromPath(tmpTar.Name(), &tag)
+	if err != nil {
+		return fmt.Errorf("load image from tar: %w", err)
+	}
+
+	return p.pushImage(ctx, imageRef, img, opt)
+}
```
markphelps, line 155: need to refactor this so we dont load the entire image into memory to push

PR #2779, opened 2026-02-26
fix: route List[Path] outputs through FileOutput IPC for upload

tempusfrangit:
## Summary
- When a predict function returns `List[Path]` (not an iterator), the list was processed by `output::process_output` which base64-encodes `Path` objects inline, producing multi-megabyte data URIs in the response instead of uploaded URLs
- `process_single_output` now detects list/tuple returns and iterates their items through `send_output_item`, which routes `Path` objects through the FileOutput IPC path for proper upload via `--upload-url`
## Testing
- New integration test `coglet_list_path_upload_url` verifies `List[Path]` outputs are uploaded and returned as URLs, not data URIs
- Existing `coglet_iterator_upload_url` test still passes (Iterator[Path] path unaffected)

closes: #2777


tempusfrangit:
This has been confirmed to be resolved via E2E test on replicate with a real published model.

michaeldwan (approved):

PR #2782, opened 2026-02-27
feat: build-time schema generation in Go (tree-sitter)

tempusfrangit:
## Summary
Replaces runtime Python schema generation with a **pure Go** implementation using tree-sitter, eliminating the need for a separate Rust binary (`cog-schema-gen`) or container-based schema extraction for SDK >= 0.17.0.
This is the alternative approach to #2774 (which embedded a Rust binary via `go:embed`), motivated by reviewer feedback about embedding complexity.
### Go CLI — Static Schema Generation
- **`pkg/schema/python/parser.go`** — Tree-sitter AST walker that parses Python predict/train files into `PredictorInfo`. Handles `Union[str, None]`, `typing.List`, `Literal`, enums, all default types, async iterators, etc.
- **`pkg/schema/openapi.go`** — Full OpenAPI 3.0.2 generation from `PredictorInfo` with deterministic key ordering and compact JSON (for Docker label efficiency)
- **`pkg/schema/generator.go`** — Public API: `Generate()`, `GenerateFromSource()`, `GenerateCombined()`, `MergeSchemas()`, plus `COG_OPENAPI_SCHEMA` env var bypass
- **`pkg/schema/types.go`** — Type system with Union handling, TitleCase helpers, field resolution
- **`pkg/image/build.go`** — Pre-build static schema gen (Go tree-sitter) for SDK >= 0.17.0, post-build legacy fallback (container-based) for SDK < 0.17.0
### Coglet (Rust) — Bundled Schema + Edge Validation
- **Removed runtime schema generation** — `schema()` removed from `PredictHandler` trait; coglet now reads `.cog/openapi_schema.json` from disk at startup
- **`input_validation.rs`** — Validates prediction/training inputs against the OpenAPI schema at the HTTP edge before dispatching to Python. Returns 422 with pydantic-compatible error format
- **Training route idempotency** — Training routes now have proper ID mismatch checks and existing-state returns (not just delegating to prediction routes)
- `COG_MAX_CONCURRENCY` read from env var instead of Python introspection
### Build Infrastructure
- **CGo enabled globally** — Required for `go-tree-sitter` (C bindings). Uses zig for Linux cross-compilation, native clang for macOS
- **Release builds on `macos-14`** — Single ARM64 macOS runner cross-compiles all 4 targets
- **Dockerfile changes** — `cogEnvVars()` injects `COG_PREDICT_TYPE_STUB`, `COG_TRAIN_TYPE_STUB`, `COG_MAX_CONCURRENCY`; coglet install is conditional
### Tests
- **100+ Go unit tests** across parser, OpenAPI generator, and schema generator
- **9 input validation unit tests** (Rust)
- **2 integration tests** — `static_schema_gen.txtar` (end-to-end with Docker label verification) and `legacy_sdk_schema.txtar` (SDK < 0.17.0 fallback)
## Design Decisions
- **Go tree-sitter over Rust binary** — Direct library call, no embed/exec/temp files. CGo is the only tradeoff.
- **Compact JSON** — `json.Marshal` not `MarshalIndent` for Docker labels
- **Belt-and-suspenders validation** — Rust validates at HTTP edge for fast rejection; Python `check_input()` still runs in worker (removed in follow-up PR)
- **SDK version gating** — Static path for >= 0.17.0, legacy container-based path for older SDKs
## Follow-up Work (Stacked PR)
- `_adt.py`/`_inspector.py` elimination + full Rust input coercion
- Dead Python SDK code deletion

### DIFF
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -351,7 +351,7 @@ func predictIndividualInputs(predictor predict.Predictor, inputFlags []string, o
 		return err
 	}
 
-	inputs, err := parseInputFlags(inputFlags, schema)
+	inputs, err := parseInputFlags(inputFlags, schema, isTrain)
 	if err != nil {
 		return err
 	}
@@ -622,7 +622,7 @@ func processFileOutputs(output any, schema *openapi3.Schema, destination string)
 	return output, nil
 }
 
-func parseInputFlags(inputs []string, schema *openapi3.T) (predict.Inputs, error) {
+func parseInputFlags(inputs []string, schema *openapi3.T, isTrain ...bool) (predict.Inputs, error) {
 	keyVals := map[string][]string{}
 	for _, input := range inputs {
 		var name, value string
@@ -644,7 +644,8 @@ func parseInputFlags(inputs []string, schema *openapi3.T) (predict.Inputs, error
 		keyVals[name] = append(keyVals[name], value)
 	}
 
-	return predict.NewInputs(keyVals, schema)
+	train := len(isTrain) > 0 && isTrain[0]
+	return predict.NewInputsForMode(keyVals, schema, train)
 }
 
 func addSetupTimeoutFlag(cmd *cobra.Command) {
--- a/pkg/predict/input.go
+++ b/pkg/predict/input.go
@@ -28,11 +28,28 @@ type Input struct {
 type Inputs map[string]Input
 
 func NewInputs(keyVals map[string][]string, schema *openapi3.T) (Inputs, error) {
+	return NewInputsForMode(keyVals, schema, false)
+}
+
+func NewInputsForMode(keyVals map[string][]string, schema *openapi3.T, isTrain bool) (Inputs, error) {
+	schemaKey := "Input"
+	if isTrain {
+		schemaKey = "TrainingInput"
+	}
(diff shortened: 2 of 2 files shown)


pkg/image/build.go
```diff
@@ -68,6 +73,48 @@ func Build(
...
+	//   - The SDK version is >= 0.17.0 (or unpinned/latest/dev)
+	//
+	// Legacy generation is needed for SDK < 0.17.0 which uses pydantic-based
+	// schemas that cannot be statically analyzed.
+	useStatic := !skipSchemaValidation && schemaFile == "" && canUseStaticSchemaGen(cfg)
+
+	// --- Pre-build static schema generation ---
+	// When using the static path, generate schema BEFORE the Docker build so we
+	// fail fast on schema errors and the schema file is in the build context.
+	var schemaJSON []byte
```
michaeldwan, line 90: wish we could break some of this out to smaller functions. I can see this important looking code mangled by accident, or lingering forever because the agents are afraid to delete stuff or break backwards compatibility the humans already forgot about

michaeldwan (approved):
looks good, much simpler than the embedded rust verison. nits about extracting some logic out into smaller functions, but not blocking

PR #2783, opened 2026-02-27
feat: move input coercion & output encoding to Rust, delete dead SDK modules

tempusfrangit:
## Summary
- **Move input coercion and output encoding from Python to Rust** in coglet-python, eliminating runtime calls to `_adt.py` and `_inspector.py`
- **Delete 20 dead Python SDK modules** (-3,071 lines) whose functionality now lives in Rust coglet
- **Simplify remaining SDK** to a thin surface: `BasePredictor`, `Input`, types, and `BaseModel`
## Commit 1: Rust coercion (`input.rs`, `output.rs`, `predictor.rs`)
**`input.rs`** — Replaced `Runtime`/`InputProcessor`/`CogInputProcessor` architecture with `prepare_input(py, input, func)`:
- `detect_file_fields()` — introspects type annotations for `cog.File`/`cog.Path` params (handles `Optional`, `list[File]`, `Union[File, None]`)
- `coerce_url_strings()` — converts URL strings to `Path.validate()` or `File.validate()` based on annotations
- `download_url_paths_into_dict()` — parallel `URLPath` downloads via `ThreadPoolExecutor`
**`output.rs`** — Pure Rust output processing replacing `cog.json.make_encodeable()` + `cog.files.upload_file()`:
- `make_encodeable()` — handles Pydantic v1/v2, dataclasses, enums, datetime, numpy, `PathLike`, collections
- `encode_files()` — recursive `Path`/`IOBase` → base64 data URL conversion
- `file_to_base64()` — MIME type guessing, seek-to-start, base64 encoding
**`predictor.rs`** — Removed `runtime`/`input_processor` fields:
- `predict_func()`/`train_func()` — getters for type annotation introspection
- `unwrap_field_info_defaults()` — patches `__defaults__`/`__kwdefaults__` to replace `FieldInfo(default=42)` wrappers with actual values
## Commit 2: SDK cleanup
**Deleted modules:** `_adt`, `_inspector`, `_schemas`, `coder`, `coders/*`, `command/*`, `config`, `errors`, `exceptions`, `files`, `json`, `logging`, `mode`, `schema`, `suppress_output`
**Simplified:**
- `input.py` — removed `Representation`, mutable default auto-conversion; `default_factory` now raises `TypeError`
- `predictor.py` — removed `get_predict`/`get_train`/`wait_for_env`/`get_healthcheck`
[...]

PR #2789, opened 2026-02-28
fix: add ExperimentalFeatureWarning compat shim

tempusfrangit:
## Summary
Existing models import `ExperimentalFeatureWarning` from `cog` to suppress experimental feature warnings. This was removed when the Python HTTP server was replaced by coglet (#2783), breaking those models at import time with `ImportError`.
## What Changed
Added a backwards-compatible shim in `python/cog/__init__.py` that:
- Provides `ExperimentalFeatureWarning` via module `__getattr__` (lazy — only triggers when actually imported)
- Prints a deprecation notice **directly to stderr** — cannot be swallowed by `warnings.filterwarnings("ignore")` or any other warnings suppression
- The class is a real `FutureWarning` subclass, so existing `warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)` calls work syntactically
- Will be removed in a future release
## Models affected
Any model that does:
```python
from cog import ExperimentalFeatureWarning
warnings.filterwarnings("ignore", category=ExperimentalFeatureWarning)
```
Including `cog-examples/hello-concurrency`, `cog-examples/hello-context`, `cog-examples/hello-replicate`.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -68,9 +68,14 @@ def _install_coglet(session: nox.Session) -> None:
 def _install_package(session: nox.Session) -> None:
     """Install the cog SDK and coglet dependency."""
     _install_coglet(session)
-    # Always use editable install for the SDK so tests run against
-    # the working tree, not a stale wheel from a previous build.
-    session.install("-e", ".")
+    whl = _find_compatible_wheel("dist/cog-*.whl")
+    if whl:
+        session.install(whl)
+    else:
+        # No pre-built wheel — editable install from source.
+        # This fails in CI (setuptools_scm needs a full git checkout),
+        # so CI must run build:sdk first.
+        session.install("-e", ".")
 
 
 @nox.session(python=PYTHON_VERSIONS)


python/cog/__init__.py
```diff
@@ -36,6 +38,43 @@ def predict(
...
+    if name == "ExperimentalFeatureWarning":
+        print(
+            "cog: ExperimentalFeatureWarning is deprecated and will be removed in a "
+            "future release. Remove `ExperimentalFeatureWarning` from your imports "
+            "and any associated `warnings.filterwarnings(...)` calls.",
+            file=_sys.stderr,
+        )
+        # Cache in module namespace so __getattr__ is not called again and
+        # the deprecation message prints at most once.
+        globals()["ExperimentalFeatureWarning"] = _ExperimentalFeatureWarning
```
markphelps, line 73: is there a way to test this with a unit tests and txtar ?

tempusfrangit: Done

python/cog/__init__.py
```diff
@@ -19,6 +19,8 @@ def predict(
             return self.model.generate(prompt, image)
 """
 
+import sys as _sys
```
meatballhat-cf, line 22: Why the `_sys` alias? To work around libraries that runtime patch `stderr`?

python/cog/__init__.py
```diff
@@ -78,4 +117,6 @@ def current_scope() -> object:
     "CancelationException",
     # Metrics
     "current_scope",
+    # Deprecated compat shims
+    "ExperimentalFeatureWarning",
```
meatballhat-cf, line 121: I think `ExperimentalFeatureWarning` _should not_ be in the `__all__` since it's not supposed to be part of the public API. I found [maybe one legit case of `from cog import *`](https://github.com/search?q=%22from+cog+import+*%22&type=code) in GitHub, so I think we can safely say the backward compatibility only applies to named access.

PR #2800, opened 2026-03-03
feat: opt-in static schema gen with legacy runtime fallback

markphelps:
## Summary

- Gate static schema generation behind `COG_STATIC_SCHEMA=1` env var for `cog build` (legacy runtime path remains default)
- Static gen always enabled for `cog train`/`cog serve`/`cog predict` (skipLabels paths) since CLI needs schema for `-i` flag parsing
- Automatic fallback to legacy runtime path when static gen fails with `ErrUnresolvableType`
- Restore Python SDK modules (`_adt.py`, `_inspector.py`, `_schemas.py`, `command/openapi_schema.py`) required by legacy fallback
- Add pydantic v2 `BaseModel` support in legacy `_inspector.py`
- Update architecture docs for dual-path model

### DIFF
--- a/pkg/image/build.go
+++ b/pkg/image/build.go
@@ -97,21 +97,23 @@ func Build(
 	case useStatic:
 		console.Debug("Generating model schema (static)...")
 		data, err := generateStaticSchema(cfg, dir)
-		if err != nil {
-			var se *schema.SchemaError
-			if !skipLabels && errors.As(err, &se) && se.Kind == schema.ErrUnresolvableType {
-				// For `cog build` only: fall back to the post-build legacy runtime
-				// schema generation which can handle types that require Python import
-				// (e.g. package __init__.py modules, pydantic v2 BaseModel subclasses).
-				console.Warnf("Static schema generation failed: %s", err)
-				console.Warn("Falling back to legacy runtime schema generation...")
-				// leave schemaJSON nil — the post-build legacy path will handle it
-			} else {
-				return "", fmt.Errorf("image build failed: %w", err)
-			}
-		} else {
+		if err == nil {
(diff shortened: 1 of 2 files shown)


pkg/image/build.go
```diff
@@ -73,28 +73,45 @@ func Build(
...
+				// (e.g. package __init__.py modules, pydantic v2 BaseModel subclasses).
+				console.Warnf("Static schema generation failed: %s", err)
+				console.Warn("Falling back to legacy runtime schema generation...")
+				// leave schemaJSON nil — the post-build legacy path will handle it
+			} else {
+				return "", fmt.Errorf("image build failed: %w", err)
+			}
+		} else {
+			schemaJSON = data
 		}
```
markphelps, line 103: we should refactor these nested if/else into early returns

pkg/image/build.go
```diff
@@ -399,11 +416,22 @@ func BuildAddLabelsAndSchemaToImage(ctx context.Context, dockerClient command.Co
...
-// Returns false (use legacy) when the SDK version is explicitly pinned < 0.17.0.
+// Note: `cog train` and `cog serve` always use static schema generation
+// regardless of this flag because they return before the post-build legacy
+// path — see the useStatic logic in Build().
+//
+// Even when opted in, returns false when the SDK version is explicitly
+// pinned < 0.17.0, since older SDKs use pydantic-based schemas that the
+// static parser cannot analyze.
 func canUseStaticSchemaGen(cfg *config.Config) bool {
+	env := os.Getenv("COG_STATIC_SCHEMA")
```
markphelps: strings.ToLower

PR #2804, opened 2026-03-04
feat: spill large inputs to disk to prevent IPC frame overflow

michaeldwan:
## Summary

- When an input payload exceeds 6 MiB, the parent spills it to a file under `/tmp/coglet/predictions/{id}/inputs/` and sends only the file path over the IPC slot socket. The worker reads the file back, parses it, and deletes it before running the prediction.
- This mirrors the existing output-spill mechanism and prevents the 8 MiB `LengthDelimitedCodec` frame limit from being exceeded by large inputs.
- Unified `MAX_INLINE_IPC_SIZE` constant (previously separate input/output constants that could drift).

### Rust (`crates/coglet/`)
- **`protocol.rs`**: `SlotRequest::Predict` now has optional `input`/`input_file` fields with `skip_serializing_if`. Added `prediction_id()` accessor and `rehydrate_input()` method that reads from inline value or spill file. Spill file is cleaned up before JSON parsing (so corrupt files don't leak). 4 unit tests + snapshot.
- **`service.rs`**: New `build_slot_request()` serializes input, checks against threshold, spills to disk if needed. Prediction dir layout changed from `/tmp/coglet/outputs/{id}` to `/tmp/coglet/predictions/{id}/{inputs,outputs}`. 3 unit tests.
- **`worker.rs`**: Uses `rehydrate_input()` with proper failure handling — sends `SlotResponse::Failed` back to parent instead of silently dropping the prediction. Removed duplicate `MAX_INLINE_OUTPUT_SIZE` in favor of shared `MAX_INLINE_IPC_SIZE`.
- **`codec.rs`**: Updated test fixture for new field layout.
- **`Cargo.toml`**: Added `tempfile` to dev-dependencies.

### Integration tests (`integration-tests/`)
- **`harness.go`**: `cmdCurl` now supports `@file` syntax to POST large bodies from files.
- **`coglet_large_input.txtar`**: New test — generates 7 MiB JSON payload, POSTs via `@file`, verifies prediction succeeds via webhook.

## Test plan

- `mise run test:rust` — 165/165 pass
- `mise run fmt:rust:fix` — clean
- `mise run lint:rust` — clean (clippy)
- `coglet_large_input` integration test — PASS
- `coglet_large_output` integration test — PASS (regression guard for dir layout change)
- `string_predictor` integration test — PASS (inline input path still works)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -328,7 +328,7 @@ jobs:
         id: version
         run: echo "version=$(grep '^version' crates/Cargo.toml | head -1 | sed 's/.*"\(.*\)"/\1/')" >> "$GITHUB_OUTPUT"
       - name: Build cog binary
-        uses: goreleaser/goreleaser-action@v6
+        uses: goreleaser/goreleaser-action@v7
         with:
           version: '~> v2'
           args: build --clean --snapshot --single-target --id cog --output cog
@@ -905,7 +905,7 @@ jobs:
       - uses: mlugg/setup-zig@v2
         with:
           version: 0.15.2
-      - uses: goreleaser/goreleaser-action@v6
+      - uses: goreleaser/goreleaser-action@v7
         with:
           version: '~> v2'
           args: check
--- a/.github/workflows/release-build.yaml
+++ b/.github/workflows/release-build.yaml
@@ -230,7 +230,7 @@ jobs:
 
       # Goreleaser builds CLI binaries and creates draft release
       - name: Build CLI and create draft release
-        uses: goreleaser/goreleaser-action@v6
+        uses: goreleaser/goreleaser-action@v7
         with:
           version: '~> v2'
           args: release --clean
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -186,6 +186,8 @@ COG_BINARY=dist/go/*/cog mise run test:integration
 7. Run `mise run docs:llm` to regenerate `docs/llms.txt` after changing `README.md` or any `docs/*.md` file
 8. Read the `./docs` directory and make sure the documentation is up to date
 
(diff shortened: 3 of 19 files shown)


integration-tests/harness/harness.go
```diff
@@ -491,7 +491,17 @@ func (h *Harness) cmdCurl(ts *testscript.TestScript, neg bool, args []string) {
...
+		if strings.HasPrefix(raw, "@") {
+			filename := raw[1:]
+			data, err := os.ReadFile(ts.MkAbs(filename))
+			if err != nil {
+				ts.Fatalf("curl: failed to read body file %q: %v", filename, err)
+			}
+			body = string(data)
+		} else {
+			body = raw
+		}
```
markphelps:
```suggestion
		body := os.Expand(args[2], ts.Getenv)
		if strings.HasPrefix(body, "@") {
			filename := body[1:]
			data, err := os.ReadFile(ts.MkAbs(filename))
			if err != nil {
				ts.Fatalf("curl: failed to read body file %q: %v", filename, err)
			}
			body = string(data)
		} 
```

or something

integration-tests/harness/harness.go
```diff
@@ -491,17 +491,15 @@ func (h *Harness) cmdCurl(ts *testscript.TestScript, neg bool, args []string) {
 	path := args[1]
 	var body string
 	if len(args) > 2 {
-		raw := os.Expand(args[2], ts.Getenv)
-		if strings.HasPrefix(raw, "@") {
-			filename := raw[1:]
+		body := os.Expand(args[2], ts.Getenv)
```
markphelps:
```suggestion
		body = os.Expand(args[2], ts.Getenv)
```

my bad

PR #2805, opened 2026-03-04
fix: address review issues in static schema generation

markphelps:
## Summary

Addresses remaining issues from the`fix/dict-list-output-schema` branch.

### Bug fixes
- **Cycle detection for recursive BaseModels** — `ResolveSchemaType` now tracks a `seen` set to prevent infinite recursion on self-referential models like `class TreeNode(BaseModel): child: Optional["TreeNode"]`. Cycles emit an opaque `{"type": "object"}` instead of stack overflowing.
- **Forward reference handling in model fields** — `parseTypeFromString` now strips quotes from forward references like `"MyType"` or `'MyType'` inside generic annotations (e.g. `Optional["TreeNode"]`). Previously these fields were silently dropped.
- **Deterministic JSON property order** — `SchemaObject.coreSchema()` now uses `orderedMapAny` instead of `map[string]any` for the `properties` field, ensuring consistent JSON output.

### Code cleanup
- **Explicit panic for unknown `SchemaTypeKind`** — replaced silent fallthrough returning `{"type": "object"}` with a panic that surfaces missing switch cases immediately.
- **Removed redundant nullable** — `SchemaObject` field loop no longer re-sets `nullable` since `jsonSchema()` already handles it.
- **Removed unused `KeyType` field** — dict key types are always strings in JSON Schema; the field was set but never read.

### Test coverage (8 new tests)
- `TestForwardReferenceOutput` / `TestForwardReferenceGeneric` — quoted string annotations
- `TestNestedBaseModelField` — `class Outer(BaseModel): inner: Inner` composition
- `TestListOfBaseModelOutput` / `TestIteratorOfBaseModelOutput` — BaseModel in generic outputs
- `TestRecursiveBaseModelDoesNotStackOverflow` — self-referential model cycle detection
- `TestOutputObjectPropertyOrderDeterministic` — JSON property ordering

### Not addressed (documented, behind opt-in gate)
- Cross-file depth-1 resolution limit
- No `__init__.py` package import support
- Fuzz tests don't exercise BaseModel/cross-file paths

### DIFF
--- a/pkg/schema/schema_type.go
+++ b/pkg/schema/schema_type.go
@@ -52,6 +52,28 @@ const (
 	SchemaConcatIterator
 )
 
+// String returns the name of the SchemaTypeKind for diagnostic messages.
+func (k SchemaTypeKind) String() string {
+	switch k {
+	case SchemaPrimitive:
+		return "SchemaPrimitive"
+	case SchemaAny:
+		return "SchemaAny"
+	case SchemaArray:
+		return "SchemaArray"
+	case SchemaDict:
+		return "SchemaDict"
+	case SchemaObject:
+		return "SchemaObject"
+	case SchemaIterator:
+		return "SchemaIterator"
+	case SchemaConcatIterator:
+		return "SchemaConcatIterator"
+	default:
+		return fmt.Sprintf("SchemaTypeKind(%d)", int(k))
+	}
+}
+
 // SchemaField is a named field within a SchemaObject.
 type SchemaField struct {
 	Type     SchemaType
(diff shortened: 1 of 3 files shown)


pkg/schema/schema_type.go
```diff
@@ -155,9 +150,13 @@ func (s SchemaType) coreSchema() map[string]any {
 			"x-cog-array-type":    "iterator",
 			"x-cog-array-display": "concatenate",
 		}
+	default:
+		// All SchemaTypeKind values must be handled above. If this is reached,
+		// it indicates a missing case (e.g. a new kind was added without updating
+		// this switch). Panic to surface the bug immediately rather than silently
+		// returning a wrong schema.
+		panic(fmt.Sprintf("unhandled SchemaTypeKind: %d", s.Kind))
```
markphelps: controversial

PR #2847, opened 2026-03-17
fix: harden integration timing and SIGTERM assertions

markphelps:
## Summary
- replace fixed sleep in `webhook_delivery_failure` with health-check polling to reduce timing flake
- tighten `async_generator_precollect` to assert `predict_time` reflects full generation duration
- strengthen `TestSIGTERMDuringSetup` to assert termination is specifically caused by `SIGTERM`

## Test Plan
- [x] Attempted: `go test -tags integration ./integration-tests/concurrent -run TestSIGTERMDuringSetup -count=1`
- [ ] Blocked locally by environment permissions while building wheels (`python/cog.egg-info` and `crates/target/release/.cargo-lock` permission denied)

### DIFF
--- a/integration-tests/concurrent/concurrent_test.go
+++ b/integration-tests/concurrent/concurrent_test.go
@@ -4,6 +4,7 @@ package concurrent_test
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


integration-tests/concurrent/concurrent_test.go
```diff
@@ -501,7 +502,22 @@ class Predictor(BasePredictor):
 
 	select {
 	case err := <-done:
-		t.Logf("Server exited: %v", err)
+		if err == nil {
+			t.Fatal("server exited cleanly after SIGTERM; expected termination by signal")
```
michaeldwan, line 506: testify assert/require can cut down a lot of this code

markphelps: #2848

PR #2925, opened 2026-04-09
Rename predict to run

markphelps:
## Summary

Renames the user-facing "predict" concept to "run" across CLI, Python SDK, cog.yaml, Rust coglet, Go schema parser, docs, and tests. Full backwards compatibility -- old names keep working with deprecation warnings.

**Scope: user-facing changes only.** Internal Go packages (`pkg/predict/`), Rust types, `COG_PREDICT_TYPE_STUB`, `predict_time` metrics, `/predictions` HTTP endpoints, and IPC wire format are all unchanged.

### What changed

- **CLI**: `cog run` is the new primary command for predictions. `cog predict` is hidden + deprecated. `cog run <cmd> [args...]` (2+ args) forwards to `cog exec` with deprecation warning.
- **cog.yaml**: `run:` key supported alongside `predict:`. Both cannot be set simultaneously. `run:` resolves into `Config.Predict` at parse time.
- **Python SDK**: `BaseRunner` is the single real class. `BasePredictor = BaseRunner` is a backwards-compatible alias. `run()` is the primary method; `predict()` is a bridge that calls `run()`. No duplicated logic between the two names.
- **Rust coglet**: Detects `run()` vs `predict()` at predictor load time via class `__dict__` MRO walk. Stored method name used for all subsequent calls.
- **Go schema parser**: Tree-sitter parser tries `run()` first, falls back to `predict()`.
- **`cog init`**: Generates `run.py` with `BaseRunner`/`Runner`/`def run()`.
- **Docs**: All examples updated. Backwards-compat notes in `yaml.md` and `python.md`.
- **Integration tests**: ~110 txtar files renamed + 5 new backwards-compat tests.

### Backwards compatibility

| Old usage | Still works? | Behavior |
|-----------|-------------|----------|
| `predict:` in cog.yaml | Yes | Resolves to same internal field |
| `BasePredictor` / `def predict()` | Yes | `BasePredictor` is an alias for `BaseRunner`; inspector detects `predict()` via MRO |
| `cog predict` | Yes (hidden) | Prints deprecation warning |
| `cog run <shell-command>` (2+ args) | Yes | Forwards to `cog exec` with warning |


michaeldwan:
Is there a benefit to having a full implementation of BasePredictor and BaseRunner with fallback logic? I _think_ we can do this by renaming BasePredictor to BaseRunner, then making an alias to BasePredictor. And we can rename more of the internals in isolation later. 

Edit: We'd also need to add an override for both run and predict methods. eg run is abstract and raises NotImplemented when called, predict calls run. I think that covers the bases and won't require duplicating and maintaining a ton of logic across 2 classes.

PR #2928, opened 2026-04-10
fix: support dict and list[dict] as predictor input types

markphelps:
dict, Dict[str, Any], list[dict], and List[Dict[str, Any]] were not supported as predictor input types, even though they worked as output types. This broke models using chat-style messages: list[dict] inputs.

The fix maps dict types to ANY (opaque JSON objects) in both the Go tree-sitter schema generator (ResolveFieldType) and the Python SDK runtime (FieldType.from_type), handling all variants: bare dict, parameterized Dict[K,V], list[dict], Optional[dict], and PEP 604 union syntax.


michaeldwan (approved):
nice

PR #2944, opened 2026-04-15
feat(test-harness): add parallel execution, better error reporting, and lint fixes

markphelps:
## Summary

- Add `--concurrency` flag (default 4) to run model builds/tests in parallel across `run`, `build`, and `schema-compare` commands
- Improve error reporting so the final error message includes per-model failure details instead of just listing names
- Harden setup command execution with `set -euo pipefail` and empty `cog.yaml` detection to catch missing tools (e.g. `yq`) immediately
- Fix all 30 golangci-lint issues across the test-harness codebase

## Parallel execution

Models are dispatched via `errgroup` with `SetLimit(concurrency)`. Each model gets its own isolated work directory (copied from the shared clone) so setup commands like `script/select.sh` don't clobber each other. In parallel mode, build output is captured rather than streamed to avoid interleaving. `--concurrency 1` preserves the original sequential behavior with real-time streaming.

## Error reporting

Before:
```
Error: 15 model(s) failed: flux-schnell, flux-dev-lora, flux-fill-dev, ...
```

After:
```
Error: 15 model(s) failed:
  x flux-schnell: Preparation failed: setup command "script/select.sh schnell" failed: exit status 127
  x flux-dev-lora: Build failed: signal: killed
  x chatterbox:
      test "text-to-speech synthesis": cog predict exited with error: exit status 1
```

## Setup hardening

Setup commands now run with `set -euo pipefail` so a missing `yq` (or any other tool) fails fast instead of silently producing an empty `cog.yaml` that later fails with a confusing `python_version is required` error.

## Lint fixes

Resolved all 30 golangci-lint issues: copyloopvar, errcheck, gocritic (octal literals, if-else chains), gosec (SSRF annotations), misspell, modernize (SplitSeq, maps.Copy, built-in min).


michaeldwan (dismissed):
looks good! might be worth noting that the skip schema validation flag (which can maybe be hidden?) should be removed once we rip out the old in-container schema generator.

PR #2949, opened 2026-04-17
fix: treat `Secret = Input(default=None)` as optional in static schema gen

markphelps:
`api_key: Secret = Input(default=None)` is a widespread, documented idiom for an optional credential that falls back to a proxy key when the caller omits it. The static schema generator was applying the "DefaultNone on a non-nullable field is not a real default" rule (originally intended for `seed: int = Input(default=None)`) uniformly, which put the Secret in `required` and dropped both `nullable: true` and the null default. The Replicate API validator then 422'd every request that didn't supply the supposedly-optional key.

Exempt `TypeSecret` with `Required` repetition and explicit `DefaultNone` from that rule. The field now emits `nullable: true` and `default: null` and is dropped from `required` — matching the `Optional[Secret] = Input(default=None)` spelling and 0.16.x behavior. `Secret` without a `default=` kwarg stays required, and `seed: int = Input(default=None)` is unchanged.

### DIFF
--- a/integration-tests/tests/static_schema_secret_optional.txtar
+++ b/integration-tests/tests/static_schema_secret_optional.txtar
@@ -1,0 +1,70 @@
+# Test that `Secret = Input(default=None)` produces an OPTIONAL/NULLABLE
+# field in the statically generated OpenAPI schema.
+#
+# This is a common and documented idiom: a BYOK-style credential that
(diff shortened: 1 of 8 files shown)


michaeldwan (approved):
nice fix

PR #2950, opened 2026-04-17
feat: make static schema generation the default, add COG_LEGACY_SCHEMA opt-out

markphelps:
## Summary

- Flip the default schema generation path from runtime to **static** (Go tree-sitter, runs before Docker build)
- Add `COG_LEGACY_SCHEMA=1` as an opt-out for users who need the legacy runtime path (pinned to SDK < 0.17.0, hitting a static-parser edge case, etc.)
- The older `COG_STATIC_SCHEMA=1` flag from 0.18.0 is now a no-op (static is already on) but still accepted so existing CI/Dockerfiles don't break
- Un-skip `cog train` integration tests that were gated behind the opt-in flag
- Add an end-to-end test for the opt-out path

## Why

Static schema generation is faster, deterministic, and doesn't require booting a container to introspect Python. It's been in the tree and stable enough that it should be the default. Keeping the runtime path around as the automatic fallback on `ErrUnresolvableType` (and as an explicit opt-out) means users don't lose any capability — only a boot cycle per build.

This also fixes a UX footgun: `cog predict`/`cog train`/`cog serve` skip the post-build runtime generation step, so under the old defaults they produced images without a bundled schema and `-i` flag parsing broke. Two integration tests (`train_basic`, `training_setup`) were skipped with a comment pointing at `COG_STATIC_SCHEMA=1`. Now they pass on the default path.

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| `cog build` (default) | runtime path | **static path**, falls back to runtime on `ErrUnresolvableType` |
| `cog build COG_STATIC_SCHEMA=1` | static path | static path (flag is a no-op) |
| `cog build COG_LEGACY_SCHEMA=1` | n/a | legacy runtime path |
| `cog build` with SDK < 0.17.0 | runtime path | legacy runtime path (SDK-version gating, unchanged) |
| `cog predict`/`serve`/`train` (default) | no schema bundled | static path bundles schema — `-i` parsing works |
| `cog predict`/etc with `COG_LEGACY_SCHEMA=1` | n/a | no schema bundled (legacy only runs during `cog build`) |

Static-error handling is unchanged: only `ErrUnresolvableType` falls back; parse errors, unsupported features (like `default_factory`), and other hard user errors still fail fast with a clear static-generator error.

## Test harness

`test-harness schema-compare` still builds each model twice to diff the two generators. The two variants swap semantics: the static build uses the new default (no env var), the runtime build sets `COG_LEGACY_SCHEMA=1`.

## Changes

- `pkg/image/build.go` — renamed `canUseStaticSchemaGen` → `useStaticSchemaGen`, inverted the check, added `legacySchemaEnvVar` const and `isTruthyEnv` helper, rewrote the big decision-block comment
- `pkg/image/build_test.go` — renamed to `TestUseStaticSchemaGen`, 16 subcases covering the default-on behavior, the opt-out, SDK-version gating, and `COG_STATIC_SCHEMA` back-compat (including the precedence case where `COG_LEGACY_SCHEMA=1` wins over `COG_STATIC_SCHEMA=1`)
- `integration-tests/tests/static_schema_gen.txtar`, `multi_file_schema.txtar`, `static_schema_fallback.txtar` — drop the explicit `env COG_STATIC_SCHEMA=1` lines
- `integration-tests/tests/legacy_sdk_schema.txtar` — comment updated to reflect that the SDK-version gate handles this case regardless of env vars
- `integration-tests/tests/train_basic.txtar`, `training_setup.txtar` — un-skipped
- `integration-tests/tests/legacy_schema_optout.txtar` (new) — asserts `COG_LEGACY_SCHEMA=1` routes through the "Validating model schema" boot-container phase and still bundles a correct schema
- `tools/test-harness/internal/runner/runner.go` + `cmd/schema_compare.go` — swap env-var semantics in the schema-compare runner
- `architecture/02-schema.md` — re-ordered the two paths, flipped the "default" labeling, updated the env-var table with `COG_LEGACY_SCHEMA` + the `COG_STATIC_SCHEMA` deprecation note
- `skills/updating-architecture-docs/SKILL.md` — updated the "static schema path is experimental" bullet to "static schema path is the default"; tightened the pydantic bullet to note that `inspect` + ADT only runs on the legacy fallback

## Verification

- `mise run test:go` — 1074 tests pass, 4 skipped (Docker not running locally)
- `mise run lint:go` — 0 issues
- `mise run docs:llm:check` — `docs/llms.txt` is current (`architecture/` isn't part of the llms.txt bundle)
- Integration tests have not been run locally yet; CI will cover them

## Follow-ups

Not in this PR:
- `architecture/02-schema.md` notes that `cog predict`/`train` with `COG_LEGACY_SCHEMA=1` produces images without a bundled schema (legacy only runs during `cog build`). If that becomes a user-visible issue we could either emit a warning when `COG_LEGACY_SCHEMA=1` + `skipLabels=true` are combined, or have coglet cold-generate a schema on first request.
- The runtime path still has the `Secret = Input(default=None)` bug fixed on [mphelps/secret-optional-schema-fix](https://github.com/replicate/cog/tree/mphelps/secret-optional-schema-fix) for the static path only. With static as the default, the runtime-path bug is far less exposed, but it's still there if someone sets `COG_LEGACY_SCHEMA=1`.

### DIFF
--- a/architecture/02-schema.md
+++ b/architecture/02-schema.md
@@ -34,23 +34,25 @@ Without the schema, consumers would have no way to know:
 
 Cog supports two schema generation paths:
 
-### Runtime Path (default)
+### Static Path (default)
 
-The **runtime path** boots the built Docker container and runs `python -m cog.command.openapi_schema` to introspect the model at runtime. This is the default for all builds. It uses Python's `inspect` module and `typing.get_type_hints()` to extract parameter types, then builds OpenAPI JSON from a hand-rolled ADT type system (dataclasses in `_adt.py`). No pydantic is involved in schema generation -- `cog.BaseModel` is a dataclass wrapper, not pydantic.
+The **static path** parses Python source code at `cog build` time using [tree-sitter](https://tree-sitter.github.io/tree-sitter/) in Go. No Python runtime is invoked, no container boots — the schema is produced from the model's source files before Docker build even begins. This makes schema generation deterministic, fast, and independent of the model's runtime dependencies. Static generation is the default for all commands and requires SDK >= 0.17.0.
 
-### Static Path (experimental)
+If the static parser encounters a type it can't resolve (for example, a `BaseModel` subclass exported from a package `__init__.py` that the file-based resolver can't find), it **falls back to the runtime path** automatically with a warning — so builds don't break due to static-parser limitations. Hard user errors (parse errors, unsupported features like `default_factory`) still fail fast with a clear static-generator error rather than being swept into the fallback.
 
-The **static path** parses Python source code at `cog build` time using [tree-sitter](https://tree-sitter.github.io/tree-sitter/) in Go. No Python runtime is invoked. The goal is to replace the runtime path entirely -- making schema generation deterministic, fast, and independent of the model's dependencies. It's not there yet.
+### Runtime Path (legacy fallback)
 
-Enable it for testing with the `COG_STATIC_SCHEMA` environment variable:
+The **runtime path** boots the built Docker container and runs `python -m cog.command.openapi_schema` to introspect the model at runtime. It uses Python's `inspect` module and `typing.get_type_hints()` to extract parameter types, then builds OpenAPI JSON from a hand-rolled ADT type system (dataclasses in `_adt.py`). No pydantic is involved in schema generation -- `cog.BaseModel` is a dataclass wrapper, not pydantic.
+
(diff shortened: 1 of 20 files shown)


markphelps:
Rebased onto `mphelps/secret-optional-schema-fix` (which itself sits on `mphelps/fix-static-schema-gen`) so the static path picks up the "default None / casing" fixes and the Secret-optional fix before they land on main. Also changed this PR's base to `mphelps/secret-optional-schema-fix` so the diff shows only the static-default commit.

When the dependency branch merges, GitHub will auto-retarget this PR's base to `main`.

michaeldwan (approved):

PR #2951, opened 2026-04-21
ci: use mise-action as single tool installer across all workflows

markphelps:
## Summary

- Replace ~10 different GitHub Actions with `jdx/mise-action@v4` so tool versions are defined in **one place** (`mise.toml`) instead of two
- Remove the `MISE_DISABLE_TOOLS` env var entirely — mise manages all tools including rustup (needed for rustfmt/clippy components)
- Enable mise's built-in tool caching (`cache: true`, the default) with `cache_key_prefix: mise-v2`
- Update `AGENTS.md` and `.github/workflows/README.md` to reflect the new single-source-of-truth approach

## Actions removed

| Action | Jobs it was in |
|--------|---------------|
| `dtolnay/rust-toolchain@stable` | ~10 jobs |
| `taiki-e/install-action@cargo-binstall` | ~8 jobs |
| `astral-sh/setup-uv@v6` | ~7 jobs |
| `actions/setup-go@v6` | ~5 jobs |
| `golangci/golangci-lint-action@v9` | lint-go |
| `mlugg/setup-zig@v2` | build-cog, release-dry-run |
| `taiki-e/install-action` (cargo-deny, cargo-nextest) | test-rust, lint-rust-deny |
| `go install gotestsum` | test-go, test-integration |
| `actions/setup-python` | docs.yaml |

## Kept exceptions

| Tool | Why |
|------|-----|
| `PyO3/maturin-action@v1` | Builds inside a manylinux container with bundled maturin + zig |
| `Swatinem/rust-cache@v2` | Caches `crates/target/` compilation artifacts (not tools) |
| `actions/setup-python@v6` in check-stubs | Needs shared-library Python (libpython3.x.so) for PyO3 auto-initialize |

## Net result

- **-170 lines, +75 lines** across 7 files
- Tool version updates now only need to touch `mise.toml`
- `cache_key_prefix: mise-v2` busts stale caches from before this PR

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -97,16 +97,15 @@ Runs CodeQL security scanning for Go, Python, and Rust.
 
 ## Caching Strategy
 
-### Rust Cache
+### Tool Cache (mise)
+- `jdx/mise-action@v4` with `cache: true` (default) caches all installed tools
+- Tool versions are defined in `mise.toml` — CI and local dev use the same versions
+- Cache key includes platform, mise version, and config file hashes
+
+### Rust Target Cache
 - **Save**: Only on `main` branch pushes (to avoid PR cache pollution)
 - **Restore**: On all runs (PRs restore from main's cache)
 - Uses `Swatinem/rust-cache@v2` with workspace path `crates -> target`
-
-### Go Cache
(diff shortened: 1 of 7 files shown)


michaeldwan (approved):
nice!

PR #2974, opened 2026-04-28
feat: experimental managed weights

michaeldwan:
Adds a managed weights pipeline that stores model weights as separate OCI artifacts alongside the container image.

Four commands drive the pipeline:

1. **`cog weights import`** -- inventories weight sources (local dirs or HuggingFace repos), content-addresses each file, packs them into OCI layers, pushes weight manifests to the registry, and generates `weights.lock`.
2. **`cog weights pull`** -- pulls weight layers from the registry into a local content-addressed store.
3. **`cog push`** -- builds the model image, HEAD-checks that weight manifests exist, and assembles an OCI Image Index pointing to both image and weight manifests.
4. **`cog predict` / `cog train`** -- bind-mounts weights from the local store into the container at the configured target paths.

Weight sources support `file://` (local directories) and `hf://` (HuggingFace repos). Each weight entry in `cog.yaml` gets a name, target mount path, and source URI. The lockfile tracks per-file digests and a set-digest fingerprint -- build, push, predict, and train all fail fast if `weights.lock` is stale.

The OCI index uses `application/vnd.cog.weight.v1` for weight manifests, with annotations per the draft spec in `specs/draft-weights.md`.

Also in this PR: `cog inspect` removed -- replaced by `cog weights status`.


pkg/cli/predict.go
```diff
@@ -248,12 +263,17 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
-		Env:     env,
-	}, false, dockerClient)
+	predictor, err := predict.NewPredictor(ctx, predict.PredictorOptions{
+		RunOptions: command.RunOptions{
+			GPUs:    gpus,
+			Image:   imageName,
+			Volumes: volumes,
+			Env:     env,
+		},
+		IsTrain:       false,
```
markphelps, line 273: 🚋

pkg/model/pusher.go
```diff
@@ -91,22 +69,28 @@ func (p *BundlePusher) Push(ctx context.Context, m *Model, opts PushOptions) err
...
+		var descErr error
+		imgDesc, descErr = p.registry.GetDescriptor(gctx, imgArtifact.Reference)
+		if descErr != nil {
+			return fmt.Errorf("get image descriptor: %w", descErr)
 		}
+		return nil
+	})
+	g.Go(func() error {
+		var verifyErr error
+		weightDescs, verifyErr = p.verifyWeights(gctx, repo, m.Weights)
```
markphelps: should we move this verification above/before pushing the image ?

pkg/model/weight_builder.go
```diff
@@ -5,161 +5,365 @@ import (
...
+	out := make([]packedLayer, 0, len(entry.Layers))
+	for _, lk := range entry.Layers {
+		// Find the plan layer whose files match this locked layer.
+		// We compare by file digests (the file→layer mapping in
+		// entry.Files tells us which files belong to this layer).
+		want := lockedLayerKey(entry, lk.Digest)
+		lp, ok := planByKey[want]
+		if !ok {
+			return nil, fmt.Errorf("locked layer %s has no matching plan layer", lk.Digest)
 		}
```
markphelps, line 302: I think there is a correctness issue in the lockfile fast path here.
`lockedLayerKey` sorts by `Path`, but the returned key only contains file digests. `layerKey` also only contains file digests. That means paths are used only for ordering, not identity.

For example, if a weight has two separate single-file layers with identical contents:
```text
a.bin -> sha256:same
b.bin -> sha256:same
```
both layers produce the same key:
```text
sha256:same
```
Even though the file content digest is the same, the OCI layer bytes are not necessarily the same because the tar header includes the file path (`a.bin` vs `b.bin`). In `layersFromLockfile`, `planByKey[layerKey(lp)] = lp` can therefore overwrite one plan with the other, and the lockfile layer can be paired with the wrong `layerPlan`.

That can make the artifact advertise the locked digest/size for one layer while streaming tar bytes for a different path during push, which would cause digest mismatches or an invalid artifact.
I think the key needs to include both path and digest, e.g. `path + "\x00" + digest`, in both `layerKey` and `lockedLayerKey`. It would also be worth adding a regression test with two files that have identical contents but different paths.

pkg/cli/weights_status.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Format output.
+	entries := statusResultsToEntries(ws.Results())
+	out := &WeightsStatusOutput{Weights: entries}
+
+	if jsonOutput {
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", "  ")
+		return enc.Encode(out)
+	}
```
markphelps, line 139: `cog weights status --json` exits successfully after writing JSON even when weights are not ready. Text mode returns `errWeightsNotReady`, so JSON mode behaves inconsistently and may break CI readiness checks.

pkg/model/weightsource/source.go
```diff
@@ -0,0 +1,144 @@
...
+//
+// Each scheme has its own normalization rules:
+//   - file:// and bare paths → canonical file:// form (see normalizeFileURI)
+//   - hf:// and huggingface:// → canonical hf:// form (see normalizeHFURI)
+//
+// Empty strings and unsupported schemes return an error.
+func NormalizeURI(uri string) (string, error) {
+	if uri == "" {
+		return "", fmt.Errorf("empty weight source uri")
+	}
```
markphelps, line 86: The schema/validation allow weights.source to be omitted, but this returns an error if `source.uri` is empty. We should either require `source.uri` or make it truly optional

pkg/model/weightsource/huggingface.go
```diff
@@ -0,0 +1,470 @@
...
+// both).
+type hfLFSInfo struct {
+	OID  string `json:"oid"` // sha256 of the full file
+	Size int64  `json:"size"`
+}
+
+// listTree fetches the recursive tree listing at the given commit sha.
+// NOTE: the HF Hub API paginates large repos. This implementation does
+// not follow pagination yet — repos with very many files may return an
+// incomplete listing. A follow-up should add cursor-based pagination.
```
markphelps: should we go ahead and implement pagination?

pkg/weights/check_drift.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		cw := lockfile.ConfigWeight{
+			Name:   w.Name,
+			Target: w.Target,
+			URI:    uri,
+		}
+		if w.Source != nil {
+			cw.Include = sortedCopy(w.Source.Include)
+			cw.Exclude = sortedCopy(w.Source.Exclude)
+		}
```
markphelps: Drift checking sorts include/exclude patterns but does not trim whitespace, while import normalization trims them. Whitespace-padded patterns can cause persistent false drift even after re-importing.

markphelps (changes requested):

pkg/model/weightsource/huggingface.go
```diff
@@ -0,0 +1,519 @@
...
+}
+
+// newHFHTTPClient returns a standard *http.Client whose transport
+// retries on 5xx, 429, and network errors with exponential backoff.
+// The retryable behavior is provided by go-retryablehttp configured as
+// a transport — callers use the standard http.Client API.
+func newHFHTTPClient() *http.Client {
+	rc := retryablehttp.NewClient()
+	rc.RetryMax = 3
+	rc.Logger = nil // Silence default logger; errors surface via return values.
```
markphelps, line 95: im assuming this is safe and it checks logger before calling it so it doesnt panic?

pkg/model/weightsource/huggingface.go
```diff
@@ -0,0 +1,519 @@
...
+// close the body. Most callers only need the body; callers that need
+// response headers (e.g. pagination via Link) use the response directly.
+//
+// Non-retryable 4xx responses are translated into specific errors:
+//   - 401 → auth hint
+//   - 403 → permissions hint
+//   - 404 → not-found
+//   - others → raw status + snippet of body
+func (s *HFSource) doGet(ctx context.Context, url string) (io.ReadCloser, *http.Response, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
markphelps, line 517: should we add a timeout on the request context?

pkg/model/packer.go
```diff
@@ -0,0 +1,555 @@
...
+	".pt":          true,
+	".pth":         true,
+}
+
+// packOptions configures the packing algorithm.
+type packOptions struct {
+	// BundleFileMax is the size threshold separating small from large files.
+	// Files below this are bundled; files at or above get their own layer.
+	// Defaults to defaultBundleFileMax (64 MB).
+	BundleFileMax int64
```
markphelps, line 65: why not uint64 for these?

pkg/model/packer.go
```diff
@@ -0,0 +1,555 @@
...
+type packedLayer struct {
+	// Plan is the layerPlan that produced this layer's bytes. Held so
+	// fileLayer.Compressed() can reconstruct the bytes by re-running
+	// the same packing pipeline against the store.
+	Plan layerPlan
+	// Digest is the SHA256 digest of the tar bytes (the OCI blob digest).
+	Digest v1.Hash
+	// Size is the size of the tar bytes in bytes (post-compression for
+	// gzip layers).
+	Size int64
```
markphelps, line 109: uint64?

pkg/model/packer.go
```diff
@@ -0,0 +1,555 @@
...
+	Files []packedFile
+}
+
+// packedFile records a file's path, size, content digest, and which layer it
+// landed in. Used to build the config blob (§2.3) and set digest (§2.4).
+type packedFile struct {
+	// Path is the file path relative to the weight target directory.
+	Path string
+	// Size is the uncompressed file size in bytes.
+	Size int64
```
markphelps, line 132: guess what im gonna say

pkg/model/weights_status.go
```diff
@@ -0,0 +1,250 @@
...
+			return checkWeightLayers(ctx, r, repo, reg)
+		})
+	}
+
+	return g.Wait()
+}
+
+// checkWeightLayers checks each layer of a single weight against the registry
+// and populates the result's Layers and Status fields.
+func checkWeightLayers(ctx context.Context, r *WeightStatusResult, repo string, reg registry.Client) error {
```
markphelps, line 163: should we lock here as we are mutating `r` in multiple goroutines

markphelps (commented):
mostly nits, one potential issue around concurrent mutations

pkg/model/weightsource/huggingface.go
michaeldwan: some requests (blob downloads) will take a significant amount of time and we don't want to arbitrarily timeout. context cancellation is threaded through so failures in other parallel operations or sigint from the cli will do the right thing.

pkg/model/packer.go
michaeldwan: all the sizes we're interfacing with through stdlib (tar, os.File, io.Copy) as well as the OCI spec/v1 library use int64. so even though a negative isn't possible, the convention is set and I didn't want to be casting everywhere

pkg/model/packer.go
michaeldwan: same

pkg/model/packer.go
michaeldwan: same

pkg/model/weights_status.go
michaeldwan: good eye, but each goroutine is modifying its own index in a pre-sized slice which is a common pattern and safe for concurrent r/w. modifying the _same_ index would be a race caught be the race detector though!

pkg/weights/lockfile/lockfile.go
```diff
@@ -201,13 +262,56 @@ func LoadWeightsLock(path string) (*WeightsLock, error) {
...
+func atomicWriteFile(path string, data []byte, mode os.FileMode) error {
+	dir := filepath.Dir(path)
+	tmp, err := os.CreateTemp(dir, ".weights-*.lock.tmp")
+	if err != nil {
+		return fmt.Errorf("create temp file: %w", err)
+	}
+	tmpPath := tmp.Name()
+	// No-op once Rename succeeds (file is gone); covers panic/SIGINT
+	// between CreateTemp and Rename.
+	defer func() { _ = os.Remove(tmpPath) }()
```
markphelps, line 294: why not put tmp.Close in the defer before remove then we dont need the ` _ = tmp.Close()` everywhere below, unless we really care about returning the error in the last bit on close failure

markphelps (approved):
one minor potential nit, but lgtm!!

pkg/weights/lockfile/lockfile.go
michaeldwan: good call. will resolve in a followup PR

michaeldwan: said PR: https://github.com/replicate/cog/pull/2987

PR #2978, opened 2026-04-29
fix: support TypedDict schema generation

### LINKED ISSUE #2973: Schema generation does not handle TypedDict as dict
I have code which used existing TypedDict classes as predict input parameters. Neither static nor legacy schema generation treats TypedDict as a dict.

For legacy schema generation, I can include the following in my module which lets the schema generation handle TypedDict as dict.

```python
class TypedDictCoder(Coder):
    """This class lets cog accept TypedDict as a dict Input."""

    @staticmethod
    def factory(tpe: type) -> Optional["TypedDictCoder"]:
        if issubclass(tpe, dict):
            return TypedDictCoder()
        return None

    def encode(self, x: dict) -> dict:
        return x

    def decode(self, x: dict) -> dict:
        return x
```

This work around will not help with static schema generation. Both legacy and static schema generation should handle TypedDict as dict.

There are several instances of code like `tpe is dict` which probably should be something like `issubclass(tpe, dict)`. Similar equality with `dict` checks are in the Go static schema generation code.

markphelps:
Fixes: #2973 

## Summary
- fix legacy runtime schema generation so `TypedDict` inputs are treated like dicts without accepting arbitrary dict subclasses
- add static schema support for `TypedDict` inputs and outputs, including qualified imports, inheritance, and `Required` / `NotRequired` key semantics
- add regression coverage in Python, Go parser tests, and integration txtar scripts for static and legacy TypedDict paths

## Notes
- added integration txtar coverage for `typeddict_input`, `typeddict_input_legacy`, and `typeddict_output`

### DIFF
--- a/pkg/schema/python/parser.go
+++ b/pkg/schema/python/parser.go
@@ -45,9 +45,10 @@ func ParsePredictor(source []byte, predictRef string, mode schema.Mode, sourceDi
 
 	// 3. Collect schema object classes (BaseModel and TypedDict) locally first,
 	// then across local imports.
-	modelClasses, typedDicts := collectModelClasses(root, source, imports)
+	modelCtx := &modelParseContext{imports: imports, typedDicts: make(map[string]bool)}
+	modelClasses := collectModelClasses(root, source, modelCtx)
 	if sourceDir != "" {
-		resolveExternalModels(sourceDir, imports, modelClasses, typedDicts)
+		resolveExternalModels(sourceDir, modelClasses, modelCtx)
 	}
 
 	// 4. Collect Input() references from class attributes and static methods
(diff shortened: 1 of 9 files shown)


pkg/schema/python/parser.go
```diff
@@ -71,7 +72,7 @@ func ParsePredictor(source []byte, predictRef string, mode schema.Mode, sourceDi
 	isMethod := firstParamIsSelf(paramsNode, source)
 
 	// 7. Extract parameters
-	inputs, err := extractInputs(paramsNode, source, methodName, isMethod, imports, inputRegistry, moduleScope)
+	inputs, err := extractInputs(paramsNode, source, methodName, isMethod, imports, inputRegistry, moduleScope, typedDicts)
```
markphelps: this is a lot of params. we should make this take in a struct

pkg/schema/python/parser.go
```diff
@@ -43,10 +43,11 @@ func ParsePredictor(source []byte, predictRef string, mode schema.Mode, sourceDi
...
 	moduleScope := collectModuleScope(root, source)
 
-	// 3. Collect BaseModel subclasses (local file first, then cross-file)
-	modelClasses := collectModelClasses(root, source, imports)
+	// 3. Collect schema object classes (BaseModel and TypedDict) locally first,
+	// then across local imports.
+	modelClasses, typedDicts := collectModelClasses(root, source, imports)
 	if sourceDir != "" {
-		resolveExternalModels(sourceDir, imports, modelClasses)
+		resolveExternalModels(sourceDir, imports, modelClasses, typedDicts)
```
markphelps: wondering if this is getting into struct as param territory too

michaeldwan (approved):
nice fix

PR #2996, opened 2026-05-04
chore: regenerate compatability matrices for python 3.13+; fix script

markphelps:
This pull request regenerates the CUDA, PyTorch, and TensorFlow compatibility matrices and fixes the PyTorch generator so regenerated entries stay consistent with upstream PyTorch package index format.

### CUDA

- Added CUDA 13.0.3, 13.1.2, 13.2.0, and 13.2.1 development images.
- Each new CUDA version includes Ubuntu 22.04 and Ubuntu 24.04 variants.
- All added CUDA rows use cuDNN 9.
- No existing CUDA rows were removed or changed.

### TensorFlow

- Added TensorFlow 2.21.0.
- The new TensorFlow row uses CUDA 12.5 and cuDNN 9.3.
- Supported Python versions are 3.10, 3.11, 3.12, and 3.13.
- No existing TensorFlow rows were removed or changed.

### PyTorch

- Added PyTorch 2.11.0 rows for CUDA 12.6, 12.8, 12.9, 13.0, and CPU.
- Preserved upstream Python 3.14 compatibility where PyTorch publishes CPython 3.14 wheels.
- Normalized PyTorch `ExtraIndexURL` values to include trailing slashes.
- Preserved CUDA/CPU local-version suffixes in the matrix, for example `2.10.0+cu126` and `2.10.0+cpu`.
- Kept generated install requirements stable for CPU builds by stripping the `+cpu` suffix when writing `torch==...` requirements.

## Notable semantic differences

- PyTorch rows changed from 98 to 96 total rows.
- Added 5 PyTorch 2.11.0 rows.
- Removed 6 Python 2.7-only legacy rows for PyTorch 1.2.0 and 1.4.0.
- Removed the PyTorch 2.10.0 CUDA 12.9 row.
- No existing PyTorch rows changed their Python compatibility list after preserving upstream 3.14 wheels.
- 17 PyTorch rows include Python 3.14 compatibility.
- Existing PyTorch URL changes are normalization-only: `https://download.pytorch.org/whl/<variant>` became `https://download.pytorch.org/whl/<variant>/`.

## Generator fixes

- Ignore non-wheel links from PyTorch package indexes, avoiding noisy `.tar.gz` parse warnings.
- Ignore non-CPython wheel tags such as `py3` so they do not produce invalid versions like `p.y3`.
- Normalize PyTorch index URLs before package discovery to avoid trailing-slash inconsistencies.
- Join package index paths safely regardless of whether the base URL already has a trailing slash.
- Preserve future Python minor compatibility when upstream PyTorch publishes matching CPython wheels, while still filtering Python versions below Cog's minimum.


michaeldwan:
this seems sensible! I was double checking that nothing looked for the name `tf_compatibility_matrix.json` and found there's _another_ copy of generate-compat in the mise.toml tasks! just calling that out. whatever you do here should be what we keep. I'm all for the script since the mise task file is getting big.

markphelps:
> this seems sensible! I was double checking that nothing looked for the name `tf_compatibility_matrix.json` and found there's _another_ copy of generate-compat in the mise.toml tasks! just calling that out. whatever you do here should be what we keep. I'm all for the script since the mise task file is getting big.

i removed the script to go with the mise way

michaeldwan (approved):
nice!

PR #3000, opened 2026-05-05
Filter .cog/ from Docker build context, centralize build state

michaeldwan:
We want `.cog/` to be the single local state directory for a Cog project -- build staging, weight caches, mount scratch space, advisory locks. But until now, everything under the project root got sent to the Docker daemon as build context, so putting anything substantial in `.cog/` would bloat builds or require users to manually `.dockerignore` it. And the build system was already fighting this by mutating `.dockerignore` on disk mid-build to exclude weight directories -- a racy pattern that left backup files on crashes.

This PR makes `.cog/` invisible to Docker builds at the BuildKit transport layer (`fsutil.NewFilterFS` with `ExcludePatterns`), so it never reaches the daemon regardless of what the user's `.dockerignore` says. That means we can now put weight caches, build artifacts, and whatever else in `.cog/` without worrying about context size or user-facing configuration.

The other piece is `pkg/dotcog`, which owns the `.cog/` directory lifecycle -- path accessors, a project-wide advisory lock (flock), and cleanup registration. Build staging artifacts move from `.cog/tmp/build<timestamp>/` to `.cog/build/` and are referenced via a `cog_build` named build context (`COPY --from=cog_build`), so the generated Dockerfile no longer needs relative paths back into the project tree.

What got removed along the way:

- The `.dockerignore` backup/restore/rewrite dance and its pre-flight `checkCompatibleDockerIgnore` check
- `pkg/dockercontext/`, `pkg/dockerignore/`, the `NewGenerator` factory wrapper
- `lockfile.WithLock` (replaced by the project-wide `dotcog.Dir.Lock`)
- The `bad_dockerignore` integration test (the error condition no longer exists)

The Dockerfile mount to BuildKit is also now scoped to just the Dockerfile file itself, preventing build artifacts from leaking through that mount.


pkg/docker/buildkit.go
```diff
@@ -90,7 +126,11 @@ func solveOptFromImageOptions(buildDir string, opts command.ImageBuildOptions) (
 			console.Warnf("build context name collision: %q", name)
 			continue
 		}
-		localDirs[name] = dir
+		bcFS, fsErr := fsutil.NewFS(dir)
+		if fsErr != nil {
+			return buildkitclient.SolveOpt{}, fmt.Errorf("create build context fs %q: %w", name, fsErr)
+		}
+		localMounts[name] = bcFS
 		// Tell the dockerfile frontend about this build context
```
markphelps, line 94: i know there are no existing tests for buildkit.go and it might be hard to test.. but just calling out that this is a substantial ammount of code that we should probably have unit tests for if possible

integration-tests/tests/bad_dockerignore.txtar
```diff
@@ -1,20 +0,0 @@
...
-
--- .dockerignore --
-.cog
-
--- predict.py --
-from cog import BasePredictor
-
-class Predictor(BasePredictor):
-    def predict(self, s: str) -> str:
-        return "hello " + s
```
markphelps, line 20: should probably add a test that there is no error now if .cog is docker ignored?

architecture/06-cli.md
```diff
@@ -248,7 +248,7 @@ Commands delegate to packages under `pkg/`:
 **Utilities:**
 
 - `pkg/dockercontext/` -- Docker build context directory management
```
markphelps:
```suggestion
```

pkg/dockerfile/standard_generator.go
```diff
@@ -93,17 +68,7 @@ type StandardGenerator struct {
...
-	if err != nil {
-		return nil, err
-	}
-	// tmpDir, but without dir prefix. This is the path used in the Dockerfile.
-	relativeTmpDir, err := filepath.Rel(dir, tmpDir)
-	if err != nil {
-		return nil, err
-	}
-
+func NewStandardGenerator(config *config.Config, dir string, buildCacheDir string, configFilename string, command command.Command, client registry.Client, requiresCog bool) (*StandardGenerator, error) {
```
markphelps, line 71: note for later, i feel like this is staring to scream for an options struct

pkg/model/source.go
```diff
@@ -71,6 +80,11 @@ func NewSourceFromConfig(cfg *config.Config, projectDir string) *Source {
 	}
 }
 
+// Close releases resources held by the Source, including the .cog/ directory handle.
+func (s *Source) Close() error {
+	return s.DotCog.Close()
```
markphelps, line 88: is DotCog guaranteed to not be nil? we should prob do a safe nil check here

pkg/image/build.go
```diff
@@ -140,7 +185,11 @@ func Build(
 	// lockfile to the minimal runtime manifest (spec §3.3) and write it into
 	// the build context so it ends up at /.cog/weights.json in the image.
 	if len(cfg.Weights) > 0 {
-		if err := writeRuntimeWeightsManifest(dir); err != nil {
+		if err := writeRuntimeWeightsManifest(dir, bp.weightsFile); err != nil {
+			return "", err
+		}
+		// Mirror to .cog/ root for volume-mount visibility.
+		if err := writeRuntimeWeightsManifest(dir, bp.rootWeightsFile); err != nil {
```
markphelps: i feel like we should make these writes atomic so they dont leave .cog in a weird state

pkg/dotcog/dotcog.go
```diff
@@ -0,0 +1,200 @@
...
+
+// Open returns a Dir rooted at <projectDir>/.cog/, creating it if needed.
+func Open(projectDir string) (*Dir, error) {
+	root := filepath.Join(projectDir, Name)
+	if err := os.MkdirAll(root, 0o755); err != nil {
+		return nil, fmt.Errorf("create %s: %w", root, err)
+	}
+	abs, err := filepath.Abs(root)
+	if err != nil {
+		return nil, err
```
markphelps: should we wrap this error like above?

pkg/dotcog/dotcog.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil
+	}
+	d.closeOnce.Do(func() {
+		d.mu.Lock()
+		cleanups := d.cleanups
+		d.cleanups = nil
+		d.mu.Unlock()
+
+		var errs []error
+		for i := len(cleanups) - 1; i >= 0; i-- {
```
markphelps: could use [`slices.Backward`](https://pkg.go.dev/slices#Backward) here for cleaner code

pkg/docker/buildkit.go
michaeldwan: agree, though until then (custom bk frontend probably) it's indirectly tested by every test/IT that produces an image, so the inputs and outputs are at least well exercised.

markphelps (approved):
lgtm!!

PR #3003, opened 2026-05-06
chore: Remove runtime schema generation

markphelps:
## Summary
- Remove the Docker/Python runtime OpenAPI schema generation path and make build-time schema generation static-only.
- Delete legacy schema fallback integration tests and the static-vs-runtime schema comparison harness command.
- Update schema architecture docs and agent guidance to describe the static-only schema flow.


michaeldwan (approved):
nice thanks!

once this is in I'll remove the 2 pass build process

PR #3006, opened 2026-05-07
fix: sort compatibility matrices for deterministic output

markphelps:
Fixes the compatibility matrix update workflow opening PRs even when the underlying data has not changed.

## Problem

The scheduled update-compatibility-matrices workflow was creating PRs on days when no new CUDA/PyTorch/TensorFlow versions were released. This happened because the generated JSON was not deterministic.

## Root Cause

- CUDA: already sorted correctly
- PyTorch: non-deterministic because fetchCurrentTorchVersions iterates over a Go map with random iteration order
- TensorFlow: non-deterministic because it iterates over HTML table rows in document order

## Changes

- Added stable sorting to FetchTorchCompatibilityMatrix
- Added stable sorting to FetchTensorFlowCompatibilityMatrix
- Added tests verifying deterministic sorting for both generators

This should make the workflow idempotent: if the upstream data has not changed, the generated JSON will be byte-for-byte identical and no PR will be opened.

### DIFF
--- a/tools/compatgen/internal/tensorflow.go
+++ b/tools/compatgen/internal/tensorflow.go
@@ -1,7 +1,9 @@
 package internal
 
 import (
+	"cmp"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


michaeldwan (approved):
good fix

PR #3008, opened 2026-05-07
Support multi-source weights and HTTPS weight sources

michaeldwan:
Weights can now declare multiple sources that merge into a single target directory, and a new `https://` / `http://` source covers GitHub Releases, ONNX Model Zoo, and similar plain-HTTP downloads.

The driving case is models that pull `.pth` files from a few GitHub release URLs into one weights dir alongside a HuggingFace base. Today that needs a wrapper script; with this PR it's:

```yaml
weights:
  - name: realesrgan
    target: /src/weights/realesrgan
    source:
      - uri: hf://acme/base
      - uri: https://github.com/xinntao/Real-ESRGAN/releases/download/v0.1.0/RealESRGAN_x4plus.pth
```

Path conflicts resolve last-in-wins (declaration order = layering order, like CSS) and surface at info level so silent shadowing doesn't slip into a cog.yaml diff unnoticed.

### Notes for reviewers

- **Lockfile schema:** `source` (object) → `sources` (array), with `importedAt` moved up to the entry level. Version stays at `1` — managed weights are still experimental, real users are minimal, and a forced re-import on upgrade is preferable to compat shims. The lockfile-version story will get revisited as part of the broader cog ecosystem compatibility-date plan.
- **HTTP credentials in URIs are rejected** at parse time. The URI lands verbatim in `weights.lock` (which is checked into git), so embedded `user:pass@host` would leak. Auth-header support is a separate follow-up.
- **HTTP ETag is treated as a cache hint, not a content identity.** Weak ETags (`W/"..."`) are ignored. The store still verifies the sha256 on `PutFile`, so a mutable URL serving different bytes between inventory and ingress fails loudly.
- **`source: { ... }` and `source: [ ... ]` both parse**, with custom `Marshal`/`Unmarshal` for both YAML and JSON keeping single-source weights round-tripping as a mapping.

No integration test for HTTPS — would need either a fixture server in the testscript harness or live network calls. The unit tests on `httptest` cover the HTTP source thoroughly; happy to add a `weights_multisource.txtar` covering two `file://` sources as a follow-up if you want.


pkg/model/weightsource/http.go
```diff
@@ -0,0 +1,259 @@
...
+		client:   newHTTPSourceClient(),
+	}, nil
+}
+
+// parseHTTPURI parses uri and rejects anything outside the http/https
+// scheme set or with embedded credentials. Centralizes the rules so
+// that NewHTTPSource and normalizeHTTPURI cannot drift apart. The
+// "no userinfo" rule exists because the URI lands in weights.lock —
+// see NewHTTPSource for the full rationale.
+func parseHTTPURI(uri string) (*url.URL, error) {
```
markphelps, line 75: Query-string credentials like presigned S3/GCS/GitHub tokens are accepted and can be written to weights.lock or shown in output. Fix by separating fetch URL from persisted/display URL and redacting or rejecting credential-bearing query params.

pkg/model/weightsource/http.go
```diff
@@ -0,0 +1,259 @@
...
+	parsed, err := url.Parse(uri)
+	if err != nil {
+		return nil, fmt.Errorf("invalid HTTP URI %q: %w", uri, err)
+	}
+	if parsed.Scheme != "https" && parsed.Scheme != "http" {
```
markphelps, line 80: should we instead default to only allow HTTPS with maybe insecure opt-in for localhost/dev cases?

michaeldwan: not sure what we'd gain by restricting it atm. we aren't running untrusted code, so there's not the same risk as eg docker where spoofing or mitm could serve different content. this is just fetching and unpacking blobs

pkg/model/weightsource/http.go
michaeldwan: good point! I don't think we should block on that, but worth coming back to. I'll make an issue!

markphelps (dismissed):
lgtm!!

PR #3015, opened 2026-05-12
feat: Rename predict to run with backward compatibility

markphelps:
## Summary

Renames the user-facing `predict` concept to `run` across the entire stack while maintaining full backward compatibility.

### CLI
- New `cog run` command (replaces `cog predict` as primary)
- `cog predict` kept as deprecated alias with warning
- Smart dispatch: `cog run` handles both predictions and exec commands

### Python SDK
- New `BaseRunner` class (replaces `BasePredictor`)
- New `run()` method (replaces `predict()`)
- `BasePredictor` kept as compatibility alias
- `predict()` still works via two-way delegation bridge

### Configuration
- New `run:` field in `cog.yaml` (replaces `predict:`)
- `predict:` still accepted with deprecation warning
- `cog doctor --fix` can auto-migrate common patterns

### Examples & Docs
- `predict.py` → `run.py` in examples
- `Predictor` → `Runner` in examples
- README.md, docs, architecture docs updated

### Tests
- Integration tests for both legacy and new interfaces
- Doctor migration tests
- Deprecation warning tests

## Backward Compatibility

All existing models continue to work without changes:
- `cog predict` → works with warning
- `BasePredictor` → works with warning
- `predict()` → works with warning
- `predict:` in cog.yaml → works with warning


markphelps:

pkg/cli/doctor.go
```diff
@@ -30,7 +30,7 @@ func newDoctorCommand() *cobra.Command {
 NOTE: cog doctor is experimental. Behavior and checks may change in future versions.
 
 By default, cog doctor reports problems without modifying any files.
-Pass --fix to automatically apply safe fixes.`,
+Pass --fix to automatically apply safe fixes, including migrating deprecated predict names to run names when no file collision exists.`,
```
michaeldwan: Love the idea of `--fix`, but seems out of place to call out one type of automatic fix it'll do.

pkg/cli/run.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+type runDispatchMode int
+
+const (
+	runDispatchPredict runDispatchMode = iota
+	runDispatchExec
+)
+
+func runDispatchModeForArgs(args []string) runDispatchMode {
```
michaeldwan: IMO I think we can yank support for still using `run` to invoke exec. It's been deprecated with a warning for ~4-6 weeks now. The complexity of guessing isn't worth it and may have real negative side effects.

pkg/config/data/config_schema_v1.0.json
```diff
@@ -165,10 +168,15 @@
...
+      "description": "The pointer to the Runner object in your code, which defines how runs are executed on your model."
+    },
     "predict": {
       "$id": "#/properties/predict",
       "type": "string",
-      "description": "The pointer to the `Predictor` object in your code, which defines how predictions are run on your model."
+      "description": "Deprecated compatibility field for run. Use run instead."
     },
     "train": {
       "$id": "#/properties/train",
```
michaeldwan, line 182: While you're here: the `train` field description just below (line 184) still says "The pointer to the `Predictor` object in your code, which defines how predictions are run on your model." -- that's the old `predict` description that was never updated. Should describe the training interface.

crates/coglet-python/src/predictor.rs
```diff
@@ -991,11 +1039,18 @@ impl PythonPredictor {
...
+            let method_name = match &self.kind {
+                PredictorKind::Class { method_name, .. } => method_name.as_str(),
+                PredictorKind::StandaloneFunction(_) => "",
+            };
+            let coro = if method_name.is_empty() {
+                instance.call((), Some(&input_dict))
+            } else {
+                instance.call_method(method_name, (), Some(&input_dict))
+            }
+            .map_err(|e| PredictionError::Failed(format!("Failed to call predict: {}", e)))?;
```
michaeldwan: This error message says "Failed to call predict" even when the method being invoked is `run()`. `method_name` is available here -- use it:
```rust
.map_err(|e| PredictionError::Failed(format!("Failed to call {method_name}: {e}")))?;
```

integration-tests/tests/predict_legacy_predictor.txtar
```diff
@@ -0,0 +1,17 @@
+cog build -t $TEST_IMAGE
+stderr 'predict'
```
michaeldwan: `stderr 'predict'` matches any line containing "predict" -- build log noise, image names, the filename itself, etc. This would pass even if the deprecation warning were missing. Assert the actual deprecation message text instead.

Same issue in `run_legacy_predictor.txtar:2` and `run_legacy_async_predictor.txtar:2`.

integration-tests/tests/run_new_runner.txtar
```diff
@@ -0,0 +1,16 @@
+# Build and run a new-style model
+cog build -t $TEST_IMAGE
+cog run $TEST_IMAGE -i s=world
```
michaeldwan, line 3: The golden-path test for new-style Runner models should verify no deprecation noise appears. Add `! stderr 'deprecated'` to make that explicit.

pkg/doctor/check_predict_to_run_migration.go
```diff
@@ -0,0 +1,422 @@
...
+		ctx.LoadResult.Config.Predict = "run.py:Runner"
+		warnings := ctx.LoadResult.Warnings[:0]
+		for _, warning := range ctx.LoadResult.Warnings {
+			if warning.Field != "predict" {
+				warnings = append(warnings, warning)
+			}
+		}
+		ctx.LoadResult.Warnings = warnings
+	}
+	delete(ctx.PythonFiles, "predict.py")
```
michaeldwan, line 128: **blocker: tree-sitter tree leaked**

`delete(ctx.PythonFiles, "predict.py")` removes the map entry without closing the tree first. The tree's C memory leaks until GC finalizer runs.

```go
if old := ctx.PythonFiles["predict.py"]; old != nil && old.Tree != nil {
    old.Tree.Close()
}
delete(ctx.PythonFiles, "predict.py")
```

michaeldwan (commented):
## Code Review

Solid rename with comprehensive backward compatibility. 68 files, ~3400 lines -- no missed rename sites found. Test coverage is thorough across all paths (27 scenarios verified). The intent is well-executed.

### Summary

- **Blockers: 2** -- tree-sitter tree leak, test temp file leak
- **Should-fix: 9** -- non-atomic migration, diamond MRO, per-call overhead, duplicated logic, missing cycle protection, return type lie, fragile stacklevel, inconsistent test style
- **Nits: 6** -- doc comments, duplicate imports, string-literal mode check, error messages

See inline comments for details.

### Nits not attached inline (lines outside diff context)

- **`crates/coglet-python/src/predictor.rs:260`** -- doc comments on `PredictKind` variants still say `def predict(...)`. Should be `def run/predict(...)` since both are valid now.
- **`pkg/cli/init.go:163`** -- `fmt.Errorf("%w", err)` adds no context -- equivalent to `return nil, err`. Either add context or return the error directly.
- **`python/cog/predictor.py:213`** -- when `Runner` exists but is the wrong type (e.g. a string), the error says "has no Runner or Predictor" which is misleading. A more specific message like "Runner exists but is not a class or callable" would help.

crates/coglet-python/src/predictor.rs
```diff
@@ -1290,3 +1346,191 @@ impl PythonPredictor {
...
+    }
+
+    fn load_predictor_source(source: &str) -> PyResult<PythonPredictor> {
+        pyo3::Python::initialize();
+        let path = write_predictor_source(source);
+        Python::attach(|py| {
+            add_python_sdk_path(py);
+            let predictor_ref = format!("{}:Predictor", path.display());
+            let result = PythonPredictor::load(py, &predictor_ref);
+            let _ = fs::remove_file(&path);
```
michaeldwan: **blocker: test temp files and `__pycache__` leak on panic**

If `PythonPredictor::load` panics, `fs::remove_file` is skipped. Python's import machinery also creates `__pycache__/*.pyc` files in the temp dir that are never cleaned up.

Use `tempfile::tempdir()` which handles cleanup on Drop (including panics) and catches the `__pycache__` directory too:

```rust
fn load_predictor_source(source: &str) -> PyResult<PythonPredictor> {
    pyo3::Python::initialize();
    let dir = tempfile::tempdir().expect("failed to create temp dir");
    let path = dir.path().join("predictor.py");
    fs::write(&path, source).expect("failed to write test predictor");
    Python::attach(|py| {
        add_python_sdk_path(py);
        let predictor_ref = format!("{}:Predictor", path.display());
        PythonPredictor::load(py, &predictor_ref)
    })
    // dir drops here, cleaning up .py and __pycache__
}
```

pkg/doctor/check_predict_to_run_migration.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+	source := applyEdits(pf.Source, edits)
+
+	configPath := filepath.Join(ctx.ProjectDir, ctx.ConfigFilename)
+	configBytes, err := os.ReadFile(configPath)
+	if err != nil {
+		return err
+	}
+	configBytes = predictRefPattern.ReplaceAll(configBytes, []byte(`run: "run.py:Runner"`))
+	if err := os.WriteFile(configPath, configBytes, 0o644); err != nil {
```
michaeldwan: **should-fix: non-atomic migration leaves broken state on partial failure**

Fix() writes cog.yaml first (here), then run.py (line 89), then removes predict.py (line 92). If the run.py write fails, cog.yaml already points at a non-existent file. If `os.Remove` fails, both .py files exist and re-running `cog doctor --fix` fails at the "run.py already exists" preflight.

Also: `WriteFile` hardcodes `0o644`, discarding original permissions.

Consider using `os.Rename` for the predict.py -> run.py move -- it's atomic on the same filesystem, preserves permissions, and fixes the TOCTOU race between the preflight Stat and the WriteFile. Then reorder to: rename predict.py -> run.py, then rewrite cog.yaml.

python/cog/predictor.py
```diff
@@ -106,18 +120,57 @@ def record_metric(self, key: str, value: Any, mode: str = "replace") -> None:
...
 
-def load_predictor_from_ref(ref: str) -> BasePredictor:
+class BasePredictor(BaseRunner):
+    """Deprecated compatibility alias for BaseRunner."""
+
+
+def _user_method_owner(cls: type[Any], method_name: str) -> type[Any] | None:
+    for owner in inspect.getmro(cls):
+        if owner in {BaseRunner, BasePredictor, object}:
+            break
```
michaeldwan, line 151: **should-fix: MRO walk breaks at first sentinel -- wrong for diamond inheritance**

The `break` on `BaseRunner`/`BasePredictor`/`object` stops scanning the entire MRO. With diamond inheritance (`class C(A, B)` where both A and B inherit BaseRunner), if A's branch hits BaseRunner first, B's user-defined methods are never checked. The Rust version (`predictor.rs:427`) has the same `break` behavior.

Either change `break` to `continue` in both Python and Rust, or document that diamond inheritance with BaseRunner is unsupported and add a test asserting it.

python/cog/predictor.py
```diff
@@ -40,34 +40,48 @@ def setup(
...
 
         Returns:
             The prediction output.
 
         Raises:
-            NotImplementedError: If predict is not implemented.
+            NotImplementedError: If run is not implemented.
         """
-        raise NotImplementedError("predict has not been implemented by parent class.")
+        run_owner = _user_method_owner(self.__class__, "run")
```
michaeldwan: **should-fix: MRO walk on every prediction call**

`_user_method_owner()` calls `inspect.getmro()` + dict lookups on every prediction. The result never changes for a given class.

Cache it in `__init_subclass__`:
```python
class BaseRunner:
    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls._run_owner = _user_method_owner(cls, 'run')
        cls._predict_owner = _user_method_owner(cls, 'predict')

    def run(self, *args, **kwargs):
        if self.__class__._predict_owner is not None and self.__class__._run_owner is None:
            return self.predict(*args, **kwargs)
        raise NotImplementedError('run has not been implemented by parent class.')
```

crates/coglet-python/src/predictor.rs
```diff
@@ -283,6 +284,7 @@ pub enum TrainKind {
 pub enum PredictorKind {
     /// Class instance with predict() method, optionally train()
     Class {
+        method_name: String,
```
michaeldwan: **should-fix: `method_name: String` should be an enum**

Only two possible values: `"run"` and `"predict"`. A `PredictMethodName` enum would be zero-cost, self-documenting, prevent invalid values, and keep `PredictorKind` `Copy`-able:

```rust
pub enum PredictMethodName {
    Run,
    Predict,
}

impl PredictMethodName {
    fn as_str(&self) -> &'static str {
        match self {
            Self::Run => "run",
            Self::Predict => "predict",
        }
    }
}
```

crates/coglet-python/src/predictor.rs
```diff
@@ -406,6 +412,44 @@ impl PythonPredictor {
         Ok(predictor)
     }
 
+    fn selected_predict_method_name(py: Python<'_>, instance: &PyObject) -> PyResult<String> {
```
michaeldwan: **should-fix: MRO walk and validation duplicated three times**

The run-vs-predict detection is implemented independently here (Rust), in `_validate_runner_class` (predictor.py:146), and in `_selected_predict_method` (_inspector.py:100). The Rust version also doesn't emit deprecation warnings when `predict` is selected, unlike both Python paths.

Either have Rust call the Python `_validate_runner_class` to deduplicate, or at minimum add `tracing::warn!` here when `predict` is selected.

crates/coglet-python/src/predictor.rs
```diff
@@ -406,6 +412,44 @@ impl PythonPredictor {
...
+        let class = instance.bind(py).getattr("__class__")?;
+        let mro = class.getattr("__mro__")?.cast_into::<PyTuple>()?;
+        let cog_predictor = py.import("cog.predictor")?;
+        let base_runner = cog_predictor.getattr("BaseRunner")?;
+        let base_predictor = cog_predictor.getattr("BasePredictor")?;
+        let object = py.import("builtins")?.getattr("object")?;
```
michaeldwan: **nit:** `py.import("builtins")` is called twice (here and line 422). Import once, reuse:
```rust
let builtins = py.import("builtins")?;
let object = builtins.getattr("object")?;
let callable = builtins.getattr("callable")?;
```

pkg/schema/python/parser.go
```diff
@@ -1182,37 +1198,224 @@ func resolveInputReference(node *sitter.Node, source []byte, registry *inputRegi
...
+
+	parentFile, ok := loadPythonFileContext(file.sourceDir, module)
+	if !ok {
+		return nil, nil, "", false
+	}
+	parentNode := findClassByName(parentFile.root, parentFile.source, className)
+	return parentFile, parentNode, className, parentNode != nil
+}
+
+func loadPythonFileContext(sourceDir, module string) (*pythonFileContext, bool) {
```
michaeldwan: **should-fix: `loadPythonFileContext` can recurse indefinitely**

`loadPythonFileContext` -> `resolveExternalModels` (line 1377) -> `loadModelsFromModule` -> `loadPythonFileContext`. No visited-set or depth limit on the model resolution path. `collectPredictMethods` has cycle protection via `seen`, but this path doesn't. Circular imports between local .py files would stack-overflow.

Add a `visited map[string]bool` parameter or a depth limit.

python/cog/predictor.py
```diff
@@ -106,18 +120,57 @@ def record_metric(self, key: str, value: Any, mode: str = "replace") -> None:
...
+        )
+    if any(base is BasePredictor for base in inspect.getmro(cls)[1:]):
+        warnings.warn(
+            "BasePredictor is deprecated; use BaseRunner instead",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+
+
+def load_predictor_from_ref(ref: str) -> BaseRunner:
```
michaeldwan: **should-fix: return type annotation lies**

Declared `-> BaseRunner` but can return a bare callable (standalone function for training) at line 222. Change to `-> Union[BaseRunner, Callable[..., Any]]`.

python/cog/predictor.py
```diff
@@ -106,18 +120,57 @@ def record_metric(self, key: str, value: Any, mode: str = "replace") -> None:
...
+        raise ValueError(
+            f"{class_name} must define either run() or predict(), not both"
+        )
+    if not defines_run and not defines_predict:
+        raise ValueError(f"run or predict method not found: {class_name}")
+    if defines_predict:
+        warnings.warn(
+            f"{class_name}.predict() is deprecated; use run() instead",
+            DeprecationWarning,
+            stacklevel=3,
```
michaeldwan: **should-fix: fragile `stacklevel=3`**

Three different call sites hardcode `stacklevel=3` (`_validate_runner_class` here, `_inspector._selected_predict_method:115`, `_inspector._warn_if_base_predictor_in_mro:126`), each assuming a specific call depth. If any caller moves, warnings point to the wrong line. Consider documenting the expected call chain or accepting a stacklevel parameter.

python/tests/test_predictor.py
```diff
@@ -23,7 +341,7 @@ def test_default_predict_raises(self) -> None:
             predictor.predict()
             assert False, "Should have raised NotImplementedError"
```
michaeldwan: **should-fix: raw try/except instead of `pytest.raises`**

Every other test in this file uses `pytest.raises`. This one uses `try`/`assert False` which is inconsistent and produces worse failure messages:
```python
def test_default_predict_raises(self) -> None:
    predictor = BasePredictor()
    with pytest.raises(NotImplementedError, match='run has not been implemented'):
        predictor.predict()
```

pkg/schema/python/parser.go
```diff
@@ -1182,37 +1198,224 @@ func resolveInputReference(node *sitter.Node, source []byte, registry *inputRegi
...
+	for _, child := range NamedChildren(file.root) {
 		classNode := UnwrapClass(child)
 		if classNode == nil {
 			continue
 		}
 		nameNode := classNode.ChildByFieldName("name")
-		if nameNode != nil && Content(nameNode, source) == predictRef {
-			return findMethodInClass(classNode, source, predictRef, methodName)
+		if nameNode != nil && Content(nameNode, file.source) == predictRef {
+			if methodName == "predict" {
```
michaeldwan: **nit:** `methodName == "predict"` is a string-literal proxy for "is this predict mode". The `mode` variable is already in scope (line 59). Using `mode == schema.ModePredict` would be less fragile.

crates/coglet-python/src/predictor.rs
```diff
@@ -991,11 +1039,19 @@ impl PythonPredictor {
...
 
-            // Call predict - returns coroutine
+            // Call run()/predict() - returns coroutine
             let instance = self.instance.bind(py);
-            let coro = instance
-                .call_method("predict", (), Some(&input_dict))
-                .map_err(|e| PredictionError::Failed(format!("Failed to call predict: {}", e)))?;
+            let method_name = match &self.kind {
+                PredictorKind::Class { method_name, .. } => method_name.as_str(),
+                PredictorKind::StandaloneFunction(_) => "predict",
```
michaeldwan: **nit:** hardcoded `"predict"` for standalone functions. The actual dispatch (line 1049) correctly uses `instance.call()`, but a failing standalone train function's error message would say "Failed to call predict".

michaeldwan (approved):
lgtm

PR #3017, opened 2026-05-13
Refactor slot response streaming messages

markphelps:
## Summary
- Rename slot response log and output messages to explicit streaming-oriented variants.
- Add protocol version and binary chunk variants for future transport extensions.
- Include output chunk indexes and update parent/worker routing tests.

## Test Plan
- mise run fmt:rust:fix
- mise run lint:rust
- mise run test:rust

### DIFF
--- a/crates/coglet/src/bridge/protocol.rs
+++ b/crates/coglet/src/bridge/protocol.rs
@@ -294,13 +294,30 @@ pub enum MetricMode {
     Append,
 }
 
+/// Current slot response protocol version.
+///
+/// The response enum is already serde-tagged with `type`, so this constant and
+/// the optional `ProtocolVersion` message give future protocol changes an
+/// explicit marker without adding a second envelope around every message.
+pub const SLOT_RESPONSE_PROTOCOL_VERSION: u32 = 1;
+
 /// Messages from worker to parent on slot socket.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum SlotResponse {
-    Log {
+    ProtocolVersion {
+        version: u32,
+    },
+
+    LogLine {
         source: LogSource,
         data: String,
+    },
+
+    /// Raw binary payload for future WebSocket streaming.
+    BinaryChunk {
+        mime_type: String,
+        data: Vec<u8>,
     },
 
     /// Output for a file/path-like output return type or an output that exceeds the size threshold
@@ -313,9 +330,10 @@ pub enum SlotResponse {
         mime_type: Option<String>,
     },
 
-    /// Streaming output chunk (for generators).
-    Output {
+    /// Streaming output chunk for generator and iterator output.
+    OutputChunk {
         output: serde_json::Value,
+        index: u64,
     },
 
     /// User-emitted metric from the prediction.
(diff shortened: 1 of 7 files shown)


crates/coglet/src/bridge/protocol.rs
```diff
@@ -294,15 +294,32 @@ pub enum MetricMode {
...
+    LogLine {
         source: LogSource,
         data: String,
     },
 
+    /// Raw binary payload for future WebSocket streaming.
+    BinaryChunk {
+        mime_type: String,
+        data: Vec<u8>,
+    },
```
michaeldwan: `Vec<u8>` with serde's default JSON serialization produces an array of integers -- a 1 MB audio chunk becomes ~4 MB of JSON. If this will only ever transit a future binary WebSocket transport (bypassing the JSON codec), a doc comment saying so would be helpful. If it might ever go over the current JSON IPC, consider `serde_bytes` or base64.

crates/coglet/src/bridge/protocol.rs
```diff
@@ -499,20 +517,67 @@ mod tests {
...
             source: LogSource::Stdout,
             data: "Processing...".to_string(),
         };
-        insta::assert_json_snapshot!(resp);
+
+        assert_eq!(
+            serde_json::to_value(resp).unwrap(),
+            json!({
+                "type": "log_line",
+                "source": "stdout",
```
michaeldwan, line 529: The old snapshot files `slot_log_serializes.snap` and `slot_output_serializes.snap` are now orphaned since these tests were renamed and switched to inline assertions. They should be deleted.

crates/coglet/src/orchestrator.rs
```diff
@@ -953,7 +953,17 @@ async fn run_event_loop(
 
             Some((slot_id, result)) = slot_msg_rx.recv() => {
                 match result {
-                    Ok(SlotResponse::Log { source, data }) => {
+                    Ok(SlotResponse::ProtocolVersion { version }) => {
```
michaeldwan, line 956: Nothing in the worker sends `ProtocolVersion` yet -- this arm is dead code for now. A quick comment like `// TODO: worker sends ProtocolVersion on slot connect` (or wherever it'll be sent) would clarify the intent for future readers.

michaeldwan (commented):
Clean refactor. The renames from `Log`/`Output` to `LogLine`/`OutputChunk` are a clear win -- the old names were ambiguous, especially `Output` which clashed with the concept of final prediction output.

A few things to address:

- **Orphaned snapshot files.** The old `slot_log_serializes` and `slot_output_serializes` snapshots are still on disk now that the tests were renamed and switched to inline assertions. Delete them or run `cargo insta test --unreferenced=delete`.
- **`BinaryChunk` encoding.** `Vec<u8>` serializes as a JSON array of integers, which is wildly inefficient for binary payloads. If this is intentionally placeholder for a future binary transport, a doc comment saying so would prevent someone from accidentally using it over the current JSON IPC.
- **`ProtocolVersion` scaffolding.** Nothing sends this message yet. A brief comment on the intent would help future readers.

Minor nit: `send_output` burns an index even when the output spills to disk as `FileOutput` (where index is unused), creating gaps in the sequence. Not a bug since the orchestrator ignores index today, but worth knowing if index continuity ever matters.

michaeldwan (dismissed):
Approving -- the inline comments from my previous review are all minor/nits. Nothing blocking.

PR #3018, opened 2026-05-13
feat: model refs for cog push and weights commands

michaeldwan:
Today cog push targets are baked into config (`image:`) and the weights commands take an `--image` flag — there's no way to point one cog.yaml at different registries across dev/staging/prod without rewriting the file or passing flags everywhere. This PR introduces a `model:` field plus `COG_MODEL*` env vars that drive both `cog push` and the weights subcommands.

A weights-bearing cog.yaml now looks like:

```yaml
model: registry.example.com/team/llama
weights:
  - tag: alpha
    source: weights/alpha
  - tag: beta
    source: weights/beta
```

Resolution layers, lowest to highest precedence: `model:` in cog.yaml, then `COG_MODEL_REGISTRY` / `COG_MODEL_REPO` / `COG_MODEL_TAG` for piecewise override, then `COG_MODEL` for a full-ref override. `cog push` accepts no positional `[IMAGE]` arg in model mode — refs come from config and env, full stop. Pushed bundles carry `cog-image.*` and `cog-weight.*` tag conventions so the structured ref tree printed at the end is reconstructible from registry contents alone:

```
Pushing to registry.example.com/team/llama:20251114T203045Z...
  model     registry.example.com/team/llama@sha256:abc...
  ├─ image   registry.example.com/team/llama:cog-image.20251114T203045Z@sha256:f3c...
  ├─ weight  alpha  registry.example.com/team/llama:cog-weight.alpha.20251114T203045Z@sha256:d2d...
  └─ weight  beta   registry.example.com/team/llama:cog-weight.beta.20251114T203045Z@sha256:e4f...
```

`image:` and `model:` are mutex at config-parse time; `image:` + `weights:` is now also rejected (a weights-bearing model is a bundle, and bundle targets live under `model:`). The weights subcommands (`import`, `pull`, `status`) drop the `--image` flag and route through the same `model.ResolveModelRef` path as push, so a cog.yaml that `cog push` accepts also works for `cog weights *` with no extra flags.

### What's deliberately out of scope

- `Resolver.Build` re-resolves the model ref instead of accepting the already-resolved one from `validatePushArgs`. Latent (today's calls always agree), but a future change that uses validation's `ref` would silently disagree with the push tag. Tracked separately.
- The `image:` field stays in the schema for now; existing models without weights keep working unchanged.

### How to poke at it

```bash
# image-mode (unchanged)
cat > cog.yaml <<EOF
image: registry.example.com/me/old-model
predict: predict.py:Predictor
EOF
cog push

# model-mode + weights
cat > cog.yaml <<EOF
model: registry.example.com/me/bundle-model
predict: predict.py:Predictor
weights:
  - tag: w1
    source: weights/w1
EOF
cog push                                              # uses cog.yaml
COG_MODEL_TAG=v2 cog push                             # pin the tag
COG_MODEL=registry.example.com/me/other:v1 cog push   # override entirely
```

### DIFF
--- a/pkg/cli/weights_manager.go
+++ b/pkg/cli/weights_manager.go
@@ -1,28 +1,48 @@
 package cli
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/replicate/cog/pkg/model"
 	"github.com/replicate/cog/pkg/weights"
 )
 
-// newWeightManager resolves the repo (from imageOverride or
-// src.Config.Image) and delegates construction to weights.NewFromSource.
-// Repo resolution is CLI-input parsing, which is why it lives here and
-// not in pkg/weights.
-func newWeightManager(src *model.Source, imageOverride string) (*weights.Manager, error) {
-	repo := ""
-	if len(src.Config.Weights) > 0 {
-		imageName := imageOverride
-		if imageName == "" {
-			imageName = src.Config.Image
-		}
-		if imageName != "" {
-			parsed, err := parseRepoOnly(imageName)
-			if err != nil {
-				return nil, err
-			}
-			repo = parsed
-		}
+// weightRegistryResolutionHelp is shared help text for weight
+// subcommands so the docs can't drift between import/pull/status.
+const weightRegistryResolutionHelp = "The registry is determined from the resolved model ref ('model' in\n" +
+	"cog.yaml, optionally overridden by COG_MODEL or COG_MODEL_REPO)."
+
+func newWeightManager(src *model.Source) (*weights.Manager, error) {
+	repo, err := resolveWeightRepo(src, configFilename)
+	if err != nil {
+		return nil, err
 	}
 	return weights.NewFromSource(src, repo)
+}
+
+// resolveWeightRepo returns the bundle repository (registry/repo, no
+// tag) for weight operations. Returns ("", nil) for weight-less
+// projects — preserves the no-op Manager contract used by cog predict
+// and cog train. cfgPath is woven into user-facing errors so messages
+// reflect the actual --config flag value, not a hardcoded "cog.yaml".
+//
+// The image-in-cog.yaml branch is defense-in-depth: config validation
+// already rejects 'image:' + 'weights:'; this catches programmatic
+// callers that build a Source from a hand-rolled Config.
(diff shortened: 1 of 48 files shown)


pkg/model/pusher.go
```diff
@@ -30,57 +31,122 @@ type PushOptions struct {
...
+	// Skip the tag+cleanup pair when the build already produced an
+	// image at the cog-image ref (e.g. a caller pre-named their
+	// image): removing it would delete the only local tag and the
+	// underlying image with it.
+	if imgArtifact.Reference != imageRef {
+		if err := p.docker.Tag(ctx, imgArtifact.Reference, imageRef); err != nil {
+			return nil, err
+		}
+		defer func() {
+			if err := p.docker.RemoveImage(context.WithoutCancel(ctx), imageRef); err != nil {
```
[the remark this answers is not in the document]
michaeldwan: The `err` is bound by the `if`-init clause and used inside the block — `console.Debugf("removing local cog-image tag %q: %v", imageRef, err)` reads it on line 125. The suggested `_ = p.docker.RemoveImage(...)` would drop that debug log, so cleanup failures become silent. Leaving as-is.

pkg/cli/weights_manager.go
```diff
@@ -1,28 +1,48 @@
...
+//
+// The image-in-cog.yaml branch is defense-in-depth: config validation
+// already rejects 'image:' + 'weights:'; this catches programmatic
+// callers that build a Source from a hand-rolled Config.
+func resolveWeightRepo(src *model.Source, cfgPath string) (string, error) {
+	if len(src.Config.Weights) == 0 {
+		return "", nil
+	}
+	if src.Config.Image != "" {
+		return "", fmt.Errorf("weight commands require 'model' in %s, not 'image' — rename 'image' to 'model'", cfgPath)
```
markphelps, line 38: we should make this a check in `cog doctor` and also enable auto fixing it with `--fix`

markphelps: can do in a follow up though

markphelps (approved):
seems legit

pkg/cli/weights_manager.go
michaeldwan: oh good call! I love doctor.

PR #3019, opened 2026-05-14
feat: Expose prediction SSE streams

markphelps:
## Summary
- Reuse `POST /predictions` and `PUT /predictions/{id}` for Server-Sent Events by returning an SSE stream when requests send `Accept: text/event-stream`.
- Gate SSE behind the model author's explicit `@streaming` / `@cog.streaming` opt-in while preserving existing iterator JSON behavior for non-SSE requests.
- Preserve existing non-SSE behavior: sync prediction requests still return the blocking JSON response, and `Prefer: respond-async` without SSE still returns `202` JSON.

## Streaming behavior
- Stream prediction lifecycle events: `start`, `output`, `log`, `metric`, and terminal `completed` events.
- Replay retained events for clients that reconnect to an in-flight prediction with `PUT /predictions/{id}` and `Accept: text/event-stream`.
- Bound replay history and live subscribers to avoid unbounded memory growth.
- Add `COG_STREAM_HISTORY_CAPACITY` to configure per-prediction SSE replay history; `0` disables replay while keeping live streaming enabled.
- Return `406 Not Acceptable` for SSE requests when a model has not opted in, and for training endpoints which do not support SSE.

## CLI behavior
- This PR does not add a `cog predict --stream` option or otherwise change the predict CLI.
- Predictions made through the Cog CLI continue to default to the existing non-streaming behavior.

## Docs and tests
- Document how to produce and consume streamed predictions in the HTTP docs.
- Audit and expand `docs/environment.md` for public Cog-specific environment variables, including the new SSE replay history setting.
- Add Rust unit coverage for replay history defaults, configured capacity, invalid values, disabled replay, terminal SSE events, subscriber caps, and concurrent subscriber cancellation semantics.
- Add txtar integration coverage for configurable stream history, disabled replay, and custom metrics over SSE.


michaeldwan (changes requested):
Solid design. Main point discussed offline: streaming needs to be opt-in at the handler level via a decorator/attribute. Five concrete changes in the comments.

crates/coglet/src/prediction.rs
```diff
@@ -105,6 +180,31 @@ impl Prediction {
...
+        }
+    }
+
+    pub fn stream_receiver_count(&self) -> usize {
+        self.stream_tx.receiver_count()
+    }
+
+    fn emit_stream_event(&mut self, event: PredictionStreamEvent) {
+        if self.stream_history.len() == MAX_STREAM_HISTORY_EVENTS {
+            self.stream_history.remove(0);
```
michaeldwan: `Vec::remove(0)` is O(n) on every event after the buffer fills -- thousands of shifts per prediction for token-by-token LLM output. `VecDeque::pop_front()` is O(1).

crates/coglet/src/transport/http/routes.rs
```diff
@@ -376,7 +426,12 @@ async fn create_prediction_with_id(
 
     // Submit prediction: creates Prediction, acquires slot, registers in service
     let (handle, unregistered_slot) = match service
-        .submit_prediction(prediction_id.clone(), input.clone(), webhook_sender)
+        .submit_prediction(
+            prediction_id.clone(),
+            input.clone(),
+            webhook_sender,
+            response_mode != PredictionResponseMode::AsyncJson,
```
michaeldwan: `response_mode != PredictionResponseMode::AsyncJson` is true for `SyncJson` too. Should be `response_mode == PredictionResponseMode::AsyncSse` -- the current expression is inert for sync mode but reads as a bug.

crates/coglet/src/prediction.rs
```diff
@@ -76,12 +143,17 @@ pub struct Prediction {
...
+    stream_tx: tokio::sync::broadcast::Sender<PredictionStreamEvent>,
+    stream_history: Vec<PredictionStreamEvent>,
+    stream_history_skipped: u64,
     /// User-emitted metrics. Merged with system metrics (predict_time) in terminal response.
     metrics: HashMap<String, serde_json::Value>,
 }
 
 impl Prediction {
     pub fn new(id: String, webhook: Option<WebhookSender>) -> Self {
+        let (stream_tx, _) = tokio::sync::broadcast::channel(1024);
```
michaeldwan: Channel capacity and history cap are both 1024 but defined independently. Use a shared constant so they can't drift.

crates/coglet/src/service.rs
```diff
@@ -106,6 +107,51 @@ impl PredictionHandle {
...
+    cancel_on_stream_drop: bool,
+}
+
+impl Drop for PredictionStreamGuard {
+    fn drop(&mut self) {
+        if !self.cancel_on_stream_drop {
+            return;
+        }
+
+        if self.service.stream_receiver_count(&self.id) == 0
```
michaeldwan, line 167: The async cleanup task calls `remove_prediction` while the SSE client may still be draining. After removal, `stream_receiver_count` returns `unwrap_or(0)` and `prediction_is_terminal` returns `unwrap_or(true)` -- which happen to be the safe defaults. That invariant needs a comment, or hold an `Arc` to the entry in the guard so it doesn't depend on post-removal lookup.

integration-tests/tests/sse_streaming_output.txtar
```diff
@@ -0,0 +1,32 @@
+# Test that async generator output is available when predictions are created with SSE accept.
+
+[short] skip 'requires Docker build'
+
+cog serve --upload-url http://unused/
+
+curl -H Accept:text/event-stream PUT /predictions/sse-stream-test '{"id":"sse-stream-test","input":{}}'
+stdout 'event: output'
```
michaeldwan, line 9: Missing `stdout 'event: start'` -- should verify the full lifecycle.

markphelps:
@michaeldwan decorator added in https://github.com/replicate/cog/pull/3019/changes/8c9c9826e8a21628fe4200da23e96e63dfb0ac46

michaeldwan (approved):
lgtm

examples/streaming-text/predict.py
```diff
@@ -0,0 +1,64 @@
...
+        prompt: str = Input(description="Prompt to complete"),
+        max_new_tokens: int = Input(
+            description="Maximum number of tokens to generate",
+            default=128,
+            ge=1,
+            le=512,
+        ),
+    ) -> Iterator[str]:
+        messages = [{"role": "user", "content": prompt}]
+        text = self.tokenizer.apply_chat_template(
```
michaeldwan, line 36: **Blocker:** This method is missing `@cog.streaming`. The README (line 28-31) tells users to curl with `Accept: text/event-stream`, which will return 406 since the model doesn't opt in.

The integration test `sse_streaming_output.txtar` correctly uses `@streaming`, but this shipped example doesn't.

```python
from cog import BasePredictor, Input, streaming

    # ...
    @streaming
    def predict(self, ...) -> Iterator[str]:
```

crates/coglet/src/prediction.rs
```diff
@@ -105,6 +180,31 @@ impl Prediction {
...
+    pub fn stream_receiver_count(&self) -> usize {
+        self.stream_tx.receiver_count()
+    }
+
+    fn emit_stream_event(&mut self, event: PredictionStreamEvent) {
+        if self.stream_history.len() == STREAM_EVENT_BUFFER_CAPACITY {
+            self.stream_history.pop_front();
+            self.stream_history_skipped += 1;
+        }
+        self.stream_history.push_back(event.clone());
```
michaeldwan: **Should-fix:** This clones the event (containing `serde_json::Value`) for history, then moves the original into `broadcast::send`. For high-throughput models yielding many chunks, this deep-clones arbitrarily large JSON on every output.

Consider `Arc<PredictionStreamEvent>` for the broadcast channel type -- history and broadcast share the same allocation, and `subscribe_stream_replay()` becomes 1024 atomic increments instead of 1024 deep JSON clones under the prediction mutex.

```rust
stream_tx: broadcast::Sender<Arc<PredictionStreamEvent>>,
stream_history: VecDeque<Arc<PredictionStreamEvent>>,

fn emit_stream_event(&mut self, event: PredictionStreamEvent) {
    // ...
    let event = Arc::new(event);
    self.stream_history.push_back(Arc::clone(&event));
    let _ = self.stream_tx.send(event);
}
```

crates/coglet/src/orchestrator.rs
```diff
@@ -923,17 +925,19 @@ async fn run_event_loop(
                         }
                     }
                     None => {
-                        tracing::debug!(%prediction_id, "Cancel requested for unknown prediction (may have already completed)");
+                        tracing::debug!(%prediction_id, "Cancel requested for unknown prediction; storing pending cancellation");
+                        pending_cancellations.insert(prediction_id);
```
michaeldwan: **Should-fix:** If the cancel arrives after the prediction has already completed and been removed from `predictions`, the ID is stored here and never consumed. In a long-running server with many cancelled predictions, this is an unbounded leak.

Consider adding a size cap (e.g., 1000 entries) or a TTL, and log a warning when it's exceeded.

pkg/schema/python/parser.go
```diff
@@ -651,6 +657,39 @@ func UnwrapFunction(node *sitter.Node) *sitter.Node {
...
+			return true
+		}
+	}
+	return false
+}
+
+func decoratorIsCogStreaming(node *sitter.Node, source []byte, imports *schema.ImportContext) bool {
+	for _, child := range NamedChildren(node) {
+		switch child.Type() {
+		case "attribute":
```
michaeldwan, line 703: **Should-fix:** Hard-coded string match. `import cog as c` then `@c.streaming` won't be detected -- the content will be `"c.streaming"`, not `"cog.streaming"`.

The rest of the parser resolves aliases through `ImportContext` (e.g., `IsBaseModel`, `IsOpaque`). This should do the same:

```go
case "attribute":
    text := Content(child, source)
    parts := strings.SplitN(text, ".", 2)
    if len(parts) != 2 || parts[1] != "streaming" {
        return false
    }
    entry, ok := imports.Names.Get(parts[0])
    return ok && entry.Module == "cog" && entry.Original == "cog"
```

pkg/schema/python/parser.go
```diff
@@ -651,6 +657,39 @@ func UnwrapFunction(node *sitter.Node) *sitter.Node {
...
+			return Content(child, source) == "cog.streaming"
+		case "identifier":
+			if Content(child, source) != "streaming" {
+				return false
+			}
+			entry, ok := imports.Names.Get("streaming")
+			return ok && entry.Module == "cog" && entry.Original == "streaming"
+		case "call":
+			return false
```
michaeldwan: **Should-fix:** This rejects `@cog.streaming()` (call form), but the Python decorator works fine with either `@streaming` or `@streaming()`. A user who writes `@cog.streaming()` gets a model that builds, runs, and yields output -- but SSE returns 406 with no hint about the parentheses.

Either support the call form here (check if the callee is `cog.streaming` or imported `streaming`), or make the Python decorator raise a clear error when called with parens. The current behavior is a silent gotcha.

crates/coglet/src/service.rs
```diff
@@ -469,6 +539,46 @@ impl PredictionService {
         Some(response)
     }
 
+    pub fn subscribe_prediction_stream(
+        self: &Arc<Self>,
+        id: &str,
+    ) -> Option<PredictionStreamSubscription> {
+        let entry = self.predictions.get(id)?;
```
michaeldwan: **Should-fix:** No cap on subscriber count. The idempotent PUT endpoint allows repeated SSE connections to the same prediction ID, each creating a new broadcast receiver. An attacker opening many connections forces the sender to retain events for slow consumers.

Consider:
```rust
let pred = entry.prediction.lock().ok()?;
if pred.stream_receiver_count() >= MAX_STREAM_SUBSCRIBERS {
    return None;
}
```

crates/coglet/src/service.rs
```diff
@@ -106,6 +107,54 @@ impl PredictionHandle {
...
+impl Drop for PredictionStreamGuard {
+    fn drop(&mut self) {
+        if !self.cancel_on_stream_drop {
+            return;
+        }
+
+        // Prediction cleanup may remove the service entry before the SSE response
+        // finishes draining. Missing entries deliberately report zero receivers and
+        // terminal state so this guard cannot cancel an already-cleaned prediction.
+        if self.service.stream_receiver_count(&self.id) == 0
```
michaeldwan, line 167: **Should-fix:** `cancel()` internally calls `tokio::spawn`. If this guard is dropped outside a tokio runtime context (panic unwinding, runtime shutdown), `tokio::spawn` panics.

Safer:
```rust
if let Ok(handle) = tokio::runtime::Handle::try_current() {
    let service = Arc::clone(&self.service);
    let id = self.id.clone();
    handle.spawn(async move { service.cancel(&id); });
}
```

Or refactor `cancel()` itself to use `Handle::try_current()` before spawning.

python/cog/__init__.py
```diff
@@ -133,6 +143,7 @@ def current_scope() -> object:
     "CancelationException",
     # Metrics
     "current_scope",
+    "streaming",
```
michaeldwan, line 163: **Nit:** `streaming` is a decorator, not a metric. Move it to its own `# Decorators` section or under `# Core classes`.

python/cog/__init__.py
```diff
@@ -38,6 +40,14 @@ def predict(
     URLPath,
 )
 
+F = TypeVar("F", bound=Callable[..., object])
```
michaeldwan: **Nit:** This is importable as `from cog import F`. Prefix with underscore (`_F`) to signal it's internal.

crates/coglet/src/orchestrator.rs
```diff
@@ -353,15 +353,18 @@ pub struct OrchestratorReady {
     pub setup_logs: String,
 }
 
+type RegisterPredictionMessage = (
+    SlotId,
```
michaeldwan: **Nit:** A 4-element tuple type alias is opaque -- the ack channel's purpose is invisible at usage sites. A named struct would be clearer:

```rust
struct RegisterPredictionMessage {
    slot_id: SlotId,
    prediction: Arc<StdMutex<Prediction>>,
    idle_sender: oneshot::Sender<SlotIdleToken>,
    registered_ack: oneshot::Sender<()>,
}
```

michaeldwan (changes requested):
## Summary

Solid design. The broadcast channel + bounded replay architecture is sound, the opt-in gating via `@cog.streaming` is the right call, and the integration between Go parser → OpenAPI extension → coglet runtime is clean. The replay/live transition has no race (mutex held atomically), the bounded history buffer has no off-by-one, and `completed` events are emitted on all terminal paths.

One blocker, several should-fixes. Details below and inline.

## Blockers

- **`examples/streaming-text/predict.py` doesn't use `@cog.streaming`** -- the README tells users to curl with `Accept: text/event-stream`, which will 406. The integration test `sse_streaming_output.txtar` correctly uses `@streaming`, but the shipped example doesn't. (See inline comment.)

- **No test for SSE events during failed or cancelled predictions.** `set_failed()` and `set_canceled()` both emit `Completed` events, but no test verifies the SSE stream delivers `event: completed` with `"status":"failed"` / `"status":"canceled"`. These are critical user-facing paths.

## Should-fix

- **`decoratorIsCogStreaming` hard-codes `"cog.streaming"`** instead of resolving through `ImportContext`. `import cog as c` → `@c.streaming` won't be detected. The rest of the parser handles aliases. (See inline comment.)
- **`@cog.streaming()` with parens silently degrades.** Parser rejects call form, but Python decorator works either way. User gets a working model where SSE returns 406 with no hint about the parens.
- **No limit on SSE subscriber count per prediction.** `subscribe_prediction_stream()` creates a new broadcast receiver with no cap. Repeated SSE connections to the same prediction ID amplify memory pressure.
- **Orphaned `pending_cancellations` leak memory.** Cancel messages arriving after a prediction completes get stored in the `HashSet` and never cleaned up. (See inline comment.)
- **Double-clone on every stream event.** `emit_stream_event` clones `serde_json::Value` for history storage. `Arc<PredictionStreamEvent>` would eliminate deep clones -- history and broadcast share the same allocation. Also fixes the O(n) deep-clone in `subscribe_stream_replay()` under the mutex. (See inline comment.)
- **`PredictionStreamGuard::Drop` calls `tokio::spawn` via `cancel()`.** If dropped outside a tokio runtime context, `tokio::spawn` panics. Use `Handle::try_current()`.
- **No test for concurrent SSE subscribers.** The guard checks `stream_receiver_count() == 0` before cancelling, but no test verifies dropping one of two subscribers doesn't cancel.
- **Training endpoints silently ignore `Accept: text/event-stream`.** Returning 406 or documenting would be more honest than silent fallback to JSON.

## Nits

- `RegisterPredictionMessage` 4-element tuple → named struct.
- `streaming` listed under `# Metrics` in `__all__` -- it's a decorator.
- Module-level `F` TypeVar → `_F` to signal internal.
- `replay.into()` creates unnecessary `VecDeque` from `Vec`.
- `id.to_string()` allocated twice in `subscribe_prediction_stream`.
- Missing `require.NotNil` guard before type assertions in streaming OpenAPI tests.
- Broadcast channel capacity and history buffer both use the same 1024 constant by coincidence -- give them separate named constants.

## Verified correct

- Replay + live transition (no race -- mutex held atomically during subscribe + snapshot)
- Bounded history buffer (no off-by-one)
- `completed` event emitted on all terminal paths
- Terminal state guards prevent double-completion
- `findTargetFunction` returning `decorated_definition` handled correctly via `UnwrapFunction`
- `cog predict` CLI works fine with streaming models (uses sync JSON path)
- `docs/python.md` and `docs/llms.txt` are accurate and in sync

markphelps:

bghira:
i'm interested in working on WebSocket support after this lands, so that we can have a slightly-more-efficient means to stream responses in and out of the container (with not much idea of the limits of this until i actually begin it).

pkg/schema/python/parser.go
```diff
@@ -1450,7 +1514,7 @@ func findMethodInClass(classNode *sitter.Node, source []byte, className, methodN
 		}
 		nameNode := funcNode.ChildByFieldName("name")
 		if nameNode != nil && Content(nameNode, source) == methodName {
-			return funcNode, nil
+			return child, nil
```
michaeldwan: **Blocker:** Returning `child` here breaks decorated `train()` methods in classes. `ParsePredictor` (line 80) takes `target.node` directly without unwrapping, then line 88 calls `funcNode.ChildByFieldName("parameters")` — that field exists on `function_definition`, not `decorated_definition`, so any decorated train method fails to parse.

Reproduced locally:
```python
class Trainer(BasePredictor):
    @functools.wraps(lambda x: x)
    def train(self, n: int) -> Path: ...
```
→ `ErrParse: function has no parameters node`.

`functionSupportsStreaming` already handles a `function_definition` argument by walking `node.Parent()` (parser.go:675-681), so returning `funcNode` here is correct and streaming detection still works.

Fix:
```go
return funcNode, nil
```

Plus a regression test for a decorated `train` method — that gap is why CI missed this.

crates/coglet/src/transport/http/routes.rs
```diff
@@ -397,14 +485,21 @@ async fn create_prediction_with_id(
...
 
     // Async mode: spawn background task, return immediately
-    if respond_async {
+    if response_mode != PredictionResponseMode::SyncJson {
+        let sse_subscription = if response_mode == PredictionResponseMode::AsyncSse {
+            Some(service.subscribe_prediction_stream(&prediction_id))
+        } else {
+            None
+        };
+
```
michaeldwan, line 508: **Should-fix:** The predict task is spawned before checking whether `subscribe_prediction_stream` returned `Ok`. If subscription fails (`TooManySubscribers` on burst, `NotFound` on a cleanup race), the prediction still runs to completion, fires webhooks, consumes a slot, and is cleaned up with no client watching.

Either subscribe before spawn and gate the spawn on `Ok`, or cancel-and-await on subscription failure.

crates/coglet/src/service.rs
```diff
@@ -106,6 +120,54 @@ impl PredictionHandle {
     }
 }
 
+pub struct PredictionStreamSubscription {
+    id: String,
+    replay: std::collections::VecDeque<SharedPredictionStreamEvent>,
+    skipped: u64,
+    receiver: tokio::sync::broadcast::Receiver<SharedPredictionStreamEvent>,
+    guard: PredictionStreamGuard,
+}
```
michaeldwan, line 133: **Should-fix:** Drop order is load-bearing here: `receiver` must drop before `guard`, otherwise `stream_receiver_count(&id)` in the guard's Drop still sees this receiver (≥1) and `cancel_on_stream_drop` never fires. A future field reorder silently breaks SSE cancel-on-disconnect.

The `dropping_one_of_two_sync_stream_subscriptions` test would catch it, but a comment is what protects refactors that don't run the full suite:

```rust
// IMPORTANT: drop order matters — receiver must drop before guard so
// stream_receiver_count() returns 0 by the time the guard runs cleanup.
```

Same comment belongs on `StreamState` in `routes.rs:681-687`.

pkg/schema/openapi_test.go
```diff
@@ -657,6 +657,38 @@ func TestOutputConcatenateIterator(t *testing.T) {
...
+		Mode:              ModePredict,
+		SupportsStreaming: true,
+	}
+
+	spec := parseSpec(t, info)
+	postPath := getPath(spec, "paths", "/predictions", "post")
+	require.NotNil(t, postPath)
+	post := postPath.(map[string]any)
+	assert.Equal(t, true, post["x-cog-streaming"])
+}
```
michaeldwan, line 674: **Should-fix:** `openapi.go:220` guards `x-cog-streaming` with `!isTrain && info.SupportsStreaming`, but no test exercises `isTrain=true && SupportsStreaming=true` to confirm the extension is omitted on `/trainings`. Easy to regress if someone refactors the guard. Mirror this test with `Mode: ModeTrain` and assert the extension is absent.

pkg/schema/python/parser.go
```diff
@@ -670,6 +671,69 @@ func UnwrapFunction(node *sitter.Node) *sitter.Node {
 	return nil
 }
 
+func functionSupportsStreaming(node *sitter.Node, source []byte, imports *schema.ImportContext) bool {
```
michaeldwan, line 677: **Should-fix:** `functionSupportsStreaming` only inspects the decorator, never the return type. A user who applies `@streaming` to `def predict(self) -> str` gets a schema claiming streaming support; SSE clients receive a single chunk + `completed` and conclude the model is broken.

Enforce here: require an `Iterator[...]` / `AsyncIterator[...]` / `*ConcatenateIterator[...]` return annotation before reporting `SupportsStreaming = true`. The return type is already parsed at line 108-115 — this is the natural enforcement point. Document the constraint in `docs/python.md`.

examples/streaming-text/predict.py
```diff
@@ -0,0 +1,65 @@
...
+
+import torch
+from transformers import AutoModelForCausalLM, AutoTokenizer, TextIteratorStreamer
+
+from cog import BasePredictor, Input, streaming
+
+MODEL_NAME = "HuggingFaceTB/SmolLM2-135M-Instruct"
+
+
+class Predictor(BasePredictor):
```
michaeldwan: **Should-fix:** Uses deprecated `BasePredictor` / `predict()`. `docs/python.md:52,76` explicitly directs new code to `BaseRunner` / `run()` ("`BasePredictor`, `Predictor`, and `predict()` still work for existing models, but they are deprecated. Cog warns when it loads or inspects those legacy names."). A brand-new example landing right alongside that guidance should use the new API — silently shipping the deprecated path is the worst of both worlds for someone copy-pasting from `examples/`.

Either flip to `BaseRunner` / `run()`, or note in the README that the example is exercising the legacy path on purpose.

python/cog/__init__.py
```diff
@@ -38,6 +40,28 @@ def run(
...
+
+@overload
+def streaming(fn: None = None) -> Callable[[_F], _F]: ...
+
+
+def streaming(fn: _F | None = None) -> _F | Callable[[_F], _F]:
+    """Mark a predict handler as supporting streaming responses."""
+
+    def decorate(inner: _F) -> _F:
+        inner.__cog_streaming__ = True  # type: ignore[attr-defined]
```
michaeldwan: **Should-fix:** `inner.__cog_streaming__ = True` is dead code. SSE eligibility is decided entirely by the Go AST parser at build time (`pkg/schema/python/parser.go:674` → `pkg/schema/openapi.go:221`); nothing in Python, coglet Rust, or coglet-python reads `__cog_streaming__`. The two `test_types.py` assertions are testing an implementation detail with no observable effect.

Pick one:
- (a) Drop the attribute and let `@streaming` be a true identity decorator; update tests to assert only `decorated is predict`.
- (b) Wire the attribute into the schema generation path so dynamic decoration (`predict = streaming(predict)`) also works — the marker becomes the source of truth instead of duplicating intent across Go AST + Python attribute.

(a) is the lower-risk fix unless dynamic decoration is a stated requirement.

crates/coglet/src/service.rs
```diff
@@ -616,15 +768,7 @@ impl PredictionService {
...
-                tokio::spawn(async move {
-                    if let Err(e) = orch.cancel_by_prediction_id(&id_owned).await {
-                        tracing::error!(
-                            prediction_id = %id_owned,
-                            error = %e,
-                            "Failed to send cancel to orchestrator"
-                        );
-                    }
-                });
+                spawn_orchestrator_cancel(orch, id_owned);
```
michaeldwan, line 776: **Nit:** If `self.orchestrator.try_read()` can't acquire the lock, the worker-side cancel is silently dropped — the `CancellationToken` fires but the subprocess never gets `ControlRequest::Cancel`. Today fine (only one `.write()` at startup), but a `tracing::warn!(prediction_id = %id, "Skipped worker cancel: orchestrator lock unavailable")` would make any future regression diagnosable.

crates/coglet/src/service.rs
```diff
@@ -469,6 +552,59 @@ impl PredictionService {
...
+    ) -> Result<PredictionStreamSubscription, SubscribePredictionStreamError> {
+        let entry = self
+            .predictions
+            .get(id)
+            .ok_or(SubscribePredictionStreamError::NotFound)?;
+        let stream = {
+            let prediction = entry
+                .prediction
+                .lock()
+                .map_err(|_| SubscribePredictionStreamError::NotFound)?;
```
michaeldwan: **Nit:** Poisoned mutex maps to `SubscribePredictionStreamError::NotFound` → returns 404. A 404 for a poisoned mutex is misleading at the diagnostic level. Either use `try_lock_prediction` (which attempts recovery and fails the prediction explicitly) or add a dedicated error variant.

pkg/schema/python/parser.go
```diff
@@ -670,6 +671,69 @@ func UnwrapFunction(node *sitter.Node) *sitter.Node {
...
+	entry, ok := imports.Names.Get(parts[0])
+	return ok && entry.Module == "cog" && entry.Original == "cog"
+}
+
+func identifierIsCogStreaming(node *sitter.Node, source []byte, imports *schema.ImportContext) bool {
+	if Content(node, source) != "streaming" {
+		return false
+	}
+	entry, ok := imports.Names.Get("streaming")
+	return ok && entry.Module == "cog" && entry.Original == "streaming"
```
michaeldwan, line 734: **Nit:** `from cog import streaming as stream` then `@stream` is not detected — this rejects anything where the local identifier isn't literally `"streaming"`. The attribute form (line 720-727) already handles aliasing via `entry.Original == "cog"`, so the asymmetry is mildly surprising.

One-line fix mirrors the attribute form:
```go
func identifierIsCogStreaming(node *sitter.Node, source []byte, imports *schema.ImportContext) bool {
    entry, ok := imports.Names.Get(Content(node, source))
    return ok && entry.Module == "cog" && entry.Original == "streaming"
}
```

examples/streaming-text/predict.py
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        thread = Thread(target=self.model.generate, kwargs=generation_kwargs)
+        thread.start()
+
+        for chunk in streamer:
+            if chunk:
+                yield chunk
+
+        thread.join()
```
michaeldwan: **Nit:** `thread.join()` is unreachable if the consumer of the `Iterator[str]` (the coglet worker) stops iterating early — e.g. prediction cancellation. The `for chunk in streamer:` loop exits via GeneratorExit, the join is skipped, and the generation thread keeps running until `max_new_tokens` completes, holding GPU memory.

Best practice for an example users will copy:
```python
try:
    for chunk in streamer:
        if chunk:
            yield chunk
finally:
    thread.join()
```
(or a comment noting the limitation.)

docs/python.md
```diff
@@ -263,11 +263,14 @@ Cog models can stream output as the `run()` method is running. For example, a la
 
 To support streaming output in your Cog model, add `from typing import Iterator` to your `run.py` file. The `typing` package is a part of Python's standard library so it doesn't need to be installed. Then add a return type annotation to the `run()` method in the form `-> Iterator[<type>]` where `<type>` can be one of `str`, `int`, `float`, `bool`, or `cog.Path`.
 
+To allow clients to receive chunks as server-sent events with `Accept: text/event-stream`, decorate the prediction method (`run()` or `predict()`) with `@cog.streaming`, `@cog.streaming()`, `@streaming`, or `@streaming()` imported from `cog`. Without the decorator, iterator outputs still work in normal JSON responses, but SSE requests return `406 Not Acceptable`.
```
michaeldwan: **Nit:** Listing all four spellings (`@cog.streaming`, `@cog.streaming()`, `@streaming`, `@streaming()`) reads like the user has a meaningful choice between four things. Consider: "decorate with `@cog.streaming` (or `@streaming` if imported directly); the parenthesized form `@streaming()` is also accepted." Same change in `docs/llms.txt:1982`.

michaeldwan (changes requested):
Previous review feedback is fully addressed — every blocker and should-fix from 2026-05-19 is in with regression tests. The architecture (broadcast + bounded replay + `Arc<PredictionStreamEvent>` + receiver-before-guard drop order) holds up under a fresh pass.

One regression to fix before merge: a decorated `train()` method in a class now fails to parse (inline). The should-fixes below are real but none are correctness-critical for the SSE happy path.

## Blocker
- **`findMethodInClass` returns `decorated_definition`** — breaks decorated `train` methods. Reproduced locally with `ErrParse: function has no parameters node`. Inline.

## Should-fix
- SSE subscription failure leaves prediction running unobserved (inline on `routes.rs`)
- Drop-order invariant in `StreamState` / `PredictionStreamSubscription` is load-bearing but undocumented (inline)
- No test for train mode + `SupportsStreaming=true` — `openapi.go:220` guards with `!isTrain && SupportsStreaming`, no coverage for the omit path (inline on `openapi_test.go`)
- `@streaming` on a non-iterator return type silently produces broken SSE — the parser only inspects the decorator, never the return annotation (inline on `parser.go`)
- `examples/streaming-text/predict.py` uses deprecated `BasePredictor` / `predict()` even though `docs/python.md` directs new code to `BaseRunner` / `run()` (inline)
- `__cog_streaming__` attribute is dead code — nothing reads it, but two unit tests assert it (inline)

## Nits
- `PredictionService::cancel` silently drops worker-side cancel on `try_read()` contention (inline)
- `subscribe_prediction_stream` maps poisoned mutex to `NotFound` → misleading 404 (inline)
- Aliased `from cog import streaming as stream` then `@stream` is not detected; attribute form already handles aliasing — asymmetric (inline)
- Example's `thread.join()` is unreachable on early consumer exit → GPU memory leak in the copy-paste case (inline)
- Docs list four spellings as if the user has a meaningful choice (inline on `docs/python.md`)

markphelps:

markphelps:
please @michaeldwan

crates/coglet/src/service.rs
```diff
@@ -50,6 +53,18 @@ pub enum CreatePredictionError {
     AtCapacity,
 }
 
+const MAX_STREAM_SUBSCRIBERS: usize = STREAM_CHANNEL_CAPACITY;
```
michaeldwan, line 56: just calling out that these both share the same value (1024) but seem intended for different needs. max subscribers is the client limit, and max capacity is the queue size. these might diverge someday, but 1024 is probably high enough for a cog model anyway, so nbd.

michaeldwan (approved):
Awesome. looks great. thanks for iterating on the feedback!

two things worth calling out, neither are blocking. good to merge.

- this was being built midway through the predictor to runner refactor and there's some stale references, like the example using predict.py. NBD we need to take a pass over the cog code, examples, and docs anyway, so it'll get sorted in the end.
- I left a comment in service.rs about 2 related but distinct constants sharing a value, probably fine but wanted to point out incase you intended otherwise and an agent just merged them because "same value duplication bad"

LGTM

PR #3020, opened 2026-05-19
fix: when pushing a model with a version tag, emit URL without the tag

bghira:
When I'm pushing my simpletuner/z-image cogs with the version tag supplied, the push works using the correct local cached artifacts. But, the URL that it gives me would suffix with the tag `:latest`, causing a 404 unless I manually remove that every time.

This uses a small helper to remove the tag from the URL, making it more reliable.


markphelps (approved):
thank you @bghira !

PR #3027, opened 2026-05-20
Refactor static schema parser into phase pipeline

markphelps:
## Summary

This refactors the static Python schema parser from one large implementation file into an explicit phase pipeline with smaller, responsibility-focused parser components.

The branch also updates target resolution so `run()` is the primary prediction entry point, while preserving legacy `predict()` fallback behaviour where it is still allowed. Standalone function targets continue to resolve correctly, and class targets now share the same run-first semantics across local, inherited, and imported cases.

## What changed

- Split the Python parser into focused files for annotations, imports, inputs, literals, model discovery, module scope, node helpers, parser state/phases, and target selection.
- Added a `ParseState`/`parserPhase` pipeline so parser phases run in a clear order and fail early if called out of sequence.
- The branch parses the primary source once and reuses that syntax tree across explicit parser phases. It also improves imported-module caching for model and target resolution, but it does not fully unify all imported-file parsing behind one global syntax-tree cache.
- Made `run()` the preferred prediction method for class targets, with deprecated `predict()` fallback support and an option to disable that fallback.
- Preserved standalone target behaviour so explicit function targets and default standalone `run` resolution still work.
- Improved cross-file handling by caching loaded Python module contexts and resolving inherited targets/models across imports more consistently.
- Extended schema representation and generation paths needed by the parser refactor, including richer object/dict/list handling.
- Updated schema parser architecture docs to describe the phase pipeline and target-resolution rules.
- Cleaned up review feedback by removing unused parser state fields.
- Added focused unit coverage for parser subcomponents plus additional end-to-end parser pipeline tests.
- Expanded fuzz coverage beyond full-parser fuzzing to exercise import/module-scope/input-registry collection and default-expression resolution.

## Why this is useful

The old parser had grown into a large, tightly coupled file where target selection, import resolution, input parsing, model discovery, literal parsing, and output resolution were hard to reason about independently. That made the parser difficult to review and risky to change.

This refactor makes the parser easier to maintain by giving each parsing concern a smaller boundary and making the full parse flow explicit. The phase checks help catch accidental ordering bugs, and the new focused tests make regressions easier to diagnose than failures in one large end-to-end test suite.

The `run()`-first behaviour also aligns the static schema parser with the newer predictor interface while retaining legacy compatibility for existing `predict()`-based predictors.

## Testing

- `go test ./pkg/schema/python`
- `go test ./pkg/schema/...`
- `mise run lint:go`
- `mise run test:go`
- Short fuzz verification runs for:
  - `FuzzParsePredictor`
  - `FuzzParseTypeAnnotation`
  - `FuzzCollectImportsAndModuleScope`
  - `FuzzParseDefaultExpression`


pkg/schema/python/models.go
```diff
@@ -0,0 +1,544 @@
...
+			if Content(nameNode, source) == "total" && strings.TrimSpace(Content(valueNode, source)) == "False" {
+				requiredByDefault = false
+			}
+		}
+	}
+
+	return isTypedDict, requiredByDefault, parents
+}
+
+func extractClassAnnotations(classNode *sitter.Node, source []byte, imports *schema.ImportContext, isTypedDict bool, requiredByDefault bool) []schema.ModelField {
```
markphelps, line 438: this argument signature is repeated a few times. id like to extract types wherever possible to conslidate args instead of passing in 5-6 args to each func

markphelps:

PR #3031, opened 2026-05-25
fix: show doctor remediation for unfixed findings

immanuwell:
Small paper cut fix.

`cog doctor --fix` could print `(no auto-fix available)` even when the finding already had a remediation message. So the command knew what to say, it just wasnt showing it.

Repro:
1. Make a `cog.yaml` with:
   ```yaml
   build:
     python_version: "3.12"
   ```
2. Do not create `predict.py`
3. Run `cog doctor --fix`
4. Before this patch, `Predict reference` ended with `(no auto-fix available)`
5. Now it prints the actual remediation, like `Create predict.py or update the predict field in cog.yaml`

Also adds a unit test and a small integration test, nice and tidy.

### DIFF
--- a/integration-tests/tests/doctor_fix_shows_remediation.txtar
+++ b/integration-tests/tests/doctor_fix_shows_remediation.txtar
@@ -1,0 +1,11 @@
+# Test that cog doctor --fix prints the remediation text for unfixed findings.
+
+! cog doctor --fix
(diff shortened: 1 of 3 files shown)


markphelps (approved):
looks great! thank you for the contribution @immanuwell !!

PR #3033, opened 2026-05-28
Send layer media type for Cog weight uploads

markphelps:
## Summary
- Add an opt-in `OCI-Layer-Media-Type` finalize header path to registry layer uploads.
- Send the header only for Cog managed weight layers, preserving image-layer size enforcement.
- Cover weight header propagation, image non-propagation, and commit request header behavior in tests.

## Test Plan
- [x] `go test -count=1 ./pkg/model ./pkg/registry`


meatballhat-cf (approved):
🎬

PR #3034, opened 2026-05-29
fix: handle PEP 563 string annotations in predictor validator

immanuwell:
Two bugs in `_validate_setup` / `_validate_run_or_predict` when the predictor file uses `from __future__ import annotations` (PEP 563).

With PEP 563, all annotations are stored as strings at compile time. So `-> None` becomes the string `"None"`, not the `None` value.

Bug 1 - false positive that blocks valid code: `_validate_setup` checks `spec.annotations.get("return") is not None`, which is `True` for `"None"`, so a correctly annotated `setup(self) -> None` raises `ValueError: setup() must return None`.

Bug 2 - false negative: `_validate_run_or_predict` checks `is None`, which is `False` for `"None"`, so `run(self, ...) -> None` incorrectly passes validation instead of being rejected.

**Reproduce bug 1:**

```python
# predict.py
from __future__ import annotations
from cog import BasePredictor

    def setup(self) -> None:   # <- valid, but cog crashes here
        pass
    def predict(self, text: str) -> str:
        return text
```

Running `cog build` against this gives `ValueError: setup() must return None`.

Fix: compare against both `None` and the string `"None"` when checking the return annotation, consistent with PEP 563 behavior.

### DIFF
--- a/python/cog/_inspector.py
+++ b/python/cog/_inspector.py
@@ -70,7 +70,10 @@ def _validate_setup(f: Callable[..., Any]) -> None:
         raise ValueError("setup() must not have keyword-only args")
     if spec.kwonlydefaults:
         raise ValueError("setup() must not have keyword-only defaults")
-    if spec.annotations.get("return") is not None:
+    # With `from __future__ import annotations` (PEP 563), `-> None` is stored as
+    # the string "None" rather than the None value, so check both forms.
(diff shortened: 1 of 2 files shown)


markphelps (approved):
lgtm! thank you @immanuwell !!

Issue #1344, opened 2023-10-20
Does cog support multiple Path input? e.g. List[Path] as input?


peter65374:
```
  File "/root/.pyenv/versions/3.9.18/lib/python3.9/site-packages/cog/predictor.py", line 239, in get_input_type
    raise TypeError(
TypeError: Unsupported input type list for parameter `model_faces`. Supported input types are: str, int, float, bool, cog.File, cog.Path.

ⅹ Failed to get type signature: exit status 1
```
I got error as above when I try to build the below Predictor(...)
```
def predict(
        self,
        target_image: Path = Input(
            description="Faces in target image would be changed.",
        ),
        model_faces: list[Path] = Input(
            description="Model Faces would be swapped into target image.", 
            default=None,
        ),
        inference_mode: str = Input(
            default="swap",
            choices=["swap", "detect"],
            description="Face swap mode or detection mode. Default is swap.",
        ),
    ) -> Union[List[tuple], List[Path]]:
        """Run a single prediction on the model"""
```
Does cog support a List of Path as input in any method? In my scenario, I wish to make multiple faces processing in one HTTP API calling instead of 6-7 times HTTP latency.

yorickvP:
Which cog version are you using? You may have more luck with one of the 0.9 beta's, which support list and union inputs.

alexblattner:
@yorickvP how do you use that?

peter65374:
> Which cog version are you using? You may have more luck with one of the 0.9 beta's, which support list and union inputs.

0.8.3. I am going to upgrade cog to 0.9 beta. Thank you.😄

alexblattner:
@peter65374 @yorickvP it didn't work for me...

Version: 0.9.0b9

```
TypeError: Unsupported input type typing.List for parameter `liked_images`. Supported input types are: str, int, float, bool, cog.File, cog.Path.
ⅹ Failed to get container status: exit status 1
```

```
TypeError: Unsupported input type list for parameter `liked_images`. Supported input types are: str, int, float, bool, cog.File, cog.Path.
ⅹ Failed to get container status: exit status 1
```

jonhilgart22:
I'm also running into this, but I swear it was working yesterday...

jonhilgart22:
This should work

```❯ sudo curl -o /usr/local/bin/cog -L "https://github.com/replicate/cog/releases/download/v0.9.0-beta11/default.cog_darwin_arm64"```

alexblattner:
[...]

PR #3036, opened 2026-05-30
fix: recurse through nested JSON file inputs

immanuwell:
`cog predict --json` only rewrites top level `@file` values rn. If those paths sit inside arrays or nested objects, they go through as raw `@...` strings, so `list[Path]` style inputs via JSON are kinda busted.

This makes the rewrite recursive for JSON arrays and objects. Added a unit test and an integration test too.

Repro:
`cog build -t $TEST_IMAGE`
`cog predict $TEST_IMAGE --json '{"paths": ["@1.txt", "@2.txt"]}'`

Before this, the CLI sends literal `@1.txt` strings and the prediction falls over.
After this, those entries become data URLs and it works.

Related: #1344

### DIFF
--- a/pkg/cli/predict.go
+++ b/pkg/cli/predict.go
@@ -159,33 +159,59 @@ func transformPathsToBase64URLs(inputs map[string]any) (map[string]any, error) {
 	result := make(map[string]any)
 
 	for key, value := range inputs {
-		if strValue, ok := value.(string); ok && strings.HasPrefix(strValue, "@") {
-			// This is a file path, convert to base64 data URL
-			filePath := strValue[1:]
-
-			// Read file
-			data, err := os.ReadFile(filePath)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to read file %q: %w", filePath, err)
-			}
-
-			// Get MIME type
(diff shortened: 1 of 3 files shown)


pkg/cli/predict.go
```diff
@@ -159,33 +159,59 @@ func transformPathsToBase64URLs(inputs map[string]any) (map[string]any, error) {
...
-			}
+	return result, nil
+}
 
-			// Get MIME type
-			mimeType := mime.TypeByExtension(filepath.Ext(filePath))
-			if mimeType == "" {
-				mimeType = "application/octet-stream"
-			}
+func transformJSONValuePathsToBase64URLs(value any) (any, error) {
```
markphelps, line 172: you didnt introduce this, and its called out in the docs that `@` is reserved for file paths, but Im thinking in the future we should somehow make this more resilient or atleast make the error message better , since this will fail if a user tries to pass in `@string` thats not a file path. like:

```
{"metadata":{"handle":"@alice"}} or {"tags":["@foo"]}
```

immanuwell: @markphelps done, I addressed this in the latest commit

PR #3047, opened 2026-06-04
Kong CLI parity via shared command logic

markphelps:
## Summary

Builds the side-by-side `cmd/cog-kong` CLI to **full command parity** with the existing Cobra CLI (`cmd/cog`), while extracting parser-independent command behaviour into shared, exported functions under `pkg/cli`.

Both CLIs now execute the **same core code**: command behaviour, validation, Docker/model/provider setup, output formatting, and errors live in shared `Run*` functions and `*Options` structs. Cobra owns flag registration/parsing; Kong owns struct tags + DI; neither owns behaviour.

All 13 commands reach parity: `build`, `push`, `serve`, `exec`, `predict`, `run`, `train`, `init`, `login`, `doctor`, `debug`, `weights` (import/pull/status), `base-image` (dockerfile/build).

## Approach

Each task extracts one command's behaviour into shared code, then wires both Cobra and Kong adapters into it (TDD: failing parse/behaviour test → shared runner → thin adapters):

| Task | Shared runner(s) added |
|------|------------------------|
| build | `RunBuild`, `BuildFlagsOptions`, `ResolveBuildImageName` |
| push | `RunPush`, `ResolvePushTarget`, `PushCommandOptions` |
| serve/exec | `RunServe`, `RunExec`, `RuntimeBuildOptions` |
| init/login/doctor/debug | `RunInit`, `RunLogin`, `RunDoctor`, `RunDebug` |
| predict/run/train | `RunPrediction`, `RunTrain` |
| weights | `RunWeightsImport`, `RunWeightsPull`, `RunWeightsStatus` |
| base-image | `RunBaseImageDockerfile`, `RunBaseImageBuild` |

## Deviations from the original plan

The plan was speculative in a few spots; the implementation matches the real code:

- Dropped non-existent `OCIIndex` / `OCIIndexEnabled()` from build options.
- Used the **real** base-image flags (`--cuda`/`--python`/`--torch`), not the plan's guessed `--*-version`.
- Removed `short:"v"` from Kong's `--version` — it diverged from Cobra (which has no `-v`) and collided with the weights commands' `-v` (verbose).
- Moved the weights experimental warning from Cobra's `PersistentPreRun` into the shared runners, so it fires once per subcommand from a single source.

## Verification

- `mise run test:go` — all pass (2018 tests)
- `mise run lint:go` — 0 issues
- `mise run fmt:go` — clean
- Both `cog` and `cog-kong` compile; no binaries committed
- New `cmd/cog-kong/main_test.go` asserts the full command surface (registration, nested commands, root globals, per-command flag parsing) without requiring Docker

## Notes

- Base branch is `kong` (this stacks on the existing side-by-side scaffold).
- `cmd/cog` remains the production Cobra binary, unchanged in behaviour.
- Draft: opening for review of the shared-code seams before finalising.


cmd/cog-kong/baseimage.go
```diff
@@ -0,0 +1,62 @@
...
+
+	NoCache  bool   `name:"no-cache" help:"Do not use cache when building the image."`
+	Progress string `name:"progress" default:"${progress_default}" enum:"auto,plain,tty,quiet" help:"Set type of build progress output: ${enum}."`
+}
+
+func (cmd *BaseImageDockerfileCmd) Run(ctx context.Context) error {
+	opts := cmd.options()
+	opts.NoCache = cmd.NoCache
+	opts.ProgressOutput = cmd.Progress
+	return cli.RunBaseImageDockerfile(ctx, opts)
```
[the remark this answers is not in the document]
markphelps: Declining this one — it would actually *break* parity. The Cobra `base-image build` command (`newBaseImageBuildCommand`) only calls `addBaseImageFlags(cmd)`, which does **not** register `--no-cache` or `--progress`. Those flags are added only to `base-image dockerfile` (via the separate `addNoCacheFlag`/`addBuildProgressOutputFlag` calls). So the Kong `BaseImageBuildCmd` correctly omits them to match Cobra. `RunBaseImageBuild` reading `opts.NoCache`/`opts.ProgressOutput` just falls back to zero values in both CLIs, which is the existing Cobra behavior.

The other 5 comments (hidden predict/train/weights/debug, dead `os.Exit(1)`) are addressed in bf8a2499.

cmd/cog-kong/main.go
```diff
@@ -84,6 +82,32 @@ func main() {
...
+		kong.Name("cog"),
+		kong.Description("Containers for machine learning."),
+		kong.Vars{
+			"version":          fmt.Sprintf("cog version %s (built %s)", version, buildTime),
+			"commit":           commit,
+			"progress_default": progressDefault(),
+			"registry_default": global.DefaultReplicateRegistryHost,
+		},
+		kong.UsageOnError(),
+		kong.NoDefaultHelp(),
```
anish-sahoo: `--help` doesn't short-circuit. With `NoDefaultHelp()`, help is handled by the `cli.Help` check after `parser.Parse()` (L71), but `Parse()` has already run `Validate()`, arg-count checks, enum validation and the `existingfile` mapper — so validation errors before help is ever shown:

```
cog exec --help                                                  # exit 80 (requires 1 arg)
cog push --use-cog-base-image --use-cuda-base-image=false --help # mutex error
cog build --progress bogus --help                                # enum error
cog build --dockerfile missing --help                            # stat error
```

Expected (Cobra): all print help and exit 0. Make `--help` fire before validation — e.g. use Kong's built-in help (a `BeforeReset` hook that runs before `Validate`) instead of `NoDefaultHelp()` + a post-parse bool.

cmd/cog-kong/train.go
```diff
@@ -0,0 +1,34 @@
...
+
+// TrainCmd implements the hidden, deprecated "cog train" command.
+type TrainCmd struct {
+	RuntimeFlags `embed:""`
+	EnvFlag      `embed:""`
+
+	Image        string   `arg:"" optional:"" name:"image" help:"Image to train. If omitted, builds from cog.yaml in the current directory."`
+	Input        []string `name:"input" short:"i" help:"Inputs, in the form name=value. If value is prefixed with @, it is read from a file on disk, e.g. -i path=@image.jpg."`
+	Output       string   `name:"output" short:"o" default:"weights" help:"Output path."`
+	SetupTimeout uint32   `name:"setup-timeout" default:"300" help:"The timeout for a container to setup (in seconds)."`
```
anish-sahoo: `--setup-timeout` is registered here, but Cobra's `train` doesn't have it (`addSetupTimeoutFlag` is only wired into predict/run):

```
cog train --setup-timeout 5   # cobra: "unknown flag: --setup-timeout" | kong: accepted
```

Expected: rejected, matching Cobra. Drop this field (and the `SetupTimeout:` line in the `RunTrain` call below), or add the flag to Cobra's `train` if it's intentional.

cmd/cog-kong/push.go
```diff
@@ -25,62 +21,11 @@ func (cmd *PushCmd) Validate() error {
 	return cmd.ValidateMutualExclusivity()
```
anish-sahoo: Cobra's `push` has no mutual-exclusivity check (only `build`/`exec` call `checkMutuallyExclusiveFlags`), so this diverges:

```
cog push --use-cog-base-image --use-cuda-base-image=false
# cobra: accepted (proceeds to build) | kong: rejected
```

Expected: accepted, matching Cobra. Remove this `Validate()` method, or add the check to Cobra's `push` if the stricter behaviour is intended.

cmd/cog-kong/main.go
markphelps: Fixed in 89369f17. Switched to Kong's built-in `--help` flag (removed `NoDefaultHelp()` and the post-parse `cli.Help` check). Help now fires in a `BeforeReset` hook before `Validate()`/arg-count/enum/file mapping, so `cog exec --help`, the mutex case, `--progress bogus --help`, and `--dockerfile missing --help` all print help and exit 0, matching Cobra. (`--dockerfile` also dropped `type:"existingfile"` so the file check defers to build time like Cobra.)

cmd/cog-kong/train.go
markphelps: Fixed in 89369f17. Dropped the `--setup-timeout` field (and the `SetupTimeout:` line in the `RunTrain` call) from the Kong `train` command to match Cobra, which doesn't wire `addSetupTimeoutFlag` into `train`.

cmd/cog-kong/push.go
markphelps: Fixed in 89369f17. Removed the `Validate()` mutual-exclusivity check from `PushCmd` to match Cobra, where only `build`/`exec` call `checkMutuallyExclusiveFlags`.

cmd/cog-kong/flags.go
```diff
@@ -33,42 +34,44 @@ type BuildFlags struct {
 	Progress         string   `name:"progress" default:"${progress_default}" enum:"auto,plain,tty,quiet" help:"Set type of build progress output: ${enum}."`
 	UseCudaBaseImage string   `name:"use-cuda-base-image" default:"auto" enum:"auto,true,false" help:"Use Nvidia CUDA base image, 'true' (default) or 'false' (use python base image)."`
```
anish-sahoo, line 35: **Parity gap (behavioral): invalid `--use-cuda-base-image` value.** Kong enforces `enum:"auto,true,false"` at parse time, so an invalid value is rejected before any work:
```
cog build --use-cuda-base-image maybe
#   main → exit 0, silently BUILDS the image (ignores "maybe")
#   kong → exit 80, "must be one of auto,true,false but got maybe"
```
Same command, divergent outcome. Kong is arguably more correct, but this changes behavior for anyone relying on Cobra's lax handling — worth a deliberate decision + changelog note rather than a silent change.

cmd/cog-kong/flags.go
```diff
@@ -33,42 +34,44 @@ type BuildFlags struct {
 	Progress         string   `name:"progress" default:"${progress_default}" enum:"auto,plain,tty,quiet" help:"Set type of build progress output: ${enum}."`
```
anish-sahoo, line 34: **Parity gap: `--progress` fails fast vs fails late.** Kong's `enum:` rejects an invalid value at parse (exit 80, no work done); Cobra accepts it and only fails *after starting the Docker build*:
```
cog build --progress bogus
#   kong → exit 80 at parse
#   main → exit 1, but only after "Building Docker image..." → "invalid progress mode bogus"
```
Different exit code and different side effects (main begins building first).

cmd/cog-kong/main.go
```diff
@@ -76,6 +69,7 @@ func main() {
 		parser.FatalIfErrorf(err)
```
anish-sahoo, line 69: **Two systematic divergences from Cobra flow through `FatalIfErrorf` here:**

- **Exit code:** parse/usage errors exit `80` (Kong's `ParseError.ExitCode()`); the Cobra CLI exits `1`. Anything asserting on exit codes (CI, scripts, integration tests) will see the difference.
- **stderr prefix:** Kong prints `cog: error: <msg>` (from `kong.Name("cog")` + `UsageOnError()`); Cobra prints the bare message. Body + exit code otherwise match, e.g.:
```
kong : cog: error: cog.yaml not found in <dir> ...
main : cog.yaml not found in <dir> ...
```
Worth a decision on whether parity requires matching Cobra's exit `1` + bare text.

anish-sahoo, line 69: **Remaining `--help` short-circuit gap after this commit.** Dropping `NoDefaultHelp` correctly makes enum/mutex/file errors short-circuit to help, but **extra positional args still don't** — `cog <cmd> <extra> --help` errors instead of printing help:
```
cog predict a b --help   # kong → 80 "unexpected argument b";  cobra → 0 (help)
cog login extra --help   # kong → 80 "unexpected argument extra"; cobra → 0 (help)
```
The rescue at line 63 only matches `strings.HasPrefix(parseErr.Error(), "expected")`, so `unexpected argument …` parse errors fall through to `FatalIfErrorf` and exit 80. This is the only bucket still firing in the flag fuzzer (`HELP_VS_ARGS`, ~50 unique combos). Likely the same fix should also defer positional-arity validation past Kong's help hook.

anish-sahoo (commented):
End-to-end parity re-test after 89369f17 (kong-branch vs main, Dockerfile-generation diff + command validation + real build/predict + flag fuzz).

**What this commit fixed (verified):** `--help` now short-circuits past enum/mutex/file validation (e.g. `build --progress bogus --help` and `build --use-cuda-base-image maybe --help` both exit 0), `train --setup-timeout` is gone, and the push mutual-exclusivity check is removed. The fuzzer's `HELP_VS_FLAGVALUE` / `HELP_VS_MUTEX` buckets are now clean.

**Still diverging from Cobra** — inline comments below:
1. Parse-time enum strictness on `--use-cuda-base-image` / `--progress` (behavioral, no `--help`).
2. Usage/parse-error exit code `80` vs Cobra `1`.
3. `cog: error:` stderr prefix vs Cobra's bare message.
4. Remaining `--help` gap: extra positional args still exit 80 instead of showing help.

Build pipeline + predictions themselves are identical: byte-identical Dockerfiles across fixtures/examples, identical image config/OpenAPI labels, identical prediction output.

cmd/cog-kong/flags.go
markphelps: yeah im kind of ok with this being not compatible as we should reject invalid values so we could address this in a changelog

cmd/cog-kong/flags.go
markphelps: same as above tbh

PR #3048, opened 2026-06-04
feat: support JSON-native union inputs

markphelps:
## What

Adds support for JSON-native **union input types**, e.g.:

```python
def predict(self, value: str | float) -> str: ...
def predict(self, value: str | float | None = Input(default=None)) -> str: ...
def predict(self, nums: list[int] | list[float]) -> str: ...
```

Unions are restricted to JSON-native members (`str`, `int`, `float`, `bool`, `dict`/`Any`, `list[T]`, `None`) so request validation happens at the HTTP edge against the OpenAPI schema. Unions involving `Path`, `File`, `Secret`, custom coders, and `BaseModel` are rejected at build/schema-generation time. Output unions remain unsupported.

## How

- **`pkg/schema`** — recursive `InputType` model + resolver; emits OpenAPI `anyOf` for union inputs. Multi-variant nullable unions stay `required` when no default is supplied; non-union optionals are unchanged.
- **`pkg/predict`** — CLI `-i` parsing inspects `anyOf` members: numeric values parse as numbers for unions that accept one (`schemaAcceptsNumber`), and a failed numeric parse falls back to the string member for number-first unions like `float | str` (`schemaAcceptsString`).
- **`python/cog/_adt`** — deterministic union normalisation with strict per-member compatibility (bool is not int/float, scalars don't pass through `dict`/`Any`, list unions validate each element); `anyOf` `json_type` emission.
- **docs** — `architecture/02-schema.md` and `docs/python.md` document supported members, `anyOf`, and nullable semantics.

## Behavioral notes

- **No runtime primitive coercion.** Input validation happens at the HTTP edge via the OpenAPI schema; the worker does not re-coerce primitives. So `-i value=1` for `str | float` stays an `int` (consistent with how a plain `float` field already receives a Python `int` for `-i num=10`); only fractional values become floats.
- **Nullable means "may be omitted", not "accepts explicit null".** `nullable: true` lets an omitted optional fall back to its default. An explicit JSON `null` is still validated against the field type and rejected at the HTTP edge — identical to every other optional field today.

## Testing

- Go unit tests: `pkg/schema/...`, `pkg/predict/...`
- Python regression tests: `python/tests/test_adt.py` (union ordering, bool/int/float overlap, list-element validation, optional/required `None` handling, empty list, mixed scalar+list)
- End-to-end txtar integration tests: `union_input_http`, `union_input_cli` (both member orderings), `union_input_list_http`

All passing locally; `lint:go` clean; `ruff` clean.

### DIFF
--- a/architecture/02-schema.md
+++ b/architecture/02-schema.md
@@ -102,15 +102,15 @@ class Runner(BaseRunner):
 
 The resolver handles local imports relative to the predictor file and project root:
 
-| Import Style                         | File Resolved                                               |
-| ------------------------------------ | ----------------------------------------------------------- |
-| `from output_types import X`         | `<project>/output_types.py`                                 |
-| `from .output_types import X`        | `<predictor-dir>/output_types.py`                           |
-| `from models.output import X`        | `<project>/models/output.py`                                |
-| `from .models.output import X`       | `<predictor-dir>/models/output.py`                          |
-| `from output_types import X as Y`    | `<project>/output_types.py` (alias tracked)                 |
-| `from .output_types import X as Y`   | `<predictor-dir>/output_types.py` (alias tracked)           |
-| `from . import output_types`         | `<predictor-dir>/output_types.py` (module alias tracked)    |
+| Import Style                       | File Resolved                                            |
+| ---------------------------------- | -------------------------------------------------------- |
+| `from output_types import X`       | `<project>/output_types.py`                              |
+| `from .output_types import X`      | `<predictor-dir>/output_types.py`                        |
+| `from models.output import X`      | `<project>/models/output.py`                             |
+| `from .models.output import X`     | `<predictor-dir>/models/output.py`                       |
+| `from output_types import X as Y`  | `<project>/output_types.py` (alias tracked)              |
+| `from .output_types import X as Y` | `<predictor-dir>/output_types.py` (alias tracked)        |
+| `from . import output_types`       | `<predictor-dir>/output_types.py` (module alias tracked) |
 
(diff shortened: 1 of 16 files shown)


architecture/02-schema.md
```diff
@@ -186,8 +186,14 @@ Each `SchemaType` produces its JSON Schema fragment via `JSONSchema()`:
 | `cog.Secret`                          | `{"type": "string", "format": "password", "x-cog-secret": true}` | Masked in logs             |
 | `list[T]`                             | `{"type": "array", "items": {...}}`                              |                            |
 | `Optional[T]`                         | Type T + not in `required`                                       | Input fields only          |
+| `A \| B` / `Union[A, B]`               | `{"anyOf": [A, B]}`                                              | Input-only, JSON-native unions only |
+| `A \| None` / `Optional[A]`            | Type A + `nullable: true`                                        | Input fields may still be required if no default is supplied |
```
dennnnnnnnnis: so it's always better to include the default = None if an input filed is optional right? even if we have Optional[A] defined, otherwise the schema might not pick them up as optional? 
I don't recall what I did with optional fields previously but I always had default=something, guess it's not a problem with me but just asking

dennnnnnnnnis (dismissed):

pfrohlich (dismissed):

architecture/02-schema.md
markphelps: Good question, and you've actually hit on a real subtlety.

Short answer: `Optional[T] / T | None` does not need `default=None`. What it needs is some `= Input(...)` assignment. The `Input(...)` is what makes the field omittable; the `None` default is implicit for optionals.

BUT doing `value: Optional[str]` (bare, no Input) generates the openAPI schema correctly (not required, nullable: true), but it fails at runtime with no argument with `TypeError: missing 1 required positional argument`.. so that is actually a preexisting bug that we've been papering over with `=Input()` which we should fix!

But for now will update the docs so we are consistent everywhere and look into what it will take to actually fix or atleast make it more clear as to why it doesnt work as expected with a better error message or something

pfrohlich (approved):

PR #3055, opened 2026-06-11
Move examples from cog-examples to cog

anish-sahoo:
Moves the example models from the separate `replicate/cog-examples` repo into this repo under `examples/`, migrates them to the `BaseRunner`/`run` API, and adds `cog.Secret` input support to coglet.

- Add example models using `run.py`/`BaseRunner`: `blur`, `canary`, `hello-concurrency`, `hello-context`, `hello-image`, `hello-replicate`, `hello-train`, `hello-world`, `notebook`, `z-image-turbo`
- Migrate the existing `streaming-text` example to `run.py`/`BaseRunner`
- Replace `examples/resnet` with the classic torchvision ResNet50 classifier (`BaseRunner`, ImageNet weights bundled with torchvision -- no managed weights or import step), ported and modernized from `replicate/cog-examples`
- Move the previous managed-weights ResNet example to `examples/experimental/resnet-managed-weights` (renamed `model: resnet-managed-weights`)
- Update `managed-weights` `cog.yaml` to use the `model:` field (replacing `image: <your-registry>/...`) and refresh its `weights.lock`
- `hello-replicate` demonstrates `Secret` usage by calling the Replicate API

### coglet
- Add `cog.Secret` input support: classify and coerce `Secret`, `Optional[Secret]`, and `Secret | None` fields, wrapping string values in `cog.types.Secret` (`list[Secret]` is intentionally not coerced)
- Ignore two pyo3 0.27 advisories in `cargo-deny` (`RUSTSEC-2026-0176`, `RUSTSEC-2026-0177`) with justification; pinned due to numpy/pyo3-async-runtimes lag, and neither code path is reachable in coglet

### test harness
- Add a `base_dir` field for `repo: local` models so local models resolve relative to the manifest (defaults to `fixtures/models/`)
- Point all example models in `manifest.yaml` at the local `examples/` directory instead of `replicate/cog-examples`
- Migrate fixture models from `predict.py`/`BasePredictor` to `run.py`/`BaseRunner`
- Only pass `--setup-timeout` to `cog predict` (`cog train` does not support it)

PR #3057, opened 2026-06-16
feat: add cog.Secret input support to coglet

anish-sahoo:
Split out from #3055 (example migration) per review feedback: coglet `Secret` input coercion is new runtime behavior and deserves its own focused review and tests.

## coglet
- Add `cog.Secret` input support: classify and coerce `Secret`, `Optional[Secret]`, and `Secret | None` fields, wrapping plain string values in `cog.types.Secret`. `list[Secret]` is intentionally **not** coerced.
- Rename `coerce_url_strings` to `coerce_typed_inputs` to reflect that it now handles File/Path/Secret coercion.
- Add end-to-end coglet-python pytest coverage asserting predictors annotated with `Secret` actually receive a `cog.Secret` whose `get_secret_value()` round-trips the submitted value — covering `Secret`, `Optional[Secret]`, `Secret | None`, plus negatives (plain `str` stays `str`, omitted `Optional[Secret]` stays `None`).

## examples
- Add `hello-replicate`, which demonstrates `Secret` usage by calling the Replicate API (`replicate_api_token: Secret`). The README recommends passing the token via an environment variable rather than a literal on the command line.

## Notes
- The `cargo-deny` pyo3 advisory ignores (`RUSTSEC-2026-0176`/`RUSTSEC-2026-0177`) and the `hello-replicate` test-harness wiring depend on #3055. If this PR merges first, those may need to come along to keep CI green.


crates/coglet-python/src/input.rs
```diff
@@ -630,6 +665,51 @@ mod tests {
...
+        assert!(
+            c.secret_fields.is_empty(),
+            "non-Secret types incorrectly detected as secret: {:?}",
+            c.secret_fields
+        );
+    }
+
+    #[test]
+    #[ignore] // Requires cog Python package in PYTHONPATH
+    fn detect_direct_secret() {
```
markphelps, line 684: This adds useful Secret annotation coverage, but these tests are all `#[ignore]` and only validate classification. The behavior this PR changes is runtime coercion in `coerce_typed_inputs`, so default CI still would not catch a predictor receiving a plain `str` instead of `cog.Secret`. Could you add active coglet-python pytest coverage that asserts a predictor annotated with `Secret` receives a `cog.Secret` whose `get_secret_value()` returns the submitted value? Ideally cover `Secret`, `Optional[Secret]`, `Union[Secret, None]` / `Secret | None`, plus negative cases like non-secret `str` staying `str` and optional `None` staying `None`.

crates/coglet-python/src/input.rs
```diff
@@ -100,22 +102,27 @@ struct FieldClassification {
...
     path_fields: HashSet<String>,
+    /// Fields typed as `cog.Secret` (or `Optional[Secret]`, etc.)
+    /// These wrap plain string values in `cog.types.Secret`.
+    secret_fields: HashSet<String>,
 }
 
 /// Inspect a Python function's type annotations to find parameters typed as
-/// `cog.File` or `cog.Path` (including `list[...]`, `Optional[...]`,
-/// `... | None`, etc.).
+/// `cog.File`, `cog.Path`, or `cog.Secret` (including `list[...]`,
```
markphelps: This wording now overstates Secret support: File/Path support list forms, but the implementation intentionally does not classify or coerce `list[Secret]` (`classify_type` only handles list elements for File/Path). Could you narrow these comments to say Secret supports direct `Secret`, `Optional[Secret]`, and `Secret | None` only, or otherwise make the list behavior explicit?

examples/hello-replicate/README.md
```diff
@@ -0,0 +1,36 @@
...
+    ...
+```
+
+`cog.Secret` redacts its value in logs and string representations. Read the
+underlying value with `get_secret_value()`.
+
+## Run it
+
+```sh
+cog predict -i image=@cat.png -i replicate_api_token=r8_...
```
markphelps: Because this example is specifically about secret handling, I’d avoid recommending a literal token on the command line: it can leak through shell history and sometimes process listings before `cog.Secret` ever gets a chance to redact it. Could you show a safer flow, e.g. reading from an environment variable/stdin or another non-literal-token pattern, and mention that `cog.Secret` redacts model logs but cannot protect secrets exposed by the user’s shell?

markphelps (commented):
Adversarial review found a few fixes worth doing before merge; comments inline.

crates/coglet-python/src/input.rs
anish-sahoo: Added active pytest coverage in `TestSecretInput` (Secret/Optional/PEP604 + negatives) — done in 6340d2f3.

crates/coglet-python/src/input.rs
anish-sahoo: Narrowed the doc comments to make explicit that `list[Secret]` is not classified/coerced — done in 6340d2f3.

examples/hello-replicate/README.md
anish-sahoo: Switched to an env-var flow and added a note that `cog.Secret` can't protect shell-exposed secrets — done in 6340d2f3.

markphelps (approved):
lgtm!!

PR #3064, opened 2026-06-18
docs: Document how to run Cog models as Docker images

### LINKED ISSUE #1804: Document how to run Cog models as Docker images
It's described in one code example in the readme, and there is some HTTP reference documentation, but this could do with a step-by-step guide.

It's also not documented anywhere how to pass input files. There are probably some other little things like this that could do with documenting.

anish-sahoo:
Expand deploy.md with a step-by-step guide covering:
- Building and running Docker images (docker run, cog serve, cog run)
- Making predictions via the HTTP API with the correct input wrapper
- Passing file inputs via URLs and base64 data URLs
- Getting output files (data URLs and output_file_prefix)

Removed stale docs that referred to `cog serve my-model` - this functionality isn't supported anymore

Closes #1804


docs/deploy.md
```diff
@@ -39,33 +41,141 @@ docker run -d -p 5001:5000 --gpus all my-model
...
+the response contains a base64-encoded data URL by default:
+
+```json
+{
+    "status": "succeeded",
+    "output": "data:image/png;base64,iVBORw0KGgo..."
+}
+```
+
+To have the server upload output files to external storage instead,
```
markphelps, line 152: I think this section is documenting behavior the current server doesn’t support. `output_file_prefix` isn’t part of the HTTP prediction request shape in coglet, so this field will be ignored and users will still get data URLs back.

From the code, file uploads are configured at server startup with `--upload-url` instead. Could we change this section to show the supported flow, e.g. `cog serve --upload-url https://example.com/upload/`, and for raw Docker show overriding the command with `python -m cog.server.http --upload-url ...`?

If we do want per-request `output_file_prefix`, I think that needs an implementation + test before documenting it.

docs/deploy.md
```diff
@@ -39,33 +41,141 @@ docker run -d -p 5001:5000 --gpus all my-model
...
+    -H "Content-Type: application/json" \
+    -d '{
+        "input": {"prompt": "a cat"},
+        "output_file_prefix": "https://example.com/upload"
+    }'
 ```
 
-For more details about the HTTP API,
-see the [HTTP API reference documentation](http.md).
+The server uploads the file via HTTP PUT and returns the resulting URL:
```
markphelps: This is close, but tied to the unsupported `output_file_prefix` example above. The upload behavior is currently controlled by the server’s startup `--upload-url`, not by a field in each prediction request.

I’d reframe this as “If the server was started with `--upload-url`, file outputs are uploaded and the response contains the uploaded URL.” That keeps the docs aligned with how Cog behaves today.

docs/deploy.md
```diff
@@ -39,33 +41,141 @@ docker run -d -p 5001:5000 --gpus all my-model
...
+use `cog run` with the image name:
+
+```console
+cog run my-model -i image=@input.jpg
+```
+
+This starts the container, runs a single prediction, and prints the result.
+File inputs are passed with `@` prefix (e.g. `-i image=@photo.jpg`).
+
+## Make a prediction
```
markphelps, line 60: This section follows examples for both `docker run` and `cog serve`, but all the curl examples use `localhost:5001`, which only matches the Docker example above (`-p 5001:5000`).

Someone using the default `cog serve` path will have the server on `8393`, so these commands would fail as written. Could we add a short note here like:

“The examples below use `localhost:5001`, matching the Docker command above. If you used `cog serve`, use `localhost:8393` by default, or the port you passed with `-p`.”

docs/deploy.md
```diff
@@ -39,33 +41,141 @@ docker run -d -p 5001:5000 --gpus all my-model
...
+```
+
+By default the server runs on port 8393.
+Use `-p` to choose a different port:
+
+```console
+cog serve -p 5000
+```
+
+### cog run
```
markphelps: I’d separate `cog run` from the server-starting options. The Docker and `cog serve` examples leave an HTTP server running, but `cog run` starts a container for a single prediction and then exits.

Since the next section walks through calling `/predictions`, grouping `cog run` here makes it sound like it also starts a server for those HTTP examples. Maybe move this into a separate “Run a one-off prediction” subsection after the HTTP flow, or explicitly say it’s an alternative to running a server.

docs/deploy.md
```diff
@@ -77,6 +187,16 @@ curl http://localhost:5001/health-check
...
 
+## Stop the server
+
+If you started the container with `docker run -d`, stop it with:
+
+```console
+docker kill <container-id>
+```
+
+If you used `cog serve` or `cog run`, press `Ctrl+C` in the terminal.
```
markphelps: Small follow-up to the `cog run` point above: `cog run` usually runs one prediction and exits, so “press Ctrl+C” doesn’t really fit the same way it does for `cog serve`.

Could we make this just about stopping `docker run -d` and `cog serve`, or clarify that `cog run` normally doesn’t leave a server running?

docs/deploy.md
anish-sahoo: Fixed

docs/deploy.md
anish-sahoo: Fixed. Reframed as 'with an upload URL configured, file outputs are uploaded and the response contains the uploaded URL.'

docs/deploy.md
anish-sahoo: Fixed

docs/deploy.md
anish-sahoo: Fixed. Moved cog run into its own 'Run a one-off prediction' section after the HTTP flow, noting it runs a single prediction and exits without starting a server.

docs/deploy.md
anish-sahoo: Fixed

docs/deploy.md
anish-sahoo: removed mentions of `output_file_prefix` to not confuse users. I can open an issue for a future implementation of this if needed

markphelps (approved):
nice

PR #3065, opened 2026-06-18
docs: Consolidate install instructions in README

### LINKED ISSUE #2749: Consolidate install instructions
Getting started guide has different info to readme. Maybe should consolidate in an install doc, and link everywhere to that.

anish-sahoo:
Closes #2749.

Moves all installation instructions into the README with platform accordions, and updates the getting-started guides to link there instead of duplicating the steps.


docs/wsl2/wsl2.md
```diff
@@ -156,12 +156,7 @@ Open Windows Terminal and enter your WSL 2 VM:
 wsl.exe
 ```
 
-Download and install `cog` inside the VM:
-
-```bash
-sudo curl -o /usr/local/bin/cog -L https://github.com/replicate/cog/releases/latest/download/cog_`uname -s`_`uname -m`
-sudo chmod +x /usr/local/bin/cog
-```
+Download and install `cog` inside the VM by following the [Linux install instructions in the README](../../README.md#install).
```
markphelps, line 159: so the docs get build using mkdocs, and published to github pages. does that work with relative links like this? because I dont think its gonna be able to link to the README once published unless it replaces this with an absolute link at build time

markphelps (changes requested):

markphelps:
we should just make CI generate docs and LLM docs and not have these silly checks

docs/wsl2/wsl2.md
anish-sahoo: added fix

markphelps (approved):
nice!

PR #3069, opened 2026-06-22
test: cover CUDA minor version base image selection

### LINKED ISSUE #813: Pick latest CUDA version when specifying patch version
```
build:
  python_version: "3.10"
  cuda: "11.6"
  python_packages:
   - "torch==1.13.0"
```

This automatically selects CUDA 11.6.0 for the base image. It should pick the latest patch version and use 11.6.2.

sarptandoven:
## What does this PR do?

Adds regression coverage for CUDA minor-version base image selection.

Issue #813 describes the case where specifying a minor CUDA version like `11.6` should select the latest available patch version. Current `main` already appears to do this correctly, resolving `CUDA: "11.6"` and `CuDNN: "8"` to:

```text
nvidia/cuda:11.6.2-cudnn8-devel-ubuntu20.04
```

This PR adds a focused test so that behavior stays covered.

Fixes #813.

## Testing

```bash
go test ./pkg/config
```

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -468,6 +468,21 @@ func TestCUDABaseImageTag(t *testing.T) {
 	require.Equal(t, "nvidia/cuda:11.8.0-cudnn8-devel-ubuntu22.04", imageTag)
 }
 
+func TestCUDABaseImageTagChoosesLatestPatchForMinorCUDA(t *testing.T) {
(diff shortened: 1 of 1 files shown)


sarptandoven:
Closing this since the change only added coverage for behavior that already works on current main. I’m going to take a deeper pass and avoid leaving maintainers with a low-impact PR.

PR #3076, opened 2026-06-24
fix: add ignore for tar directories

anish-sahoo:
adds an ignore for tar directories in the weights pulling step. this reduces unnecessary `a606: skipping unexpected tar entry "text_encoder/" (typeflag 53)` style logs.


michaeldwan (approved):
nice fix

PR #3086, opened 2026-06-30
feat: add cog playground browser UI

### LINKED ISSUE #649: Local web interface
It'd be neat if Cog served a Replicate-style web interface when you ran the web server for testing locally.

## User data
- As requested by @postera-james

anish-sahoo:
Closes #649 

Adds a `cog playground` command: a local web server that serves a schema-driven, Postman-like browser UI and reverse-proxies to a running model API (e.g. `cog serve`). It reflects the model's OpenAPI schema and supports sync, streaming (SSE), and async (webhook) predictions, with form or raw-JSON input.

The browser only talks to the playground's own origin, which proxies to the target chosen at runtime (via `X-Cog-Target`) — sidestepping CORS and keeping SSE intact — and hosts a webhook sink that relays async events back over SSE.

```mermaid
flowchart LR
    Browser["Browser UI"]

    subgraph Playground["cog playground (host)"]
        Static["Static UI assets<br/>(go:embed)"]
        Proxy["Reverse proxy<br/>/proxy/*"]
        Sink["Webhook sink + relay<br/>/webhook/{token} → /events"]
    end

    Model["Target model API<br/>(e.g. cog serve)"]

    Browser -->|"load UI"| Static
    Browser -->|"schema, predictions, SSE"| Proxy
    Proxy -->|"X-Cog-Target"| Model
    Model -->|"webhook (async)"| Sink
    Sink -->|"SSE events"| Browser
```

**Usage:** run `cog serve` in one terminal and `cog playground` in another.

The UI is plain HTML/JS embedded via `go:embed` (no build step; Ace vendored for JSON editing). Documented in `docs/cli.md` and `architecture/06-cli.md`. Also bumps the `streaming-text` example to Python 3.13.

image
image
image


pkg/cli/playground/schema.js
```diff
@@ -0,0 +1,125 @@
+// Pure OpenAPI-schema helpers (no DOM, no Lit) so they can be unit-reasoned
```
michaeldwan, line 1: The one thing I'd most want to see added: a test + lint gate on the JS. Right now `mise run fmt` covers go/rust/python/docs, prettier's installed but only runs against `docs/cli.md`, and there's no linter or JS test runner -- so about a third of the diff has no automated checking. It's not a knock on the code; it's just missing the safety net the rest of the repo has, and it's where the bugs in the other comments slipped through.

Good news is it stays buildless -- no bundler:
- **Format** -- point the vendored prettier at `pkg/cli/playground/**/*.js`, ignore `vendor/`. One task, no new deps.
- **Test** -- `node --test` is built into the Node we already manage in `mise.toml`, and it imports these ES modules directly, same as the browser does. You've already written these helpers pure ("so they can be unit-reasoned" -- love that) -- `parseSSEEvent`, `resolveEnum`, `effectiveProp`, `fieldKind`, `coerceEnum`, `defaultInput` are ready to test as-is. Add a `test:js` task and wire it into the aggregate `test`.

pkg/cli/playground/vendor/ace/ace.js
```diff
@@ -0,0 +1,23 @@
....off("changeScrollTop",this.$onScrollTopChange),this.session.off("changeScrollLeft",this.$onScrollLeftChange);var n=this.session.getSelection();n.off("changeCursor",this.$onCursorChange),n.off("changeSelection",this.$onSelectionChange)}this.session=e,e?(this.$onDocumentChange=this.onDocumentChange.bind(this),e.on("change",this.$onDocumentChange),this.renderer.setSession(e),this.$onChangeMode=this.onChangeMode.bind(this),e.on("changeMode",this.$onChangeMode),this.$onTokenizerUpdate=this.onTokenizerUpdate.bind(this),e.on("tokenizerUpdate",this.$onTokenizerUpdate),this.$onChangeTabSize=this.renderer.onChangeTabSize.bind(this.renderer),e.on("changeTabSize",this.$onChangeTabSize),this.$onChangeWrapLimit=this.onChangeWrapLimit.bind(this),e.on("changeWrapLimit",this.$onChangeWrapLimit),this.$onChangeWrapMode=this.onChangeWrapMode.bind(this),e.on("changeWrapMode",this.$onChangeWrapMode),this.$onChangeFold=this.onChangeFold.bind(this),e.on("changeFold",this.$onChangeFold),this.$onChangeFrontMarker=this.onChangeFrontMarker.bind(this),this.session.on("changeFrontMarker",this.$onChangeFrontMarker),this.$onChangeBackMarker=this.onChangeBackMarker.bind(this),this.session.on("changeBackMarker",th
```
michaeldwan, line 1: Want to open a conversation about this one rather than just ask for a change. It's 456KB, minified, with no LICENSE, pinned version, or source URL -- and we gate Rust deps through `cargo-deny` and Go through `go.mod`, so a hand-vendored blob is a bit of an outlier for the repo.

The deeper question is whether we need a full editor at all. The job is format + edit JSON in, format JSON out -- two jobs, and only one really needs an editor:
- Output is read-only -- right now it's an Ace instance with the cursor hidden, basically a code editor standing in for a viewer. A highlighted `<pre>` over `JSON.stringify(v, null, 2)` covers format + highlight for ~0 bytes. The only thing you'd lose is folding on big outputs -- worth deciding if that's actually needed or just came for free with Ace.
- Input is where an editor earns its place. [CodeJar](https://github.com/antonmedv/codejar) (~2.5KB) plus a small highlight grammar gets you highlighting in a vendorable ES module. No folding, but for JSON textareas that's likely fine.

(I checked what the frontend repo uses -- it's Monaco, but that's heavier than Ace and React/worker/bundler-bound, so it can't come into a buildless embed. Not a fit here, just closing the loop.)

If folding does matter for large payloads, keeping one lightweight editor for both sides is reasonable -- just the lightest thing that folds, and I'd still want Ace to justify 456KB against a CodeMirror single-file build. If it stays either way, adding the LICENSE + version + a regen note is the minimum.

pkg/cli/playground.go
```diff
@@ -0,0 +1,333 @@
...
+			_, _ = fmt.Fprint(w, ": keep-alive\n\n")
+			flusher.Flush()
+		}
+	}
+}
+
+// writeSSEData emits a payload as a single SSE event, prefixing every line with
+// "data: ". This preserves embedded newlines without letting them terminate the
+// event early or inject additional SSE fields (e.g. a spoofed "event:" line).
+func writeSSEData(w io.Writer, msg []byte) {
```
michaeldwan, line 213: Small bug here: a lone `\r` slips past the injection guard. This splits on `\n`, but SSE treats CR, LF, and CRLF all as line terminators -- so a `\r` in a webhook body survives the split and everything after it gets emitted as an unprefixed field. A payload like `x\revent: custom\r...` can inject an `event:` field. I confirmed it empirically.

`TestPlaygroundWebhookRelayPreservesNewlines` only exercises `\n` so it doesn't catch this. Fix is easy: normalize `\r\n` and lone `\r` to `\n` before framing, and add a `\r` case to the test. The framing approach itself is right -- just needs the CR handling.

pkg/cli/playground/app.js
```diff
@@ -0,0 +1,587 @@
...
+    reportRunError(err);
+  } finally {
+    setRunning(false);
+    renderLive(); // final render without the streaming cursor
+  }
+}
+
+// runAsync submits with Prefer: respond-async and a webhook pointing at the
+// playground server's sink, then observes delivered events over /events (SSE).
+async function runAsync(input) {
```
michaeldwan, line 448: Edge case worth handling: an async run can hang if the "completed" filter is unchecked. `finishAsync()` only fires when a terminal status comes over the stream, so without the completed envelope the spinner never stops and the EventSource stays open. Same category -- `es.onerror` isn't handled, so a dropped `/events` connection quietly auto-reconnects with the spinner stuck.

Force-including `completed` in the filter (or a client-side terminal fallback) plus an `onerror` -> `finishAsync()` covers both.

pkg/cli/playground/form.js
```diff
@@ -0,0 +1,327 @@
...
+      } catch (err) {
+        error.textContent = "Invalid JSON: " + err.message;
+        return "";
+      }
+    },
+  };
+}
+
+// arrayWidget renders a growable list of item widgets.
+function arrayWidget(root, items, initial) {
```
michaeldwan, line 284: This one's subtle and worth catching now: edits to an optional array-of-object field get silently dropped. The include checkbox auto-ticks on `input`/`change` or an editor's `onChange`, but `arrayWidget` doesn't expose an `onChange`, and Ace edits inside array rows don't bubble DOM events -- so editing an optional `list[dict]` never ticks the box, and `collect()` leaves it out. From the user's side they fill it in, hit Run, and it's just missing. Threading an `onChange` through the array items fixes it.

integration-tests/harness/harness.go
```diff
@@ -480,40 +492,153 @@ func (h *Harness) cmdCogServe(ts *testscript.TestScript, neg bool, args []string
 	}
 }
 
+// cmdCogPlayground implements background 'cog playground' for testscript.
+// It starts a cog playground process in the background and waits for it to be
+// ready. Usage: cog playground [flags]
+// Exports $PLAYGROUND_URL environment variable with the playground address.
+func (h *Harness) cmdCogPlayground(ts *testscript.TestScript, neg bool, args []string) {
```
michaeldwan, line 499: `cmdCogPlayground` is close to a copy of `cmdCogServe` -- port allocation, the env-build loop, proc-map + mutex tracking, wait-for-ready -- and `stopPlaygroundByWorkDir` is `stopServerByWorkDir` minus the `/shutdown` call. Totally reasonable way to get it working; I'd just pull a shared background-process helper the two commands parameterize before it drifts.

One real consequence of the copy: playground has no `/shutdown`, so cleanup falls to `Process.Kill()` (SIGKILL), which doesn't cancel the cobra context -- so the graceful-shutdown goroutine never runs and every playground test waits the full 5s before the kill. Sending SIGINT/SIGTERM first, then falling back to Kill, fixes the wait and gets the graceful path covered as a bonus.

pkg/cli/playground/output.js
```diff
@@ -0,0 +1,128 @@
...
+}
+
+function stringNode(value) {
+  const media = mediaNode(value);
+  if (media) return media;
+  if (value.startsWith("data:")) {
+    return el("a", { href: value, download: "output", text: "Download file" });
+  }
+  if (value.startsWith("http://") || value.startsWith("https://")) {
+    return el("a", { href: value, target: "_blank", text: value });
```
michaeldwan, line 110: Tiny nit, take it or leave it: add `rel="noopener noreferrer"` here. It's a model-controlled URL in a new tab, so the opened page gets a live `window.opener`. Modern browsers default to noopener for `target=_blank`, so this is just defense-in-depth rather than a real hole.

michaeldwan (commented):
Really nice first cut -- this is a genuinely useful tool and the Go side is already at the repo's bar. The command follows the `serve.go` conventions, the proxy/webhook/SSE design holds up well, and the eventHub locking is correct, which is easy to get wrong. Fun to review.

Most of my notes are on the frontend, and they're about long-term maintenance rather than anything being broken. The shape of it: we're adding ~1,600 lines of hand-written JS plus a 456KB vendored editor, and right now nothing in CI looks at any of it. The Go shell is well-tested; the JS -- which is where most of the logic lives -- ships untested and ungated, and that's where the few real bugs turned out to be.

To be clear up front: buildless is the right call, don't second-guess that instinct. The asks are (1) give the frontend the same format+test floor every other language here has, (2) let's talk about whether Ace needs to be 456KB for what we're doing, and (3) a bit of harness dedup. Then a few concrete bugs below. None of this is a hard blocker -- it's the stuff worth closing before we put more human eyes on it, and honestly it's a strong starting point to iterate from.

PR #3089, opened 2026-07-01
feat: concurrency decorator

### LINKED ISSUE #2811: Configure concurrency in predictor rather than in cog.yaml
It always felt a bit funny to me that concurrency was declared in cog.yaml, when it felt like a thing that was associated with the predictor. It's the `async` on `predict()` that marks whether a model can be concurrent, then max concurrency is sort like a default concurrency for that function.

This becomes more clear when we consider `train`, or other arbitrary functions if that's ever a thing. It's each function that has a concurrency, not the model as a whole.

Anyway, here's what Claude thinks it could look be if it was a decorator...

## Proposed API

```python
import cog

# Class-based predictor
    def setup(self):
        self.model = load_model()

    @cog.concurrent(max=5)
    async def predict(self, prompt: str = Input()) -> str:
        return await self.model.generate(prompt)

# Function-based predictor
@cog.concurrent(max=5)
async def predict(prompt: str = Input()) -> str:
    return await model.generate(prompt)
```

### Decorator behavior
- `@cog.concurrent(max=N)` attaches `__cog_concurrent_max__ = N` metadata to the decorated function.
- Applying `@cog.concurrent(max=N)` where `N > 1` to a synchronous function raises `TypeError` at decoration time.
- `@cog.concurrent` (bare, without arguments) is also supported and defaults to `max=1`.

### Precedence (highest to lowest)
1. `COG_MAX_CONCURRENCY` environment variable at runtime (operator override)
2. `cog.yaml` `concurrency.max` → baked into Dockerfile as `COG_MAX_CONCURRENCY` (deprecated, see below)
3. `@cog.concurrent(max=N)` decorator on the predict function
4. Default: `1`

Since `cog.yaml` `concurrency.max` works by setting the env var at build time, cases 1 and 2 are effectively the same mechanism. The decorator becomes the new default path for model authors.

### Deprecation of `cog.yaml` `concurrency` field
- When `concurrency` is present in `cog.yaml`, emit a deprecation warning during validation suggesting users move to `@cog.concurrent`.
- The field continues to work (no breaking change) — it still generates the `COG_MAX_CONCURRENCY` env var, which takes precedence over the decorator.

### Python SDK (`python/cog/`)
- New file `python/cog/concurrent.py` with the decorator implementation.
- Export `concurrent` from `python/cog/__init__.py`.

### Rust coglet (`crates/coglet-python/`)
- In `predictor.rs`: after loading the predictor, read `__cog_concurrent_max__` from the predict function (via `getattr`).
- In `lib.rs`: change `read_max_concurrency()` to accept the decorator value as a fallback, with env var taking priority.
- Thread the decorator value from `PythonPredictor::load()` through to `OrchestratorConfig`.

### Go CLI (`pkg/config/`)
- In `validate.go`: add a deprecation warning when `concurrency` is present in `cog.yaml`.

### Tests
- Python unit tests for the decorator (sync rejection, metadata attachment).
- Rust tests for the new resolution logic (env var > decorator > default).
- New integration test (`.txtar`) verifying end-to-end: decorator sets concurrency without `cog.yaml` field.

### Docs
- Document the `@cog.concurrent` decorator.
- Update concurrency docs to recommend the decorator as the primary approach.
- Regenerate `docs/llms.txt`.

## What does NOT change
- The `PermitPool`, slot transport, and orchestrator logic in `crates/coglet/` — they just receive `num_slots` from a different source.
- Async/sync detection logic in `predictor.rs`.
- Existing `cog.yaml` configs with `concurrency.max` continue to work.

anish-sahoo:
## Summary
- add @cog.concurrent for configuring async runner concurrency
- statically extract decorator concurrency and bake it through COG_MAX_CONCURRENCY with existing precedence
- enforce async requirements at decorator, build, and runtime; deprecate cog.yaml concurrency
- update concurrency docs, llms.txt, examples, and integration fixtures

## Example
```py
import cog

class Runner(cog.BaseRunner):
    @cog.concurrent(max=4)
    async def run(self, prompt: str = cog.Input()) -> str:
        return await generate(prompt)
```

## Testing
- go test ./pkg/schema/python ./pkg/config ./pkg/image
- golangci-lint run ./pkg/schema/python ./pkg/config ./pkg/image
- uv run --with pytest pytest python/tests/test_concurrent.py
- uv run --with ruff ruff check python/cog/__init__.py python/tests/test_concurrent.py
- PYO3_PYTHON="$(uv run python -c 'import sys; print(sys.executable)')" cargo test -p coglet-python parse_max_concurrency
- PYO3_PYTHON="$(uv run python -c 'import sys; print(sys.executable)')" cargo clippy --manifest-path crates/Cargo.toml -p coglet-python -- -D warnings
- prettier --check docs/deploy.md docs/environment.md docs/python.md docs/yaml.md examples/hello-concurrency/README.md
- git diff --check

Fixes #2811

### DIFF
--- a/crates/coglet-python/src/worker_bridge.rs
+++ b/crates/coglet-python/src/worker_bridge.rs
@@ -101,11 +101,12 @@ pub struct PythonPredictHandler {
     async_loop: Mutex<Option<Py<PyAny>>>,
     /// Handle to the asyncio loop thread for joining on shutdown.
     async_thread: Mutex<Option<JoinHandle<()>>>,
+    max_concurrency: usize,
 }
 
 impl PythonPredictHandler {
     /// Create a handler in prediction mode.
-    pub fn new(predictor_ref: String) -> Result<Self, SetupError> {
+    pub fn new(predictor_ref: String, max_concurrency: usize) -> Result<Self, SetupError> {
         let (loop_obj, thread) = Self::init_async_loop()?;
         Ok(Self {
             predictor_ref,
@@ -114,6 +115,7 @@ impl PythonPredictHandler {
             mode: HandlerMode::Predict,
             async_loop: Mutex::new(Some(loop_obj)),
             async_thread: Mutex::new(Some(thread)),
+            max_concurrency,
         })
     }
 
@@ -122,7 +124,7 @@ impl PythonPredictHandler {
     /// NOTE: For bug-for-bug compatibility with cog mainline, use new() instead.
     /// Cog mainline's training routes incorrectly use a predict-mode worker.
     #[allow(dead_code)]
-    pub fn new_train(predictor_ref: String) -> Result<Self, SetupError> {
+    pub fn new_train(predictor_ref: String, max_concurrency: usize) -> Result<Self, SetupError> {
         let (loop_obj, thread) = Self::init_async_loop()?;
         Ok(Self {
             predictor_ref,
@@ -131,6 +133,7 @@ impl PythonPredictHandler {
             mode: HandlerMode::Train,
             async_loop: Mutex::new(Some(loop_obj)),
             async_thread: Mutex::new(Some(thread)),
+            max_concurrency,
         })
     }
 
@@ -281,6 +284,11 @@ impl PredictHandler for PythonPredictHandler {
 
             let pred = PythonPredictor::load(py, &self.predictor_ref)
                 .map_err(|e| SetupError::load(e.to_string()))?;
+            if self.max_concurrency > 1 && !pred.is_async() {
+                return Err(SetupError::setup(
+                    "COG_MAX_CONCURRENCY > 1 requires an async run() or predict() method",
+                ));
+            }
 
             // Detect SDK implementation
             let sdk_impl = match py.import("cog") {
--- a/crates/coglet-python/src/worker_bridge.rs
+++ b/crates/coglet-python/src/worker_bridge.rs
@@ -711,3 +711,107 @@ fn output_to_json(output: coglet_core::PredictionOutput) -> serde_json::Value {
         coglet_core::PredictionOutput::Stream(v) => serde_json::Value::Array(v),
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use std::path::PathBuf;
+
+    use pyo3::types::PyList;
+    use tempfile::TempDir;
+
+    fn add_python_sdk_path(py: Python<'_>) {
+        py.run(
+            c"\
+import sys, types
+coglet = types.ModuleType('coglet')
+coglet.CancelationException = Exception
+sys.modules.setdefault('coglet', coglet)
+requests = types.ModuleType('requests')
+sys.modules.setdefault('requests', requests)
+",
+            None,
+            None,
+        )
+        .expect("failed to install coglet test stub");
+
+        let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        let sdk_path = manifest_dir
+            .parent()
+            .and_then(|p| p.parent())
+            .expect("crate should live under crates/coglet-python")
+            .join("python");
+        let sys = py.import("sys").expect("sys should import");
+        let path = sys
+            .getattr("path")
+            .expect("sys.path should exist")
+            .cast_into::<PyList>()
+            .expect("sys.path should be a list");
+        path.insert(0, sdk_path.to_string_lossy().as_ref())
+            .expect("failed to prepend SDK path");
+    }
+
+    fn write_predictor(source: &str) -> (TempDir, String) {
+        let dir = tempfile::tempdir().expect("failed to create temp dir");
+        let path = dir.path().join("predictor.py");
+        std::fs::write(&path, source).expect("failed to write test predictor");
+        let predictor_ref = format!("{}:Predictor", path.display());
+        (dir, predictor_ref)
+    }
+
+    // Runtime guard: an operator/env override of COG_MAX_CONCURRENCY > 1 must
+    // fail setup when the predictor is synchronous, mirroring the static
+    // parser and build-time checks.
+    #[tokio::test]
+    async fn setup_rejects_sync_predictor_with_concurrency() {
+        pyo3::Python::initialize();
+        Python::attach(add_python_sdk_path);
+
+        let (_dir, predictor_ref) = write_predictor(
+            r#"
+from cog import BaseRunner
+
+class Predictor(BaseRunner):
+    def run(self) -> str:
+        return "ok"
+"#,
+        );
+
+        let handler =
+            PythonPredictHandler::new(predictor_ref, 2).expect("handler should initialize");
+        let err = handler
+            .setup()
+            .await
+            .expect_err("setup should fail for a sync predictor with concurrency > 1");
+        assert!(
+            err.to_string()
+                .contains("COG_MAX_CONCURRENCY > 1 requires an async run() or predict() method"),
+            "unexpected error: {err}"
+        );
+    }
+
+    // The same guard must allow an async predictor with concurrency > 1.
+    #[tokio::test]
+    async fn setup_allows_async_predictor_with_concurrency() {
+        pyo3::Python::initialize();
+        Python::attach(add_python_sdk_path);
+
+        let (_dir, predictor_ref) = write_predictor(
+            r#"
+from cog import BaseRunner
+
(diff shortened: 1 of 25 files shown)


pkg/image/build.go
```diff
@@ -124,6 +125,10 @@ func Build(
 	// Generate schema before the Docker build so schema errors fail fast and the
 	// schema file is available in the build context.
 	var schemaJSON []byte
+	predictorInfo, err := generatePredictorMetadata(cfg, dir)
```
markphelps: This now runs before the `needsSchema` / `schemaFile` / `skipSchemaValidation` switch, so builds that intentionally provide an external schema or skip schema validation can still fail on static source parsing or target lookup. That feels like a behavioral regression for the opt-out paths. Could we either only require this metadata parse when we need decorator-derived concurrency, or make it best-effort/non-fatal when schema validation is explicitly skipped or supplied externally?

pkg/image/build.go
```diff
@@ -159,6 +164,11 @@ func Build(
 		}
 	}
 
+	dockerfileCfg, err := configWithDecoratorConcurrency(cfg, predictorInfo)
```
markphelps, line 178: This effective config is what the image build uses, but later the final image label still marshals the original `cfg`. If concurrency came from `@cog.concurrent`, the runtime env is right, but anything reading the Cog config label will see concurrency as unset. I think we should either marshal the effective config in the label path or add explicit effective-concurrency metadata so labels and runtime do not drift.

docs/python.md
```diff
@@ -128,7 +128,20 @@ class Runner(BaseRunner):
         return "hello world";
 ```
 
-Models that have an async `run()` function can run concurrently, up to the limit specified by [`concurrency.max`](yaml.md#max) in cog.yaml. Attempting to exceed this limit will return a 409 Conflict response.
+Models that have an async `run()` function can run concurrently. Use `@cog.concurrent(max=N)` to configure the default maximum concurrency for the function:
```
markphelps: Because this sits under the existing “Added in cog 0.14.0” note, this newly-added `@cog.concurrent` API now reads as if it has existed since 0.14.0. Could we split this up or reword it, e.g. async runners were added in 0.14.0 and `@cog.concurrent` was added in the current release? `docs/llms.txt` should be regenerated after the docs wording changes.

docs/yaml.md
```diff
@@ -189,8 +189,9 @@ build:
 ## `concurrency`
 
 > Added in cog 0.14.0.
+> Deprecated: use [`@cog.concurrent(max=N)`](python.md#async-runners-and-concurrency) on your async `run()` method instead.
 
-This stanza describes the concurrency capabilities of the model. It has one option:
+This stanza describes the concurrency capabilities of the model. It is still supported for backwards compatibility, but new models should use `@cog.concurrent(max=N)`. It has one option:
```
markphelps, line 194: For migration, it would help to explicitly say that this deprecated field still takes precedence over `@cog.concurrent`. Otherwise users can add the decorator and keep wondering why the YAML value is still the one that gets baked into the image. Maybe add: “If both are set, `concurrency.max` takes precedence; remove it when migrating to `@cog.concurrent`.”

python/cog/__init__.py
```diff
@@ -64,6 +65,41 @@ def decorate(inner: _F) -> _F:
...
+        if max > 1 and not (
+            _inspect.iscoroutinefunction(inner) or _inspect.isasyncgenfunction(inner)
+        ):
+            raise TypeError("concurrent max greater than 1 requires an async function")
+        inner.__cog_concurrent_max__ = max  # type: ignore[attr-defined]
+        return inner
+
+    if fn is None:
+        return decorate
+    return decorate(fn)
```
markphelps, line 104: The static parser rejects positional usage like `@concurrent(2)`, but the runtime API currently falls through with `fn=2` and produces an `AttributeError` when `decorate()` tries to set `__cog_concurrent_max__`. Could we add a `callable(fn)` check before this return and raise a clearer `TypeError`, something like `concurrent must be used as @concurrent or @concurrent(max=...)`?

crates/coglet-python/src/worker_bridge.rs
```diff
@@ -281,6 +284,11 @@ impl PredictHandler for PythonPredictHandler {
 
             let pred = PythonPredictor::load(py, &self.predictor_ref)
                 .map_err(|e| SetupError::load(e.to_string()))?;
+            if self.max_concurrency > 1 && !pred.is_async() {
```
markphelps, line 287: This runtime guard is the last line of defense for operator/env overrides, but I don't see a direct test that sets `COG_MAX_CONCURRENCY=2` with a sync predictor and asserts setup fails with this message. The parser/build tests cover related paths, but a small coglet-python or integration test for this worker setup failure would make the runtime contract less fragile.

pkg/schema/python/annotations.go
```diff
@@ -243,6 +244,73 @@ func functionSupportsStreaming(node *sitter.Node, source []byte, imports *schema
 	return false
 }
 
+func functionIsAsync(node *sitter.Node, source []byte) bool {
+	return strings.HasPrefix(strings.TrimSpace(Content(node, source)), "async def ")
```
markphelps: This async detection is a bit brittle because it depends on the exact text prefix `async def `. Valid Python with different whitespace, like `async  def`, would be parsed as non-async and then fail `@concurrent(max > 1)`. Could we use the tree-sitter structure/token instead, or at least a regexp like `^async\s+def\s+`?

markphelps (commented):
I did a focused pass on the concurrency decorator changes. The happy path looks coherent; these are the spots I think are worth tightening before merge.

pkg/image/build.go
anish-sahoo: Good catch. The predictor metadata parse still runs before the schema switch, but it is now only fatal when \`needsSchema\` is true. When schema validation is skipped or an external schema is supplied, a parse failure is logged at debug and we fall back to \`nil\` (cog.yaml-only concurrency) rather than failing the build. Fixed in cf334653.

pkg/image/build.go
anish-sahoo: Fixed in cf334653: the config label now marshals the effective \`dockerfileCfg\` instead of the original \`cfg\`, so decorator-derived concurrency in the label stays consistent with the baked \`COG_MAX_CONCURRENCY\` runtime env.

docs/python.md
anish-sahoo: Reworded in cf334653: the note now reads "Async runners were added in cog 0.14.0. The \`@cog.concurrent\` decorator was added in cog 0.21.0.", and the decorator paragraph also calls out 0.21.0. Regenerated \`docs/llms.txt\`.

docs/yaml.md
anish-sahoo: Added in cf334653: "If both \`concurrency.max\` and \`@cog.concurrent(max=N)\` are set, \`concurrency.max\` takes precedence and is the value baked into the image. Remove \`concurrency.max\` when migrating to \`@cog.concurrent\`." Regenerated \`docs/llms.txt\`.

python/cog/__init__.py
anish-sahoo: Fixed in cf334653: added a \`callable(fn)\` guard that raises \`TypeError("concurrent must be used as @concurrent or @concurrent(max=...)")\` before the \`decorate(fn)\` return, so positional usage like \`@concurrent(2)\` gives a clear error instead of an \`AttributeError\`. Added \`test_concurrent_rejects_positional_argument\`.

crates/coglet-python/src/worker_bridge.rs
anish-sahoo: Added in cf334653: a \`tests\` module in \`worker_bridge.rs\` with \`setup_rejects_sync_predictor_with_concurrency\` (creates a handler with \`max_concurrency=2\` + a sync predictor, calls \`setup().await\`, and asserts it fails with the guard message) plus \`setup_allows_async_predictor_with_concurrency\` for the passing case.

pkg/schema/python/annotations.go
anish-sahoo: Fixed in cf334653: \`functionIsAsync\` now uses the tree-sitter parse tree (\`UnwrapFunction\` + scanning for the anonymous \`async\` token) instead of the \`"async def "\` text prefix. Verified this handles whitespace variants like \`async  def\`.

markphelps (approved):
nice!

PR #3094, opened 2026-07-06
reject file path input defaults

### LINKED ISSUE #606: Disallow setting `default` for files/paths
For example, this ain't gonna work:

```
def predict(self, file: Path = Input(default=Path("image.png")):
```

The thing consuming the schema (e.g. Replicate) can't get that default. We should just not allow this until we've figured out a better solution here.

As an aside: just setting the default to string is allowed but shouldn't be. Maybe we should validate default values to check they're actually the correct type.

Reported by @yael-vinker (thank you!)

sarptandoven:
summary
- reject non-`None` defaults for `Path` and deprecated `File` inputs during static schema parsing
- keep `default=None` working for optional file/path inputs
- track unresolved `Input(default=...)` expressions so `Path("...")` defaults no longer silently become `None`
- add regression coverage for string, constructor, and list defaults

why
- closes #606
- generated schemas cannot reliably represent file/path defaults today
- silently accepting these defaults can make the schema claim a usable default that downstream consumers cannot provide

validation
- `go test ./pkg/schema/python -run 'TestPathAndFileInputsRejectDefaults|TestOptionalInputPipeNone|TestMultipleInputsWithDefaults' -count=1`
- `go test ./pkg/schema/... -count=1`
- `go test ./pkg/config ./pkg/util -count=1`
- `git diff --check`

note
- `go test ./pkg/... -count=1` was also attempted. The schema/config/util packages passed, but Docker-dependent packages failed because this machine cannot connect to a Docker daemon at `unix:///var/run/docker.sock` / rootless Docker is unavailable.

### DIFF
--- a/pkg/schema/python/inputs.go
+++ b/pkg/schema/python/inputs.go
@@ -9,15 +9,16 @@ import (
 )
 
 type inputCallInfo struct {
-	Default     *schema.DefaultValue
-	Description *string
-	GE          *float64
-	LE          *float64
-	MinLength   *uint64
-	MaxLength   *uint64
-	Regex       *string
-	Choices     []schema.DefaultValue
-	Deprecated  *bool
+	Default           *schema.DefaultValue
+	DefaultUnresolved bool
+	Description       *string
+	GE                *float64
+	LE                *float64
+	MinLength         *uint64
+	MaxLength         *uint64
+	Regex             *string
+	Choices           []schema.DefaultValue
+	Deprecated        *bool
 }
 
 type inputMethodInfo struct {
@@ -216,6 +217,11 @@ func resolveInputReference(node *sitter.Node, source []byte, registry *inputRegi
 			case "default":
 				if val, ok := parseDefaultValue(callNode, source); ok {
 					resolved.Default = &val
+					resolved.DefaultUnresolved = false
+				} else {
+					none := schema.DefaultValue{Kind: schema.DefaultNone}
+					resolved.Default = &none
+					resolved.DefaultUnresolved = true
 				}
 			case "description":
 				if s, ok := parseStringLiteral(callNode, source); ok {
@@ -283,6 +289,14 @@ func inputFieldWithInfo(name string, order int, inputType schema.InputType, fiel
 	field.Choices = info.Choices
 	field.Deprecated = info.Deprecated
 	return field
+}
+
+func validateInputFieldWithInfo(field schema.InputField, info inputCallInfo) error {
+	if info.DefaultUnresolved && field.InputType != nil && schema.InputTypeContainsFileOrPath(*field.InputType) {
+		return schema.WrapError(schema.ErrDefaultNotResolvable,
+			fmt.Sprintf("parameter '%s': Path and File defaults cannot be statically resolved", field.Name), nil)
+	}
+	return schema.ValidateInputField(field)
 }
 
 func firstParamIsSelf(params *sitter.Node, source []byte) bool {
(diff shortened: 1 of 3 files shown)


pkg/schema/python/inputs.go
```diff
@@ -216,6 +217,11 @@ func resolveInputReference(node *sitter.Node, source []byte, registry *inputRegi
 			case "default":
 				if val, ok := parseDefaultValue(callNode, source); ok {
 					resolved.Default = &val
+					resolved.DefaultUnresolved = false
+				} else {
+					none := schema.DefaultValue{Kind: schema.DefaultNone}
+					resolved.Default = &none
```
markphelps, line 223: Nice catch overall. I think this branch needs one small fix here before it lands: this `else` changes behavior outside the Path/File case.

`resolved` starts as a copy of `methodInfo.BaseInfo`, so it can already have a default from the helper method. Before this PR, if a call-site default was not a literal, we left that base default alone. Now we replace it with `None`.

I verified this with a non-file input like:

```python
MY = "base.png"

class Helper:
    @staticmethod
    def img(default: str) -> Input:
        return Input(description="Image", default="base.png")

    def predict(self, image: str = Helper.img(default=MY)) -> str:
        return "ok"
```

On this branch the parsed default becomes `None`; before the PR it stayed `"base.png"`. That is a silent schema change for non-file inputs.

Could we keep the new unresolved marker without clobbering the base default?

```go
} else {
    // Preserve the base default for non-file types; file/path types are still
    // rejected by validateInputFieldWithInfo via DefaultUnresolved.
    resolved.DefaultUnresolved = true
}
```

I checked this locally: it preserves the old non-file behavior, and file/path defaults through this path still get rejected because `validateInputFieldWithInfo` sees `DefaultUnresolved`.

pkg/schema/python/parser_test.go
```diff
@@ -361,6 +361,66 @@ class Predictor(BasePredictor):
 	require.False(t, scale.IsRequired())
 }
 
+func TestPathAndFileInputsRejectDefaults(t *testing.T) {
```
markphelps, line 364: Can we add one regression test for the static-method/reference path too? The new table is good for direct `Input(...)`, but it would not have caught the default-clobbering bug above because that lives in `resolveInputReference`.

A useful test would be the non-file helper-method case from the other comment: helper returns `Input(default="base.png")`, caller passes `default=MY`, and we assert the parsed default is still `"base.png"`.

It would also be worth adding one Path/File case through this path, e.g. `image: Path = Helper.img(default=Path("x.png"))`, to prove the new `DefaultUnresolved` rejection still works for registry references.

pkg/schema/python/parser_test.go
```diff
@@ -361,6 +361,66 @@ class Predictor(BasePredictor):
...
+        return "ok"
+`,
+			wantErr: schema.ErrUnsupportedType,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			se := parseErr(t, tt.source, "Predictor", schema.ModePredict)
+			require.Equal(t, tt.wantErr, se.Kind)
```
markphelps, line 419: Small optional tightening: these assertions would be a little stronger if the table also checked a message substring.

The `Kind` assertion is meaningful, but a `wantMsg` column would pin the two user-facing paths too:

- literal Path/File default: `defaults are not supported on Path or File inputs`
- unresolvable constructor default: `Path and File defaults cannot be statically resolved`

Not a blocker, just a cheap way to make the regression tests more explicit.

pkg/schema/types.go
```diff
@@ -245,9 +245,38 @@ func ValidateInputField(field InputField) error {
...
 		}
 	}
+	if inputFieldContainsFileOrPath(field) && field.Default != nil && field.Default.Kind != DefaultNone {
+		return errUnsupportedType("defaults are not supported on Path or File inputs")
+	}
 	return nil
 }
 
+// InputTypeContainsFileOrPath reports whether a recursive input type contains
+// a Path or File primitive. It is used by parser frontends that need to reject
```
markphelps, line 255: Tiny doc nit: this helper is also used by `ValidateInputField` through `inputFieldContainsFileOrPath`, not just by parser frontends.

Maybe:

```go
// It is used by ValidateInputField and parser frontends that need to reject
// unsupported file/path defaults after annotation resolution.
```

Optional, but it would make the comment match the actual call sites.

markphelps (changes requested):
Thanks for the fix — the main Path/File default handling looks pointed in the right direction, and the direct `Input(...)` cases are covered well. I found one regression in the static-method/reference path that changes behavior for non-file inputs, so I think this needs a small tweak before merge. The other comments are test/doc follow-ups that should make this safer to maintain.

PR #3095, opened 2026-07-07
feat: support local package artifacts in requirements

### LINKED ISSUE #1945: Installing local packages during cog build
Hello,

I am trying to install a local package during the cog build process. Specifically, I would like to install a package from a local file like this:

`pip install mylibpackage.zip`

I tried adding it to `requirements.txt` and also specifying it in the run command inside my `cog.yaml` file, like so:

```
run:
  - pip install mylibpackage.zip
```

But no luck so far. 😕

Is there a proper way to install local packages during the cog build? Maybe I'm missing something or it’s just not supported?

Any help or direction would be greatly appreciated!

Thank you!

anish-sahoo:
## Summary

- Add support for bare local wheel and source archive paths in `build.python_requirements`.
- Resolve paths relative to the requirements file and stage only referenced artifacts through `cog_build` before dependency installation.
- Keep config loading and custom Dockerfile builds unchanged by applying local artifact validation only to Cog-generated Dockerfiles.
- Preserve requirement filenames, confine symlink and staging access to project/build roots, and restore Cog's managed runtime after user artifacts are installed.
- Document the supported syntax and add unit and integration coverage for wheels, ZIP source archives, tarballs, and paths containing spaces.

Fixes #1945

## Usage

Local artifacts can be listed directly in a requirements file:

```txt
./dist/mylib-0.1.0-py3-none-any.whl
./vendor/helper library.zip
./packages/localpkg.tar.gz
```

Cog supports `.whl`, `.zip`, `.tar.gz`, `.tgz`, `.tar.bz2`, and `.tar.xz`. Paths may contain spaces, are resolved relative to the requirements file, and must point to regular files inside the project after symlink resolution.

Remote URLs, remote direct references, and remote requirements options continue to pass through unchanged.

## Limits

- `build.run: pip install ./artifact.zip` does not stage local source files.
- Local package directories are not supported.
- Local file URLs and local direct references such as `name @ path` are rejected.
- Local `--find-links` directories and recursive local requirements includes are rejected.
- Inline markers, extras, hashes, and options on local artifact lines are rejected.
- Any `cog` or `coglet` distribution installed by a local artifact is overridden by Cog's configured runtime. Use `build.sdk_version`, `COG_SDK_WHEEL`, or `COGLET_WHEEL` to select those packages.

## Testing

- `gotestsum -- -short -timeout 1200s -parallel 5 ./...`
- `golangci-lint run ./...`
- `go test -tags integration -run "TestIntegration/local_python_requirement_artifact" ./integration-tests/...`


markphelps (approved):
lgtm x 3

PR #3096, opened 2026-07-07
fix: delete coglet-managed output files after upload

### LINKED ISSUE #1434: Delete output paths after upload
Currently, you hand off the output file to cog and then none of your code runs and the file gets left in place.
However, there's a class of bug where models can return outputs for previous predictions if some overwrite fails.

This also stops us from using tmp files as output files, since they never get deleted. Cog should probably delete those files (or even give the user a temp directory/overlay to work in).

anish-sahoo:
Fixes #1434.

- Treats every predictor-returned `Path` as an ownership handoff. Coglet consumes the source entry during staging, including paths outside temporary storage, while preserving symlink targets.
- Deletes managed output files after inline encoding or a successful upload. Prediction-directory cleanup handles files left by failures and cancellation.
- Preserves structured outputs with nested files and their original upload names. Reserved filename characters are escaped as one URL path segment without changing signed query parameters.
- Streams completed file uploads before prediction completion while preserving protocol output order. Pending tasks are still joined or aborted before terminal state and ID reuse.
- Validates and atomically reserves prediction IDs. Idempotent prediction and training requests return the existing prediction during concurrent races.
- Binds cancellation to the exact prediction instance so delayed SSE or sync guards cannot cancel a later prediction that reused the same public ID.
- Documents the returned-path ownership contract and cleanup behavior.


crates/coglet/src/service.rs
```diff
@@ -797,9 +796,20 @@ impl PredictionService {
...
     }
 
-    /// Remove a prediction from the DashMap after completion.
+    /// Remove a prediction from the DashMap after completion and clean up
+    /// its on-disk prediction directory as a backstop for any output files
+    /// that were not deleted individually (e.g. aborted uploads, cancelled
+    /// predictions).
     pub fn remove_prediction(&self, id: &str) {
         self.predictions.remove(id);
+        let dir = prediction_dir_for(id);
```
michaeldwan: `id` comes from `PredictionRequest.id` without validation. An absolute ID makes `PathBuf::join` discard `/tmp/coglet/predictions`, so `POST /predictions` with `{"id":"/some/writable/directory"}` reaches `remove_dir_all` on that directory. The route calls cleanup even when `predict()` fails. We need to reject path separators and traversal at the API boundary, and keep a defense here that verifies the result is a direct child of the prediction root before recursively deleting it.

crates/coglet/src/service.rs
```diff
@@ -797,9 +796,20 @@ impl PredictionService {
...
+        match std::fs::remove_dir_all(&dir) {
```
michaeldwan: This cleanup is keyed only by the public prediction ID. `submit_prediction` currently uses an unconditional `DashMap::insert`, and POST doesn’t reject duplicate active IDs. With concurrency greater than one, the first of two same-ID predictions to finish removes the second prediction’s map entry and shared working directory while it may still be writing. Registration needs an atomic check-and-insert, or the working directory needs a per-submission identity, before we add recursive cleanup here.

crates/coglet-python/src/predictor.rs
```diff
@@ -195,7 +195,7 @@ fn send_output_item(
             .and_then(|p| p.extract())
             .map_err(|e| PredictionError::Failed(format!("Failed to get fspath: {}", e)))?;
         slot_sender
-            .send_file_output(std::path::PathBuf::from(path_str), None)
+            .send_file_output(std::path::PathBuf::from(path_str), None, false)
```
michaeldwan: Passing `false` means a predictor-returned `Path` is never deleted. That’s the exact file #1434 describes: the predictor hands it to Cog, can’t run cleanup afterward, and a later failed overwrite can return stale output. It also contradicts `docs/python.md`, which says the returned file is automatically deleted. Cleaning IOBase copies and oversized spills is useful, but it doesn’t close #1434. We need a safe ownership design for returned paths -- such as a Cog-owned output directory -- or we should narrow this PR’s scope and stop claiming it closes that issue. The same applies to the call in `process_single_output`.

michaeldwan (changes requested):
Thanks for taking this on. Cleaning up Coglet-created spill files is worthwhile, and making ownership explicit in the protocol is a useful direction.

I’m requesting changes because the recursive cleanup currently trusts a client-controlled prediction ID, duplicate IDs can delete a live prediction’s state, and predictor-returned paths -- the files described in #1434 -- remain intentionally undeleted. I’ve left details inline.

Please also add behavioral coverage around cleanup. The snapshots prove that `managed` serializes, but they don’t prove that managed files are deleted, that returned paths follow the intended contract, or that one prediction can’t remove another prediction’s directory.

crates/coglet/src/service.rs
anish-sahoo: Updated. Prediction IDs are validated at every API entry point and again before cleanup, rejecting empty IDs, path separators, and the `.`/`..` path components. `remove_prediction` verifies normal directories are direct children of the prediction root, safely unlinks final-component symlinks without following them, and retains the ID reservation if cleanup fails. Coverage includes absolute/traversal targets, symlink targets, cleanup failures, and accepted safe dotted IDs.

crates/coglet/src/service.rs
anish-sahoo: Updated. `submit_prediction` now atomically reserves the ID before acquiring a slot permit, so duplicate active IDs return `CreatePredictionError::DuplicateId` / HTTP 409 without consuming capacity. Cleanup holds that reservation until the prediction directory is removed, preventing a reused ID from losing new files to an older cleanup. Tests verify duplicate rejection, preserved permit capacity, and cleanup-failure reservation behavior.

crates/coglet-python/src/predictor.rs
anish-sahoo: Updated. Returning a `Path` now explicitly transfers ownership to Coglet: the worker copies it into the per-prediction managed output directory, removes the original only after the copy succeeds, and marks the managed copy for deletion immediately after the parent reads it. This fixes the stale fixed-path case from #1434 while working across filesystems. Coverage verifies Python `Path` handoff, source removal, stale-path rejection, managed deletion, and unmanaged preservation.

anish-sahoo:
Follow-up summary for Michael’s review:

- Returned `Path` values now transfer ownership to Coglet. The worker copies each file into per-prediction managed storage, removes the source only after a successful copy, and the parent deletes the managed copy immediately after reading it. This directly covers the stale fixed-path case in #1434.
- Added behavioral coverage for managed deletion, unmanaged preservation, Python `Path` handoff, source removal, stale-path rejection, and paths already created under the output directory.
- Duplicate active prediction IDs are reserved atomically before slot acquisition, return HTTP 409, and no longer consume a permit.
- Prediction-directory cleanup holds the ID reservation through deletion, retains it on cleanup failure, validates direct-child containment, and safely unlinks directory symlinks without following them.
- Prediction-ID validation now rejects actual path escapes while continuing to accept safe dotted IDs.
- Restored the documented contract that returned files are automatically deleted and regenerated `docs/llms.txt`.

Validation completed:

- Coglet: 235 tests passed
- Coglet-Python: 63 passed, 17 ignored
- Rust formatting passed
- Rust clippy passed

crates/coglet/src/service.rs
```diff
@@ -797,9 +807,53 @@ impl PredictionService {
         }
     }
 
-    /// Remove a prediction from the DashMap after completion.
+    /// Remove a prediction from the DashMap after completion and clean up
+    /// its on-disk prediction directory as a backstop for any output files
+    /// that were not deleted individually (e.g. aborted uploads, cancelled
+    /// predictions).
     pub fn remove_prediction(&self, id: &str) {
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 45d06415. Cleanup now holds the DashMap entry while renaming the prediction directory to a UUID path under the cleanup root. The ID is released only after a successful detach or confirmation that no directory exists, and recursive deletion runs against the detached path. If detach fails, the ID remains reserved so a later prediction cannot reuse stale files. Tests cover detach failure, regular files, directories, and symlinks.

crates/coglet/src/worker.rs
```diff
@@ -217,12 +223,40 @@ impl SlotSender {
...
+        if let Err(e) = std::fs::remove_file(&path) {
+            let _ = std::fs::remove_file(&dest);
+            return Err(e);
+        }
+        self.send_file_output(dest, mime_type, true)
+    }
+
     /// Send a user metric to the parent process.
     pub fn send_metric(
         &self,
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 45d06415. Returned files are now consumed when both the directory entry and resolved target are under a scratch root. Coglet uses rename when possible, with copy-then-unlink across filesystems and for symlinks. Files outside scratch roots are copied so bundled assets and shared caches survive. Tests cover temp source removal, stale-path rejection, external-file reuse, relative symlinks, and the Python-to-Rust Path handoff. I also updated the PR description to state this policy.

crates/coglet/src/worker.rs
```diff
@@ -199,30 +241,202 @@ impl SlotSender {
...
+                ),
+            ));
+        }
+
+        let extension = path
+            .extension()
+            .and_then(|e| e.to_str())
+            .unwrap_or("bin")
+            .to_string();
+        let upload_filename = path
```
markphelps, line 312: Preserving the original basename makes this user-controlled now. `upload_file()` later concatenates this directly onto the endpoint, so otherwise-valid names like `result?draft.png`, `result#1.png`, or `report%20final.txt` are interpreted as URL syntax rather than an object-key segment. That can upload to the wrong key (or fail) and return the wrong URL.

Can we append it as one escaped URL path segment instead? A regression test with reserved characters in a returned filename would catch this.

crates/coglet/src/orchestrator.rs
```diff
@@ -1105,35 +1205,40 @@ async fn run_event_loop(
...
-                                                    && let Some(mut p) = try_lock_prediction(&pred)
-                                                {
-                                                    p.append_output(serde_json::Value::String(url));
-                                                }
-                                            }
-                                            Err(e) => {
-                                                tracing::error!(error = %e, "Failed to upload file output");
-                                            }
-                                        }
+                                        let url = upload_file(
```
markphelps, line 1273: This changes streaming behavior for file outputs. The task now only returns a `ResolvedOutput`; we keep it in `pending_outputs` and do not call `append_output_chunk` until the worker sends `Done` and every output task has completed.

For a generator that yields a `Path` and then keeps working, the SSE client will not see that yielded value until the prediction finishes. Before this change, a completed file upload was appended as soon as it completed. The architecture doc still says yielded `FileOutput`s stream immediately.

Could we publish each successful resolved output as its task completes, while still retaining and joining the task handles before terminal state? Please add an SSE regression test that yields a path, blocks before `Done`, and asserts the output arrives during the block.

crates/coglet/src/transport/http/routes.rs
```diff
@@ -488,6 +530,19 @@ async fn create_prediction_with_id(
             )
                 .into_response();
         }
+        Err(CreatePredictionError::DuplicateId) => {
```
markphelps: There is still a check-then-act race for idempotent `PUT`s: two requests can both miss the entry at `get_prediction_response`, then one reserves the ID and the other reaches this branch. The second request gets `409`, even though an idempotent retry should return the prediction that just won the race.

Could the idempotent handlers re-read and return the existing prediction on `DuplicateId` (including the SSE path), or move this into an atomic get-or-create service operation? This applies to the training idempotent endpoint too. A concurrent `PUT /predictions/{id}` regression test would be useful.

crates/coglet/src/worker.rs
anish-sahoo: Fixed in 6d9bd832. Upload filenames are appended as one escaped URL path segment, and signed query parameters are preserved. The regression test uses `result?draft.png` with a signed query.

crates/coglet/src/orchestrator.rs
anish-sahoo: Fixed in 6d9bd832. Completed uploads now publish immediately through an ordered output publisher, so SSE keeps protocol order without waiting for `Done`. A subprocess test yields a path, blocks before completion, and confirms the output arrives first.

anish-sahoo:
Pushed `6d9bd832` with the latest review fixes. Returned paths are now always consumed, file outputs stream in order before completion, signed upload URLs preserve their query parameters, and idempotent requests and cancellation are safe across ID reuse. The new tests cover the live SSE path, concurrent PUT races, reserved filenames, and stale cancellation guards.

crates/coglet/src/transport/http/routes.rs
anish-sahoo: Fixed in 6d9bd832. Duplicate reservations now carry a stable handle to the existing prediction, so cleanup cannot make the retry fall back to `409`. Barrier-based tests cover concurrent prediction and training PUTs.

crates/coglet-python/tests/test_coglet.py
```diff
@@ -363,23 +365,118 @@
...
+    uploads = []
+
+    class UploadHandler(BaseHTTPRequestHandler):
+        def do_PUT(self) -> None:
+            length = int(self.headers.get("Content-Length", "0"))
+            uploads.append((self.path, self.rfile.read(length)))
+            self.send_response(200)
+            self.send_header(
+                "Location",
+                f"http://127.0.0.1:{self.server.server_port}{self.path}",
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 6d9bd832. The test server now quotes the request path before using it in the `Location` header, so request data cannot split the response headers.

PR #3097, opened 2026-07-09
feat: model input validation before build and run

anish-sahoo:
Validate `cog predict`, `cog run`, and `cog train` inputs against the model's OpenAPI schema before building the image or starting the container. Bad inputs fail immediately instead of after a full build or container start.

- Generate the schema statically from `cog.yaml` in Go — no Docker or Python needed.
- For existing images, read the schema from the `openapi_schema` label after pull, before start.
- Images without a usable label fall back to the runtime schema after start (unchanged).
- Reject unknown input names to catch typos; the runtime only warns.
- Report missing required inputs.
- Report type mismatches, e.g. `expected integer, got string`.
- Reject invalid enum values and list the allowed ones.
- Enforce numeric bounds, string length, and regex.
- Reject explicit `null` for typed fields, matching runtime semantics.
- Coerce `-i` string values to schema types before validating.
- Name the offending input on file-read failures.
- Add `pkg/schema/openapi.GenerateSchema`, one config-aware entry point shared by build and CLI preflight.
- Enforce a minimum SDK version for static schema generation.
- Rename `pkg/schema/openapi.go` → `openapi_spec.go`.
- Add unit and integration tests; update architecture docs.

---

Notes:
- currently has code to support openapi 3.1, so a few extra if statements that are marked as defensive checks for now. later if we add more things to the type system these checks will get used
- right now the only change in behavior between api and cli is that in the cli unknown inputs are rejected. this is an intentional change because we want the cli to reject invalid input keys (and warn about missing proper inputs). In the api it ignores all unknown fields.
- refactored out the openapi schema generation and spec validation so that we can later extend this to json schema/ajv if needed


pkg/cli/train.go
```diff
@@ -58,20 +58,18 @@ func cmdTrain(cmd *cobra.Command, args []string) error {
...
-	}
-
 	imageName := ""
 	volumes := []command.Volume{}
 	gpus := gpusFlag
+	var preparedInputs predict.Inputs
+	inputsPrepared := false
+	var dockerClient command.Command
+	var err error
+	var m *model.Model
```
markphelps:
```suggestion
var (
    imageName  = ""
    volumes = []command.Volume{}
    gpus = gpusFlag
    preparedInputs predict.Inputs
	inputsPrepared = false
    dockerClient command.Command
    err error
    m *model.Model
)
```

nit style: i generally prefer to use var blocks if declaring two or more variables right near each other, and also prefer not mixing `var` and implicit syntax

pkg/cli/predict.go
```diff
@@ -332,62 +371,96 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
 }
 
-func isURI(ref *openapi3.Schema) bool {
-	return ref != nil && ref.Type.Is("string") && ref.Format == "uri"
+func prepareIndividualInputs(inputFlags []string, schema *openapi3.T, isTrain bool) (predict.Inputs, error) {
+	inputs, err := parseInputFlags(inputFlags, schema, isTrain)
+	if err != nil {
+		return nil, err
+	}
+	if err := predict.ValidateInputsForMode(inputs, schema, isTrain); err != nil {
```
markphelps, line 426: I think this new preflight validation exposes a boolean regression for `-i` inputs. `NewInputsForMode` does not currently coerce schema boolean values, so `-i flag=true` stays a string and this returns `expected boolean, got string` before runtime gets a chance to normalize it. I would add bool coercion (`strconv.ParseBool`) and cover `cog run -i flag=true` / existing-image `cog predict -i flag=false`.

markphelps, line 426: Same area: repeated `-i` values for numeric arrays look like they will now fail preflight. `NewInputsForMode` stores repeated values as raw strings, so `-i nums=1 -i nums=2` for `list[int]` validates as `["1", "2"]`. That feels like existing CLI behavior we should preserve by coercing repeated scalar values against the array item schema, or very explicitly making numeric arrays JSON-only.

pkg/cli/predict.go
```diff
@@ -332,62 +371,96 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
 	jsonInputs, err := parseJSONInput(jsonInput)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	if err := predict.ValidateInputNamesForMode(jsonInputs, schema, isTrain); err != nil {
+		return nil, err
 	}
-
 	transformedInputs, err := transformPathsToBase64URLs(jsonInputs)
```
markphelps, line 441: Could we include the input name in JSON-mode file read errors here? The `-i image=@missing.png` path now reports the offending input name, but `--json {"image":"@missing.png"}` comes back as just `Failed to read file ...`. Since this function has the key, something like `input "image": failed to read file ...` would make the two modes consistent.

pkg/cli/predict.go
```diff
@@ -260,11 +288,22 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
-		m, err := resolver.Pull(ctx, ref)
+		m, err = resolver.Pull(ctx, ref)
 		if err != nil {
 			return err
 		}
 
+		// Preflight-validate inputs when the image ships an OpenAPI schema
+		// label. Older images without the label fall back to the runtime
+		// schema fetched from the container after it starts.
+		if m.Schema != nil {
```
markphelps: I would make this fallback a little less binary. `m.Schema != nil` only proves the label parsed, not that it has the `Input` component this validator needs. For older/third-party images with a minimal or malformed-but-parseable `run.cog.openapi_schema` label, we will fail here instead of falling back to the runtime schema. Could we check for the relevant input component first, and only set `inputsPrepared` when the label is actually usable? Same thought applies in `train.go`.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,243 @@
...
+			return fmt.Errorf("unknown input %q; model does not accept inputs", unknown[0])
+		}
+		quoted := make([]string, len(unknown))
+		for i, key := range unknown {
+			quoted[i] = fmt.Sprintf("%q", key)
+		}
+		return fmt.Errorf("unknown inputs %s; model does not accept inputs", strings.Join(quoted, ", "))
+	}
+	if len(unknown) == 1 {
+		return fmt.Errorf("unknown input %q; valid inputs are: %s", unknown[0], strings.Join(valid, ", "))
```
markphelps: Optional/product decision: this intentionally turns unknown inputs into hard errors, while the runtime currently prints `WARNING unknown input field ignored` and continues. I am okay with the stricter behavior if that is what we want, but it is a user-visible behavior change. Worth documenting and ideally aligning runtime behavior later so CLI preflight and server behavior do not diverge.

pkg/cli/predict.go
```diff
@@ -332,62 +371,96 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
+		}
+		preparedInputs, needsJSON, err = prepareInputs(inputFlags, inputJSON, schema, false)
+		if err != nil {
+			return err
 		}
+	}
+
+	return runPrediction(*predictor, preparedInputs, outPath, false, needsJSON)
+}
+
```
markphelps, line 389: Optional maintainability thought: local-source runs now generate/validate the OpenAPI schema here, then the build path generates/validates it again. That is probably fine short-term, but it creates two chances for drift in one invocation. Longer term, it would be cleaner to generate once and pass the schema bytes/doc into the build path, or have build return the parsed schema used for labels.

pkg/cli/predict.go
```diff
@@ -332,62 +371,96 @@ func cmdPredict(cmd *cobra.Command, args []string) error {
...
+	if err := predict.ValidateInputNamesForMode(jsonInputs, schema, isTrain); err != nil {
+		return nil, err
 	}
-
 	transformedInputs, err := transformPathsToBase64URLs(jsonInputs)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	if err := predict.ValidateInputMapForMode(transformedInputs, schema, isTrain); err != nil {
```
markphelps, line 445: Optional compatibility audit: this validates JSON inputs strictly as OpenAPI JSON, which is sometimes stricter than runtime normalization. Examples worth deciding on explicitly are numeric strings for numeric fields and `Any` inputs whose schema is currently object-shaped. If the goal is fail-fast with the same semantics as runtime, we may need to loosen/coerce before this point; if the goal is stricter CLI JSON, let us document that.

integration-tests/tests/string_predictor.txtar
```diff
@@ -8,7 +8,7 @@ stdout 'hello world'
 
 # Missing required input fails
 ! cog predict $TEST_IMAGE -i wrong=value
-stderr 's: Field required'
+stderr 'unknown input "wrong"'
```
markphelps, line 11: Optional test coverage: this changed the old missing-required assertion into an unknown-input assertion. The new unknown-input coverage is useful, but I would add a separate integration case for omitting a required input entirely (`! cog run` / `! cog predict $TEST_IMAGE`) and assert `missing required input ...` plus the expected no-build/no-start behavior.

pkg/cli/train.go
anish-sahoo: Done in bf16d884 — consolidated into a single `var` block. Used `imageName string` rather than `= ""` to keep `ineffassign` happy (string zero value).

pkg/cli/predict.go
anish-sahoo: Good catch — fixed in c73635c8. Added an `Input.Bool` field and a `boolean` case in `NewInputsForMode` (`strconv.ParseBool`), so `-i flag=true` coerces to a real bool before validation. Covered `-i flag=true`/`flag=false` in `TestNewInputsForMode_CoercesBool`.

anish-sahoo: Fixed in c73635c8. `NewInputsForMode` now coerces both repeated (`-i nums=1 -i nums=2`) and single (`-i nums=1`) values against the array item schema, and `toMap` preserves the typed elements instead of flattening to `[]string`. So `list[int]` reaches the runtime as `[1, 2]` and passes preflight. Covered in `TestNewInputsForMode_CoercesRepeatedIntArray`/`_CoercesSingleIntArray`.

pkg/cli/predict.go
anish-sahoo: Fixed in c73635c8. JSON-mode file read errors now include the key: `input "image": failed to read file "..."`, consistent with the `-i` path.

pkg/cli/predict.go
anish-sahoo: Fixed in c73635c8. Added `predict.HasInputComponent`; preflight now only runs when the schema label actually carries the `Input`/`TrainingInput` component. A minimal or malformed-but-parseable label falls back to the runtime schema after start instead of failing here. Applied in both `predict.go` and `train.go`.

pkg/predict/input_validation.go
anish-sahoo: Keeping the stricter behavior intentionally — rejecting unknown inputs catches typos before an expensive build/pull/start, whereas the runtime silently strips + warns. I documented this as a deliberate divergence in a comment on `validateKnownInputs` (327b15d5). Agree it is worth aligning the runtime later so the two do not drift; happy to open a follow-up issue for that.

pkg/cli/predict.go
anish-sahoo: Done in 327b15d5. Local-source `predict`/`train` now generate the schema once and thread the bytes into the build via the new `BuildOptions.OpenAPISchema`; the build reuses them instead of regenerating, so preflight validation and the image label share one source of truth (no drift). Pulled the schema-selection logic into `resolveBuildSchema` with unit tests for the precedence.

pkg/cli/predict.go
anish-sahoo: Investigated this one closely. The runtime validates inputs as **strict JSON Schema** (Rust `jsonschema` against the bundled `Input` component) — not pydantic — so it does not normalize/coerce. That means the two examples here already match: a numeric string for a numeric field is rejected by both, and `Any` is `{"type": "object"}` in both (must be a JSON object).

The one real divergence I found was explicit `null`: kin-openapi honors the OpenAPI `nullable` keyword and *accepted* an explicit null for an optional field, while the runtime ignores `nullable` and *rejects* it (422). Fixed in 327b15d5 — preflight now rejects explicit null (`rejectExplicitNulls`), matching the runtime; omit the input to get a null default.

integration-tests/tests/string_predictor.txtar
anish-sahoo: Added in c73635c8: `run_missing_required_input_before_build.txtar` (local source — asserts `missing required input "prompt"` with no build/start) and a missing-required assertion on the existing-image path in `string_predictor.txtar`.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,283 @@
...
+
+// rejectExplicitNulls rejects inputs whose value is an explicit JSON null.
+//
+// The runtime validates inputs as strict JSON Schema and ignores the OpenAPI
+// `nullable` keyword, so it rejects an explicit null for every field, including
+// optional ones — the way to get a null default is to omit the input entirely.
+// kin-openapi honors `nullable`, so without this check it would accept a null
+// that the server would then reject with a 422. Rejecting here keeps preflight
+// consistent with runtime.
+func rejectExplicitNulls(input map[string]any) error {
```
markphelps: One more edge case here: this only checks top-level input keys, but the comment above says runtime rejects explicit null “for every field.” A nested value like `{"opts":{"scale":null}}` can still get through this preflight check and then fail later at runtime because coglet treats the schema as strict JSON Schema and ignores OpenAPI `nullable`. Could we make this schema-aware/recursive and report the full path, e.g. `invalid input "opts.scale": must not be null`? I would avoid blanket-recursing through unconstrained `Any` / `additionalProperties` if null is valid there.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+func formatSchemaValidationError(err error) string {
+	var schemaErr *openapi3.SchemaError
+	if errors.As(err, &schemaErr) {
+		reason := schemaErr.Reason
+		if reason == "" {
+			reason = fmt.Sprintf("doesn't match schema %q", schemaErr.SchemaField)
+		}
+		if missingInput, ok := missingRequiredInput(reason); ok {
```
markphelps: This rewrite happens before we look at the JSON pointer, so nested missing-required errors lose their parent path. For example, if `opts.scale` is required and `opts` is `{}`, this can report `missing required input "scale"`, which reads like a top-level input. Could we preserve/include the path here, e.g. `missing required input "opts.scale"`, and add an exact wording test for nested required fields?

pkg/predict/input.go
```diff
@@ -51,121 +52,128 @@ func NewInputsForMode(keyVals map[string][]string, schema *openapi3.T, isTrain b
...
-									continue
-								}
+						}
+					}
+					// A single repeated-style value (e.g. `-i nums=1`) becomes a
+					// one-element array; coerce it to the array's item type.
+					arr := []any{coerceScalarValue(val, arrayItemSchema(propertySchema))}
+					input[key] = Input{Array: &arr}
+					continue
+				case propertySchema.Type.Is("boolean"):
```
markphelps: Optional follow-up: bool coercion is fixed for direct boolean schemas, but boolean unions still look incomplete. Numeric unions have `schemaAcceptsNumber` / `schemaAcceptsFloat` fallbacks below; there is not an equivalent `schemaAcceptsBool`. For `str | bool`, `-i flag=true` may stay as the string branch if that is the first resolved type. That might be an acceptable ambiguity, but I think we should either add bool-union coercion/tests or document that `--json` is required for ambiguous bool/string unions.

anish-sahoo: Fixed in 7153eedb. Boolean union coercion now handles both member orders and mixed numeric/boolean unions, accepts only JSON `true`/`false`, and validates a typed candidate against union constraints before choosing it over a valid string branch. Added coverage for both orders, constraints, aliases, and mixed unions.

pkg/predict/input_validation.go
anish-sahoo: Fixed in 7153eedb. Missing-required formatting now uses the SchemaError JSON pointer, so nested failures report the complete path such as `missing required input "opts.scale"`. Added an exact-wording regression test.

pkg/predict/input_validation.go
anish-sahoo: Fixed in 7153eedb. Explicit-null validation now walks nested declared properties, constrained additional properties, arrays, and composed schemas while preserving the full path. Unconstrained additional values and schemas that permit null without OpenAPI nullable are left alone. Added nested, allOf, and additionalProperties regression tests.

pkg/predict/input.go
```diff
@@ -184,32 +144,33 @@ func (inputs *Inputs) toMap() (map[string]any, error) {
 			// Single file handling: read content and convert to a data URL
 			dataURL, err := fileToDataURL(*input.File)
 			if err != nil {
-				return keyVals, err
+				return keyVals, fmt.Errorf("input %q: %w", key, err)
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 2618c482 — `toMap` now returns `nil` (not the partially-populated `keyVals` map) on error. Callers already discard the result when `err != nil`, but this removes the brittle partial-map footgun.

pkg/predict/input.go
```diff
@@ -184,32 +144,33 @@ func (inputs *Inputs) toMap() (map[string]any, error) {
...
+			// with '@') or values already coerced to their schema type.
+			values := make([]any, len(*input.Array))
 			for i, elem := range *input.Array {
 				if str, ok := elem.(string); ok && strings.HasPrefix(str, "@") {
-					filePath := str[1:] // Remove '@' prefix
-					dataURL, err := fileToDataURL(filePath)
+					dataURL, err := fileToDataURL(str[1:]) // strip '@' prefix
 					if err != nil {
-						return keyVals, err
+						return keyVals, fmt.Errorf("input %q: %w", key, err)
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 2618c482 — same fix on the array element (`@file`) path: returns `nil` on error instead of the partial map.

pkg/predict/input.go
```diff
@@ -33,6 +33,12 @@ func NewInputsForMode(keyVals map[string][]string, schema *openapi3.T, isTrain b
 		schemaKey = "TrainingInput"
 	}
 	var inputComponent *openapi3.SchemaRef
+	// Defensive: callers guard against a schema without components (build paths
+	// generate+validate it, existing images gate on HasInputComponent), but
+	// guard here too so a future caller can't trip a nil-map deref.
+	if schema == nil || schema.Components == nil {
+		return Inputs{}, nil
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 7f82a618. When the schema is nil or has no components, `NewInputsForMode` still parses values as strings/files (and repeated flags as string arrays) instead of returning an empty map. Covered by `TestNewInputsForModeWithoutComponentsPreservesValues`.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,475 @@
...
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+// HasInputComponent reports whether the schema has the Input (or TrainingInput)
+// component needed by preflight validation, so callers can fall back to the
+// runtime schema for images with minimal or missing labels.
+func HasInputComponent(schema *openapi3.T, isTrain bool) bool {
+	_, err := inputComponentForMode(schema, isTrain)
+	return err == nil && schema.Validate(context.Background()) == nil
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 7f82a618. `HasInputComponent` now only checks that the Input/TrainingInput component exists; it no longer runs full-document `schema.Validate`. Incomplete labels (e.g. missing `paths`) can still drive preflight when the component is present. Covered by `TestHasInputComponentAcceptsIncompleteDocument`.

Also guarded `resolved.Type` nil checks in `newSingleInput` in the same commit so oneOf/composition schemas without a concrete wrapper type do not panic.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,474 @@
...
+// ValidateInputsForMode validates CLI inputs after schema-directed coercion.
+func ValidateInputsForMode(inputs Inputs, schema *openapi3.T, isTrain bool) error {
+	component, err := inputComponentForMode(schema, isTrain)
+	if err != nil {
+		return err
+	}
+	if err := validateKnownInputNames(inputs, component); err != nil {
+		return err
+	}
+	inputMap, err := inputs.toMap()
```
markphelps: This materializes file inputs during validation, but the returned `Inputs` still contains the original `File`/`@...` values. `Predictor.Predict` calls `toMap()` again, so file-backed inputs are read twice. That breaks one-shot sources such as `-i image=@/dev/stdin` (the request gets EOF after preflight) and can block on a FIFO; it also means the bytes we validate can differ from the bytes we send if a normal file changes. Can we materialize these inputs once and reuse that representation for both validation and the request? A regression test using stdin or a single-write FIFO would catch this.

pkg/predict/input_validation.go
```diff
@@ -0,0 +1,474 @@
...
+	"strings"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+// HasInputComponent reports whether the schema has the Input (or TrainingInput)
+// component needed by preflight validation, so callers can fall back to the
+// runtime schema for images with minimal or missing labels.
+func HasInputComponent(schema *openapi3.T, isTrain bool) bool {
+	_, err := inputComponentForMode(schema, isTrain)
```
markphelps: This treats any non-nil `Input` component as usable, even when the label contains an incomplete object schema with no `properties`. In that case we take the preflight path and reject every supplied input as unknown rather than falling back to the runtime schema:

```json
{"components": {"schemas": {"Input": {"type": "object"}}}}
```

An explicit empty properties object can still mean the model intentionally has no inputs, but an absent/nil properties map looks like the minimal/incomplete-label case this fallback is intended to cover. Could `HasInputComponent` require a resolved properties map before opting into preflight validation, with coverage for this label shape?

markphelps (changes requested):
A couple of compatibility issues to address before this is ready.

pkg/predict/input_validation.go
anish-sahoo: Fixed by materializing file inputs once after successful validation and reusing those values for the request.

pkg/predict/input_validation.go
anish-sahoo: Fixed by falling back when properties is absent while still accepting an explicit empty object.

markphelps (approved):

Issue #389, opened 2022-01-13
`cog run` does not work with A100


mitchellgordon95:
Trying to do inference on a machine with an A100 GPU. Here's my cog file:

```
image: "myimage"
build:
  gpu: true
  python_version: "3.8"
  cuda: "11.3"
  python_packages:
    - "transformers==4.15.0"
    - "torch==1.10.1"
    - "ray[serve]==1.5.0"
    - "kfserving==0.6.1"
```

When I do `sudo cog run python model.py` I get the following error:

```
/root/.pyenv/versions/3.8.12/lib/python3.8/site-packages/torch/cuda/__init__.py:106: UserWarning:
NVIDIA A100-PCIE-40GB with CUDA capability sm_80 is not compatible with the current PyTorch installation.
The current PyTorch install supports CUDA capabilities sm_37 sm_50 sm_60 sm_70.
If you want to use the NVIDIA A100-PCIE-40GB GPU with PyTorch, please check the instructions at https://pytorch.org/get-started/locally/

  warnings.warn(incompatible_device_warn.format(device_name, capability, " ".join(arch_list), device_name))
RuntimeError: CUDA error: no kernel image is available for execution on the device
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
> /src/model.py(24)predict()
-> output_ids = self.model.generate(input_ids.cuda())
```

So apparently my CUDA installation is missing some capabilities. Looking at stack overflow, it seems like doing

`pip3 install torch==1.10.1+cu113 -f https://download.pytorch.org/whl/cpu/torch_stable.html`

might fix that problem? But I can't pin `1.10.1+cu113` since cog doesn't allow us to specify custom pip archives via `pip -f`

zeke:
Thanks for reporting and following up, @mitchellgordon95 👍🏼

pranay-ar:
Hi @mitchellgordon95 , I noticed ray in your cog yaml. I assume you're trying to serve the inference using ray serve and use cog for packaging the model + environment? Curious to know your experience to how it's been?

pranay-ar:
Hi @mitchellgordon95, I noticed Ray in your Cog YAML. I assume you're using Ray Serve for inference and Cog for packaging the model and environment? Curious to hear about your experience—how has it been?

PR #3104, opened 2026-07-14
feat: detect when the resolved torch/CUDA ships no kernels for the local GPU

golithe:
- Cog resolves CUDA from the framework pin and never consults the GPU it runs on
- On a Blackwell card (sm_120), `torch==2.4.1` builds green, and every kernel launch fails with `CUDA error: no kernel image is available for execution on the device`
- Hard to notice: `torch.cuda.is_available()` returns `True`, and PyTorch's own diagnostic is a `UserWarning`
- There's no fallback either, since the wheel ships neither sm_12x cubins nor PTX to JIT from
- #389 is the same failure class on an A100 in 2022 (closed when the reporter worked around it); Blackwell makes it current again

This PR adds: 
- `GPUCompatibilityCheck` to `cog doctor`, modelled on `DockerCheck`
-  Reads compute capability from `nvidia-smi --query-gpu=compute_cap` (every distinct GPU, not just the weakest, since a mixed-GPU host must clear each device's floor) and compares the resolved torch/CUDA against the oldest release known to ship kernels for it.

`cog doctor` output when it fires:
  ```
   ⚙  Environment
   ✔ Docker
   ✔ Python version
   ⚠ GPU compatibility
   ⚙    torch==2.4.1 (CUDA 12.4) ships no kernels for sm_120, the compute capability of
        this machine's GPU. The image will build, but every CUDA operation in it will
        fail at runtime with "no kernel image is available for execution on the device".

   ⚙  Found 1 warning.
  ```
Notes for review:

- The floors are measured, not read off release notes. I installed each wheel and read `torch._C._cuda_getArchFlags()`, which works without a GPU
- Each floor is bracketed: the named version ships the kernels, the release below it doesn't. sm_120/sm_100 need torch>=2.7.0 + CUDA>=12.8; sm_90 needs torch>=2.0.1 + CUDA>=11.8
- Both bounds are load-bearing: a 2.7 build against an old CUDA (cu118) has no Blackwell kernels. The CUDA compared is the one from the wheel Cog actually resolves and installs (same resolution as `PythonRequirementsForArch`), not the raw requirement, so it's correct in both directions: `torch==2.7.0+cu118` with `cuda: "12.8"` resolves to the `cu128` wheel and stays silent, while an explicit `--extra-index-url=.../cu118` installs the cu118 wheel regardless of `cuda:` and warns against 11.8
- Deliberately no rows below sm_90. Every probed wheel already covers Turing/Ampere/Ada, so those floors can't be bracketed; unknown-old devices produce no finding rather than a guess
- `SeverityWarning`, matching `PythonVersionCheck`: the image is valid and runs fine on other hardware; it only fails when executed on this machine's GPU
- Silent when there's no GPU, no torch pin, or `gpu: false`; `COG_SKIP_GPU_CHECK=1` skips it when building for different hardware than the local card. Doctor-only, no build-path changes
- The torch pin is matched PEP 503-style, so `Torch==` and `torch[extra]==` resolve rather than silently skipping the check.

Testing:

- the comparison is a pure function (`evaluateGPUCompat`), table-tested without a GPU, including `+cu128` local-tag pins and the new-torch/old-CUDA case
- Verified by hand on an RTX 5070 Ti (sm_120): fires on `torch==2.4.1`, silent on `2.7.1` and on `gpu: false`

Out of scope (deliberately): 
- the floors are hand-entered from measurement (precedent: `MinimumTorchVersion` and friends in `pkg/dockerfile/base.go`)
- A probe script in `tools/` to regenerate them on demand could be a follow-up; full `compatgen` integration seems like a poor fit, since new capability majors ship every couple of years and the probe needs multi-GB wheel installs. Happy to add the script here or in a follow-up PR if useful

### DIFF
--- a/pkg/doctor/check_env_gpu_compat.go
+++ b/pkg/doctor/check_env_gpu_compat.go
@@ -1,0 +1,199 @@
+package doctor
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/replicate/cog/pkg/util/version"
+)
+
+// GPUFloor is the oldest torch (and CUDA) release that ships executable kernels for a
+// given compute capability.
+//
+// CUDA cubins are binary-compatible only within a compute capability major; crossing majors
+// requires embedded PTX for the driver to JIT, which older torch wheels do not ship. Both
+// bounds matter: torch 2.7.0+cu118 is a genuine 2.7 build with no Blackwell kernels.
+type GPUFloor struct {
+	MinTorch string
+	MinCUDA  string
+}
+
+// gpuFloors maps compute capability (major, minor) to its floor. Lookup takes the highest
+// entry the device meets or exceeds; a device below every entry yields no finding.
+//
+// Measured by installing each wheel and reading torch._C._cuda_getArchFlags(), which works
+// without a GPU. Each floor is bracketed: the named version ships kernels for the
+// capability, the release below it does not.
+//
+//	torch 2.7.0+cu128   cubins: sm_75 sm_80 sm_86 sm_90 sm_100 sm_120   PTX: compute_120
+//	torch 2.6.0+cu126   cubins: sm_50 ... sm_90                         PTX: none
+//	torch 2.4.1+cu124   cubins: sm_50 ... sm_90                         PTX: none
+//	torch 2.0.1+cu118   cubins: sm_37 ... sm_90                         PTX: none
+//	torch 1.13.1+cu117  cubins: sm_37 ... sm_86                         PTX: none
+//
+// No rows below sm_90: every probed wheel already covers Turing/Ampere/Ada, so those floors
+// cannot be bracketed. When extending, bracket the new row the same way rather than guess.
+//
+// sm_89 (Ada) appears in no cubin list yet RTX 40xx works: sm_86 cubins run on sm_89 (same
+// major). sm_120 gets no such fallback from sm_90.
+var gpuFloors = map[[2]int]GPUFloor{
+	{12, 0}: {MinTorch: "2.7.0", MinCUDA: "12.8"}, // Blackwell consumer, RTX 50xx
+	{10, 0}: {MinTorch: "2.7.0", MinCUDA: "12.8"}, // Blackwell datacenter, B100/B200
+	{9, 0}:  {MinTorch: "2.0.1", MinCUDA: "11.8"}, // Hopper, H100
+}
+
+// GPUCompatibilityCheck verifies that the resolved torch/CUDA versions can execute on this
+// machine's GPU. Cog derives CUDA from the framework pin without consulting the device, so
+// a build can succeed and produce an image where every CUDA operation fails at runtime with
+// "no kernel image is available for execution on the device".
+type GPUCompatibilityCheck struct{}
+
+func (c *GPUCompatibilityCheck) Name() string        { return "env-gpu-compat" }
+func (c *GPUCompatibilityCheck) Group() Group        { return GroupEnvironment }
+func (c *GPUCompatibilityCheck) Description() string { return "GPU compatibility" }
+
+func (c *GPUCompatibilityCheck) Check(ctx *CheckContext) ([]Finding, error) {
+	if os.Getenv("COG_SKIP_GPU_CHECK") != "" {
+		return nil, nil
+	}
+	if ctx.Config == nil || !ctx.Config.Build.GPU {
+		return nil, nil
+	}
+
+	torchVersion, hasTorch := ctx.Config.TorchVersion()
+	if !hasTorch {
+		return nil, nil
+	}
+
+	capability, ok := detectComputeCapability(ctx.ctx)
+	if !ok {
+		// No GPU on this machine (or no driver); nothing to compare against.
+		return nil, nil
+	}
+
+	return evaluateGPUCompat(capability, torchVersion, ctx.Config.Build.CUDA), nil
+}
+
+// evaluateGPUCompat compares a resolved torch/CUDA pair against the floor for the given
+// compute capability. Kept free of I/O so it is testable without a GPU.
+func evaluateGPUCompat(capability [2]int, torchVersion string, cuda string) []Finding {
+	floor, ok := floorFor(capability)
+	if !ok {
+		// Older than any floor we know; say nothing rather than guess.
+		return nil
+	}
+
+	if torchVersion == "" {
+		// torch listed but unpinned; pip resolves a current release. Nothing to compare.
+		return nil
+	}
+
+	torchOK := version.GreaterOrEqual(torchVersion, floor.MinTorch)
+	cudaOK := cuda == "" || version.GreaterOrEqual(cuda, floor.MinCUDA)
+	if torchOK && cudaOK {
+		return nil
+	}
+
+	sm := fmt.Sprintf("sm_%d%d", capability[0], capability[1])
+	// Warning, not error: the image is valid and runs on other hardware; it only fails on
+	// this machine's GPU (same taxonomy as PythonVersionCheck).
+	return []Finding{{
+		Severity: SeverityWarning,
+		Message: fmt.Sprintf(
+			"torch==%s (CUDA %s) ships no kernels for %s, the compute capability of this machine's GPU. "+
+				"The image will build, but every CUDA operation in it will fail at runtime with "+
+				"\"no kernel image is available for execution on the device\".",
+			torchVersion, cudaDisplay(cuda), sm,
+		),
+		Remediation: fmt.Sprintf(
+			"%s requires torch>=%s built against CUDA>=%s. Pin a newer torch, or set "+
+				"COG_SKIP_GPU_CHECK=1 if you are building for a different GPU than the one in this machine.",
+			sm, floor.MinTorch, floor.MinCUDA,
+		),
+	}}
+}
+
+func (c *GPUCompatibilityCheck) Fix(_ *CheckContext, _ []Finding) error {
+	return ErrNoAutoFix
+}
+
+func cudaDisplay(cuda string) string {
+	if cuda == "" {
+		return "unset"
+	}
+	return cuda
+}
+
+// detectComputeCapability returns the lowest compute capability among the machine's GPUs:
+// the image has to run on the weakest device present.
+func detectComputeCapability(ctx context.Context) ([2]int, bool) {
+	execCtx, cancel := context.WithTimeout(ctx, envCheckTimeout)
+	defer cancel()
+
+	out, err := exec.CommandContext(execCtx,
+		"nvidia-smi", "--query-gpu=compute_cap", "--format=csv,noheader").Output()
+	if err != nil {
+		return [2]int{}, false
+	}
+
+	var caps [][2]int
+	for line := range strings.SplitSeq(string(out), "\n") {
+		if c, ok := parseCapability(line); ok {
+			caps = append(caps, c)
+		}
+	}
+	if len(caps) == 0 {
+		return [2]int{}, false
+	}
+
+	sort.Slice(caps, func(i, j int) bool {
+		if caps[i][0] != caps[j][0] {
+			return caps[i][0] < caps[j][0]
+		}
+		return caps[i][1] < caps[j][1]
+	})
+	return caps[0], true
+}
+
+// parseCapability parses a single nvidia-smi compute_cap value, e.g. "12.0".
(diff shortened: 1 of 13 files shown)


pkg/doctor/check_env_gpu_compat.go
```diff
@@ -0,0 +1,199 @@
...
+		return [2]int{}, false
+	}
+
+	sort.Slice(caps, func(i, j int) bool {
+		if caps[i][0] != caps[j][0] {
+			return caps[i][0] < caps[j][0]
+		}
+		return caps[i][1] < caps[j][1]
+	})
+	return caps[0], true
```
anish-sahoo: Could we evaluate every detected capability, or select the capability with the strictest applicable floor? Sorting ascending and returning only `caps[0]` checks the numerically lowest GPU, which is not necessarily the least compatible one. On a host with `sm_90` and `sm_120`, torch 2.4.1/CUDA 12.4 passes the `sm_90` floor and produces no finding even though operations on `sm_120` will fail. A mixed-GPU regression test would help cover this.

pkg/doctor/check_env_gpu_compat.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (c *GPUCompatibilityCheck) Check(ctx *CheckContext) ([]Finding, error) {
+	if os.Getenv("COG_SKIP_GPU_CHECK") != "" {
+		return nil, nil
+	}
+	if ctx.Config == nil || !ctx.Config.Build.GPU {
+		return nil, nil
+	}
+
+	torchVersion, hasTorch := ctx.Config.TorchVersion()
```
anish-sahoo: `TorchVersion()` does not establish that this is an exact pin: the current requirement parsing discards the comparator and returns the first version token. As a result, `torch<2.7.0` is treated as `2.7.0`, so this check can pass even though pip installs an older, incompatible release. Could we evaluate only normalized exact `==` pins and skip ranges, direct URLs, and unresolved requirements? This should also have regression coverage.

pkg/doctor/check_env_gpu_compat.go
```diff
@@ -0,0 +1,199 @@
...
+		// Older than any floor we know; say nothing rather than guess.
+		return nil
+	}
+
+	if torchVersion == "" {
+		// torch listed but unpinned; pip resolves a current release. Nothing to compare.
+		return nil
+	}
+
+	torchOK := version.GreaterOrEqual(torchVersion, floor.MinTorch)
```
anish-sahoo: `version.GreaterOrEqual` includes local metadata in equality, so `2.7.0+cu128` is not considered greater than or equal to `2.7.0`: the release is not greater, and equality fails because the metadata differs. That produces a false warning at the exact supported floor. Could we strip the local modifier for the torch release comparison, retain it separately for CUDA resolution, and add an exact-boundary test?

pkg/doctor/check_env_gpu_compat_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// Complete populates the requirements content that TorchVersion reads; without it the
+	// check skips before reaching the comparison.
+	require.NoError(t, cfg.Complete(t.TempDir()))
+	return &CheckContext{ctx: context.Background(), ProjectDir: t.TempDir(), Config: cfg}
+}
+
+func TestGPUCompatibilityCheck_RunsWithoutError(t *testing.T) {
+	// A GPU may or may not be present in the test environment; just ensure no panic or error.
+	ctx := gpuContext(t, true, "2.4.1", "12.4")
+	_, err := (&GPUCompatibilityCheck{}).Check(ctx)
```
anish-sahoo: This test only verifies that the check returns no error and ignores its findings, so the production detection path is not exercised deterministically. Could we test it with a fake `nvidia-smi` and cover multiple GPUs, malformed output, command failure, and registration through the doctor runner? That would also catch the mixed-GPU issue noted above.

pkg/doctor/gpu_compatibility.go
```diff
@@ -0,0 +1,199 @@
...
+// a build can succeed and produce an image where every CUDA operation fails at runtime with
+// "no kernel image is available for execution on the device".
+type GPUCompatibilityCheck struct{}
+
+func (c *GPUCompatibilityCheck) Name() string        { return "env-gpu-compat" }
+func (c *GPUCompatibilityCheck) Group() Group        { return GroupEnvironment }
+func (c *GPUCompatibilityCheck) Description() string { return "GPU compatibility" }
+
+func (c *GPUCompatibilityCheck) Check(ctx *CheckContext) ([]Finding, error) {
+	if os.Getenv("COG_SKIP_GPU_CHECK") != "" {
```
anish-sahoo, line 61: Could we document `COG_SKIP_GPU_CHECK` in `docs/environment.md` alongside `COG_SKIP_DOCKER_CHECK`, then regenerate `docs/llms.txt`? This is a new user-facing environment variable and its intended use when building for different hardware should be discoverable outside the source.

pkg/doctor/gpu_compatibility.go
```diff
@@ -0,0 +1,199 @@
+package doctor
```
anish-sahoo, line 1: Could we rename these files to `gpu_compatibility.go` and `gpu_compatibility_test.go`? We prefer subject-based filenames here and spell out `compatibility` rather than shortening it to `compat`.

anish-sahoo (changes requested):
Thanks for putting this together. This is a useful diagnostic and the overall approach looks promising. I found several compatibility edge cases, along with some test and documentation gaps, that need additional fixes before we can work toward merging this. Details are inline.

pkg/doctor/gpu_compatibility.go
golithe: done

pkg/doctor/gpu_compatibility.go
golithe: done

pkg/doctor/check_env_gpu_compat.go
golithe: the check now evaluates every distinct capability rather than only the lowest. your described scenario (sm_90 + sm_120 with torch 2.4.1 is the regression test `TestGPUCompatibilityCheck_MixedGPUsFireForStrictestFloor`

wdyt?

pkg/doctor/check_env_gpu_compat.go
golithe: - added `requirements.ExactVersion` and `Config.TorchExactVersion` where only a single `==` specifier counts
- ranges/wildcards/direct URLs/unpinned all produce no finding rather than a guess
- tested in `TestExactVersion` and `TestGPUCompatibilityCheck_SkipsNonExactPin`

pkg/doctor/check_env_gpu_compat.go
golithe: - the torch comparison now strips the local tag before `version.GreaterOrEqual`, so 2.7.0+cu128 passes at the exact floor
- CUDA bound is still checked separtely from Build.CUDA
- boundari case in `TestEvaluateGPUCompat` ("sm_120 exact floor with local tag passes)

pkg/doctor/check_env_gpu_compat_test.go
golithe: - replaced the `RunsWithoutError` test entirely
- the nvidia-smi probe is now a package-level functin variable that tests stub
- went with that over fake bin on PATH to keep it hermetic, same intent
- now covers multiple GPUs, command failure, malformed output, non-exact pins, and runner registration

golithe:
@anish-sahoo ty for your thorough review! i addressed your comments, PTAL

anish-sahoo (changes requested):
Thanks for addressing the first review round. The mixed-GPU evaluation, exact-pin handling for ordinary `torch==...` requirements, local-tag boundary case, deterministic probe tests, docs, and rename all look good to me.

I found two remaining issues that need to be addressed before merge:

1. `gpu_compatibility.go` strips the local wheel tag and then relies only on `build.cuda`. This can silence a real Blackwell incompatibility: `torch==2.7.0+cu118` with an explicit `cuda: "12.8"` passes the check because it compares the stripped `2.7.0` and base CUDA 12.8 against the floor. But Cog retains the `+cu118` wheel, and that wheel has no Blackwell kernels regardless of its CUDA base image. Please preserve/derive the wheel CUDA tag for compatibility evaluation, rather than losing it when normalizing the torch release, and add a full `Config.Complete -> Check` regression test for this override combination.

2. `Config.TorchExactVersion()` compares `requirements.PackageName(pkg)` directly to lowercase `"torch"`. Exact valid requirements such as `torch[extra]==2.4.1` and `Torch==2.4.1` therefore return no torch pin and skip the GPU check. Please normalize the base distribution name (strip extras and normalize casing, ideally PEP 503-style) before matching, with an end-to-end check regression test.

golithe:
> `gpu_compatibility.go` strips the local wheel tag and then relies only on `build.cuda`.

good catch. now derives the wheel's CUDA from the `+cuXXX` local tag (via the exported `config.CUDAVersionFromTorchLocalTag`) and compares that against the floor instead of `build.cuda`, so `torch==2.7.0+cu118` with `cuda: "12.8"` fires. regression: `TestGPUCompatibilityCheck_LocalCUDATagOverridesBuildCUDA`

> `Config.TorchExactVersion()` compares `requirements.PackageName(pkg)` directly to lowercase `"torch"`.

fixed. added `requirements.NormalizePackageName` (PEP 503: strip extras, collapse -_., lowercase) and match on it in` pythonPackageExactVersion`, so `torch[extra]==2.4.1` and `Torch==2.4.1` resolve instead of being skipped. Regression: `TestNormalizePackageName` + end-to-end `TestGPUCompatibilityCheck_NormalizesTorchName`.

golithe:
@anish-sahoo adressed both, PTAL

pkg/doctor/gpu_compatibility.go
```diff
@@ -0,0 +1,250 @@
...
+	if torchVersion == "" {
+		// torch listed but unpinned; pip resolves a current release. Nothing to compare.
+		return nil
+	}
+
+	// A wheel's +cu118 tag is authoritative over the build's base CUDA: torch==2.7.0+cu118
+	// ships no Blackwell kernels even under cuda: "12.8", so evaluate against 11.8, not 12.8.
+	// Fall back to the build CUDA only when there is no cuXYZ tag to derive from.
+	effectiveCUDA := cuda
+	if wheelCUDA, _ := config.CUDAVersionFromTorchLocalTag(torchVersion); wheelCUDA != "" {
```
anish-sahoo: I re-tested this and need to correct my earlier assumption. The GPU check and the build resolve CUDA differently:

- `torch==2.7.0+cu118` with `cuda: "12.8"` builds with the cu128 wheel, but the check uses cu118 and warns incorrectly.
- An explicit cu118 index builds with the cu118 wheel, but the check uses `cuda: "12.8"` and does not warn.

Could the check use the final resolved requirement and index from `PythonRequirementsForArch`? That would ensure it checks the wheel Cog will actually install.

anish-sahoo (changes requested):
The GPU compatibility check can evaluate a different wheel than Cog actually installs, which causes both false positives and false negatives. Everything else LGTM.

pkg/doctor/gpu_compatibility.go
golithe: good catch again, and you're right. 

- the check now resolves the wheel via the same path as `PythonRequirementsForArch` (new `Config.ResolvedTorchWheel`) and derives CUDA from the resolved package index instead of the requirement's local tag, so it matches what Cog installs in both directions: `torch==2.7.0+cu118` + `cuda: "12.8"` resolves to cu128 and stays silent, and an explicit `--extra-index-url=.../cu118` fires on 11.8. 
- Regressions for both in `TestResolvedTorchWheel` and the two `GPUCompatibilityCheck` tests. PTAL.

anish-sahoo (approved):
Thank you for contributing. LGTM!

golithe:
> Thank you for contributing. LGTM!

Thank you for reviewing! cheers

anish-sahoo:
note to @golithe - this will be merged soon, we are a bit swamped right now, I'm waiting for one of the repo admins to override the codeql check.

golithe:
hi @anish-sahoo, is there any action required from my side regarding the failing CI? i rebased on main to have a clear diff, the failing CI steps mention too many requests.

cheers

michaeldwan (approved):
nice, thanks!

michaeldwan:
@golithe GH is stuck waiting on CodeQL checks that passed weeks ago. Going to push an empty commit to trigger it and get this merged. Sorry for the wait!

golithe:
@michaeldwan thanks!

PR #3106, opened 2026-07-14
5. Add Cog playground API client

anish-sahoo:
- cog api client for config, health, schema, predictions, sse, and file conversion
- includes http/sse/api unit tests
- cog-playground stack: #3109, #3108, #3114, #3110/#3106, #3105/#3111/#3107, #3113, #3112

### DIFF
--- a/playground/src/services/cog/CogApi.ts
+++ b/playground/src/services/cog/CogApi.ts
@@ -1,0 +1,132 @@
+import type { HealthResponse } from "@/types/health";
+import type { OpenAPIDocument } from "@/types/openapi";
+import type { PredictionEnvelope, StreamEvent } from "@/types/prediction";
+import { parseJSONResponse, parsePredictionResponse, responseError } from "@/services/cog/http";
+import { readSSE } from "@/services/cog/sse";
+
+const PROXY_PREFIX = "/proxy";
+
+type SubmitOptions = {
+  target: string;
+  endpoint: string;
+  id?: string;
+  input: Record<string, unknown>;
+  signal: AbortSignal;
+  async?: boolean;
+  webhook?: string;
+  webhookEvents?: string[];
+  onResponse?: (response: Response) => void;
+};
+
+type StreamOptions = Pick<
+  SubmitOptions,
+  "target" | "endpoint" | "id" | "input" | "signal" | "onResponse"
+>;
+
+/** Routes model requests through the same-origin proxy rather than exposing cross-origin access. */
+export class CogApi {
+  /** Loads the playground's optional target, webhook, and version configuration. */
+  async config(
+    signal?: AbortSignal,
+  ): Promise<{ target?: string; webhookBase?: string; cogVersion?: string }> {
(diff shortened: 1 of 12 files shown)


playground/src/services/cog/files.ts
```diff
@@ -0,0 +1,9 @@
+/** Converts a browser file to a data URI and rejects if the file cannot be read. */
+export function fileToDataURI(file: File): Promise<string> {
```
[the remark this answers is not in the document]
anish-sahoo: Fixed in 9ee1410a: local file conversion rejects files over 16 MiB before constructing FileReader. A regression test verifies the reader is not created.

anish-sahoo:
Follow-up disposition: validated against the Streams Standard and a local Web Streams probe. `cancel()` on an explicitly acquired reader does not release its lock, so the `releaseLock()` calls in `sse.ts` and `http.ts` are required and do not mask the intended errors. Both attached review threads are resolved.

markphelps (commented):
I spotted two type-boundary issues in the new API client. Both are small structurally, but worth fixing before callers start depending on these public types.

playground/src/services/cog/http.ts
```diff
@@ -0,0 +1,74 @@
...
+  const detail = Array.isArray(body.detail) ? body.detail : undefined;
+  const message =
+    (typeof body.error === "string" && body.error) ||
+    (typeof body.detail === "string" && body.detail) ||
+    text ||
+    `HTTP ${response.status}`;
+  return new HttpError(message, response.status, detail);
+}
+
+/** Reads JSON under the response-size limit without performing runtime shape validation. */
```
markphelps: `T` is only an assertion here: `JSON.parse` returns untrusted data, and `as T` makes any successful response look like a valid `HealthResponse`, `OpenAPIDocument`, or `PredictionEnvelope`. For example, a target can return `null`, an array, or fields with the wrong types and callers will still receive the declared type. Could we return `unknown` from this helper and validate/narrow it at each API boundary (AJV is already in the playground), rather than exposing a generic parser that cannot establish `T`?

playground/src/services/cog/CogApi.ts
```diff
@@ -0,0 +1,132 @@
...
+import { parseJSONResponse, parsePredictionResponse, responseError } from "@/services/cog/http";
+import { readSSE } from "@/services/cog/sse";
+
+const PROXY_PREFIX = "/proxy";
+
+type SubmitOptions = {
+  target: string;
+  endpoint: string;
+  id?: string;
+  input: Record<string, unknown>;
```
markphelps: Since this client serializes `input` itself, `Record<string, unknown>` is too broad for the wire contract. It accepts values that `JSON.stringify` cannot represent (`bigint`, cycles) or silently drops (functions). We already have `JsonObject` in `@/types/json`; could this be `input: JsonObject` so invalid prediction input is rejected by TypeScript before we attempt to make the request?

markphelps (approved):

PR #3107, opened 2026-07-14
8. Add playground prediction controls and output

anish-sahoo:
- sync, streaming, and async prediction controls with bounded output tracing
- includes toolbar, webhook, output, hook, and trace unit tests
- cog-playground stack: #3109, #3108, #3114, #3110/#3106, #3105/#3111/#3107, #3113, #3112


anish-sahoo:
Reviewed the reported EventSource race. `eventSourceReady` installs `onerror` synchronously in its Promise executor; connection failures are awaited in `run()`'s `try`/`catch`, and the post-open handler is attached before a later browser task can deliver an error. No unhandled rejection path was found, so no code change is needed. Existing coverage includes post-connect failure and connection timeout.

This is a top-level comment, so GitHub provides no review thread to resolve.

markphelps (approved):

PR #3131, opened 2026-08-03
chore(deps): dependabot

anish-sahoo:
adds dependabot fixes from the last month


.github/workflows/ci.yaml
```diff
@@ -136,7 +136,7 @@ jobs:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
-      - uses: jdx/mise-action@v4
+      - uses: jdx/mise-action@v4.2.3
```
markphelps: why do we have to pin the minor/patch version? shouldnt pinning the major always use the latest minor/patch?

.github/workflows/codeql.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v4
+      uses: github/codeql-action/init@v4.37.3
```
markphelps: same q here

.github/workflows/codeql.yml
```diff
@@ -56,7 +56,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, Go, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v4
+      uses: github/codeql-action/autobuild@v4.37.3
```
markphelps: and here, basically everywhere

.github/workflows/ci.yaml
anish-sahoo: good point - I've updated all of them to stay pinned to the major version + updated the dependabot config for github actions to only track major version changes

markphelps (approved):

PR #3154, opened 2026-08-13
feat: add --playground flag to cog serve

anish-sahoo:
Adds `--playground` to `cog serve` so it starts the playground alongside the model and prints both URLs:

```
Serving at http://127.0.0.1:8393
Playground at http://127.0.0.1:9000
```

The playground binds its own port (`--playground-port`, default 9000, `0` = free port) and never collides with the model port.

- `pkg/cli/serve.go`: new `--playground` and `--playground-port` flags; validates the playground port doesn't overlap the model port; binds the playground before the container starts so a bind failure aborts fast; prints the playground URL next to `Serving at`; runs it with graceful shutdown. `cmdServe` now cancels its context on signal, so Ctrl+C exits cleanly instead of erroring.
- `pkg/cli/playground.go`: pulls the standalone command's startup into a reusable `startPlayground(ctx, playgroundConfig)` helper. `cog playground` is unchanged.
- Tests for the port validation, host overlap, the target URL helper, and the new helper.

Only `serve` supports `--playground` for now; `predict`/`run`/`train` run the container headlessly with no public port.

### DIFF
--- a/pkg/cli/serve.go
+++ b/pkg/cli/serve.go
@@ -1,12 +1,17 @@
 package cli
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 
@@ -19,9 +24,11 @@ import (
 )
 
 var (
-	serveHost = command.DefaultHostIP
-	port      = 8393
-	uploadURL = ""
+	serveHost           = command.DefaultHostIP
+	port                = 8393
+	uploadURL           = ""
+	servePlaygroundFlag = false
+	servePlaygroundPort = 9000
 )
 
 func newServeCommand() *cobra.Command {
@@ -42,6 +49,9 @@ the Docker port mapping is published on.`,
 
   # Start on a custom port
   cog serve -p 5000
+
+  # Start the playground alongside the model
+  cog serve --playground
 
   # Listen on all interfaces (e.g. to expose to the network)
   cog serve --host 0.0.0.0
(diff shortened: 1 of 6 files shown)


pkg/cli/serve.go
```diff
@@ -200,6 +257,24 @@ func cmdServe(cmd *cobra.Command, arg []string) error {
...
 	serveURL := formatServeURL(serveHost, port)
 
+	// Start the playground before the container so a bind failure aborts fast,
+	// before the multi-minute image build and run. It targets the model's
+	// actual bind host so the UI reaches the server being served.
+	var playgroundURL string
+	var playgroundSrv *http.Server
+	var playgroundLn net.Listener
+	if servePlaygroundFlag {
+		playgroundURL, playgroundSrv, playgroundLn, err = startPlayground(ctx, playgroundConfig{
```
markphelps, line 273: This makes async mode unusable for the integrated playground on Linux. `playgroundHost` is still the standalone command default (`127.0.0.1`), while the UI tells the model to send webhooks to `host.docker.internal`. The Linux host-gateway mapping is only added when `--upload-url` is set, and even with that mapping a container cannot reach a loopback-only listener through the gateway. Could the embedded playground bind on a container-reachable interface and add `host.docker.internal:host-gateway` whenever `--playground` is enabled? An end-to-end async webhook test here would catch this.

pkg/cli/serve.go
```diff
@@ -105,8 +117,53 @@ func formatServeURL(host string, port int) string {
 	return url
 }
 
+// validateServePorts errors when the model and playground would bind the same
+// port on overlapping interfaces. A playground port of 0 asks for a free port,
+// so it never collides.
+func validateServePorts(serveHost string, port, playgroundPort int, playgroundHost string) error {
+	if playgroundPort == 0 {
```
markphelps, line 126: Port 0 does not guarantee that the playground avoids the model port: the playground listener is opened before Docker publishes the model port, so for a user-selected model port in the ephemeral range the OS can assign that exact port to the playground. Docker will then fail to start the model. After binding, please compare the resolved playground port with the model port (and retry or fail clearly if they overlap), rather than treating `0` as collision-free.

markphelps (changes requested):
Thanks for factoring the playground startup out cleanly. I found two edge cases that need fixing before this is ready.

pkg/cli/serve.go
anish-sahoo: The embedded playground now binds to 0.0.0.0 and adds the host-gateway alias whenever --playground is enabled.

pkg/cli/serve.go
anish-sahoo: The bound playground port is now checked before Docker starts, and Cog closes the listener and returns an error if it collides with the model port.

markphelps (approved):

PR #3160, opened 2026-08-21
Fix uv-managed CUDA base compatibility

### LINKED ISSUE #3159: CUDA 13 Cog base fails Python dependency install under PEP 668
## Problem

A clean build with `r8.im/cog-base:cuda13.0-python3.13-torch2.11.0` fails while installing `python_requirements`.

Cog generates:

```text
uv run pip install --cache-dir /root/.cache/pip -r /tmp/requirements.txt
```

The base image uses a uv-managed Python installation, so pip rejects the install:

```text
error: externally-managed-environment
This Python installation is managed by uv and should not be modified.
```

## Reproduction

Using `cog version 0.22.0`:

```yaml
build:
  gpu: true
  cuda: "13.0"
  python_version: "3.13"
  python_requirements: requirements.txt
run: "run.py:Runner"
```

```text
# requirements.txt
torch==2.11.0
requests==2.32.5
```

Run:

```bash
cog build --no-cache
```

The build selects the published CUDA 13 Cog base, then fails on the requirements install shown above. `cog debug` confirms that the command does not include `--break-system-packages`.

## Cause

The fix in #3073 only enables `--break-system-packages` on the fallback `nvidia/cuda` path:

```go
g.Config.Build.GPU && g.useCudaBaseImage && !g.IsUsingCogBaseImage()
```

The CUDA 13 Cog base also uses uv-managed Python, but `IsUsingCogBaseImage()` excludes it from the fix.

## Expected

Cog should install requirements successfully when it selects a published Cog base image.

## Workaround

`--use-cog-base-image=false` adds the required flag, but gives up the prebuilt image and rebuilds from `nvidia/cuda`.

Related: #2994 contains an earlier report in its comments, but mainly tracks CUDA 13 base-image availability.

anish-sahoo:
- Pass `--break-system-packages` when CUDA 13+ Cog bases provide uv-managed Python.
- Add the `python` alias expected by Cog runtime and image inspection commands.
- Apply the alias to both prebuilt CUDA 13 bases and the direct CUDA path.
- Update `h2` to `0.4.16` for the current RustSec advisory.
- Keep existing behavior for older Cog bases and non-CUDA builds.

Closes #3159

### DIFF
--- a/crates/Cargo.lock
+++ b/crates/Cargo.lock
@@ -503,7 +503,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


michaeldwan (approved):
looks good, thanks!

